PR #98, opened 2026-04-21
feat: integrate glossary arcade module

tetheusz:
Após o comentário do kauanet incentivando a ideia decidi tentar agregar mais com projeto, adicionei a parte de gameficação ao app.


thomgabriel:
Closing per review. The glossary dataset and game mechanics are genuinely valuable, but this PR ships fake client-side XP (client-chosen amounts, exceeds the 100/lesson cap) and a parallel dead i18n path, and removes Sanity fail-fast out of scope. Salvage path post-launch: games without XP (or wired through a server route modeled on /api/quests/daily) + glossary served server-side. Thanks @tetheusz — the dataset is worth keeping.

PR #171, opened 2026-06-24
fix(security): stop passing ADMIN_SECRET into the client admin component (P0-B6)

### LINKED ISSUE #151: [P0-B6] Stop passing ADMIN_SECRET into the client component

`admin/page.tsx` reads `process.env.ADMIN_SECRET` and passes it as a prop to the `"use client"` AdminClient — serialized into page source/React props (XSS-stealable, post-auth).

- [ ] Remove the secret prop; client POSTs the user-entered secret to `/api/admin/auth` and relies on the HMAC `admin_session` cookie only.
- [ ] Rotate `ADMIN_SECRET` after the fix (feeds D-2 / G-1).

**Done when:** `ADMIN_SECRET` no longer appears in any client bundle or page payload. **Refs:** `apps/web/src/app/[locale]/admin/page.tsx`.

thomgabriel:
Closes #151

### The leak
`apps/web/src/app/[locale]/admin/page.tsx` read `process.env.ADMIN_SECRET` and passed it as an `adminToken` prop into the `"use client"` `AdminClient`. Next.js serializes client-component props into the page's React/RSC payload and inlines them into the client bundle, so the admin secret was shipped to the browser and was XSS-stealable. The secret was then re-sent on every admin call as an `Authorization: Bearer <secret>` header.

### Auth-flow change (header/prop path → cookie path)
The HMAC `admin_session` cookie machinery already existed (`POST /api/admin/auth` mints it; middleware gates `/admin` on it). This PR makes the admin **API routes** authorize via that cookie too, and removes the client-held secret entirely:

- **`admin/page.tsx`** — no longer reads `ADMIN_SECRET`; gates on the cookie via `isValidAdminSession(session)`; renders `<AdminClient />` with **no token prop**.
- **`AdminClient` + `course-sync-table` + `achievement-sync-table` + `data-resync-panel`** — drop the `adminToken` prop and the `Authorization` header from every `fetch`. The `admin_session` cookie (httpOnly) is sent automatically on same-origin requests.
- **`lib/admin/auth.ts`** — `requireAdminAuth(req)` now validates the `admin_session` cookie (reusing the existing `isValidAdminSession` HMAC + expiry check) instead of the `Bearer` header.
- **`api/admin/resync/route.ts`** — replaced its inline header-based `verifyAdminToken` with the shared cookie-based `requireAdminAuth` (the other admin routes already call `requireAdminAuth`, so they're covered by the helper change).

Login UX is unchanged: enter secret → `POST /api/admin/auth` → `admin_session` cookie set → admin actions work via the cookie.

### Canary-grep proof (the Done-when)
Built with a sentinel `ADMIN_SECRET=LEAKCANARY-ADMIN-deadbeef1234`, then grepped the client bundle:

```
$ grep -rl "LEAKCANARY-ADMIN-deadbeef1234" apps/web/.next/static
# (no matches) → PASS: canary absent from client bundle
$ grep -rl "LEAKCANARY-ADMIN-deadbeef1234" apps/web/.next
# (no matches anywhere in .next) → secret not baked into any artifact;
#   route handlers read process.env.ADMIN_SECRET at runtime
```
grep correctness self-tested (writing the canary to a temp file under `.next` is found by the same `grep -rl`). `.next/static` contained 124 client files; the canary is in none.

### Verification
- `pnpm --filter @superteam-lms/web typecheck` → exit 0
- `pnpm --filter @superteam-lms/web build` → success (placeholder env vars)
- `next lint` / lint-staged (eslint --fix + prettier) → clean
- `git grep adminToken` / `Bearer ${adminToken}` over `apps/web/src` → no matches

### Follow-up (human ops)
**`ADMIN_SECRET` rotation is a required follow-up ops step (D-2/G-1).** This PR stops *exposing* the secret but does not rotate it; the previously-shipped value should be rotated out of band.

### DIFF
--- a/apps/web/src/app/[locale]/admin/admin-client.tsx
+++ b/apps/web/src/app/[locale]/admin/admin-client.tsx
@@ -55,11 +55,7 @@ interface AdminStatus {
   achievements: AchievementStatus[];
 }
 
-interface AdminClientProps {
-  adminToken: string;
-}
-
-export function AdminClient({ adminToken }: AdminClientProps) {
+export function AdminClient() {
   const [status, setStatus] = useState<AdminStatus | null>(null);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState<string | null>(null);
@@ -68,9 +64,8 @@ export function AdminClient({ adminToken }: AdminClientProps) {
     setLoading(true);
     setError(null);
     try {
-      const res = await fetch("/api/admin/status", {
-        headers: { Authorization: `Bearer ${adminToken}` },
-      });
+      // Authorized via the httpOnly admin_session cookie (sent automatically).
+      const res = await fetch("/api/admin/status");
       if (!res.ok) {
         setError("Failed to fetch status");
         return;
@@ -82,7 +77,7 @@ export function AdminClient({ adminToken }: AdminClientProps) {
     } finally {
       setLoading(false);
     }
-  }, [adminToken]);
+  }, []);
 
   useEffect(() => {
     void fetchStatus();
@@ -158,7 +153,6 @@ export function AdminClient({ adminToken }: AdminClientProps) {
           ) : (
             <CourseSyncTable
               courses={courses}
-              adminToken={adminToken}
               onRefresh={() => void fetchStatus()}
             />
           )}
@@ -178,7 +172,6 @@ export function AdminClient({ adminToken }: AdminClientProps) {
           ) : (
             <AchievementSyncTable
               achievements={achievements}
-              adminToken={adminToken}
               onRefresh={() => void fetchStatus()}
             />
           )}
@@ -191,7 +184,7 @@ export function AdminClient({ adminToken }: AdminClientProps) {
           Data Resync
         </h2>
         <div className="rounded-lg border border-border bg-card p-4 shadow-card">
-          <DataResyncPanel adminToken={adminToken} />
+          <DataResyncPanel />
         </div>
       </section>
     </div>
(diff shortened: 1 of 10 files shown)


thomgabriel:
**Handing off for human review** — `area:security` → human-gated. Gates are finishing (CI/review running); the implementer pre-verified: **canary-grep confirms `ADMIN_SECRET` is absent from the client bundle** (0/124 client files), typecheck + build pass. Required follow-up after merge: **rotate `ADMIN_SECRET`** (D-2/G-1) — this PR stops exposing it but does not rotate the already-shipped value. @thomgabriel — confirm final CI, then review + merge.

thomgabriel:
## Adversarial review found a CSRF regression → fixed
An independent skeptical review confirmed the secret-removal + HMAC cookie crypto are **sound** (no forgery), but found a **[HIGH] CSRF regression**: moving admin-route auth from a `Bearer` header (CSRF-immune) to an ambient cookie with only `SameSite=Lax` left the admin routes — which sign **on-chain mutations with the platform authority key** — CSRF-exposed.

**Fixed in `b2fe024`:** `SameSite=Strict` + a `Sec-Fetch-Site`/`Origin` same-origin check in `requireAdminAuth` for state-changing methods; plus malformed-cookie → clean 401 (was a 500), and a shared Node/Edge session-format module to prevent drift. typecheck + build pass; login flow verified end-to-end, cross-origin POSTs now rejected. Still needs final human sign-off + the **`ADMIN_SECRET` rotation** (D-2/G-1).

@claude re-review — fixes in `b2fe024`.

PR #179, opened 2026-06-24
fix(security): stop shipping lesson solutions/hidden tests + cap challenge XP (P0-C4)

### LINKED ISSUE #150: [P0-C4] Stop shipping lesson solutions + hidden tests; server-side challenge validation

`getLessonBySlug` projects `tests` (incl. `hidden:true`) and `solution` into the page payload; the client only filters hidden tests for *display*. Answers are in the network response and "passed" is browser-trusted (gate G1 — credential proves nothing).

- [ ] Project the GROQ query to send only `code`, `hints`, `tests where hidden==false`.
- [ ] Move hidden-test checking + solution reveal to a server route.
- [ ] `/api/lessons/complete` must distinguish challenge lessons and not trust a browser "passed" (interim: per-user daily XP ceiling + capped challenge XP in SQL).
- [ ] Regression test: solution + hidden tests absent from the lesson network payload.

**Done when:** devtools shows no `solution`/`hidden:true` tests in the payload and challenge completion is server-validated. **Refs:** `apps/web/src/lib/sanity/queries.ts:86-113`, `lesson-client.tsx:493-509,323`.

thomgabriel:
Closes #150

## Problem
`getLessonBySlug` (and the shared `moduleWithLessonsFields` projection used by `getCourseBySlug`/`getCourseById`) projected `solution` and **all** `tests` — including `hidden: true` — into the page payload. The client only filtered hidden tests for *display*, so the answer key rode in the network response and challenge "passed" was browser-trusted → forgeable XP/credentials (gate G1: a credential proved nothing).

### 1. Stop shipping answers (the core fix)
- **`apps/web/src/lib/sanity/queries.ts`** — `getLessonBySlug` and `moduleWithLessonsFields` now project `"tests": tests[hidden != true]{ id, description, input, expectedOutput }` and **no longer project `solution`**. `tests[hidden != true]` keeps tests where `hidden` is `false`/undefined and excludes only `true`; re-projecting the fields drops the `hidden` flag itself. The AI-chat route (the other `getLessonBySlug` consumer) only reads `title`/`content`, so it is unaffected.
- **`packages/types/src/course.ts`** — split the client `TestCase` (no `hidden`) from a server-only `AdminTestCase` (`hidden?`). Removed `solution` from the client `ChallengeLesson`; added server-only `AdminChallengeLesson` (solution + hidden tests). Exported the two new types.
- **`lesson-client.tsx` / `challenge-interface.tsx` / `editor/types.ts`** — removed the now-redundant client-side `hidden` filter (the server already excludes them) and made `ChallengeInterface.solution` optional; the "Show solution" affordance hides when the solution is absent.

### 2. Interim XP integrity (defense-in-depth) — `supabase/schema.sql` `award_xp()`
A forged client "passed" is bounded so it cannot mint unbounded XP:
- **Hard per-award cap** (2000) — clamps any single award.
- **Per-user/day ceiling** (5000) across the learning XP path (lesson/challenge/bonus/achievement). Community XP is **excluded** from this sum because it already has its own 50/day cap in `award_community_xp`.
- Clamps rather than rejects, so legitimate one-time awards (e.g. a course-completion bonus up to 2000) still apply. All constructs mirror the existing `award_community_xp` (param reassignment, `reason LIKE 'community:%'`, `(CURRENT_DATE)::timestamptz`, `COALESCE(SUM(...))`).
- ⚠️ **Needs a Supabase apply** — this SQL is in the schema file but is not applied to any running DB by this PR (cannot run it here).

> Honest framing: today `award_xp` is only ever called from the Helius webhook with an **on-chain-derived** amount (`event.xpEarned`, from the course's admin-set `xp_per_lesson`) — the browser cannot send an arbitrary XP number. These caps are genuine defense-in-depth (they bound blast radius if on-chain config were misconfigured or the award path were abused); they are **not** a mechanism that detects a forged "passed". The real residual risk a forged pass leaves — marking a challenge complete without solving it — is closed only by true server-side validation (see Follow-up).

### 3. Server-side validation primitive (no untrusted-code execution)
- **`getChallengeAnswerKey` (SERVER-ONLY, `queries.ts`)** — returns the full test set (incl. hidden) + solution, fetched uncached (`revalidate=0`, never via the public CDN). This is the deliberately-withheld data, retrievable server-side so it never needs to ship to the browser.
- **`POST /api/lessons/validate-challenge`** — auth-gated; loads the answer key server-side and returns only pass/fail-style metadata (authoritative hidden/visible test counts, `serverValidated`, `reason`). It **never** serializes tests or the solution to the client.

## Regression proof (the Done-when)
`apps/web/src/lib/sanity/__tests__/queries-answer-leak.test.ts` (Vitest, 5 cases) mocks `sanityFetch` to capture the GROQ string and asserts:
- `getLessonBySlug` projects **no** `solution` and uses `"tests": tests[hidden != true]{ id, description, input, expectedOutput }` (and not a bare `tests,`).
- `getCourseBySlug` (course-detail payload) is likewise free of `solution`/hidden tests.
- A simulated returned payload serializes with **no** `solution`/`hidden` keys.
- `getChallengeAnswerKey` (server-only) *does* include the answer key **and** is uncached (`revalidate === 0`).

**How verified locally (worktree, placeholder env for build):**
- `pnpm --filter @superteam-lms/web typecheck` → exit 0
- `pnpm --filter @superteam-lms/web test` → **28/28 pass** (incl. the 5 new)
- `pnpm --filter @superteam-lms/web build` → exit 0; `/api/lessons/validate-challenge` registered in the route manifest
- `pnpm --filter @superteam-lms/web lint` → exit 0 (no new findings)
- SQL: docker unavailable here, so no live parse. Validated structurally (balanced `$$` / `BEGIN…END`) and every construct has in-file precedent in the already-deployed `award_community_xp`.

## Follow-up / out of scope (deferred)
- **Full server-side challenge execution.** Per the issue, this PR does **not** run untrusted learner code in a trusted context — that is a separate sandboxing problem. `/api/lessons/validate-challenge` therefore returns `serverValidated: false` (`reason: "server_execution_unavailable"`) for code challenges today; it is the home for that logic once a sandbox exists.
- **Wire `/api/lessons/complete` to require server validation** for challenge lessons (refuse XP/finalize on an unverified pass) once the sandbox lands. The XP caps are the interim guard until then.
- Apply the `award_xp` change to Supabase (migration/`db push`).

> NOTE: SENSITIVE (security + `schema.sql`). Flagged for independent adversarial review before any human merge.


thomgabriel:
## Handing off for human merge — SENSITIVE (security + `schema.sql`)

Both gates passed (CI green, `claude[bot]` = no blocking), then per our SENSITIVE-PR discipline I ran an **independent adversarial review** (a skeptic told to *break* the two claims, not confirm them). Summary for whoever merges:

### Adversarial verdict — no critical/high; both claims hold
- **Answer-key leak closed on every client-reaching path** — all GROQ projections (lesson + course + `moduleWithLessonsFields`) use `tests[hidden != true]{…}` and drop `solution`; the AI routes (`api/ai/chat`, `api/ai/suggest`) never fetch or echo the key; the lesson/course RSC props no longer carry `solution`/hidden tests. The sole full-key fetcher (`getChallengeAnswerKey`) is server-only, `revalidate=0`, never serialized to a client.
- **XP forging bounded** — `/api/lessons/complete` doesn't trust a client "passed"; XP is minted server-side from the on-chain event and clamped by `award_xp` (≤2000/award, ≤5000/day learning path). No new forgeable write path.

### In-scope review fixes applied (commit `b24fa99`)
Both reviewers flagged these, in functions/files this PR already touches:
1. **`award_xp` race** → per-user `pg_advisory_xact_lock` so the read-then-insert daily cap is atomic.
2. **`award_xp` daily window** → pinned to UTC midnight (`date_trunc('day', now() AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'`), session-TZ-independent.
3. **Regression test** → removed the misleading `{0}` quantifier (now `/\btests,/`).

Verified: `typecheck` + the 5 answer-leak cases pass.

### Filed as tracked follow-ups (not merge blockers)
- **#181 (P1)** — wire **real** server-side challenge validation into completion. `validate-challenge` is server-only scaffolding today (unwired); a forged "passed" still produces a completion record (XP is bounded, but credential integrity isn't enforced). This is #150's deferred hard half — needs a sandboxed executor.
- **#182 (P2)** — `award_community_xp` has the same race + TZ pattern; harden for consistency.

### ⚠️ MUST-DO on merge (operational)
**Apply the `award_xp` change to the live DB.** Editing `supabase/schema.sql` is inert until the function is re-created on the running instance — the caps don't exist in prod just because this merges. Confirm the Supabase deploy path runs it (and ideally fold #182 in while you're there).

@claude please re-review the hardened commit `b24fa99`.

Marking `needs-human-review` — a human signs off + merges (SENSITIVE: `schema.sql` + `area:security`). 🤖

Issue #107, opened 2026-06-23
[P0-A3] Enforce a cumulative XP ceiling in `reward_xp`


Potolski:
> **Code legend / glossary:** see [`docs/TASK-CODES.md`](https://github.com/solanabr/superteam-academy/blob/main/docs/TASK-CODES.md) for how to read task codes (e.g. `P0-A1`), severities, and area labels.

- [ ] Add optional `max_total_xp` to `MinterRole`; enforce `total_xp_minted + amount <= max_total_xp`.
- [ ] Test: minter blocked after cap.

**Done when:** a compromised minter key can't mint unbounded XP. **Refs:** `instructions/reward_xp.rs:14-19,33-37`.

thomgabriel:
**Scope expansion (from #185's adversarial review).** While hardening `reward_xp` for the cumulative XP ceiling, ALSO add the missing authorization gate: `reward_xp` (`instructions/reward_xp.rs:50-84`) currently mints **arbitrary XP to an arbitrary recipient gated only by an active `MinterRole`** — no `backend_signer` check. Because `register_minter` can register any pubkey and `MinterRole ≠ config.backend_signer`, an active minter alone can mint arbitrary XP, violating #106's program-wide invariant.

Add the identical gate used everywhere else — `backend_signer: Signer<'info>` + `constraint = backend_signer.key() == config.backend_signer @ AcademyError::Unauthorized` (copy `award_achievement.rs:185-188`) — plus the IDL entry, the `rewardXp()` call-site in `academy-program.ts:413-457` (already signs with the backend keypair → ~1-line account add), and a wrong-`backend_signer`-rejected test. So #107's PR delivers BOTH: the cumulative ceiling AND the backend-signer gate, fully hardening `reward_xp`.

PR #185, opened 2026-06-24
fix(onchain): require backend_signer + eligibility proof in award_achievement (P0-A2)

### LINKED ISSUE #106: [P0-A2] Add eligibility proof to `award_achievement`

- [ ] Require `backend_signer` (or dedicated role) check.
- [ ] Pass + validate relevant `Enrollment` receipt(s) on-chain where feasible.
- [ ] Tests: non-eligible recipient rejected; only authorized signer can mint.

**Done when:** an active minter alone can't mint arbitrary achievements/XP to arbitrary recipients. **Refs:** `instructions/award_achievement.rs:15-27`.

thomgabriel:
Closes #106

## The hole

`award_achievement` previously gated **only** on:
- a `minter: Signer`
- an active `MinterRole` PDA (`minter_role.minter == minter && minter_role.is_active`)
- an active `AchievementType` + supply check

There was **no `backend_signer` authorization and no recipient-eligibility proof**. Any single active minter could mint **any** achievement — and its `xp_reward` — to **any arbitrary recipient**.

## Authorization model now enforced

Added a `backend_signer: Signer<'info>` account constrained `== config.backend_signer @ AcademyError::Unauthorized`, mirroring `complete_lesson` / `issue_credential` / `finalize_course` / `upgrade_credential`. Every XP-minting / NFT-issuing instruction in the program now requires the rotatable backend signer to co-sign. **An active minter alone can no longer mint** — satisfying the issue's "Done when" for *all* achievement types.

## Eligibility proof — partial / deferred (no link fabricated)

Checklist item 2 (pass + validate an `Enrollment` receipt on-chain) is **not generally feasible** here, and I deliberately did not fake it:

- `AchievementType` (`state/achievement_type.rs`) carries **no on-chain link to a qualifying action** — no `course`, no criterion, no reference to any receipt. Fields are: `achievement_id, name, metadata_uri, collection, creator, max_supply, current_supply, xp_reward, is_active, created_at`.
- Achievements are generic. Of the 15 defined, most have **no course at all** (streaks: Week Warrior / Monthly Master / Consistency King; community: Helper / First Comment / Top Contributor; special: Early Adopter / Bug Hunter / Perfect Score). There is no `Enrollment` (or any PDA) that proves "recipient earned *this* achievement."
- Eligibility is computed **off-chain** (`apps/web/src/lib/gamification/achievements.ts`, `UNLOCK_CHECKS`) and the award is enqueued server-side.

Passing an `Enrollment` would only fit the handful of course-completion achievements and would require a schema change (adding a course/criterion link to `AchievementType`) plus a migration — out of scope for a P0 authorization fix, and meaningless for the majority of achievement types. So per-recipient on-chain proof is **deferred**; the `backend_signer` co-signature is the authorization boundary, exactly as for every other XP-minting instruction. The `AchievementReceipt` PDA already prevents double-awards on-chain.

> If desired as a follow-up: add an optional `qualifying_course: Option<Pubkey>` to `AchievementType` and, when set, require a matching completed `Enrollment` for the recipient. Tracked separately rather than fabricated here.

## Frontend / IDL changes (live award flow kept working)

The live caller already signs as both `minter` and `payer` with the backend keypair (`getBackendSigner()`), which is auto-registered as a minter at `initialize` and **is** `config.backend_signer`. So:

- `apps/web/src/lib/solana/academy-program.ts` — `awardAchievement()` now passes `backendSigner: signer.publicKey`. No `.signers()` change (the NodeWallet already signs it). No call-site changes needed (`onchain-queue.ts` / helius `event-handlers.ts` pass only `achievementId` + wallet).
- `apps/web/src/lib/solana/idl/superteam_academy.json` — added the `backend_signer` (signer) account. **Account order verified identical** to the Anchor-generated `target/idl/onchain_academy.json`.

## Tests

- **New:** `award_achievement rejects wrong backend_signer (active minter alone cannot mint)` — `minter` is the legitimately-registered active backend minter, but `backend_signer` is an attacker key. Asserts `Unauthorized` **and** that no `AchievementReceipt` was created (nothing minted). This is the direct P0-A2 regression guard.
- Happy path still succeeds (NFT soulbound, attributes, +200 XP, supply increment).
- 5 existing `awardAchievement()` call-sites updated to pass `backendSigner` (the IDL now requires it).

## Verification (run locally — actual results)

| Step | Result |
|------|--------|
| `anchor build` | ✅ compiled; generated IDL `award_achievement` account order matches checked-in IDL |
| `cargo fmt --check` | ✅ clean (no diff) |
| `cargo clippy -- -W clippy::all` | ✅ 0 errors (only pre-existing `anchor-debug` cfg warnings from the `#[program]`/`#[derive(Accounts)]` macros, present on all 17 instructions) |
| `cargo test` (rust unit) | ✅ **77 passed; 0 failed** |
| `anchor test` (localnet) | ✅ **63 passing** (incl. the new wrong-backend-signer test) |
| `program_autofixer` (changed Rust) | ✅ `{ issues: [], suggestions: [], require_another_tool_call_after_fixing: false }` |
| `tsc --noEmit` (apps/web) | ✅ clean |

> `anchor test` was run on localnet with a throwaway signer + temporary local program id (the real `wallets/*` keypairs are gitignored and absent from the worktree). Those local-only edits to `Anchor.toml` / `lib.rs` were reverted; the commit touches **only** the 4 intended files.

## Uncertainty flagged honestly

- **Eligibility proof feasibility:** I'm confident it is not generally feasible without a schema change — `AchievementType` has no qualifying-action link and most achievements have no course. The adversarial reviewer should confirm whether a partial course-only proof is wanted now vs. deferred.
- **Off-chain signing flow:** unchanged in behavior — `backend_signer == minter == getBackendSigner()` in production. The only new requirement is that the backend keypair equals `config.backend_signer`, which is already the invariant for `complete_lesson` etc.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,8 +65,29 @@ BUILD_SERVER_URL=https://academy-build-server-HASH.a.run.app
 # API key for X-API-Key header (same value as ACADEMY_API_KEY on Cloud Run)
 BUILD_SERVER_API_KEY=                      # PRIVATE — server-only
 
-# ── App URL (for sitemap, OG tags, NFT metadata URI) ─────────────────────────
+# ── App URL (for sitemap, OG tags, NFT metadata URI fallback) ────────────────
 NEXT_PUBLIC_APP_URL=http://localhost:3000
+
+# ── Arweave / Irys — permanent credential metadata (server-only) ─────────────
+# At credential mint, certificate metadata is pinned to Arweave via Irys so the
+# on-chain Metaplex Core asset URI resolves independently of app uptime
+# (apps/web/src/lib/solana/arweave.ts). The uploader returns an Irys GATEWAY URL
+# (https://gateway.irys.xyz/<id>) — NOT https://arweave.net/<txid> — and that is
+# the exact string written on-chain. On mainnet the data settles on Arweave, so
+# the same <id> is also reachable via https://arweave.net/<id> and ar://<id>.
+# If unset, mint falls back to the app-served /api/certificates/metadata URL
+# (logs a warning) — minting still works, but issued credentials will NOT
+# resolve if the app is offline.
+#
+# Value: 64-element JSON array of a SOLANA keypair's secret-key bytes (same
+# shape as BACKEND_SIGNER_SECRET). Irys is funded with this wallet's SOL — NOT
+# an Arweave JWK. The human must fund it:
+#   - devnet (NEXT_PUBLIC_SOLANA_NETWORK=devnet): uses https://devnet.irys.xyz,
+#     free, but uploads are pruned after ~60 days — NOT truly permanent.
+#   - mainnet (NEXT_PUBLIC_SOLANA_NETWORK=mainnet-beta): uses
+#     https://uploader.irys.xyz, permanent, paid in SOL from this wallet's Irys
+#     balance (top up at irys.xyz before mainnet minting).
+ARWEAVE_UPLOADER_SECRET=
 
 # ── Admin operations (server-only — never expose to browser) ──────────────────
 # JSON array of authority keypair bytes (64 elements), e.g. [12,34,...]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,11 @@ on:
     branches: ["**"] # every branch — gives fork branches CI before the upstream PR
   pull_request:
     branches: [main]
+    # `labeled` lets the verifiable-build job (gated on the `verifiable-build`
+    # label) re-run when the label is added mid-PR. Other jobs ignore the label.
+    types: [opened, synchronize, reopened, labeled]
+  # Manual trigger for on-demand reproducible build + program-hash publishing.
+  workflow_dispatch:
 
 # Cancel superseded runs on the same ref to save CI minutes.
 concurrency:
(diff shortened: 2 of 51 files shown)


thomgabriel:

Both gates passed (CI green, `claude[bot]` Approved), then per SENSITIVE discipline an **independent adversarial review** stress-tested the authorization claim.

### The `award_achievement` fix is correct + tight ✅
- `backend_signer: Signer` + `constraint == config.backend_signer` is a real struct-level gate (enforced at account resolution) — can't pass without a backend signature; the wrong-signer test confirms `Unauthorized` + no receipt created.
- IDL `backend_signer` at index 11 (signer=true) matches the Rust struct order exactly; `awardAchievement()` co-signs. No silent-skip, no order mismatch.
- `config.backend_signer` only rotatable by the Config authority; no overflow/double-award regression (`AchievementReceipt` init-PDA still blocks re-awards).
- Eligibility-proof deferral is well-justified (no on-chain course link on `AchievementType`; backend co-sign is the trust boundary, same as every other mint).

### ⚠️ Adversarial review found a SIBLING hole (the gate missed it)
`reward_xp` (`instructions/reward_xp.rs`) mints **arbitrary XP to an arbitrary recipient gated only by an active `MinterRole` — NO `backend_signer`**. So #106's broad "Done when" isn't met program-wide by this PR alone. **Folded into #107** (which already hardens `reward_xp`) — I expanded its scope to add the identical gate. **Recommend merging #185 (the `award_achievement` portion of #106) and ensuring #107 lands the `reward_xp` gate** before considering the minter-authorization class fully closed.

### Non-blocking (triaged)
- minter/backend_signer redundancy → addressed in spirit by #107's reward_xp gate; no separate cleanup now.
- "assert recipient XP stayed 0" test → declined: the receipt-PDA null check already proves the instruction didn't complete.
- eligibility deferral → accepted as a known, justified limitation (would need an `AchievementType` schema change).

Issue #118, opened 2026-06-23
[P1-2] Rotate `backend_signer` to a distinct hot key at mainnet init


Potolski:
> **Code legend / glossary:** see [`docs/TASK-CODES.md`](https://github.com/solanabr/superteam-academy/blob/main/docs/TASK-CODES.md) for how to read task codes (e.g. `P1-3`), severities, and area labels.

Via `update_config`; scope/deactivate the auto-registered unlimited minter; multisig authority ≠ hot signer.

**Done when:** mainnet `Config.backend_signer ≠ authority`, no unlimited minter. **Refs:** `instructions/initialize.rs:79-95`. *(Feeds G-7; coordinate ops with @Potolski.)*

thomgabriel:
**Cross-link (from #186's adversarial review).** This key-split is also a security *prerequisite*, not just a mainnet-init nicety: in the current config `config.backend_signer == config minter == the one backend key` (the frontend passes `signer.publicKey` for both `minter` and `backendSigner` in `academy-program.ts:456-457`). So the `backend_signer` co-signature added to the XP-minting instructions (#185/#186) is **not a real second factor** until `backend_signer` is split into distinct custody from the minter key. Until then, a single key compromise defeats both controls.

thomgabriel:
**Blocked** on #186 merging — #118 touches `Config`/`initialize`/`update_config` which #186 modified (added `max_total_xp`). Also relates to the single-hot-key custody finding (GAP-3, see #188/#118 thread). Unblock once #186 is merged.

thomgabriel:
Mainnet-init operational step — the program already supports it (rotate via `update_config`'s `new_backend_signer`; scope/deactivate the auto-registered unlimited minter via `update_minter`/`revoke_minter`, both shipped). **No code change is needed.** The Done-when is a *mainnet* `Config` state (`backend_signer ≠ authority`, no unlimited minter), so it can only be performed by a human at mainnet initialization, behind the mainnet-flip gate (#145). Labeling `blocked:needs-human` — it's an init-runbook action, not autonomous work.

thomgabriel:
Closing per triage — mainnet-cutover step, handled at deploy time, not tracked as a standing issue. Checklist lives in docs/LAUNCH-READINESS.md; rotation plan in docs/SECRET-ROTATION.md. Re-open at cutover if useful.

PR #186, opened 2026-06-24
fix(onchain): cumulative XP ceiling + backend_signer gate in reward_xp (P0-A3)

### LINKED ISSUE #107: [P0-A3] Enforce a cumulative XP ceiling in `reward_xp`

thomgabriel:
Closes #107

Hardens `reward_xp` against a compromised minter key with **two** independent controls. SENSITIVE on-chain change — please give it an adversarial review.

## Part A — cumulative XP ceiling (#107)
Previously `reward_xp` enforced only a per-call cap (`max_xp_per_call`) and tracked `total_xp_minted`, but had **no cumulative ceiling** — a compromised minter key could mint unbounded XP across many calls.

- Added `max_total_xp: u64` to `MinterRole` (`0` = unlimited).
- Enforced in the handler, **before** the `mint_xp` CPI, alongside the existing per-call check:
  ```rust
  let new_total = role.total_xp_minted.checked_add(amount).ok_or(Overflow)?;
  if role.max_total_xp > 0 {
      require!(new_total <= role.max_total_xp, AcademyError::MinterCapExceeded);
  }
  ```
  (`new_total` is reused to update `total_xp_minted`, so the checked-add isn't computed twice.)
- New error variant `MinterCapExceeded` (code **6019**).
- Threaded `max_total_xp` into `register_minter` (`RegisterMinterParams`, default `0`) and the `MinterRegistered` event. `initialize`'s auto-registered "backend" role sets it to `0` (unlimited), matching its `max_xp_per_call`.

### MinterRole size / migration decision (no size change, no migration)
`max_total_xp` **consumes the existing `_reserved: [u8; 8]` field in place** — it sits in the exact same byte offset (after `created_at`, before `bump`). Therefore:
- `MinterRole::SIZE` is **unchanged (110 bytes)** — verified by `minter_role_size_constant_is_correct`.
- **Existing on-chain `MinterRole` accounts deserialize unchanged**: their previously-zeroed reserved bytes now read as `max_total_xp = 0` (= unlimited), preserving current behavior until an admin registers a new minter with a cap. There is **no account migration** and no realloc.
- This byte-level guarantee is pinned by a new Rust test, `legacy_zeroed_reserved_bytes_deserialize_as_unlimited_cap`.

> Note: only `register_minter` sets `max_total_xp` — there is **no `update_minter` instruction** in the program, so an existing role's cap can't be changed in place (it would have to be revoked + re-registered). Flagging in case a settable cap is desired later.

## Part B — backend_signer authorization gate (folded in from #185's adversarial review)
`reward_xp` was gated **only** by an active `MinterRole`. Since `register_minter` can register any pubkey and `MinterRole` ≠ `config.backend_signer`, an active minter alone could mint arbitrary XP — violating #106's invariant that XP mints require the backend signer.

- Added `backend_signer: Signer<'info>` to `RewardXp` with `constraint = backend_signer.key() == config.backend_signer @ AcademyError::Unauthorized` — copied verbatim from the pattern in `complete_lesson.rs` / `award_achievement.rs`.
- `reward_xp` now requires **both** an active `MinterRole` for `minter` **and** a signature from `config.backend_signer`.

## Frontend / IDL changes
- **IDL** (`apps/web/src/lib/solana/idl/superteam_academy.json`): regenerated from `anchor build`. Structural diff vs. the previous IDL is *exactly* the intended changes — `backend_signer` on `reward_xp`, `max_total_xp` on `MinterRole`/`RegisterMinterParams`/`MinterRegistered`, and the new `MinterCapExceeded` error (which shifts subsequent error codes 6019→6027).
- **Call-site** (`academy-program.ts` `rewardXp()`): one-line account add — `backendSigner: signer.publicKey`. The backend keypair already signs the tx and is both `minter` and `backend_signer`, so no extra signer plumbing. The two other callers (`/api/quests/daily`, `onchain-queue.ts`) go through this same wrapper, so they need no change.
- **Error map** (`program-errors.ts`): inserted `MinterCapExceeded` at 6019 and renumbered `LabelTooLong`…`InvalidXpReward` (6019→6020 … 6026→6027) to stay in sync with `errors.rs`. This map is hand-maintained and keyed by numeric code, so it *had* to shift.
- **i18n**: added `programErrors.minterCapExceeded` to all three locales (en, pt-BR, es).
- No frontend `registerMinter` JS caller exists (it lives only in the IDL + on-chain), so nothing else needed updating for Part A.

## Tests
On-chain integration (`onchain-academy.ts`, block "15. Minter Roles" — test minter registered with `maxXpPerCall=1000`, `maxTotalXp=1200`):
- happy path still works (`reward_xp` 500 → success);
- **`reward_xp` fails without backend_signer** (active minter alone insufficient) → `Unauthorized`;
- **`reward_xp` fails with wrong backend_signer** → `Unauthorized`, state untouched;
- **cumulative cap blocks** the over-cap call (`500 + 800 > 1200`) → `MinterCapExceeded`, state untouched;
- **boundary**: `500 + 700 == 1200` succeeds (total hits the cap exactly);
- **at-cap** `+1` → `MinterCapExceeded`;
- existing per-call / amount=0 / revoke tests still pass (each updated to pass `backendSigner`).

Rust unit (`test_minter_role.rs`): size constant unchanged, serialization round-trips the new field, and the new legacy-deserialization test pins the zero-init migration guarantee.

## Verification (run locally — actual results)
- `anchor build` → ✅ compiles; IDL regenerated and synced to frontend.
- `cargo fmt --check` → ✅ clean.
- `cargo clippy -- -W clippy::all` → ✅ no new warnings (only pre-existing Anchor-macro `anchor-debug`/glob-reexport noise; none reference the changed files).
- `cargo test --manifest-path tests/rust/Cargo.toml` → ✅ **77 passed; 0 failed**.
- `anchor test` (local validator, program preloaded at the declared ID + mpl_core fixture) → ✅ **67 passing**, including all new/updated minter-role tests.
- `program_autofixer` on `reward_xp.rs`, `register_minter.rs`, `minter_role.rs` → ✅ no issues, `require_another_tool_call_after_fixing: false`.
- `apps/web` `tsc --noEmit` → ✅ no errors.

### DIFF
--- a/apps/web/src/lib/solana/academy-program.ts
+++ b/apps/web/src/lib/solana/academy-program.ts
@@ -454,6 +454,7 @@ export async function rewardXp(
       xpMint: xpMint,
       recipientTokenAccount: recipientTokenAccount,
       minter: signer.publicKey,
+      backendSigner: signer.publicKey,
       tokenProgram: TOKEN_2022_PROGRAM_ID,
     })
     .rpc();
--- a/apps/web/src/lib/solana/idl/superteam_academy.json
+++ b/apps/web/src/lib/solana/idl/superteam_academy.json
@@ -856,6 +856,10 @@
           "signer": true
         },
         {
+          "name": "backend_signer",
+          "signer": true
+        },
+        {
           "name": "token_program",
           "address": "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb"
         }
@@ -1242,41 +1246,46 @@
     },
     {
       "code": 6019,
+      "name": "MinterCapExceeded",
+      "msg": "Cumulative minted XP would exceed minter's total cap"
+    },
+    {
+      "code": 6020,
       "name": "LabelTooLong",
       "msg": "Minter label exceeds max length"
     },
     {
-      "code": 6020,
+      "code": 6021,
       "name": "AchievementNotActive",
       "msg": "Achievement type is not active"
     },
     {
-      "code": 6021,
+      "code": 6022,
       "name": "AchievementSupplyExhausted",
       "msg": "Achievement max supply reached"
     },
     {
-      "code": 6022,
+      "code": 6023,
       "name": "AchievementIdTooLong",
       "msg": "Achievement ID exceeds max length"
     },
     {
-      "code": 6023,
+      "code": 6024,
       "name": "AchievementNameTooLong",
       "msg": "Achievement name exceeds max length"
     },
     {
-      "code": 6024,
+      "code": 6025,
       "name": "AchievementUriTooLong",
       "msg": "Achievement URI exceeds max length"
     },
     {
-      "code": 6025,
+      "code": 6026,
       "name": "InvalidAmount",
       "msg": "Amount must be greater than zero"
     },
     {
-      "code": 6026,
+      "code": 6027,
       "name": "InvalidXpReward",
       "msg": "XP reward must be greater than zero"
     }
@@ -1973,6 +1982,10 @@
             "type": "u64"
           },
           {
+            "name": "max_total_xp",
+            "type": "u64"
+          },
+          {
             "name": "timestamp",
             "type": "i64"
           }
@@ -2035,10 +2048,14 @@
             "type": "i64"
           },
           {
-            "name": "_reserved",
-            "type": {
-              "array": ["u8", 8]
-            }
+            "name": "max_total_xp",
+            "docs": [
+              "Cumulative XP ceiling for this role. 0 = unlimited.",
+              "Occupies the 8 bytes formerly reserved, so the account size is",
+              "unchanged and pre-existing roles deserialize with this field = 0",
+              "(unlimited), preserving prior behavior until an admin sets a cap."
+            ],
+            "type": "u64"
           },
           {
             "name": "bump",
@@ -2062,6 +2079,11 @@
           },
           {
             "name": "max_xp_per_call",
+            "type": "u64"
+          },
+          {
+            "name": "max_total_xp",
+            "docs": ["Cumulative XP ceiling for this role. 0 = unlimited."],
             "type": "u64"
           }
         ]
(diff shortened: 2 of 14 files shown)


thomgabriel:

Both gates passed (CI green, `claude[bot]` approved), then an **independent adversarial review** stress-tested the cap + gate claims. The CODE is correct and tight; the SECURITY CLAIMS need follow-ups (filed).

### Correct + verified ✅
Cap arithmetic (`total_xp_minted + amount <= max_total_xp`, `checked_add`, check-before-CPI); `backend_signer` constraint (real `Signer` + `== config.backend_signer`); `MinterRole` byte-migration (reused reserved bytes → size unchanged, no realloc); error-code renumbering coherence (`errors.rs` ↔ IDL ↔ `program-errors.ts` ↔ i18n all agree). cargo test 77, anchor test 67, tsc clean.

### ⚠️ Ships the cap MECHANISM but does NOT yet deliver #107's "a compromised minter can't mint unbounded XP" — tracked as **P0 #188**
- **GAP-1:** the live backend minter is auto-registered with `max_total_xp = 0` (unlimited) and there's **no `update_minter`** → the cap is inert on the only prod minter until #188 adds a way to set it.
- **GAP-2:** `award_achievement` bumps `total_xp_minted` but **doesn't check `max_total_xp`** → a capped minter can still mint unbounded XP via the achievement path (the cap covers 1 of 4 `mint_xp` paths). #188 mirrors the check there.
- **GAP-3 (custody):** in prod `minter == backend_signer == the one key`, so the co-signature isn't a real second factor until the keys are split — tracked in **#118**.

**Recommendation:** merge #186 as the cap mechanism (it satisfies #107's checklist and is a correct, verified building block), but **do not consider the minter-authorization invariant closed until #188 lands** (+ the #118 key-split for a genuine second factor).

PR #190, opened 2026-06-25
fix(db): data-integrity constraints — NOT NULL user_id, unique tx_signature, FK indexes (P1-9)

### LINKED ISSUE #125: [P1-9] Data-integrity constraints

Add `NOT NULL` to `user_id` FKs (`schema.sql:27,38,60,80`); unique on `tx_signature`; FK indexes on `thread_views`/`flags`.

**Done when:** no orphan-able rows, no dup tx, FK cascades indexed.

thomgabriel:
Closes #125

## P1-9 — Data-integrity constraints

Adds a new migration `supabase/migrations/20260624193000_data_integrity_constraints.sql` and mirrors it into the generated `supabase/schema.sql`. Every statement is idempotent (`ALTER … SET NOT NULL` and `CREATE [UNIQUE] INDEX IF NOT EXISTS` are all no-ops when already applied), so a partial/re-apply is safe.

### 1. `NOT NULL` on ownership `user_id` FK columns (9 tables)
A `NULL` `user_id` makes a row un-ownable: RLS `auth.uid() = user_id` can never match it and it survives the `ON DELETE CASCADE` from `profiles` → orphan-able. Applied to:

`enrollments`, `user_progress`, `user_xp`, `xp_transactions`, `user_achievements`, `certificates`, `deployed_programs`, `pending_onchain_actions`, `user_daily_quests`.

Every legitimate writer (Helius webhook handlers, the retry queue, admin resync, and the `award_xp` / `unlock_achievement` / `get_daily_quest_state` SECURITY DEFINER RPCs) resolves a concrete `user_id` and returns early when it is null, so these columns are always set in practice. Columns already `NOT NULL` (`votes.user_id`, `flags.reporter_id`, `threads.author_id`, `answers.author_id`, and `thread_views.user_id` via its PK) are intentionally left untouched.

### 2. Unique `tx_signature` — `certificates` only (partial)
```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_certificates_tx_signature_unique
  ON certificates (tx_signature) WHERE tx_signature IS NOT NULL;
```
Partial so multiple `NULL` signatures (off-chain / resync rows) stay allowed while no two real mint signatures can collide. `certificates` has a clean **one mint = one signature = one (user, course)** invariant: every non-null writer (`certificates/mint`, `onchain-queue`, the finalize handler) uses a freshly generated mint signature, and the admin-resync writer inserts `NULL`.

**Deliberately NOT added to `enrollments`, `user_progress`, `user_achievements`, or `xp_transactions`.** The Helius webhook decodes *every* Anchor event from a transaction and dispatches each with that transaction's single signature (`lib/helius/event-decoder.ts` + the per-event loop in `app/api/webhooks/helius/route.ts`). So one tx can legitimately write the **same** `tx_signature` to multiple rows of those tables:
- `xp_transactions` — a `CourseFinalized` tx calls `award_xp` twice (learner bonus + creator reward) with the same signature. Dedup there is the **existing** unique `(user_id, idempotency_key)` index, not the signature.
- `user_achievements` — a tx can award more than one achievement.
- `enrollments` / `user_progress` — a tx can batch multiple `enroll` / `complete_lesson` instructions.

A per-row unique on `tx_signature` for those tables would reject legitimate rows and break batched/finalize flows, so it is omitted by design. (This narrows the issue's original "unique on `tx_signature`" to the one table where it is actually safe.)

### 3. FK indexes for cascade / lookup paths
- `idx_thread_views_thread_id` — the PK `(user_id, thread_id)` only serves `user_id`-leading lookups, so deleting a thread (cascade on `thread_id`) would seq-scan.
- `idx_flags_thread_id`, `idx_flags_answer_id` (partial), `idx_flags_reporter_id`, `idx_flags_resolved_by` (partial) — the existing flag unique indexes lead with `reporter_id`, so they don't serve standalone `thread_id` / `answer_id` cascades, and `resolved_by` had no index at all.

## Migration safety
- `ALTER … SET NOT NULL` and `CREATE UNIQUE INDEX` **fail and roll back** if existing data violates them — the safe failure mode (abort loudly, don't corrupt). Postgres could not be run in this environment (no docker/psql), so the constraints were verified **structurally**, not against live data.
- ⚠️ **Depends on existing data being clean.** On apply, this migration will error if:
  - any of the 9 tables already holds a row with `user_id IS NULL`, or
  - `certificates` already holds two rows with the same non-null `tx_signature`.

  Offender-finding queries are included as comments in the migration header. Both are expected to be empty given the writers above, but should be confirmed on the target DB before/at apply.

## Verification performed (no Postgres available)
- Valid SQL shape; parens balanced (migration 26/26, `schema.sql` 505/505).
- Exactly 9 `ALTER … SET NOT NULL` + 6 `CREATE [UNIQUE] INDEX IF NOT EXISTS` (all idempotent).
- New migration timestamp `20260624193000` sorts strictly after the latest existing (`20260624190000`).
- Every referenced table/column confirmed to exist in `schema.sql`.
- `schema.sql` diff is exactly the 9 `NOT NULL` + 6 index additions, nothing else.

## Requires a Supabase apply on merge
This adds a migration — run the Supabase migration apply against the target database after merge. (Pre-commit `lint-staged` only covers JS/TS/JSON/CSS/MD and was skipped: SQL-only change, deps not installed in the loop worktree.)

### DIFF
--- a/supabase/migrations/20260624193000_data_integrity_constraints.sql
+++ b/supabase/migrations/20260624193000_data_integrity_constraints.sql
@@ -1,0 +1,99 @@
+-- ============================================
+-- Data-integrity constraints (P1-9)
+-- ============================================
+-- Three classes of hardening, all idempotent so re-running this migration (or
+-- applying it on top of a partially-migrated DB) is a no-op:
+--
+--   1. NOT NULL on ownership user_id FK columns that were nullable. A NULL
+--      user_id makes a row un-ownable (RLS "auth.uid() = user_id" can never
+--      match it, and it survives the ON DELETE CASCADE from profiles), so such
+--      a row is orphan-able. Every legitimate writer (Helius webhook handlers,
+--      retry queue, admin resync, SECURITY DEFINER RPCs) resolves a concrete
+--      user_id and returns early when it is null, so these columns are always
+--      set in practice.
+--
+--   2. A partial UNIQUE index on certificates.tx_signature so the same mint
+--      transaction can never be recorded as two certificate rows. Partial
+--      (WHERE tx_signature IS NOT NULL) because off-chain/legacy and resync rows
+--      legitimately have no signature (multiple NULLs must stay allowed).
+--
+--   3. FK indexes on thread_views.thread_id and on flags.thread_id / answer_id /
+--      resolved_by so their ON DELETE CASCADE / SET NULL paths and reverse
+--      lookups use an index instead of a sequential scan.
+--
+-- ── Migration safety ──────────────────────────────────────────────────────
+-- ALTER ... SET NOT NULL and CREATE UNIQUE INDEX both FAIL (and roll back) if
+-- existing data violates them. That is the safe failure mode: it aborts loudly
+-- rather than corrupting. If an apply fails here it means a row already holds a
+-- NULL user_id or a duplicate non-null tx_signature that must be cleaned first;
+-- find offenders with, e.g.:
+--   SELECT id FROM <table> WHERE user_id IS NULL;
+--   SELECT tx_signature FROM certificates WHERE tx_signature IS NOT NULL
+--     GROUP BY tx_signature HAVING count(*) > 1;
+--
+-- Why tx_signature UNIQUE is added ONLY to certificates and NOT to enrollments,
+-- user_progress, user_achievements, or xp_transactions:
+--   The Helius webhook decodes EVERY Anchor event from a transaction and
+--   dispatches each with that transaction's single signature (see
+--   lib/helius/event-decoder.ts + the per-event loop in the webhook route).
+--   So one transaction can legitimately write the SAME tx_signature to several
+--   rows of those tables:
+--     * xp_transactions  — a CourseFinalized tx awards both a learner bonus and
+--       a creator reward (two award_xp calls, same signature). xp_transactions
+--       is instead de-duplicated by the existing unique (user_id,
+--       idempotency_key) index.
+--     * user_achievements — a tx can award more than one achievement.
+--     * enrollments / user_progress — a tx can batch multiple enroll /
+--       complete_lesson instructions.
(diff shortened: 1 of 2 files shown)


thomgabriel:
## Handing off for human merge — SENSITIVE (db / schema.sql)

Both gates passed (CI green, `claude[bot]` no blocking), then an **independent adversarial DB review** stress-tested the constraints and the deliberate scoping decision.

### Adversarial verdict: safe ✅
- **`certificates`-only `tx_signature` unique is CORRECT** — verified the Helius webhook legitimately writes one tx's signature to multiple rows of `xp_transactions` (learner bonus + creator reward on `CourseFinalized`, deduped by `(user_id, idempotency_key)`) / `user_achievements` / `enrollments`. A blanket unique would have broken finalize/batch flows. `certificates` is the only table with a true one-mint=one-sig invariant.
- **No dup-tx hole missed** — `xp_transactions` replay is prevented by the existing `(user_id, idempotency_key)` `ON CONFLICT DO NOTHING`; certificates upsert on `(user_id, course_id)` so no second row.
- **`NOT NULL` safe on all 9 tables** — every writer (JS handlers + SECURITY DEFINER RPCs) resolves a concrete `user_id` and early-returns on null. Migration is idempotent and sorts last.

### Non-blocking finding — fixed
The FK-index sweep had missed `threads.accepted_answer_id` (FK → `answers.id ON DELETE SET NULL`). Added `idx_threads_accepted_answer_id` (partial, commit `6920276`) to complete the sweep.

### ⚠️ MUST-DO before apply (operational)
Run the offender queries (embedded in the migration header) against the target DB first: confirm **no NULL `user_id`** in the 9 tables and **no duplicate non-null `certificates.tx_signature`**. The migration fails loud (rolls back) on dirty data — apply to Supabase on merge.

PR #192, opened 2026-06-25
fix(security): harden build-server isolation — gate /metrics + container hardening (P1-8)

### LINKED ISSUE #124: [P1-8] Harden build-server isolation

Confirm Cloud Run egress disabled + dropped caps + read-only FS; gate `/metrics` behind API key/IAM; treat `BLOCKED_PATTERNS` as defense-in-depth.

**Done when:** untrusted compilation can't reach network, `/metrics` not public. **Refs:** `apps/build-server/src/{build.rs:25,225,middlewares/auth.rs}`, `Dockerfile:26`.

thomgabriel:
Closes #124

## Summary
Hardens the Anchor build server (`apps/build-server`, Rust/Axum), which compiles **untrusted user-submitted Rust**. The SBF target is the primary sandbox; this PR adds layered defense so a compromise during compilation can't read operational telemetry, escalate to root, or reach the network.

**Done-when status (verified):** `/metrics` is no longer public; egress lockdown is wired into the deploy config + documented.

### 1. Gate `/metrics` behind the API key (the core fix)
`src/middlewares/auth.rs` previously skipped auth for **both** `/health` and `/metrics`, so anyone with the URL could scrape build counts, durations, and cache stats. Now only `/health` is exempt (needed for liveness probes, exposes nothing sensitive); `/metrics` requires `X-API-Key` like `/build` and `/deploy`.

- Updated `tests/integration_test.rs`: `test_metrics_requires_auth` (expects **401** without key) + `test_metrics_endpoint_with_auth` (expects 200 with key).
- Fixed stale `docs/ARCHITECTURE.md` endpoint table (`/metrics` auth: None → X-API-Key).

### 2. Dockerfile container hardening
- Non-root user pinned to a **fixed numeric UID `10001`** (so Cloud Run / k8s `runAsNonRoot` can enforce it) with a `nologin` shell.
- Server binary now copied with explicit `--chown` (was landing as root).
- **All setuid/setgid bits stripped image-wide** (`find / -xdev -perm /6000 -exec chmod a-s`) so a read-only-rootfs + no-new-privileges sandbox can't be escalated via a stray privileged binary.
- Documented that `/tmp` (`BUILDS_DIR`) is the only runtime-writable path → enables a read-only root FS with just `/tmp` mounted writable.

### 3. Egress lockdown wired into deploy config
`deploy/deploy.sh` + `deploy/cloudbuild.yaml` now apply **Direct VPC egress** (`--network`/`--subnet`/`--network-tags`/`--vpc-egress=all-traffic`) when VPC vars are provided, routing all outbound traffic through a VPC where a deny-egress firewall drops it. Gated on the VPC vars being set so the pipeline still deploys before the VPC exists (with a loud WARNING otherwise). `cargo-build-sbf` already runs `--offline`, so a legitimate build needs **zero egress**.

New `deploy/HARDENING.md` documents the threat model, every control, and the exact one-time VPC + firewall commands.

## What REQUIRES a Cloud Run deploy setting (not enforceable in the image alone)

- **Egress disable (the real "no network" control):** there is **no single flag**. Requires the VPC + a deny-egress firewall rule (exact `gcloud compute` commands in `HARDENING.md`), then deploy with the flags above. **This must be applied at deploy time to fully close the Done-when** — the in-repo `--offline` + `BLOCKED_PATTERNS` are defense-in-depth, not the primary control.
- **Drop capabilities / read-only root FS:** Cloud Run (fully managed) does **not** expose `--cap-drop` / `securityContext` as `gcloud run` flags — it enforces a fixed minimal-privilege contract (no privileged mode, no host mounts, `no-new-privileges`). The image-level hardening (non-root, stripped setuid, `/tmp`-only writes) is the in-repo expression. GKE/Knative equivalents are documented for a future migration.
- **IAM-gated ingress (optional, stronger):** the service stays on `--no-invoker-iam-check` because the caller (Vercel API route) authenticates via `X-API-Key`, **not** GCP IAM. Switching to `--no-allow-unauthenticated` would 403 the legitimate Vercel→Cloud Run call unless the caller mints a Google ID token — documented as a follow-up, intentionally not flipped here to avoid breaking prod.

## Verification (actual results)
- `cargo build` → **success** (`apps/build-server`).
- `cargo clippy --all-targets` → **clean, zero warnings**.
- `cargo fmt --check` → **clean**.
- Ran the server locally (`ACADEMY_API_KEY=… PORT=8099`) and probed:
  - `GET /health` (no auth) → **200**
  - `GET /metrics` (no key) → **401** `{"error":"Invalid or missing API key"}`  ← was 200
  - `GET /metrics` (wrong key) → **401**
  - `GET /metrics` (correct key) → **200** + Prometheus body
- `deploy.sh` `bash -n` + `cloudbuild.yaml` YAML parse → both valid.

## Honest scope: code vs deploy-config
- **Enforced by this PR (code):** `/metrics` gating, Dockerfile non-root/UID/setuid-strip, writable-`/tmp` contract.
- **Wired but requires deploy action:** VPC egress lockdown (flags in repo; VPC + firewall must be provisioned at deploy time).
- **Cannot be enforced in-repo:** the actual network-egress block depends on the Cloud Run VPC + firewall; cap-drop/read-only-FS are platform-contract on managed Cloud Run, not repo flags.


thomgabriel:
## Handing off for human merge — SENSITIVE (area:security)

Both gates passed, then an **independent adversarial review** stress-tested both claims.

### ✅ Claim B (`/metrics` not public) — fully code-enforced
`/metrics` now behind `X-API-Key` (constant-time compare, fail-closed on missing/empty); all routes under the auth layer; `/build`/`/deploy` still gated; the Vercel caller is unaffected. Verified (15/15 tests incl. 401-without-key).

### ⚠️ Claim A (untrusted compilation can't reach network) — DEPLOY-GATED, not code-enforced
The egress block depends on a Cloud Run **VPC + deny-egress firewall** (manual `gcloud`), not code. A deploy WITHOUT the VPC vars has **zero egress protection** (only a console warning). So #124's network Done-when is **not closed by merging this PR** — tracked as **#193** (provision VPC + firewall + verify egress is actually blocked). `BLOCKED_PATTERNS` is correctly defense-in-depth only. Don't read the green checks as "egress blocked."

### Review fixes applied (commit `5a88fce`)
- **Fail-closed egress tag** — deploy now ERRORS if the VPC path is taken with an empty `--network-tags` (was a silent un-hardened deploy — the worst footgun).
- **Setuid strip verified** — Dockerfile fails the build if any setuid/setgid binary survives the strip (was silent).
- **Public `/health` trimmed** — removed `cache_entries`/`active_builds`/`total_builds` (build telemetry now only on the gated `/metrics`).
- Container hardening (non-root UID, stripped setuid) sound. cargo build/clippy/fmt + 15/15 tests clean.

**Recommendation:** merge the code (correct + hardened), but keep #124's egress invariant open against **#193** until the VPC + firewall are provisioned and egress is verified blocked.

PR #194, opened 2026-06-25
feat(credentials): pin credential metadata to Arweave at mint (P1-11)

### LINKED ISSUE #127: [P1-11] Pin credential metadata to permanent storage (Arweave)

Credential metadata is served from `/api/certificates/metadata` (`certificates/mint/route.ts:175`). Pin to Arweave at mint, point Metaplex Core asset URI at the immutable URL.

**Done when:** issued mainnet credentials resolve independent of app infra. *(Feeds PLAN K-4; coordinate asset-URI shape with @thomgabriel.)*

thomgabriel:
Closes #127

## What

At credential mint, certificate metadata is now pinned to **Arweave** (via Irys) and the on-chain Metaplex Core asset URI points at the **immutable** Arweave gateway URL instead of `/api/certificates/metadata`. This decouples issued credentials from app uptime — the goal of #127.

## Upload flow

`apps/web/src/app/api/certificates/mint/route.ts`:
1. Build the metadata JSON (unchanged).
2. Insert it into the Supabase `nft_metadata` row (**kept** as the app-served fallback).
3. `uploadCertificateMetadata(metadataJson)` → new module `apps/web/src/lib/solana/arweave.ts`:
   - Builds a UMI instance (`createUmi` + `keypairIdentity` + `irysUploader`) — UMI is already in the dep tree.
   - `umi.uploader.uploadJson(...)` settles the JSON on Arweave and returns the permanent gateway URL.
   - Picks the Irys node by `NEXT_PUBLIC_SOLANA_NETWORK` (`mainnet-beta` → `https://uploader.irys.xyz`, else devnet `https://devnet.irys.xyz`).
4. The on-chain `issueCredential(...)` call and the `certificates.metadata_uri` record both use the Arweave URL when available, else the app URL.

New dependency: `@metaplex-foundation/umi-uploader-irys@^1.5.0` (resolves to the Node plugin server-side).

## New env var (needs funding — human prerequisite)

`ARWEAVE_UPLOADER_SECRET` (server-only) — documented in `.env.example` + `CLAUDE.md`.

- **Value:** 64-element JSON array of a **Solana** keypair's secret-key bytes (same shape as `BACKEND_SIGNER_SECRET`). Irys is funded with this wallet's **SOL** (Irys's Solana currency) — **not** a raw Arweave JWK. I chose the Solana-keypair approach (over the issue's `ARWEAVE_KEY_JWK` suggestion) because the whole stack is Solana + UMI and Irys funds natively from SOL; flagging the naming/approach for your call below.
- The **code reads + uses** the key. The **human provides + funds** it:
  - devnet → free, but Irys devnet uploads are **pruned after ~60 days** (not truly permanent — fine for testing).
  - mainnet → permanent, paid in SOL from this wallet's Irys balance (top up at irys.xyz before mainnet minting).

## Fallback behavior (mint never hard-crashes)

If `ARWEAVE_UPLOADER_SECRET` is **absent** or the upload **fails** (unfunded wallet, Irys outage), `uploadCertificateMetadata` returns `null` and mint falls back to the existing `/api/certificates/metadata?id=…` URL, logging a clear server-side warning. So mint keeps working before Arweave is provisioned. `/api/certificates/metadata` is unchanged and still serves already-issued credentials.

## ⚠️ Needs maintainer sign-off (@thomgabriel)

The issue says to coordinate the asset-URI shape with @thomgabriel. I made a documented default:

- **Asset-URI shape:** whatever `umi.uploader.uploadJson` returns — an Irys/Arweave gateway URL (`https://arweave.net/<txid>` form). **Please confirm this is the final on-chain URI shape before mainnet** (e.g. if you'd prefer `ar://<txid>` or a specific gateway host, that's a one-line change).
- **Env var name/approach:** `ARWEAVE_UPLOADER_SECRET` as a Solana keypair vs the issue's `ARWEAVE_KEY_JWK` — confirm the funding model you want.

## Verification

- `pnpm --filter @superteam-lms/web typecheck` ✅
- `pnpm --filter @superteam-lms/web build` ✅ (`/api/certificates/mint` compiles with the new import)
- eslint clean on both changed files.

**Untested:** the **live Arweave upload** — it requires a funded `ARWEAVE_UPLOADER_SECRET`, which is a human prerequisite. Verified: it compiles, the Node Irys plugin resolves at runtime, and the **fallback path** (no key → app URL + warning) is exercised by the existing build. The happy-path upload itself has not run against a live Irys node.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,8 +65,25 @@ BUILD_SERVER_URL=https://academy-build-server-HASH.a.run.app
 # API key for X-API-Key header (same value as ACADEMY_API_KEY on Cloud Run)
 BUILD_SERVER_API_KEY=                      # PRIVATE — server-only
 
-# ── App URL (for sitemap, OG tags, NFT metadata URI) ─────────────────────────
+# ── App URL (for sitemap, OG tags, NFT metadata URI fallback) ────────────────
 NEXT_PUBLIC_APP_URL=http://localhost:3000
+
+# ── Arweave / Irys — permanent credential metadata (server-only) ─────────────
+# At credential mint, certificate metadata is pinned to Arweave via Irys so the
+# on-chain Metaplex Core asset URI resolves independently of app uptime
+# (apps/web/src/lib/solana/arweave.ts). If unset, mint falls back to the
+# app-served /api/certificates/metadata URL (logs a warning) — minting still
+# works, but issued credentials will NOT resolve if the app is offline.
+#
+# Value: 64-element JSON array of a SOLANA keypair's secret-key bytes (same
+# shape as BACKEND_SIGNER_SECRET). Irys is funded with this wallet's SOL — NOT
+# an Arweave JWK. The human must fund it:
+#   - devnet (NEXT_PUBLIC_SOLANA_NETWORK=devnet): uses https://devnet.irys.xyz,
+#     free, but uploads are pruned after ~60 days — NOT truly permanent.
+#   - mainnet (NEXT_PUBLIC_SOLANA_NETWORK=mainnet-beta): uses
+#     https://uploader.irys.xyz, permanent, paid in SOL from this wallet's Irys
+#     balance (top up at irys.xyz before mainnet minting).
+ARWEAVE_UPLOADER_SECRET=
 
 # ── Admin operations (server-only — never expose to browser) ──────────────────
 # JSON array of authority keypair bytes (64 elements), e.g. [12,34,...]
--- a/.env.example
+++ b/.env.example
@@ -71,9 +71,13 @@ NEXT_PUBLIC_APP_URL=http://localhost:3000
 # ── Arweave / Irys — permanent credential metadata (server-only) ─────────────
 # At credential mint, certificate metadata is pinned to Arweave via Irys so the
 # on-chain Metaplex Core asset URI resolves independently of app uptime
-# (apps/web/src/lib/solana/arweave.ts). If unset, mint falls back to the
-# app-served /api/certificates/metadata URL (logs a warning) — minting still
-# works, but issued credentials will NOT resolve if the app is offline.
+# (apps/web/src/lib/solana/arweave.ts). The uploader returns an Irys GATEWAY URL
+# (https://gateway.irys.xyz/<id>) — NOT https://arweave.net/<txid> — and that is
+# the exact string written on-chain. On mainnet the data settles on Arweave, so
+# the same <id> is also reachable via https://arweave.net/<id> and ar://<id>.
+# If unset, mint falls back to the app-served /api/certificates/metadata URL
+# (logs a warning) — minting still works, but issued credentials will NOT
+# resolve if the app is offline.
 #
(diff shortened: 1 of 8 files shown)


thomgabriel:
Fixed in `bfa55fe`: **(1)** `getUploaderKeypair()` now wraps the parse/validate in try/catch and returns `null` on a malformed `ARWEAVE_UPLOADER_SECRET` — same graceful app-URL fallback as the no-key path, so mint no longer 500s or orphans the `nft_metadata` row on a misconfigured key. **(2)** the **webhook credential path** (`event-handlers.ts` `tryIssueCredential`, the primary/automatic path) now also pins to Arweave (was only the manual mint route). Added `arweave.test.ts` (3/3: absent / non-JSON / wrong-length secret all fall back without throwing); typecheck + build clean.

@claude re-review

thomgabriel:
## Handing off for human merge — SENSITIVE (area:onchain)

Full discipline: `claude[bot]` reviewed (2 blocking → fixed), then an **independent adversarial review** verified integrity and caught a doc-accuracy issue both gate reviews missed.

### Adversarial verdict: integrity solid ✅
- **On-chain ↔ DB ↔ served metadata are consistent** — both mint paths (manual `mint/route.ts` + automatic webhook `event-handlers.ts`) compute the URI ONCE (`arweaveUri ?? appMetadataUri`), fully await the upload before the CPI, and write the same string on-chain + to `certificates.metadata_uri`. No race, no path that pins one URL but writes another.
- **Fallback is honest + loud** (no silent "claimed Arweave but used app URL"); **secret is strictly server-only** (no client/bundle/edge exposure, errors log only messages); the prior orphan-row bug is genuinely fixed; devnet-pruning caveat surfaced.

### Findings fixed (commit `41010d7`)
- **Doc accuracy** — the on-chain URI is actually `https://gateway.irys.xyz/<id>`, NOT `arweave.net/<txid>` as the PR/docs claimed. Corrected `arweave.ts`/`.env.example`/`CLAUDE.md` to describe the real string.
- **URL guard** — a malformed (non-`https://`) upload return now falls back instead of being pinned irreversibly on-chain (+2 tests).

### ⚠️ YOUR DECISION (per the issue's "coordinate URI shape with @thomgabriel") — irreversible per credential
The on-chain asset-URI host. Pick before mainnet (one-line in `arweave.ts`):
- **`https://gateway.irys.xyz/<id>`** (current) — resolves on devnet AND mainnet, but depends on the Irys gateway.
- **`https://arweave.net/<id>`** — canonical Arweave gateway, more infra-independent — but **mainnet-only** (devnet Irys uploads don't settle to Arweave, so this won't resolve on devnet).
- **`ar://<id>`** — protocol/wallet-resolved, gateway-agnostic.

### ⚠️ Human prerequisites (Done-when only holds on the Arweave path)
Fund `ARWEAVE_UPLOADER_SECRET` (a Solana keypair; Irys funds from SOL). Until funded, credentials pin to the mutable app URL.

Non-blocking (pre-existing): `image: ""` is permanent once issued; `NEXT_PUBLIC_SOLANA_NETWORK` read via `process.env` (style).

PR #195, opened 2026-06-25
feat(security): server-side challenge validation gates completion (#181)

### LINKED ISSUE #181: [P0-C4 follow-up] Wire server-side challenge validation into lesson completion
## Context
PR #179 closed the #150 interim (answer key no longer ships to the browser; XP bounded by `award_xp` caps) and added a server-only `POST /api/lessons/validate-challenge`. **But nothing calls it** — `serverValidated` is hard-coded `false` and the route is dead scaffolding today.

## The gap
`/api/lessons/complete` accepts a completion for ANY lesson (including an unattempted challenge) from any authenticated+enrolled user — body is only `{lessonId, courseId}`, no proof of solving. It then submits the on-chain `completeLesson` tx; the Helius webhook writes the completion record + awards (capped) XP, and a full set of completions makes the learner eligible for `finalize_course` → credential NFT.

So a forged "passed" cannot mint XP beyond the caps (good), but **lesson-completion and credential integrity are not enforced** — you can earn a credential without solving the challenges.

## Done when
- Challenge-type lessons require a PASSING server-side validation before `/api/lessons/complete` will submit `completeLesson`.
- `validate-challenge` is wired into the completion flow and `serverValidated` reflects a real check (run the submission against the server-held hidden tests).
- A forged browser "passed" cannot produce a completion record for a challenge lesson.

## Notes
- The hard part is a **sandboxed executor** for untrusted user code (explicitly deferred from #150). Design that first (resource/time limits, no network/FS).
- Refs: `apps/web/src/app/api/lessons/validate-challenge/route.ts`, `apps/web/src/app/api/lessons/complete/route.ts`, `apps/web/src/components/.../lesson-client.tsx`.
- Source: #179 review (claude[bot] finding #1 + independent adversarial review).

thomgabriel:
Closes #181

## The gap

PR #179 added `POST /api/lessons/validate-challenge` but nothing called it (`serverValidated` was hard-coded `false`), and `/api/lessons/complete` accepted a completion for any lesson — including an unattempted challenge — with **no proof of solving**. A forged browser "passed" earned a real completion record (and credential eligibility). The hidden tests + reference solution are server-held (`getChallengeAnswerKey`), but were never used to actually grade anything.

## What this does

Challenge-type lessons now require a **passing server-side validation** before `/api/lessons/complete` submits the on-chain `completeLesson`. The browser runner stays for UX (visible tests), but **the server is authoritative**.

### 1. Secure executor for untrusted JS — `src/lib/challenge/executor.ts`

Untrusted learner JS runs inside a **real V8 isolate** via [`isolated-vm`](https://github.com/laverdet/isolated-vm) — **not** Node `vm` / `eval` / `new Function` (those share the host realm and are not isolation boundaries). Guarantees, all enforced by V8 itself, not by string filtering:

- **Separate heap, hard memory limit** (64 MB) — allocation past the cap terminates the script; host memory can't be exhausted.
- **In-isolate wall-clock timeout** (5 s) — kills CPU-bound runaways (`while(true){}`).
- **Host-side wall-clock guard** (6 s) + `isolate.dispose()` — defense-in-depth for the one case the CPU timeout can't catch: a *quiescent* never-settling promise (`await new Promise(()=>{})`). There are no timers/event-loop primitives in the isolate (`setTimeout` is `undefined`), so such a promise never settles; the host guard abandons the eval and tears down the isolate.
- **No ambient host objects** — `process`, `require`, `module`, Node `globalThis`, filesystem, and network simply don't exist in the isolate. Verified by tests (`typeof process === 'undefined'`, `typeof require === 'undefined'` inside the sandbox). We never bridge a host function/object in (no `Reference`).
- **Only two values cross the boundary**: a code **string** goes in; a JSON-serialisable **string** result is copied out. The mock Solana SDK + mock console live entirely inside the isolate, assembled from source text — a faithful server port of the existing browser-worker contract in `components/editor/challenge-runner.tsx`, so a submission is judged the same way it is in-browser, except the **server also runs the hidden tests**.
- **Degrade-closed**: `isolated-vm` is a native addon; if it can't load, the executor reports `available: false` and the completion gate **denies** the completion (503). A forged pass can never slip through an unavailable executor.

### 2. Shared validation — `src/lib/challenge/validate.ts`

Single source of truth used by **both** routes (no duplicated logic). Loads the answer key server-side, runs the submission through the executor, returns a typed verdict: `validated{passed}` / `non_js_challenge` / `not_a_challenge` / `executor_unavailable`.

### 3. Wiring

- **`validate-challenge` route** now runs a REAL check; `serverValidated` reflects actual execution against all tests. Answer key / hidden tests / solution are never serialised into the response (P0-C4 preserved).
- **`complete` route** — for challenge lessons: requires `submittedCode`, runs `validateAgainstAnswerKey`, and **rejects with 403 if not all tests pass — before** the on-chain `completeLesson`. `executor_unavailable` → 503 (degrade closed). Non-challenge lessons are unaffected (no answer key → gate skipped).
- **Client** threads the editor's code through the `superteam:lesson-complete` event → `completeLessonAPI` → request body.
- `getChallengeAnswerKeyById` added (the complete route works in Sanity `_id`s); answer-key projection extended with `buildType`.
- `isolated-vm` added to `serverComponentsExternalPackages` so webpack keeps the native addon external (build output confirms it's emitted as `import("isolated-vm")`, not bundled).

## Test — `src/lib/challenge/__tests__/executor.test.ts` (13 tests, run the REAL isolate)

Proves the boundary:
- Correct submission **passes** all tests (visible + hidden actually executed).
- **WRONG and EMPTY** submissions are **rejected**.
- **Forged-pass is blocked**: code that hard-codes the visible answer (`if (a===2 && b===3) return 5`) passes the visible test but **fails the hidden test → overall not passed**. This is exactly the "all green in the browser" shape; the server runs the hidden test and rejects. `validateAgainstAnswerKey` returns `passed:false`, so the complete route would 403.
- Runaway `while(true)` killed by the in-isolate timeout; never-settling promise abandoned by the host guard (neither hangs).
- Host escapes denied: `process` / `require` unreachable inside the isolate.

All 41 web tests pass (`pnpm --filter @superteam-lms/web test`).

## Verify

- `pnpm --filter @superteam-lms/web typecheck`: **my code is clean.** The only errors are 55 pre-existing `sanity/schemas/*` errors that are also present on a clean `main` (unrelated to this PR).
- `pnpm --filter @superteam-lms/web build`: **`✓ Compiled successfully`**, and the build output confirms `isolated-vm` stays external and the executor logic is bundled into both lesson routes. The build then fails at **page-data collection** because this sandbox has no real `NEXT_PUBLIC_*` / on-chain secret env vars (`Invalid public environment variables`, then `/api/admin/*` keypair envs) — a pre-existing env gate identical on clean `main`, not introduced here.

## ⚠️ Honest limitation — read before merging

**Is the sandbox genuinely escape-resistant?** The *isolation model* is real: `isolated-vm` is a true V8 isolate, and the tests demonstrate no host-object access, enforced memory/CPU/wall-clock limits, and the forged-pass being blocked. That is a real security boundary, not a `vm`/`Function` shim.

**But there is a deployment caveat that must not be glossed over:**

1. **Native-module / Node-version coupling.** `isolated-vm` is a native addon and is tightly bound to the Node/V8 ABI. In this environment (**Node 24**), `isolated-vm@5` (which declares Node 18+) **fails to compile**, and `isolated-vm@7` requires **Node ≥26**. I pinned **`isolated-vm@6.1.2`** (Node ≥22), which builds and runs cleanly here — but the repo's `package.json` engines say `node >=18`. **Whoever deploys must ensure the runtime Node version matches a buildable `isolated-vm` and that the prebuilt/compiled binary is present.** On a platform where it can't load, the executor degrades closed (completion is denied) — safe, but challenge completion is then **blocked** rather than insecure.

2. **Vercel serverless** (the documented frontend target) **generally cannot run native addons like `isolated-vm`** in the standard Node functions runtime. On that platform this executor will report `available:false` and **deny all JS-challenge completions** until it's relocated.

**Recommended production architecture:** move untrusted execution into a **dedicated sandboxed microservice** — mirroring the existing **Rust build-server** (Docker-per-execution on Cloud Run) that this repo already uses for the other untrusted-execution path. The code is structured for exactly that swap: `runJsSubmission` is a clean async interface behind `validateAgainstAnswerKey`; pointing it at an HTTP sandbox service (like `lib/rust/execute.ts` → `/api/rust/execute`) requires no route-logic changes. **I would not call the in-process executor production-secure on serverless as-is** — the isolation is sound, but the deployment surface needs the microservice.

3. **Scope:** only **JS/TS** challenges are gated by this executor. **Rust** and **buildable** challenges (`non_js_challenge`) are validated by the Rust playground / build server (a separate trust boundary) and are explicitly out of scope here — the gate requires a submission for them but defers correctness to those services.

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -114,6 +114,12 @@ const nextConfig = {
   // (validates env vars). Stable in Next 15; opt-in in Next 14.
   experimental: {
     instrumentationHook: true,
+    // `isolated-vm` is a native addon used by the server-side challenge
+    // executor (lib/challenge/executor.ts). Keep it external so webpack does
+    // not attempt to bundle/trace the .node binary; it is required at runtime
+    // on the Node.js server. The module loads lazily and degrades closed if the
+    // prebuilt binary is unavailable in the host environment.
+    serverComponentsExternalPackages: ["isolated-vm"],
   },
   transpilePackages: ["@superteam-lms/types", "@superteam-lms/sanity"],
   images: {
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -48,6 +48,7 @@
     "canvas-confetti": "^1.9.4",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.0",
+    "isolated-vm": "6.1.2",
     "isomorphic-dompurify": "2.36.0",
     "next": "^14.2.0",
     "next-intl": "^3.22.0",
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -48,7 +48,6 @@
     "canvas-confetti": "^1.9.4",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.0",
-    "isolated-vm": "6.1.2",
     "isomorphic-dompurify": "2.36.0",
     "next": "^14.2.0",
     "next-intl": "^3.22.0",
@@ -69,6 +68,9 @@
     "tailwindcss-animate": "^1.0.7",
     "tweetnacl": "^1.0.3",
     "zod": "^4.4.3"
+  },
+  "optionalDependencies": {
+    "isolated-vm": "6.1.2"
   },
   "devDependencies": {
     "@tailwindcss/typography": "^0.5.19",
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
@@ -163,12 +163,13 @@ interface CompletionResponse {
 
 async function completeLessonAPI(
   lessonId: string,
-  courseId: string
+  courseId: string,
+  submittedCode?: string
 ): Promise<CompletionResponse> {
   const res = await fetch("/api/lessons/complete", {
     method: "POST",
     headers: { "Content-Type": "application/json" },
-    body: JSON.stringify({ lessonId, courseId }),
+    body: JSON.stringify({ lessonId, courseId, submittedCode }),
   });
   if (!res.ok) {
     throw new Error("Failed to complete lesson");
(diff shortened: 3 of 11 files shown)


thomgabriel:
## Handing off — SENSITIVE (area:security) + an ARCHITECTURE DECISION for you

Full discipline: `claude[bot]` reviewed, then an **independent adversarial review** found a **P0 the gate under-rated**, and it's now fixed + regression-tested.

### Adversarial review → both holes fixed (commit `849904b`)
- **[P0 — validator bypass, FIXED]** the executor reused ONE isolate context for all tests; user code in test #1 could reassign `globalThis.Function` (the gate saw this but rated it "low") to force-pass all hidden tests → complete any JS challenge without solving. Fix: **fresh isolate context per test** + the harness now captures pristine `Function`/`JSON.stringify`/`JSON.parse` before user code runs (a *sibling* readout-poison vector the fix surfaced). Proven: exploit went `passed:true` → `passed:false`; 4 taint vectors (`Function`/`Object.prototype`/`Array.prototype`/`JSON`) regression-tested.
- **[High — non-JS forge, FIXED]** the `non_js_challenge` (Rust) branch fell through to on-chain completion with no validation. Now **fail-closed (503)**. Real Rust handshake tracked in **#197**.
- **Verified sound:** real isolation (no host escape), CPU timeout + wall-clock kill, **degrade-closed airtight**, gate is server-authoritative (client can't mislabel a challenge), P0-C4 answer-key secrecy intact. 47 tests pass.

### ⚠️ THE DECISION: secure, but blocked in production until a microservice exists
`isolated-vm` is a native addon that **Vercel serverless can't run** (and CI Node 20 can't build) → it's an `optionalDependency` and the executor **degrades closed**. So on the deployed app, **JS challenge completion is BLOCKED (503), not forgeable** — and non-JS is fail-closed too. Net: merging this makes challenge completion *secure* but *non-functional in production* until the **execution microservice (#196)** is built (Docker-per-exec on Cloud Run, like the build-server; #195's logic is structured for that swap).

**Your call:** merge now (challenges secure-but-blocked in prod, unblocked when #196 lands) **or** hold #195 until #196 is ready. The forgeable-completion hole is closed either way once this merges (everything is validated-or-blocked, never forged).

PR #201, opened 2026-06-25
fix(ci): pin indexmap to 2.11.4 for Solana rustc 1.79 (anchor build)

Potolski:
The Anchor program Docker build runs Solana's bundled rustc 1.79, but indexmap bumped its MSRV to 1.82 at 2.12.0 (transitively pulled via toml_edit -> proc-macro-crate -> borsh-derive), breaking the build:

  indexmap@2.13.0 requires rustc 1.82

Pin to 2.11.4 — the latest indexmap with MSRV 1.63, well within 1.79 — via cargo update --precise. Lockfile-only change (downgrades its hashbrown dep to 0.15.2); no source or version-range changes.

### DIFF
--- a/onchain-academy/Cargo.lock
+++ b/onchain-academy/Cargo.lock
@@ -951,12 +951,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "hashbrown"
-version = "0.16.1"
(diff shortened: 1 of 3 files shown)


Potolski:
@claude the issue we are trying to fix is CI/verifiable build breaking.

Issue #197, opened 2026-06-25
[#181 follow-up] Server-side validation for non-JS (Rust/buildable) challenges


thomgabriel:
## Context
#195 made the `non_js_challenge` (Rust/buildable) completion path **fail-closed** (returns 503) — previously it silently fell through to on-chain `completeLesson` with NO validation (a forge hole found by #195's adversarial review, HOLE 2). Fail-closed is safe but means Rust challenges can't currently be completed.

## Done when
Non-JS challenge completion is server-validated via a handshake with the Rust playground / build-server (e.g. run the submission against the tests in the build-server, or a signed-completion token) — both FUNCTIONAL and not forgeable.

## Refs
`apps/web/src/app/api/lessons/complete/route.ts` (the `non_js_challenge` 503 + TODO). Source: #195 adversarial review (HOLE 2).

thomgabriel:
**Blocked** on #195 merging — #197 builds on the `non_js_challenge` fail-closed path #195 added to `complete/route.ts`. Drafting against `main` now would conflict. Unblock once #195 is merged.

thomgabriel:
The Rust-Playground approach attempted in #216 is **forgeable** (proven live): untrusted Rust reads its own source to recover the grading nonce and prints passing markers for wrong answers, and can `process::exit(0)` to suppress the harness. There is **no** non-forgeable grading within the single-file Playground model.

Non-JS (Rust function) validation therefore needs the same **isolated sandboxed-execution microservice** as #196 — a separate trust boundary where the grading driver + verdict are out-of-band from the untrusted submission (Docker-per-execution on Cloud Run, mirroring `apps/build-server`). Marking `blocked:needs-human` on that infra. Until it exists, rust challenges stay fail-closed (503), same as buildable/Anchor. Full repro in closed #216.

thomgabriel:
Verified against current `main` — server-side validation for non-JS challenges **is implemented and non-forgeable**:

[...]

Issue #196, opened 2026-06-25
[#181 follow-up] Build sandboxed challenge-execution microservice (run server-side validation in production)


thomgabriel:
## Context
#195 added server-side challenge validation via in-process `isolated-vm`. But `isolated-vm` is a native addon that **Vercel serverless cannot run** (and the repo CI Node 20 can't build it either). So in production the executor **degrades CLOSED** — ALL JS challenge completions are BLOCKED (HTTP 503), safe but non-functional.

## Done when
Challenge submissions are validated by a dedicated **sandboxed execution microservice** (Docker-per-execution on Cloud Run, mirroring the existing Rust `apps/build-server`), so JS challenge completion works *and* stays non-forgeable on the deployed app. #195's `executor.ts` logic is structured for this swap with no route-logic changes.

## Also harden
The in-process mock grader is heuristic (arg-setup by parameter name, `expectedOutput` eval'd). The #195 adversarial review noted a determined submission could game the GRADING shape-matching (a grading-fidelity issue, not a sandbox escape — the cross-test-taint/forge class is closed). The microservice should use a robust grader.

## Refs
`apps/web/src/lib/challenge/executor.ts`, `apps/build-server` (reference architecture). Source: #195 adversarial review.

thomgabriel:
**Blocked — needs a human architecture decision.** Building a sandboxed execution microservice is a major design+infra undertaking (new Cloud Run service, Docker-per-exec, grader). Worth scoping deliberately rather than auto-drafting overnight. Unlabel + (optionally) hand to the loop once the approach is decided.

thomgabriel:
Verified against current `main` — the microservice premise here is **obsolete**.

The JS executor was migrated from `isolated-vm` (native addon, couldn't run on Vercel) to **QuickJS-emscripten (single-file WASM)** — `apps/web/src/lib/challenge/executor.ts`, deps `quickjs-emscripten-core` + `@jitl/quickjs-singlefile-cjs-release-sync` in `apps/web/package.json`. Pure WASM (no native addon), so it runs **in-process on Vercel serverless**. The sandbox is escape-resistant and prod-viable:
- a **fresh QuickJS runtime + context per test** (no state, prototype, `Function`, or `JSON` poisoning carries between tests, including hidden ones);
- pass/verdict cross the boundary via **captured `Function`/`JSON` intrinsics** (a submission that reassigns them can't substitute the wrapper or forge the verdict);
[...]

PR #216, opened 2026-06-26
feat(challenge): server-side validation for non-JS (Rust function) challenges (#197)

thomgabriel:
## Summary
Implements non-forgeable, server-side validation for **Rust function** challenges, so they can be completed (previously the `non_js_challenge` path fail-closed with 503 — #195 HOLE 2). Mirrors the JS isolate executor's contract with a new `runRustSubmission`.

## Design
- **New `apps/web/src/lib/challenge/rust-executor.ts`** — `runRustSubmission(code, tests)`, the Rust analog of `runJsSubmission`. Untrusted Rust cannot be sandboxed in-process, so it executes **server-to-server in the Rust Playground sandbox** (`play.rust-lang.org/execute`), never on our server.
- **`validate.ts` routing**: `buildable` (Anchor) → `non_js_challenge` (still fail-closed); `language === "rust"` && !buildable → graded by the Rust executor → `validated`; everything else → JS isolate. Both executable paths degrade closed (`executor_unavailable` → 503), never a silent pass.

## Why this is non-forgeable
1. **Per-run nonce**: the server synthesises a test harness emitting `TEST_<nonce>_<i>_PASS` markers with an unpredictable per-run nonce, and grades by scanning the **real** Playground stdout. A submission cannot pre-`println!` a passing marker because it cannot guess the nonce.
2. **Hidden tests stay server-side**: hidden `input`/`expectedOutput` are injected into the harness here; they never reach the browser.
3. **Degrade-closed**: Playground unreachable / non-2xx / timeout / malformed / no gradable function ⇒ `available: false` ⇒ deny. `passed` is true only if the server observed every test's PASS marker.

## Scope — partially addresses #197
- ✅ **Rust function challenges** — fully implemented + graded non-forgeably.
- ⏸️ **`buildable` (Anchor) challenges** — intentionally still fail-closed. The Cloud Run build server is **compile-only**; grading "it compiled" as "passed" would be forgeable. Deferred until a build-server "run tests against submission" endpoint exists (overlaps #196). The route + verdict comments document this precisely.

## Tests
`apps/web/src/lib/challenge/__tests__/rust-executor.test.ts` (10) + extended `executor.test.ts` — pass, fail, fail-closed-on-unavailable/timeout/non-2xx, hidden-test gating, nonce anti-forgery. Local: typecheck clean, 33 challenge tests green.

Partially addresses #197 (Rust-function path). Buildable/Anchor validation tracked as follow-up.

### DIFF
--- a/apps/web/src/app/api/lessons/complete/route.ts
+++ b/apps/web/src/app/api/lessons/complete/route.ts
@@ -124,15 +124,18 @@ export async function POST(request: NextRequest) {
             { status: 503 }
           );
         case "non_js_challenge":
-          // FAIL CLOSED. Rust / buildable challenges must be graded by the Rust
-          // playground / build server (a separate trust boundary), but that
-          // validation handshake is NOT wired up yet. Until it is, the server
-          // cannot prove the submission is correct, so it MUST NOT submit the
-          // on-chain completeLesson — a bare fall-through here would grant a
-          // completion (and credential eligibility) for any unverified
-          // Rust/buildable submission. Deny exactly like executor_unavailable.
-          // TODO(#195 follow-up): wire a real Rust/build-server validation
-          // handshake, then gate this branch on its passing verdict.
+          // FAIL CLOSED. Reaches here ONLY for `buildType: "buildable"` (Anchor)
+          // challenges: the Cloud Run build server is compile-only and cannot
+          // run a submission against the hidden tests, so there is no
+          // non-forgeable way to prove correctness (grading "it compiled" as
+          // "passed" would be forgeable). A bare fall-through would grant a
+          // completion (and credential eligibility) for any compiling Anchor
+          // submission. Deny exactly like executor_unavailable. Rust *function*
+          // challenges are now graded by the Rust executor (validate.ts) and
+          // return `validated`, so they no longer land here.
+          // TODO(#197 follow-up): a build-server "run tests against submission"
+          // endpoint (or #196-style microservice) would let buildable be graded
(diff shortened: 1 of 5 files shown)


thomgabriel:
Closing — the independent adversarial review found a **CRITICAL, proven-live forge** that this approach cannot fix within the Rust Playground model.

**The forge:** untrusted Rust reads its own compiled source — `include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/src/main.rs"))` (compile-time) or `std::fs::read_to_string` (runtime) — to recover the per-run nonce, then prints the correctly-tagged `TEST_<nonce>_<i>_PASS` markers for every test (visible **and** hidden) while returning wrong answers. Verified live: a wrong `add` returning `999` graded as a full pass.

**Why it's not patchable here:** the Playground compiles everything into one `src/main.rs`, so the untrusted submission shares the artifact (reads any nonce/secret we embed) and the output channel (can `std::process::exit(0)` to suppress the harness and become the sole, well-formed output source). Any grading that trusts submission-visible markers is defeatable; denylisting `include_str!`/`std::fs`/`env!` is bypassable (macros/path tricks).

Non-forgeable Rust grading needs an **out-of-band isolated executor** (separate compilation unit + verdict via a channel the submission can't read or write) — the #196-style sandboxed microservice. Rerouting #197 there; rust challenges stay fail-closed (like buildable) until it exists. The implementation's own tests only exercised *blind* nonce-guessing and went green — the mandatory adversarial pass is what caught this.

PR #219, opened 2026-06-26
feat(onchain): minting kill-switch via Config.paused (#128)

### LINKED ISSUE #128: [P1-12] Minting kill-switch for incident response

Add operator pause (deactivate `MinterRole`s via `revoke_minter`, or `Config.paused` gating `reward_xp`/`award_achievement`/`issue_credential`).

**Done when:** an operator can pause/resume minting on mainnet without redeploy. *(Feeds PLAN K-3; David writes runbook step.)*

thomgabriel:
## Summary
Adds an authority-controlled **minting kill-switch** for incident response: when `Config.paused == true`, `reward_xp`, `award_achievement`, and `issue_credential` are blocked (`MintingPaused`). Toggled live via `update_config` — no redeploy to pause/resume.

## Design — no account resize (critical)
`Config` gains `pub paused: bool` by consuming the **first** former reserved byte: `_reserved: [u8; 8]` → `paused` + `_reserved: [u8; 7]`. `Config::SIZE` stays **113** — existing on-chain Config accounts need **no migration**, and because their reserved bytes are zeroed, they deserialize as `paused = false` (verified by `legacy_zeroed_reserved_byte_deserializes_as_not_paused`).

- **Gating**: `require!(!config.paused, AcademyError::MintingPaused)` in all three mint handlers.
- **Toggle**: `UpdateConfigParams.paused: Option<bool>` (authority-gated by the existing `update_config` handler) → sets `config.paused` + emits `MintingPauseSet { paused, timestamp }`. `None` leaves it unchanged.
- `MintingPaused` is appended last in `errors.rs` (existing error codes unchanged).

## Tests (98 rust unit tests pass; fmt + clippy `-D warnings` clean)
`test_kill_switch.rs` + extended `test_initialize.rs`: paused blocks each of the 3 instructions; pause→resume re-enables; non-authority cannot toggle (`non_authority_cannot_toggle_pause`); fresh Config is not paused; `Some/None` apply/leave-unchanged; `Config::SIZE == 113`; legacy-account backward-compat; event/params serialization roundtrips. Plus a TS integration case in `onchain-academy.ts`.

## Activation (human gate)
**Requires a human devnet deploy to take effect** — no account migration needed (Config size unchanged). Per loop policy I did not deploy. IDL regen + frontend `superteam_academy.json` sync is a follow-up before activation (added in a follow-up commit if the build lands here; otherwise a pre-deploy step).

Closes #128.

### DIFF
--- a/onchain-academy/programs/onchain-academy/src/errors.rs
+++ b/onchain-academy/programs/onchain-academy/src/errors.rs
@@ -64,4 +64,6 @@ pub enum AcademyError {
     EnrollmentFinalized,
     #[msg("Account is not a valid Course PDA")]
     InvalidCourseAccount,
+    #[msg("Minting is paused")]
+    MintingPaused,
 }
--- a/onchain-academy/programs/onchain-academy/src/events.rs
+++ b/onchain-academy/programs/onchain-academy/src/events.rs
@@ -144,3 +144,9 @@ pub struct CourseClosed {
     pub course_id: String,
     pub timestamp: i64,
 }
+
+#[event]
+pub struct MintingPauseSet {
+    pub paused: bool,
+    pub timestamp: i64,
+}
--- a/onchain-academy/programs/onchain-academy/src/instructions/award_achievement.rs
+++ b/onchain-academy/programs/onchain-academy/src/instructions/award_achievement.rs
@@ -13,6 +13,8 @@ use crate::state::{AchievementReceipt, AchievementType, Config, MinterRole};
 use crate::utils::mint_xp;
 
 pub fn handler(ctx: Context<AwardAchievement>) -> Result<()> {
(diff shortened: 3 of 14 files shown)


thomgabriel:
**B1 (blocking) resolved in `7abb18a`.** Added `require!(!config.paused, AcademyError::MintingPaused)` to `complete_lesson::handler` and `finalize_course::handler` (both already load `config` read-only — no account-struct change), and updated the kill-switch tests to cover all five mint paths. Verified: `cargo fmt` + `clippy -D warnings` clean, **98 rust tests pass**, and a grep confirms exactly 5 `MintingPaused` guards (complete_lesson, finalize_course, reward_xp, award_achievement, issue_credential). This was independently caught by the mandatory adversarial review as well. Non-blocking S1/S2 tracked in https://github.com/solanabr/superteam-academy/issues/220.

PR #225, opened 2026-06-27
fix(web): force-dynamic on cookie/auth API GET routes (Vercel build)

Potolski:
### Symptom
`next build` fails / logs `DYNAMIC_SERVER_USAGE`:
> Route /api/quests/daily couldn't be rendered statically because it used `cookies`.

### Root cause
Next.js statically prerenders `GET` route handlers at build time. These routes call the Supabase **server** client (`createClient()` → reads cookies) or do other per-request work, which is invalid during prerender → Next throws `DYNAMIC_SERVER_USAGE`. The routes' own `try/catch` then swallows it and logs it as an "Unexpected error", and the route risks being statically cached as an error/stale response.

### Fix
Add `export const dynamic = "force-dynamic"` to the 11 dynamic GET API routes so Next always renders them at request time (never prerenders/caches):
`auth/nonce`, `auth/callback`, `quests/daily`, `deploy/[uuid]`, `deploy/save`, `admin/status`, `leaderboard`, `certificates/metadata`, `community/threads`, `community/threads/[id]`, `community/search`.

POST-only routes are unaffected (Next doesn't prerender them). Typecheck passes; verified each directive sits after imports and before the handler.

### DIFF
--- a/apps/web/src/app/api/admin/status/route.ts
+++ b/apps/web/src/app/api/admin/status/route.ts
@@ -27,6 +27,9 @@ import {
   getMissingAchievementFields,
 } from "@/lib/admin/sync-diff";
 
+// Auth/cookie + per-request DB access — never statically prerender (DYNAMIC_SERVER_USAGE).
+export const dynamic = "force-dynamic";
+
 export async function GET(req: NextRequest): Promise<NextResponse> {
   try {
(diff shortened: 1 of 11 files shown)


Potolski:
@claude

PR #237, opened 2026-06-30
fix(db): make award_community_xp daily cap atomic + UTC-pin the window (#182)

### LINKED ISSUE #182: [db] award_community_xp daily cap is non-atomic + uses session-TZ rollover
## Context
PR #179 hardened `award_xp()` with (a) a per-user `pg_advisory_xact_lock` so the read-then-insert daily cap is atomic, and (b) a UTC-pinned daily window (`date_trunc('day', now() AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'`).

`award_community_xp()` has the **same two issues** and was intentionally left untouched to keep #179 focused:
1. **Non-atomic daily cap** — `SELECT SUM(...) INTO ... ; INSERT` with no per-user lock; concurrent awards can both read a sub-cap total and exceed the 50/day community ceiling. Bounded + low impact (cap is 50/day), but inconsistent with the now-hardened `award_xp`.
2. **Session-TZ rollover** — the daily window uses `CURRENT_DATE`-style boundary instead of UTC midnight.

## Done when
Apply the same two fixes to `award_community_xp()` for consistency.

## Notes
- Severity is genuinely low (community cap is 50/day), filed for consistency/tracking per review discipline.
- Refs: `supabase/schema.sql` → `award_community_xp` (~lines 1190–1230).
- Source: #179 independent adversarial review.

thomgabriel:
## Summary

Ports the two hardening fixes from PR #179 (`award_xp`) to `award_community_xp`, which had the same two defects. Closes #182.

- **Non-atomic daily cap**: Added `pg_advisory_xact_lock(hashtext('award_community_xp:' || p_user_id::text)::bigint)` immediately after the `p_amount <= 0` guard. Concurrent community awards for the same user now serialize before the read-then-insert cap check, so two concurrent calls can no longer both read a sub-50 total and both insert. Uses a distinct `award_community_xp:` namespace to avoid contending with `award_xp`'s `award_xp:` lock (the two functions count disjoint transaction rows: `community:%` vs `NOT community:%`).
- **Session-TZ rollover**: Replaced both `(CURRENT_DATE)::timestamptz` occurrences (in the `v_daily_total` query and the `v_daily_vote_total` query) with `date_trunc('day', now() AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'`, pinning the daily window to UTC midnight regardless of DB session timezone.

Streak logic (`CURRENT_DATE` for `last_activity_date` in `user_xp` upsert) is intentionally unchanged, matching `award_xp` behaviour.

## Files changed

- `supabase/schema.sql` — patched `award_community_xp` body in place (canonical source of truth)
- `supabase/migrations/20260630165536_harden_award_community_xp.sql` — idempotent `CREATE OR REPLACE FUNCTION` migration with re-stated REVOKE/GRANT

## Verification

- Structural diff between the function body in `schema.sql` and the migration: **byte-identical** (`diff` returned no output).
- No `(CURRENT_DATE)::timestamptz` remaining in the `award_community_xp` body in either file (only appears in the migration comment header explaining the old pattern).
- UTC-pinned pattern `date_trunc('day', now() AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'` appears exactly 2 times in the SQL body of each file.
- Advisory lock line `pg_advisory_xact_lock(hashtext('award_community_xp:' || p_user_id::text)::bigint)` present once in each file.
- `$$` delimiters balanced (2 occurrences = 1 pair). `SECURITY DEFINER SET search_path = ''` preserved. REVOKE/GRANT re-stated.
- No local Supabase CLI or Postgres available (Docker daemon not running); validated by structural review against the known-valid `award_xp` constructs already in production.

## Deploy note

Apply `supabase/migrations/20260630165536_harden_award_community_xp.sql` to the production Supabase at merge (SECURITY DEFINER function replacement). Migration is idempotent — safe to re-run.

### DIFF
--- a/supabase/migrations/20260630165536_harden_award_community_xp.sql
+++ b/supabase/migrations/20260630165536_harden_award_community_xp.sql
@@ -1,0 +1,123 @@
+-- Migration: harden_award_community_xp
+-- Fixes two defects in award_community_xp() that were already fixed in
+-- award_xp() by PR #179. Refs: #182 (this fix), #179 (the reference fix).
+--
+-- Fix 1 — Non-atomic daily cap:
+--   The previous implementation did SELECT SUM(...) INTO v_daily_total; INSERT
+--   with no per-user lock. Concurrent community awards could both read a
+--   sub-cap total and both insert, exceeding the 50/day (and 10/day-vote)
+--   ceiling. Added pg_advisory_xact_lock() to serialize concurrent calls for
+--   the same user, making the read-then-insert atomic.
+--   Uses a distinct key namespace ('award_community_xp:') so it does not
+--   contend with award_xp's lock ('award_xp:') — the two functions count
+--   disjoint xp_transactions rows (community:% vs NOT community:%).
+--
+-- Fix 2 — Session-TZ rollover:
+--   The daily window used (CURRENT_DATE)::timestamptz, which depends on the
+--   DB session timezone. Replaced with:
+--     date_trunc('day', now() AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'
+--   to pin the boundary to UTC midnight regardless of session timezone.
+--   Both occurrences patched (v_daily_total query and v_daily_vote_total query).
+--
+-- Scope: ONLY the cap window + atomicity. Streak logic (last_activity_date =
+-- CURRENT_DATE) and user_xp upsert are intentionally left unchanged, matching
+-- award_xp() which also keeps CURRENT_DATE for streak day.
+--
+-- This is idempotent (CREATE OR REPLACE FUNCTION). Safe to re-run.
+
+CREATE OR REPLACE FUNCTION award_community_xp(
+  p_user_id UUID,
+  p_amount INTEGER,
+  p_reason TEXT,
+  p_idempotency_key TEXT DEFAULT NULL
+) RETURNS BOOLEAN
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path = ''
+AS $$
+DECLARE
+  v_daily_total INTEGER;
+  v_daily_vote_total INTEGER;
+  v_is_vote_xp BOOLEAN;
+BEGIN
+  IF p_amount <= 0 THEN RETURN FALSE; END IF;
+
+  -- Serialize concurrent community awards for the same user so the read-then-
+  -- insert daily cap below is atomic (mirrors the lock added to award_xp in
+  -- #179). A distinct key namespace avoids contending with award_xp's lock —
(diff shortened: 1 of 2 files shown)


thomgabriel:
## ✅ Both gates pass + independent adversarial review — flagging for human merge

**CI:** all checks green (Frontend lint·typecheck·test, On-chain, Build server, Anchor localnet, Dependency audit, Lighthouse; Verifiable-build skipped as configured).

**Gate 1 — `claude[bot]` auto-review:** *No blocking issues. Approved after human sign-off.*

**Gate 2 — independent adversarial review** (fresh skeptic tasked to *break* the atomicity/UTC/scope claims): **SOUND.** All attack points held:
- Advisory lock genuinely serializes same-user awards — every caller goes through `admin.rpc()` (service_role); each PostgREST RPC is its own auto-commit txn; lock is acquired **before** the daily-total SELECT.
- No deadlock: `revoke_community_xp` takes no advisory lock, and no code path acquires both `award_xp:` and `award_community_xp:` locks in one transaction (the votes route interleaves them as *separate* RPCs).
- UTC expression resolves to the correct UTC-midnight instant, byte-identical to the proven `award_xp` pattern.
- No behavioral regression (exactly 3 hunks vs baseline); `SECURITY DEFINER` / `search_path=''` / REVOKE+GRANT posture preserved; schema.sql↔migration byte-identical; timestamp correctly ordered; idempotent.

### Non-blocking findings — triaged (all declined, with reasons)
1. **32-bit lock key space** (`hashtext()::bigint`) — declined: identical to the `award_xp` production pattern; false-contention ≈ 1 in 2³² per user-pair, negligible. Diverging here would break consistency with the reference.
2. **`CURRENT_DATE` streak boundary vs UTC cap window** — declined: pre-existing and intentional (matches `award_xp`); Supabase runs in UTC, so `CURRENT_DATE = now()::date` operationally. Out of scope for #182 (cap window + atomicity only).
3. **No per-award hard ceiling (unlike `award_xp`'s 2000)** — declined: redundant here. The existing `IF v_daily_total + p_amount > 50 THEN p_amount := 50 - v_daily_total` clamp already bounds any single community award to ≤50; `award_xp` needs its 2000 ceiling only because its 5000/day cap permits large single awards.

### ⛔ needs human review — NOT auto-merging
Touches `supabase/schema.sql` + `area:db` (SENSITIVE) → a human signs off per project policy.

**Deploy note:** on merge, apply `supabase/migrations/20260630165536_harden_award_community_xp.sql` to the production Supabase (idempotent `CREATE OR REPLACE` SECURITY DEFINER replacement). Merging the PR alone does not change prod behaviour until the migration runs.

PR #238, opened 2026-06-30
test(onchain): harden kill-switch — pin MintingPaused=6031 + assert all 5 gated mint paths blocked (#220)

### LINKED ISSUE #220: [onchain][test] Kill-switch (#128) test follow-ups: TS blocked-path assertions + pin MintingPaused code
Non-blocking test follow-ups from the #219 (issue #128) adversarial + claude[bot] review. The kill-switch logic itself is complete and gated on all five mint paths (merged); these tighten the tests.

**S1 — TS integration: assert blocked-path behaviour.** `onchain-academy/tests/onchain-academy.ts` verifies the pause/resume toggle state machine but never calls a minting instruction while `paused == true` and asserts `MintingPaused`. Add end-to-end "blocked while paused" assertions for all five gated instructions (`complete_lesson`, `finalize_course`, `reward_xp`, `award_achievement`, `issue_credential`). NOTE: these run under the `anchor-localnet` CI job, currently gated off / red (#126/#209) — pairs with enabling that job.

**S2 — pin the MintingPaused error code.** `tests/rust/src/test_kill_switch.rs` `pause_errors_have_distinct_stable_codes` asserts `minting_paused_code > unauthorized_code` (true for any non-zero position). Hardcode the exact expected value (`MintingPaused` is appended last) so a future error-list reorder fails CI immediately.

thomgabriel:
Closes #220

## What

Test-only hardening of the already-merged minting kill-switch (#128/#219). **No program/source logic changed** — only the two test files.

### S2 — pin the `MintingPaused` error code (`tests/rust/src/test_kill_switch.rs`)
`pause_errors_have_distinct_stable_codes` previously asserted only the *relative* ordering (`minting_paused_code > unauthorized_code`). Added `assert_eq!(minting_paused_code, 6031)` (the existing relative assertion is kept). `MintingPaused` is the 32nd/last `AcademyError` variant (index 31 → 6000 + 31 = 6031), cross-checked against the canonical IDL (`apps/web/src/lib/solana/idl/superteam_academy.json`, which carries an explicit `"code": 6031`). Pinning the absolute value makes any future error-list reorder fail CI immediately instead of silently shifting downstream clients.

### S1 — assert blocked-path behaviour for all 5 gated instructions (`tests/onchain-academy.ts`)
Section 2's pause/resume tests verified the toggle state machine but never invoked a mint instruction while `paused == true`. Added a self-contained section 22 that, with the config **paused**, drives all five gated instructions and asserts each fails with `MintingPaused` (code 6031):

- `complete_lesson` — fresh enrollment, lesson 0 left uncompleted
- `finalize_course` — separate enrollment, all lessons done, not finalized
- `reward_xp` — fresh active minter with full cap headroom
- `award_achievement` — fresh achievement type + recipient/receipt + asset
- `issue_credential` — finalized enrollment, no credential yet

Every fixture is built **unpaused** in `before()` and chosen so the instruction would *succeed* if not paused, making the pause gate the only possible failure reason. The pause → assert → resume body is wrapped in `try/finally` so a failed assertion cannot leave the chain paused for later suites.

## Verification

All commands run locally in the worktree. Honest pass/fail:

| Command | Result |
| --- | --- |
| `cargo test --manifest-path onchain-academy/tests/rust/Cargo.toml` | **PASS** — 98 passed, 0 failed (incl. `pause_errors_have_distinct_stable_codes` with the new 6031 pin) |
| `anchor build` | **PASS** — program compiles, IDL + TS types regenerated |
| `pnpm exec tsc --noEmit … tests/onchain-academy.ts` (typecheck, `--skipLibCheck`) | **PASS** — exit 0, zero type errors (verified `AnchorError.error.errorCode.number` is a typed field) |
| `anchor test --provider.cluster localnet` (local validator + `mpl_core` genesis) | **PASS** — 89 passing, 0 failing, incl. **`22. Kill-switch blocks all gated mint paths › rejects all five gated mint instructions with MintingPaused while paused`** |
| `prettier --check onchain-academy/tests/onchain-academy.ts` (root prettier 3.8.1) | **PASS** — Prettier-clean |

Notes:
- S1 ran for real on a **local validator** (not just CI). The `anchor-localnet` CI job will re-execute it.
- The `tests/rust` workspace has **pre-existing rustfmt drift on `main`** (`lib.rs` mod ordering, `test_achievement.rs` imports) that is *not* part of this issue; it was deliberately left untouched. Only the intended 3-line addition to `test_kill_switch.rs` is included, and that addition is itself rustfmt-clean.
- Diff vs `main` is exactly the two target files (`+476 / −0`).

### DIFF
--- a/onchain-academy/tests/onchain-academy.ts
+++ b/onchain-academy/tests/onchain-academy.ts
@@ -5110,4 +5110,477 @@ describe("onchain-academy", () => {
       expect(recreated.bump).to.equal(migCourseBump);
     });
   });
+
+  // ===========================================================================
+  // 22. Kill-switch: paused blocks every gated mint path (issue #220)
+  //
+  // Sections 2's pause/resume tests only verify the toggle state machine; they
+  // never call a mint instruction while paused == true. This section closes
+  // that gap end-to-end: with the config paused, each of the FIVE gated
+  // instructions — complete_lesson, finalize_course, reward_xp,
+  // award_achievement, issue_credential — must fail with MintingPaused (anchor
+  // code 6031). Every fixture is built UNPAUSED in `before()` and chosen so the
+  // instruction would SUCCEED if not paused, making the pause gate the only
+  // possible failure reason. The pause→assert→resume body is wrapped in
+  // try/finally so a failed assertion can't leave the chain paused for later
+  // suites.
+  // ===========================================================================
+  describe("22. Kill-switch blocks all gated mint paths", () => {
+    const MPL_CORE_PROGRAM_ID = new PublicKey(
+      "CoREENxT6tW1HoK8ypY1SxRMZTcVPm7R94rH4PZNhX7d"
+    );
+
+    const KS_COURSE_ID = "killswitch-blocked";
+    const KS_LESSON_COUNT = 2;
+    const KS_XP_PER_LESSON = 10;
+    let ksCoursePda: PublicKey;
+    let ksCollection: PublicKey;
+
+    // complete_lesson target: lesson 0 left UNcompleted (would mint if unpaused).
+    const completeLearner = Keypair.generate();
+    let completeLearnerAta: PublicKey;
+    let completeEnrollPda: PublicKey;
+
+    // finalize_course target: all lessons done, NOT finalized (would mint if unpaused).
+    const finalizeLearner = Keypair.generate();
+    let finalizeLearnerAta: PublicKey;
+    let finalizeEnrollPda: PublicKey;
+
+    // issue_credential target: all lessons done AND finalized, no credential yet.
+    const credLearner = Keypair.generate();
+    let credLearnerAta: PublicKey;
+    let credEnrollPda: PublicKey;
+
+    // reward_xp target: a fresh active minter with full cap headroom.
+    const ksMinter = Keypair.generate();
+    let ksMinterRolePda: PublicKey;
+    const ksRewardRecipient = Keypair.generate();
+    let ksRewardRecipientAta: PublicKey;
+
+    // award_achievement target: a fresh achievement type with supply + recipient.
+    const KS_ACHIEVEMENT_ID = "killswitch-ach";
+    let ksAchievementTypePda: PublicKey;
+    let ksAchievementCollection: Keypair;
+    const ksAchRecipient = Keypair.generate();
+    let ksAchRecipientAta: PublicKey;
+    let ksAchReceiptPda: PublicKey;
+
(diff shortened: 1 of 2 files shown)


thomgabriel:
## ✅ Both gates pass + independent adversarial review — flagging for human merge

**CI:** all green, incl. the live `anchor-localnet` job (the new section 22 actually executes there) and the Rust unit suite (the 6031 pin).

**Gate 1 — `claude[bot]`:** *No blocking issues — clean test-only PR.* Verified the 6031 pin (IDL `:1428` + `errors.rs` index 31), the `try/finally` resume, and that all five handlers open with `require!(!config.paused, MintingPaused)`.

**Gate 2 — independent adversarial review** (skeptic tasked to find "passes for the wrong reason"): **MINOR.** Confirmed all 5 assertions check the EXACT `MintingPaused`/6031 code (not a generic throw), and the pause guard being the first statement in every handler guarantees right-reason failures.

### Non-blocking notes for the merger (test is correct as-is)
1. **`award_achievement` fixture vs. the section comment.** The comment says *"every fixture would succeed if not paused."* True for 4 of 5, but **not** `award_achievement`: its `minter_role` is the backend minter that section 19 leaves **deactivated** (`is_active == false`, never reactivated), and `award_achievement.rs:22` does `require!(role.is_active, MinterNotActive)` — so unpaused it would fail `MinterNotActive` (6018), not succeed. The assertion is still **right-reason**, because the pause check (`award_achievement.rs:16`) runs *before* the minter-active check (`:22`) — so this case actually verifies *pause takes precedence over the minter gate*. Optional tightening on merge: point the fixture at an active minter (e.g. the section's `ksMinter`) so all 5 are uniformly "clean-but-for-pause," or reword the comment. Not blocking. (Note: `claude[bot]`'s note #2 read this minter as usable — it didn't trace §19's deactivation; the adversarial pass + `award_achievement.rs:22` confirm it's inactive.)
2. `expectMintingPaused`'s non-`AnchorError` fallback only string-matches (no numeric check) — dead path on a localnet run, fine as-is.
3. Section 22's finalize fixture reads `creatorTokenAccount` from section 7's scope — works under Mocha's sequential execution; an implicit ordering dependency worth knowing.

### ⛔ needs human review — NOT auto-merging
`area:onchain` (SENSITIVE) → human signs off. No program/source logic changed; diff is exactly the two test files (`+476/−0`). Beyond CI, the agent ran the real `anchor test` localnet suite (89 passing incl. the new blocked-paths test) + `cargo test` (98 passing incl. the 6031 pin) locally.

PR #170, opened 2026-06-24
fix(security): add CSP + security headers (P0-C6)

thomgabriel:
Closes #153

## What

Adds an `async headers()` to `apps/web/next.config.mjs`. Previously the config had **no** `headers()`, so prod responses carried none of the standard security headers. This returns, for **all routes** (`source: "/:path*"`), the five required headers plus `Permissions-Policy`:

| Header | Value |
| --- | --- |
| `Content-Security-Policy` | enforcing policy (directives below) |
| `Strict-Transport-Security` | `max-age=63072000; includeSubDomains; preload` |
| `X-Frame-Options` | `SAMEORIGIN` (Studio is a same-origin embed) |
| `X-Content-Type-Options` | `nosniff` |
| `Referrer-Policy` | `strict-origin-when-cross-origin` |
| `Permissions-Policy` | `camera=(), microphone=(), geolocation=(), interest-cohort=()` |

## CSP directives chosen

```
default-src 'self';
script-src  'self' 'unsafe-eval' 'unsafe-inline' blob:
            https://www.googletagmanager.com https://*.posthog.com;
style-src   'self' 'unsafe-inline' https://cdn.sanity.io;
font-src    'self' data: https://fonts.gstatic.com;
img-src     'self' data: blob: https://cdn.sanity.io https://lh3.googleusercontent.com
            https://arweave.net https://*.arweave.net https://*.supabase.co;
connect-src 'self'
            https://*.supabase.co wss://*.supabase.co
            https://cdn.sanity.io https://apicdn.sanity.io https://*.api.sanity.io wss://*.api.sanity.io
            https://*.helius-rpc.com https://api.devnet.solana.com https://api.mainnet-beta.solana.com
            https://accounts.google.com https://*.googleapis.com
            https://www.googletagmanager.com https://www.google-analytics.com https://*.google-analytics.com
            https://*.posthog.com https://*.sentry.io https://*.ingest.sentry.io;
frame-src   'self' https://accounts.google.com;
worker-src  'self' blob:;
form-action 'self' https://accounts.google.com;
base-uri 'self'; object-src 'none'; frame-ancestors 'self';
```

### Allowlist rationale (origins verified by grepping the codebase)
- **Supabase** — REST + realtime (`wss://*.supabase.co`, used by `use-gamification-events.ts`).
- **Sanity** — embedded Studio at `/studio` (`cdn.sanity.io`, `apicdn.sanity.io`, `*.api.sanity.io` + listen `wss://`).
[...]

PR #246, opened 2026-07-01
fix(csp): nonce-based script-src for the app + pin wildcards (#177)

### LINKED ISSUE #177: Harden CSP: nonce-based script-src (drop unsafe-inline/unsafe-eval) + pin wildcards
Follow-up from the #170 CSP review. The CSP is functionally correct (verified live — Monaco/Studio/fonts load), but the script policy is weak: `script-src` carries `unsafe-inline` + `unsafe-eval`, so it stops little XSS.
- `unsafe-eval` is genuinely required (the in-browser code sandbox uses `new Function` in a worker).
- `unsafe-inline` is required because Next.js emits an un-nonced inline bootstrap.
To make the script CSP meaningful: add **nonce-based** hashing (per-request nonce via middleware, injected into Next inline scripts, replacing `unsafe-inline`). Also pin the broad `connect-src`/`img-src` wildcards (`*.supabase.co`, `*.helius-rpc.com`, etc.) to concrete subdomains. A `// TODO tighten` already marks this in `next.config.mjs`.
Done when: `script-src` uses nonces (no `unsafe-inline` for scripts) and wildcards are pinned, app still functioning.

thomgabriel:
Closes #177.

## What changed
Splits the CSP into two policies:
- **App routes** → a per-request **nonce** `script-src` (built in `middleware.ts` + new `lib/csp.ts`), dropping `'unsafe-inline'` for scripts.
- **`/studio/*` + `/api/*`** (excluded from middleware) → keep the static `'unsafe-inline'` fallback in `next.config.mjs` (Sanity Studio is a third-party bundle that can't be nonced; API serves JSON).

Key decisions:
- **`'unsafe-eval'` KEPT** — the challenge sandbox (`new Function`) + Monaco need it.
- **No `'strict-dynamic'`** — it would discard the host allowlist and break Monaco's jsdelivr AMD loader.
- Nonce stamped onto Next's inline bootstrap + the `next-themes` inline script; analytics (GA4/PostHog) load as **host-allowlisted external** scripts, so they need no nonce.
- **Wildcard pinning:** Supabase + the Solana RPC pinned to their configured origins, each with a wildcard fallback so a config gap can't break connectivity. Sanity/PostHog/Sentry kept as wildcards (regional/multi-subdomain ingest hosts — pinning risks breakage).

## Verification status — ⚠️ runtime check still required
- CI validates build + unit tests.
- I've **design-reviewed** this end-to-end (nonce propagation path, unsafe-eval retained, analytics external, conservative pinning) — it's sound.
- **NOT yet done:** a live-browser runtime check that Next actually propagates the nonce to its inline bootstrap and that Monaco / the challenge runner / Sanity Studio / analytics load with **zero CSP violations**. This is the real gate for a CSP change (a prior CSP change broke Monaco app-wide), so **do not merge until that live check passes.**

`needs-human-review` — high-risk CSP change; verify in a live preview first.

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -4,38 +4,28 @@ import { withSentryConfig } from "@sentry/nextjs";
 const withNextIntl = createNextIntlPlugin("./src/lib/i18n/request.ts");
 
 /**
- * Content-Security-Policy.
+ * Static Content-Security-Policy FALLBACK for routes the middleware does not
+ * run on — namely `/studio/*` (Sanity Studio) and `/api/*`. The user-facing app
+ * gets a stricter, per-request nonce CSP built in `middleware.ts` + `lib/csp.ts`
+ * (no `'unsafe-inline'` for scripts); middleware response headers override these
+ * static ones where both apply, so the nonce policy wins for matched routes.
  *
- * This is an ENFORCING policy tuned to keep every legitimate integration
- * working. Where a directive can't be locked down without breaking runtime
- * behaviour it is left permissive on purpose (see inline `unsafe-*` notes and
- * `TODO tighten` markers). Tightening any of these requires runtime QA of the
- * affected feature first.
+ * Studio KEEPS `'unsafe-inline'` + `'unsafe-eval'` for scripts: it is a
+ * third-party SPA bundle that cannot be nonced and relies on eval. `/api/*`
+ * returns JSON/binary (no inline scripts), so the script policy is moot there.
+ * Keep the non-script directives here in rough sync with `lib/csp.ts`.
  *
  * Notable allowances:
- * - `'unsafe-eval'` in script-src — REQUIRED. The in-browser code-challenge
- *   sandbox (`components/editor/challenge-runner.tsx`) runs learner code via
- *   `new Function()`, and Monaco's tokenizer/worker bootstrap also relies on
- *   it. Removing it WILL break the editor and challenge runner.
- * - `'unsafe-inline'` in script-src/style-src — Next.js injects an inline
- *   bootstrap script and inline critical CSS; nonces are not wired up.
- * - Server-only externals (Gemini, Rust Playground, build server) are NOT
- *   listed: the browser only ever talks to same-origin `/api/*` routes that
- *   proxy them.
- *
- * KNOWN DEFERRED LIMITATION: `'unsafe-inline'` + `'unsafe-eval'` keep the
- * script CSP weak — together they neutralise CSP's XSS protection for inline
- * and eval'd scripts. Closing this requires a nonce- (or hash-) based script
- * CSP wired through Next.js's inline bootstrap; that is intentionally NOT
- * attempted here. Tracked by the `TODO tighten` marker below.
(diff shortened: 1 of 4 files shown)


thomgabriel:
## Runtime verification (prod server + curl) — core CSP mechanism confirmed working ✅

Ran `pnpm start` from this branch and probed the live responses:

- **Nonce propagation works** — within a **single** request the CSP header nonce equals the nonce Next stamps on its inline scripts (e.g. header `nonce-jvdMXJKOp0ZDkJobYcUMCA==` == body `nonce="jvdMXJKOp0ZDkJobYcUMCA=="`). So Next's inline bootstrap/hydration scripts are allowed — **no hydration break**.
- **Zero un-nonced inline `<script>`** on `/en` and `/en/courses` → nothing gets CSP-blocked.
- **`script-src`** on app routes = `'self' 'nonce-…' 'unsafe-eval' blob: https://cdn.jsdelivr.net https://www.googletagmanager.com https://*.posthog.com` — **`unsafe-eval` kept**, no `'unsafe-inline'`, no `'strict-dynamic'`. Monaco's needs (`unsafe-eval`, jsdelivr host, `blob:`, `worker-src`) are all present — the exact thing whose removal broke Monaco in #170 is **not** done here.
- **Studio fallback intact** — `/studio` still serves the static `'unsafe-inline'`/`'unsafe-eval'` CSP (the third-party bundle that can't be nonced).

### Not yet done (the one remaining gate)
- A **live in-browser render of a Monaco lesson page** (editor mounts + no console CSP violations) + analytics firing. I verified the HTML/headers via curl, not a running browser executing scripts.
- **Note:** `/en/courses` returned HTTP 500 in the worktree prod build — but `/en` renders fine through the identical middleware and its scripts were still nonce'd, so this is a page-data/stale-build artifact, **not** the CSP. Worth confirming it also 500s on `main` (pre-existing) when doing the browser pass.

**Recommendation:** the risky mechanism (nonce propagation) is proven and Monaco's CSP requirements are demonstrably preserved. Remaining before merge: a ~30s browser click-test of a lesson page (Monaco) + a normal page, confirming zero CSP violations in the console. Keeping `needs-human-review` for that final live check.

thomgabriel:
## Live browser check caught a real break — fixed ✅

The lesson-page (challenge-runner Monaco) verification surfaced a genuine CSP violation:
> `Loading the stylesheet 'https://cdn.jsdelivr.net/.../editor.main.css' violates … style-src` — and a `connect-src` block on Monaco's `*.map` source maps.

Monaco loads its **CSS + source maps from jsdelivr**, but jsdelivr was only in `script-src`/`worker-src`. The blocked stylesheet made the editor render unstyled (a bare input-like box). **Fixed in `bf1e57c`** — added `https://cdn.jsdelivr.net` to `style-src` and `connect-src`. Verified live: the response header now carries jsdelivr in both.

(Note: this gap predates the PR — `main`'s CSP also lacks jsdelivr in `style-src` — so this also fixes Monaco styling on `main`. The nonce/curl checks couldn't catch it because they don't run Monaco's CDN loader; the live browser check did.)

Re-verifying: reload a lesson page, confirm the editor renders **styled** with no CSP errors in the console.

PR #311, opened 2026-07-05
feat(admin): grant/revoke teacher by wallet address instead of username

### LINKED ISSUE #310: Admin: grant/revoke teacher by wallet address instead of username
## Summary
Change teacher-role granting to look up users by **wallet address** instead of username. Usernames are auto-generated and easy to mistype; the wallet address is the stable identifier an admin actually has for a teacher.

## Scope
- `POST /api/admin/teachers`: accept `walletAddress` (not `username`); look up `profiles` by `wallet_address`. Same admin gate, same teacher↔learner-only, same admin-protection.
- `GET /api/admin/teachers`: include `wallet_address` in the teacher list.
- Admin "Teacher Roles" panel: input + list use wallet address.

## Acceptance criteria
- [ ] Admin grants/revokes teacher by wallet address from `/admin`
- [ ] Current teachers list shows wallet addresses
- [ ] `tsc` + lint + build clean

Potolski:
## Summary

Switches teacher-role granting to look up users by **wallet address** instead of username — the stable identifier an admin actually has, versus an auto-generated (easy-to-mistype) name.

## What

- **`POST /api/admin/teachers`** — accepts `walletAddress`, looks up `profiles.wallet_address`. Same admin gate (`admin_session`), teacher ↔ learner only, still refuses to touch admin accounts.
- **`GET`** — includes `wallet_address` in the teacher list.
- **Teacher Roles panel** — input + list use the wallet address (shown short-form with `@username` for reference).

## Verification

- `tsc` clean, eslint clean

Closes #310

### DIFF
--- a/apps/web/src/app/api/admin/teachers/route.ts
+++ b/apps/web/src/app/api/admin/teachers/route.ts
@@ -18,8 +18,9 @@ const ROLE_BY_ACTION = {
 
 type Action = keyof typeof ROLE_BY_ACTION;
 
-// Solana base58 addresses are 32-44 chars; keep a generous bound.
-const MAX_WALLET_LENGTH = 64;
+// The identifier is a wallet address (base58, 32-44 chars) OR a username; keep a
+// generous bound covering both.
+const MAX_IDENTIFIER_LENGTH = 64;
 
 function isAction(value: unknown): value is Action {
   return value === "grant" || value === "revoke";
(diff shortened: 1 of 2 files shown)


Potolski:
@claude Addressed the blocking regression (https://github.com/solanabr/superteam-academy/commit/cf4e913654b8bf10cc5dfbd91aada1839667e174).

The endpoint no longer looks up by wallet_address only. It now takes a single identifier matched against wallet_address first, then username, so wallet-less accounts (Google/GitHub-only sign-in, where wallet_address is null and can never be linked) stay reachable:

Grant to a wallet-less account works via its username; the 404 message now reads "No user found with that wallet address or username".
Revoke from the list revokes by wallet_address when present, else username — the per-row Revoke button is no longer disabled for null-wallet teachers.
The input accepts either form, and the response returns both username and walletAddress for display.

PR #313, opened 2026-07-05
fix(catalog): revalidate course cache on admin sync

Potolski:
## Problem
A newly approved + on-chain-synced course didn't show up in the public catalog (`/courses`) for up to an hour, even though its status matched already-visible courses.

## Root cause
The public catalog reads (`getAllCourses`, `getCourseBySlug`, `getAllLearningPaths`) go through `sanityFetch`, which uses `next: { revalidate: 3600 }` — a 1-hour ISR cache. After an admin approves + syncs a course (which writes `onChainStatus.status = "synced"` back to Sanity), the catalog keeps serving the stale cached page until the hour elapses.

## Fix
On-demand cache revalidation:
- `sanityFetch` now accepts optional `tags`, forwarded to Next's `next: { tags }`.
- The three public catalog reads are tagged with a shared `COURSES_CACHE_TAG` (`"courses"`).
- `/api/admin/courses/sync` calls `revalidateTag(COURSES_CACHE_TAG)` on every success path (created / updated / noop), right after the Sanity `onChainStatus` write-back — so the catalog reflects the change immediately.

## Note on deactivate/reactivate
Those routes only flip the on-chain `is_active` flag; they don't mutate the Sanity fields the catalog gate reads, so no revalidation is added there (it would be a no-op).

## Verification
- `tsc --noEmit` clean, eslint clean

Closes #312

PR #325, opened 2026-07-06
fix(admin): hide deactivated courses from the public catalog

### LINKED ISSUE #321: Admin: deactivating a course doesn't remove it from the catalog
## Bug
Deactivating a course from the admin panel doesn't remove it from the public catalog — the course still shows on `/courses`.

## Root cause
`POST /api/admin/courses/deactivate` calls `deactivateCoursePda(courseId)` (`lib/solana/admin-signer`), which only flips the **on-chain** `Course.is_active` flag. It writes nothing back to Sanity and does not revalidate the catalog cache.

The public catalog query `getAllCourses` (`lib/sanity/queries.ts`) gates solely on `onChainStatus.status == "synced" && publicAuthoringGate` — it never checks `is_active` (on-chain or Sanity). So a deactivated course keeps appearing.

## Fix (options)
1. On deactivate, write an "inactive"/hidden marker back to Sanity (e.g. a dedicated `isActive: false` field or an `onChainStatus.status` transition) and `revalidateTag(COURSES_CACHE_TAG)` (the tag added in #313), then have the catalog queries exclude inactive courses. Reactivate reverses it.
2. Or have the catalog read on-chain `is_active` — heavier per-request cost; option 1 (Sanity flag + revalidate) is preferred and consistent with the existing sync/revalidate flow.

Apply the same treatment to **reactivate** so it re-appears.

## Scope
- `apps/web/src/app/api/admin/courses/deactivate/route.ts` + `.../reactivate/route.ts`
- `apps/web/src/lib/sanity/queries.ts` (catalog gate) + `admin-mutations.ts` (Sanity write)
- Sanity `course` schema if a new field is added

## Done when
- [ ] Deactivating a course removes it from `/courses` promptly
- [ ] Reactivating restores it

Potolski:
## Bug
Deactivating a course from `/admin` didn't remove it from the public catalog — it kept showing on `/courses`.

## Root cause
`POST /api/admin/courses/deactivate` → `deactivateCoursePda` only flips the **on-chain** `Course.is_active` flag. It wrote nothing back to Sanity and didn't revalidate the catalog cache, and the public catalog gate keyed solely on `onChainStatus.status == "synced" && publicAuthoringGate` — never on `is_active`. So the course stayed visible.

## Fix
- **Mirror the flag into Sanity**: new `writeCourseActive(sanityId, isActive)` sets `onChainStatus.isActive`; the deactivate/reactivate routes call it after the on-chain tx and `revalidateTag(COURSES_CACHE_TAG)` (the tag from #313) so the catalog updates immediately.
- **Gate the reads**: new `activeGate = coalesce(onChainStatus.isActive, true)` folded into the shared synced/approved gate across all 12 public course queries — so detail pages, lessons, and dashboard reads hide inactive courses too, not just `/courses`. `coalesce(..., true)` keeps legacy docs (no field) visible.
- **Schema**: added `onChainStatus.isActive` (boolean) to the course schema.
- Admin queries (`getAllCoursesAdmin`) stay **ungated**, so deactivated courses remain visible and reactivatable in `/admin`.
- If the Sanity write fails, the route returns a `warning` (the on-chain tx already succeeded).

## Verification
- `tsc --noEmit` ✓, eslint ✓

Closes #321

### DIFF
--- a/apps/web/src/app/api/admin/courses/deactivate/route.ts
+++ b/apps/web/src/app/api/admin/courses/deactivate/route.ts
@@ -1,5 +1,6 @@
 import "server-only";
 
+import { revalidateTag } from "next/cache";
 import { NextRequest, NextResponse } from "next/server";
 import {
   requireAdminAuth,
@@ -7,6 +8,8 @@ import {
   AdminAuthError,
 } from "@/lib/admin/auth";
 import { deactivateCoursePda } from "@/lib/solana/admin-signer";
+import { writeCourseActive } from "@/lib/sanity/admin-mutations";
+import { COURSES_CACHE_TAG } from "@/lib/sanity/queries";
 
 export async function POST(req: NextRequest): Promise<NextResponse> {
   try {
@@ -38,5 +41,21 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
     );
   }
 
-  return NextResponse.json({ txSignature: result.signature });
+  // Mirror the on-chain flag into Sanity + purge the catalog cache so the course
+  // disappears from /courses immediately. If this write fails the course stays
+  // visible despite being deactivated on-chain, so surface a warning.
+  let warning: string | undefined;
+  try {
+    await writeCourseActive(courseId, false);
+    revalidateTag(COURSES_CACHE_TAG);
+  } catch (err) {
+    console.error("[admin/courses/deactivate] Sanity write-back failed:", err);
+    warning =
+      "Deactivated on-chain, but the catalog flag didn't update — the course may still show until re-synced.";
+  }
+
+  return NextResponse.json({
+    txSignature: result.signature,
+    ...(warning ? { warning } : {}),
+  });
 }
--- a/apps/web/src/app/api/admin/courses/reactivate/route.ts
+++ b/apps/web/src/app/api/admin/courses/reactivate/route.ts
@@ -1,5 +1,6 @@
 import "server-only";
 
+import { revalidateTag } from "next/cache";
 import { NextRequest, NextResponse } from "next/server";
 import {
   requireAdminAuth,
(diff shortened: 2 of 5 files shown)


Potolski:
Good catch — fixed in `0d9f9d6`. This was a real fail-open regression.

**Fix:** `getChallengeAnswerKey` and `getChallengeAnswerKeyById` are now **ungated** — they filter by course identity only (`_id` / `slug`), like `getCourseById`, and no longer carry `synced`/`activeGate`/`publicAuthoringGate` (and are off `catalogFetch`). So the completion gate's invariant holds again: `null` ⇔ *not a challenge lesson*, never *course is hidden*. A deactivated/unapproved course's challenge lesson is still validated before any on-chain completion.

Both the `/api/lessons/complete` gate and `/api/lessons/validate-challenge` use these, so both are covered. Grading is now independent of catalog visibility, which is the correct security posture (matches the fail-closed rule the `executor_unavailable` path already follows).

Only those two answer-key functions changed. All genuine catalog/display reads (`getAllCourses`, `getCourseBySlug`, `getLessonBySlug`, dashboard reads, learning paths, etc.) keep `activeGate`, so deactivated courses are still hidden from the catalog — the original #321 fix stands. `getCourseById` (lesson-index lookup in the same route) was already ungated.

`tsc` + eslint clean.

Potolski:
@claude Good catch — fixed in `0d9f9d6`. This was a real fail-open regression.

**Fix:** `getChallengeAnswerKey` and `getChallengeAnswerKeyById` are now **ungated** — they filter by course identity only (`_id` / `slug`), like `getCourseById`, and no longer carry `synced`/`activeGate`/`publicAuthoringGate` (and are off `catalogFetch`). So the completion gate's invariant holds again: `null` ⇔ *not a challenge lesson*, never *course is hidden*. A deactivated/unapproved course's challenge lesson is still validated before any on-chain completion.

Both the `/api/lessons/complete` gate and `/api/lessons/validate-challenge` use these, so both are covered. Grading is now independent of catalog visibility, which is the correct security posture (matches the fail-closed rule the `executor_unavailable` path already follows).

Only those two answer-key functions changed. All genuine catalog/display reads (`getAllCourses`, `getCourseBySlug`, `getLessonBySlug`, dashboard reads, learning paths, etc.) keep `activeGate`, so deactivated courses are still hidden from the catalog — the original #321 fix stands. `getCourseById` (lesson-index lookup in the same route) was already ungated.

`tsc` + eslint clean.

PR #329, opened 2026-07-06
fix(admin): reflect deactivated courses in admin table + dashboard

Potolski:
## Follow-up to #321 / #325
After deactivating a course, two surfaces still showed it. Root causes were separate from the catalog gate:

### 1) Admin table showed it as active
`/api/admin/status` and the admin table keyed only on `onChainStatus.status` (`"synced"`) — never on `is_active`. So a deactivated course still rendered as Synced with a Deactivate button, and there was no way to reactivate from the UI.

- `/api/admin/status` now returns the **authoritative on-chain `is_active`** per synced course (decoded via `fetchCourse`; defaults to active if the account is undecodable — same graceful handling as the sync route).
- The admin table shows a **"Deactivated"** badge and swaps Deactivate → **Reactivate** (calling the existing `/api/admin/courses/reactivate`) when `isActive === false`.

### 2) Dashboard still showed it
The "Continue learning" cards are built from Supabase `enrolledIds` and rendered with fallback data (`sanity?.title ?? id`). `getCoursesByIds`'s `activeGate` (from #325) only stripped the Sanity metadata — the card still rendered from the enrollment row. Now `currentCourses` is filtered to ids that still resolve from Sanity, so a deactivated (or unpublished) course drops off.

## Notes
- Admin reads on-chain `is_active` directly, so it's correct even if the Sanity mirror write lagged/failed.
- Consistent with the "deactivated = hidden everywhere" intent: lesson pages are already gated (`getLessonBySlug`), so an enrolled learner couldn't open lessons of a deactivated course anyway.

## Verification
- `tsc --noEmit` ✓, eslint ✓

Refs #321

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/dashboard/page.tsx
+++ b/apps/web/src/app/[locale]/(platform)/dashboard/page.tsx
@@ -323,19 +323,25 @@ function useDashboardData(
         // Build a lookup map: course _id -> Sanity data
         const courseMap = new Map(courseSummaries.map((c) => [c._id, c]));
 
-        const currentCourses: CurrentCourse[] = enrolledIds.map((id) => {
-          const sanity = courseMap.get(id);
-          return {
-            courseId: id,
-            title: sanity?.title ?? id,
-            slug: sanity?.slug ?? id,
-            completedLessons: completedPerCourse.get(id) ?? 0,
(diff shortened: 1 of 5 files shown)


Potolski:
@claude check if the redundant RPC call was fixed in the latest commit

Issue #314, opened 2026-07-05
Extended course: learner can't complete lessons added after completion (500) — lesson_count is immutable on-chain


Potolski:
## Summary
When a teacher adds lessons to a course that a learner has **already completed** (and been issued a certificate for), the learner can't complete the new lessons: `POST /api/lessons/complete` returns **500 Internal Server Error**.

Repro:
1. Learner completes every lesson of course X → course finalized, credential NFT issued.
2. Teacher adds one or more lessons to course X (Sanity), admin approves + re-syncs.
3. Learner opens a new lesson and marks it complete → 500.

## Root cause (on-chain limitation)
`Course.lesson_count` (`u8`) is set **once** at `create_course` and is **immutable** — `update_course` (`UpdateCourseParams`) has no field to change it (`onchain-academy/programs/onchain-academy/src/instructions/update_course.rs`).

The lesson index is derived from Sanity content order (`deriveLessonIndex` in `apps/web/src/app/api/lessons/complete/route.ts`). After the teacher extends the course, a new lesson's index is `>= course.lesson_count`, so the on-chain `complete_lesson` handler reverts:

```rust
require!(lesson_index < course.lesson_count, AcademyError::LessonOutOfBounds);
```
(`onchain-academy/programs/onchain-academy/src/instructions/complete_lesson.rs:15`)

`onChainCompleteLesson` throws on the revert and the route's catch-all returns a generic **500** (`.../lessons/complete/route.ts:280-290`). The certificate/finalize state is **not** the blocker — `complete_lesson` never checks it; the stale `lesson_count` is.

## Not a migration
The enrollment bitmap is already `[u64; 4]` = 256 bits and `lesson_count` is a `u8` (max 255), so **growing the count needs no account resize/backfill** — it fits the existing layout. Only the program logic and a redeploy are needed.

## Proposed fix (needs contract redeploy)
1. **On-chain** — add `new_lesson_count: Option<u8>` to `UpdateCourseParams`, **increase-only**:
   ```rust
   require!(new >= course.lesson_count, AcademyError::...); // never shrink — would strand
                                                            // completed flags / shift indices
   course.lesson_count = new;
   ```
   Add a dedicated error variant; cover with Rust + TS tests.
2. **Admin sync route** (`/api/admin/courses/sync`) — when Sanity `lessonCount > onChainCourse.lesson_count`, include `new_lesson_count` in the `update_course` call so a re-sync brings the on-chain count up to date.
[...]

PR #334, opened 2026-07-07
fix(onchain): sync update_course IDL to deployed program (fixes reactivate/deactivate error 102)

Potolski:
## Bug
Reactivating (and deactivating) a course failed with `AnchorError … InstructionDidNotDeserialize. Error Number: 102`. In fact **every** `update_course` call from the app was broken (course-sync field updates, collection backfill, activate/deactivate).

## Root cause — repo IDL is behind the deployed program
The deployed devnet program (`7NeJ…SgwE8V`) has a **7th field** in `UpdateCourseParams` — `new_lesson_count: Option<u8>` (the #314 lesson-count fix, deployed on-chain). The repo IDL/source only had **6 fields**, so the client serialized 15 bytes while the program expected 16 → deserialization rejected.

The earlier successful deactivate happened **before** that program upgrade, which is why the course showed `is_active=false` on-chain yet reactivate now fails.

### Confirmed by on-chain simulation
- Decoding on-chain Course accounts: all 224 bytes, one course already `is_active=false`.
- Simulating `update_course` against devnet:
  - **6-field** (repo IDL) → `Custom: 102` InstructionDidNotDeserialize
  - **7-field** (`new_lesson_count=null`) → **success**, emits `CourseUpdated`

## Fix — sync the repo to the deployed program
- **IDL** (`superteam_academy.json`): add `new_lesson_count` as the 7th `UpdateCourseParams` field (`option u8`), matching the deployed layout.
- **Client** (`admin-signer.ts`): serialize `newLessonCount` (null default) so instructions are 16 bytes again, and expose `newLessonCount` on the admin params for a future course-extension sync.
- **Program source** (`update_course.rs` + `errors.rs`): re-apply the field + increase-only guard so `anchor build` regenerates a matching 7-field IDL. **Without this, the next build would regenerate a 6-field IDL and silently re-break the app.**

Re-verified: with the committed repo IDL, the reactivate instruction now deserializes and the program returns success.

## Note for the team
This reconciles the repo with a program upgrade that was deployed to devnet but never committed (the #314 change). The source here re-implements that field with sensible increase-only semantics — please confirm it matches the exact source Thom deployed (field position/type match is what the serialization needs, and that's verified). This also means #314's client-side wiring (push `newLessonCount` on re-sync when a course is extended) can now be completed on top of this.

### DIFF
--- a/apps/web/src/lib/solana/admin-signer.ts
+++ b/apps/web/src/lib/solana/admin-signer.ts
@@ -76,6 +76,10 @@ interface UpdateCourseOnChainParams {
   newCreatorRewardXp: number | null;
   newMinCompletionsForReward: number | null;
   newCollection: PublicKey | null;
+  // 7th field — the deployed program's UpdateCourseParams includes it, so it
+  // MUST be serialized (even as null) or every update_course call fails to
+  // deserialize (AnchorError 102). Grows lesson_count when a course is extended.
+  newLessonCount: number | null;
 }
 
 interface CreateAchievementTypeOnChainParams {
@@ -133,6 +137,9 @@ export interface UpdateCourseAdminParams {
   newMinCompletionsForReward?: number;
   /** Base58 address of the Metaplex Core credential collection to set/backfill. */
   newCollection?: string;
+  /** New on-chain lesson_count (increase-only) — grows the completion bitmap
+   * bound when a course is extended after deploy. See #314. */
+  newLessonCount?: number;
 }
 
 export interface CreateAchievementAdminParams {
@@ -323,6 +330,7 @@ export async function updateCoursePda(
         params.newCollection != null
           ? new PublicKey(params.newCollection)
           : null,
+      newLessonCount: params.newLessonCount ?? null,
     };
 
     const methods = _program.methods as unknown as AdminMethods;
--- a/apps/web/src/lib/solana/idl/superteam_academy.json
+++ b/apps/web/src/lib/solana/idl/superteam_academy.json
@@ -1464,6 +1464,11 @@
       "code": 6033,
       "name": "XpAmountExceedsMax",
       "msg": "XP amount exceeds the per-mint ceiling"
+    },
+    {
+      "code": 6034,
+      "name": "LessonCountCannotDecrease",
+      "msg": "Lesson count can only increase"
     }
   ],
   "types": [
--- a/apps/web/src/lib/solana/idl/superteam_academy.json
+++ b/apps/web/src/lib/solana/idl/superteam_academy.json
@@ -9,16 +9,7 @@
   "instructions": [
     {
       "name": "award_achievement",
-      "discriminator": [
-        75,
-        47,
-        156,
-        253,
-        124,
-        231,
-        84,
-        12
-      ],
+      "discriminator": [75, 47, 156, 253, 124, 231, 84, 12],
       "accounts": [
         {
           "name": "config",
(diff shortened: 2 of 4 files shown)


Potolski:
Thanks — all four addressed.

**1. (BLOCKING) IDL errors array stale — fixed in `6979e27`.**
Added the 35th error entry to `superteam_academy.json`:
```json
{ "code": 6034, "name": "LessonCountCannotDecrease", "msg": "Lesson count can only increase" }
```
Verified it lines up: `errors.rs` has 35 variants with `LessonCountCannotDecrease` declared last → Anchor code 6034; `resolveIdlError` scans by the `code` field, so a `6034` revert now resolves to "Lesson count can only increase" instead of "Unknown program error (code 6034)". (Couldn't `anchor build`-regenerate to diff — this environment's IDL build hits an `ark-bn254`/rustc incompatibility — so the entry was hand-matched to what anchor would emit: same declaration order, code, and message.)

**2. (non-blocking) No test coverage** and **3. (non-blocking) guard not verified against the deployed program** — tracked together in #336, gated as **P1 before the growth path is wired to any caller**. Both are safe to defer right now because `admin-signer` only ever passes `newLessonCount: null` (the no-op path), so nothing exercises the guard yet. #336 covers: confirm the deployed guard/error against Thom's actual source, reconcile, add the Rust + anchor tests (revert path + `complete_lesson` accepting the newly-unlocked index), then complete the original #314 wiring.

**4. (minor) IDL reformat re-diff on next regen** — acknowledged; expect prettier noise on the next real `anchor build`. No action here.

Potolski:
@claude Thanks — all four addressed.

1. (BLOCKING) IDL errors array stale — fixed in 6979e27.
Added the 35th error entry to superteam_academy.json:

{ "code": 6034, "name": "LessonCountCannotDecrease", "msg": "Lesson count can only increase" }
Verified it lines up: errors.rs has 35 variants with LessonCountCannotDecrease declared last → Anchor code 6034; resolveIdlError scans by the code field, so a 6034 revert now resolves to "Lesson count can only increase" instead of "Unknown program error (code 6034)". (Couldn't anchor build-regenerate to diff — this environment's IDL build hits an ark-bn254/rustc incompatibility — so the entry was hand-matched to what anchor would emit: same declaration order, code, and message.)

2. (non-blocking) No test coverage and 3. (non-blocking) guard not verified against the deployed program — tracked together in https://github.com/solanabr/superteam-academy/issues/336, gated as P1 before the growth path is wired to any caller. Both are safe to defer right now because admin-signer only ever passes newLessonCount: null (the no-op path), so nothing exercises the guard yet. https://github.com/solanabr/superteam-academy/issues/336 covers: confirm the deployed guard/error against Thom's actual source, reconcile, add the Rust + anchor tests (revert path + complete_lesson accepting the newly-unlocked index), then complete the original https://github.com/solanabr/superteam-academy/issues/314 wiring.

4. (minor) IDL reformat re-diff on next regen — acknowledged; expect prettier noise on the next real anchor build. No action here.

Potolski:
Closing as **superseded by #314** (merged in `87fbe55`).

This PR diagnosed that the deployed devnet program had a 7th `update_course` field (`new_lesson_count`) that the repo IDL lacked — the cause of the reactivate/deactivate `InstructionDidNotDeserialize` (102). #314 has since landed the authoritative fix with everything this PR did and more:

- 7th `new_lesson_count` field + increase-only guard in the program source
- IDL field **and** error entry (named `LessonCountDecrease` on main — the authoritative name; this PR used `LessonCountCannotDecrease`)
- `admin-signer` client wiring (`newLessonCount` serialized)
- **Test coverage** this PR didn't have — `tests/rust/src/test_lesson_count.rs` + the TS integration test

Since main now matches the deployed program (7 fields), reactivate/deactivate (and course-sync updates) work again. Nothing here is worth keeping over #314's version, so closing and deleting the branch.

PR #338, opened 2026-07-07
fix(teacher): import image-by-URL server-side into Sanity (was CSP-blocked)

Potolski:
## Problem
Follow-up to #335 (merged). Adding a lesson image **by URL** didn't work: the URL path inserted a raw ``, and the lesson's tight `img-src` CSP blocks non-allowlisted hosts, so the image never rendered.

## Fix
The URL is now sent to `POST /api/teacher/upload-image` (JSON body), **fetched server-side and pinned into Sanity** — the returned `cdn.sanity.io` URL renders under the existing CSP and no longer depends on the source host staying up. Same outcome as an upload.

## Security — SSRF-guarded fetch (`lib/teacher/remote-image.ts`)
- **https-only**, no embedded credentials
- Host must resolve to a **public** address — private/loopback/link-local/CGNAT ranges (IPv4 + IPv6, incl. IPv4-mapped) are rejected via a pre-flight DNS lookup
- **Redirects refused** (`redirect: "error"`) so a 3xx can't bounce to an internal host
- Body **streamed with a hard 4 MB cap** (doesn't trust Content-Length); 10s timeout
- Same type allowlist as the file path (PNG/JPEG/WebP/GIF; **no SVG**)

Route still teacher-auth'd + per-user rate-limited. Updated the popover hint (en/pt-BR/es) to reflect that pasted URLs are imported too.

## Verification
- `tsc --noEmit` ✓, eslint ✓, locale key parity ✓

## Deploy note
The URL-import path needs outbound egress from the serverless function (default on Vercel) plus the `SANITY_ADMIN_TOKEN` write access the upload already relies on.

### DIFF
--- a/apps/web/src/app/api/teacher/upload-image/route.ts
+++ b/apps/web/src/app/api/teacher/upload-image/route.ts
@@ -4,6 +4,7 @@ import { NextRequest, NextResponse } from "next/server";
 import { authorizeTeacher } from "@/lib/teacher/authorize";
 import { isRateLimited } from "@/lib/rate-limit";
 import { uploadTeacherImage } from "@/lib/sanity/teacher-mutations";
+import { fetchRemoteImage, RemoteImageError } from "@/lib/teacher/remote-image";
 
 // Reads the caller's session + streams a file upload — never statically
 // prerender (DYNAMIC_SERVER_USAGE).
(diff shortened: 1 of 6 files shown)


Potolski:
Fixed in `f20c376` — the DNS-rebinding TOCTOU is closed by **pinning the validated IP for the actual connection**.

Instead of validating the hostname and then handing the URL to `fetch()` (which re-resolves independently when it opens the socket), the request now goes directly to the pre-resolved IP via `node:https`:

```ts
https.request({
  host: ip,                 // ← connect to the validated IP literal; Node does NOT re-resolve it
  servername: url.hostname, // ← SNI, so TLS still verifies against the real cert
  headers: { host: url.host },
  rejectUnauthorized: true,
  ...
})
```

So the socket can only ever reach the address we validated as public; a rebind on the second lookup can't happen because there is no second lookup. (Went with `node:https` rather than an undici `Agent`/`connect.lookup` since undici isn't a direct dependency and I didn't want a security path resting on a transitive one.) Redirects stay refused, size is still stream-capped, and idle sockets time out.

Verified end-to-end at runtime:
- `https://www.gstatic.com/webp/gallery/1.jpg` → **OK** (image/jpeg, TLS validated via SNI while connecting to the resolved IP)
- `https://127.0.0.1/…` → `blocked_host`
- `https://169.254.169.254/…` (cloud metadata) → `blocked_host`
- `https://localhost/…` (resolves to loopback) → `blocked_host`
- `http://…` → `not_https`; non-image → `bad_type`

tsc + eslint clean.

Potolski:
@claude Fixed in f20c376 — the DNS-rebinding TOCTOU is closed by pinning the validated IP for the actual connection.

Instead of validating the hostname and then handing the URL to fetch() (which re-resolves independently when it opens the socket), the request now goes directly to the pre-resolved IP via node:https:

https.request({
  host: ip,                 // ← connect to the validated IP literal; Node does NOT re-resolve it
  servername: url.hostname, // ← SNI, so TLS still verifies against the real cert
  headers: { host: url.host },
  rejectUnauthorized: true,
  ...
})
So the socket can only ever reach the address we validated as public; a rebind on the second lookup can't happen because there is no second lookup. (Went with node:https rather than an undici Agent/connect.lookup since undici isn't a direct dependency and I didn't want a security path resting on a transitive one.) Redirects stay refused, size is still stream-capped, and idle sockets time out.

Verified end-to-end at runtime:

https://www.gstatic.com/webp/gallery/1.jpg → OK (image/jpeg, TLS validated via SNI while connecting to the resolved IP)
https://127.0.0.1/… → blocked_host
https://169.254.169.254/… (cloud metadata) → blocked_host
https://localhost/… (resolves to loopback) → blocked_host
http://… → not_https; non-image → bad_type
tsc + eslint clean.

Issue #253, opened 2026-07-01
Buildable (Anchor-program) challenges can't be completed — deploy build server + wire server-side grading


Potolski:
## Summary

The 8 **buildable** challenges can't be completed. They currently surface **"Build server is not configured"** (503), and even once the build server is up, submitting one for XP would still 503 because the server-side completion gate has no grader for this challenge type.

These are the only challenges that compile a full on-chain Anchor program (vs. running a function against test cases), so they need the dedicated build server — there's no lighter substitute like the JS (QuickJS) or Rust (Playground) paths that already work.

**Affected challenges** (`buildType: "buildable"`):
- Your First Build
- Add an Instruction
- Define a Counter Account
- Wire Up Initialize
- Build the Increment
- Complete Counter Program
- Deploy to Devnet
- Deploy Your Program to Devnet

## Context

Challenge grading has three backends. The other two are fixed; buildable is the remaining gap:

| Type | Count | Backend | Status |
| --- | --- | --- | --- |
| JS/TS | 15 | QuickJS/WASM (in-process) | ✅ #252 |
| Rust snippet | 10 | Rust Playground (external HTTP) | ✅ #250 |
| **Buildable** | 8 | **Cloud Run build server** (`apps/build-server`, Docker `anchor build`) | ❌ this issue |

### 1. Deploy + configure the build server (infra)

`"Build server is not configured"` comes from the env guard in `/api/build-program` and `/api/deploy/[uuid]`:

```ts
if (!BUILD_SERVER_URL || !BUILD_SERVER_API_KEY) {
  return NextResponse.json({ error: "Build server is not configured" }, { status: 503 });
}
```

- Deploy `apps/build-server` to Cloud Run (`apps/build-server/deploy/setup-gcp.sh`, then `deploy.sh`; see `docs/DEPLOYMENT.md` GCP section + `deploy/HARDENING.md`).
- Set on Vercel (Production + Preview) and in `.env.local`:
  - `BUILD_SERVER_URL` = the real Cloud Run URL (no trailing slash)
  - `BUILD_SERVER_API_KEY` = must exactly match the service's `ACADEMY_API_KEY` (the `X-API-Key` it checks)

This unblocks the build/deploy panel actions (compile + download `.so`).

### 2. Wire buildable grading into the completion gate (validation)

Even with the build server reachable, `/api/lessons/complete` still fail-closes buildable submissions:

```ts
case "non_js_challenge":
[...]

PR #348, opened 2026-07-08
fix: unblock buildable-challenge deploy — CSP wss:// + resume binary cache (#253)

Potolski:
## Bug
Deploying a compiled program to devnet (the buildable "Deploy to Devnet" challenges) fails in the browser with:

> Connecting to `wss://devnet.helius-rpc.com/...` violates the following Content Security Policy directive: `connect-src ... https://devnet.helius-rpc.com ...`

The compile works, but the **deploy** step is blocked.

## Cause
`connect-src` (built by `solanaRpcSources()` in `lib/csp.ts`) listed only the **`https://`** RPC origins. But `@solana/web3.js`'s `Connection` opens a **WebSocket** to the RPC for subscriptions — including the signature-confirmation subscription used when deploying a program — so it needs the **`wss://`** origin too. Supabase already does exactly this (`https://…` + `wss://…`); the Solana sources didn't.

## Fix
`solanaRpcSources()` now emits each https RPC origin **and its `wss://` twin** (configured RPC + the two public fallbacks). Mirrored in the `next.config.mjs` static fallback for consistency.

For the configured `NEXT_PUBLIC_SOLANA_RPC_URL=https://devnet.helius-rpc.com/...`, `connect-src` now includes `wss://devnet.helius-rpc.com` (verified).

## Verification
- Generated sources include the wss origins; `tsc --noEmit` ✓, eslint ✓

Refs #253

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -52,6 +52,8 @@ const cspDirectives = [
     "https://*.supabase.co wss://*.supabase.co",
     "https://api.sanity.io https://cdn.sanity.io https://*.apicdn.sanity.io https://*.api.sanity.io wss://*.api.sanity.io https://media.sanity.io",
     "https://*.helius-rpc.com https://api.devnet.solana.com https://api.mainnet-beta.solana.com",
+    // web3.js opens a wss:// to the RPC for subscriptions (deploy confirmation).
+    "wss://*.helius-rpc.com wss://api.devnet.solana.com wss://api.mainnet-beta.solana.com",
     "https://accounts.google.com https://*.googleapis.com",
     "https://www.googletagmanager.com https://www.google-analytics.com https://*.google-analytics.com https://stats.g.doubleclick.net",
     "https://*.posthog.com https://*.sentry.io https://*.ingest.sentry.io",
--- a/apps/web/src/lib/csp.ts
+++ b/apps/web/src/lib/csp.ts
@@ -60,6 +60,11 @@ function supabaseSources(): { http: string; ws: string } {
  * `connect-src` to that exact origin instead of the broad `*.helius-rpc.com`
  * wildcard. The two public Solana endpoints stay allowed as fallbacks. If the
  * env var is malformed, keep the Helius wildcard so RPC is never blocked.
+ *
+ * Each https RPC origin is emitted with its `wss://` twin: web3.js's `Connection`
+ * opens a WebSocket to the RPC for subscriptions (e.g. signature confirmation
+ * when deploying a program), so a connect-src without the wss:// origin blocks
+ * the deploy ("violates … connect-src" on `wss://…`).
  */
 function solanaRpcSources(): string[] {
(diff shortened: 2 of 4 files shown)


Potolski:
Added a second fix to this branch (`0fefee2`) — it's the next thing you hit after the CSP fix.

**Deploy "Resume" 404** (`GET /api/deploy/<uuid> 404`): the build returns the `.so` inline and caches it client-side, but `fetchBinary` (in `@superteam-lms/deploy`) **deleted the cache entry on first read** and then fell back to `fetch(/api/deploy/:uuid)` — a route that no longer exists. So the initial deploy consumed the binary and Resume 404'd.

Fix: `fetchBinary` no longer deletes on read (resume re-reads the same binary) and no longer hits the removed route; a genuine miss now throws a clear "build expired — rebuild" error the panel already handles.

Both fixes are needed to complete a buildable "Deploy to Devnet" challenge end-to-end. deploy-pkg `tsc` + web `tsc` + eslint all clean.

Potolski:
Third fix on this branch (`3b47cad`): the chunk upload **stalling mid-batch** (e.g. stuck at 22/176 with no error).

**Cause:** chunk writes are sent with `maxRetries: 0`, and a rate-limited/free devnet RPC drops transactions under load. Nothing rebroadcast them, so a dropped chunk never landed and `confirmBatch` sat waiting (then timed out at 60s; resume restarted the same batch and stalled again).

**Fix:** `confirmBatch` now **rebroadcasts still-unconfirmed txs every ~4s** while polling (safe/idempotent within the blockhash window), and `BATCH_SIZE` drops 50→20 (with `SEND_DELAY` 30→50ms, timeout 60→90s) so the resend burst stays under the RPC's per-second limit and each batch confirms inside its blockhash. Applied to both the initial and resume upload loops.

Net: this branch now carries the three fixes needed to complete a buildable "Deploy to Devnet" end to end — CSP `wss://`, resume binary cache, and reliable chunk upload. deploy-pkg + web `tsc` + eslint clean.

Potolski:
Fourth fix on this branch (`62b8c1c`): the chunk upload stalling even though the **RPC returns 200**.

**Cause:** 200 means the sends are accepted, but the txs never *confirm* — devnet leaders drop the fee-less write txs under load. Rebroadcast alone can't beat that: it just re-sends txs that keep getting dropped.

**Fix:** add a **priority fee** (`ComputeBudgetProgram.setComputeUnitPrice`, 100k micro-lamports) to every deploy tx — buffer create, all chunk writes, and both deploy paths. A write ix burns only a few thousand CU, so it costs a tiny fraction of a lamport per tx while making inclusion reliable (this is what the Solana CLI's `--with-compute-unit-price` does).

This branch now has the four fixes to complete a buildable "Deploy to Devnet" end to end: CSP `wss://`, resume binary cache, chunk rebroadcast, and priority fees. deploy-pkg + web `tsc` + eslint clean.

Potolski:
@claude review this pr again and make sure nothing code breaking is happening. Review your initial notes and do a run again

PR #1, opened 2026-02-11
feat: Complete Superteam Academy LMS — 27 pages, all required + all bonus features

agentmila:
### 🔗 [Live Demo](https://superteam-academy.vercel.app) | [Repository](https://github.com/kukaklaudio/superteam-academy)

---

### ✅ All 10 Required Pages
| Page | Route | Status |
|------|-------|--------|
| Landing | `/` | ✅ |
| Course Catalog | `/courses` | ✅ |
| Course Detail | `/courses/[slug]` | ✅ |
| Lesson View + Code Editor | `/courses/[slug]/lessons/[id]` | ✅ (Monaco) |
| Code Challenge Interface | Integrated in Lesson View | ✅ |
| Dashboard | `/dashboard` | ✅ |
| Profile | `/profile`, `/profile/[username]` | ✅ |
| Leaderboard | `/leaderboard` | ✅ |
| Settings | `/settings` | ✅ |
| Certificates | `/certificates/[id]` | ✅ |

### ✅ Required Features
- **Auth:** Wallet Adapter (Phantom, Backpack, Solflare) + Google + GitHub
- **Gamification:** XP, leveling (floor(sqrt(xp/100))), streaks, achievements/badges
- **Code Editor:** Monaco Editor — Rust/TS/JSON, autocompletion, pass/fail
- **i18n:** PT-BR, ES, EN via next-intl
- **Themes:** Dark/Light mode
- **Analytics:** GA4, PostHog, Sentry
- **CMS:** Sanity v3
- **RBAC:** Admin, Professor, Student
- **Responsive:** Mobile-first

### ✅ Documentation
- README.md · ARCHITECTURE.md · CMS_GUIDE.md · CUSTOMIZATION.md

### 🏆 ALL 8 Bonus Features
1. **Admin Dashboard** — /admin/* (course mgmt, user analytics)
2. **E2E Tests** — Playwright, 7 suites
3. **Community/Forum** — /community (threads, replies, 4 categories)
4. **Onboarding Flow** — /onboarding (4-step wizard + skill assessment quiz)
5. **PWA Support** — Installable, offline-capable
6. **Advanced Gamification** — /challenges (daily/weekly challenges, seasonal events)
7. **CMS Course Creator** — /teach/* (editors, student analytics)
8. **Devnet Integration** — Token-2022 XP reads, Metaplex Core NFTs, enrollment tx builder

### Tech Stack
[...]

PR #362, opened 2026-07-09
fix(quests): durable quest-XP delivery + get_daily_quest_state hardening (CS-7)

### LINKED ISSUE #353: CS-7: Quest XP durable delivery + get_daily_quest_state hardening
Independent — start anytime. Ships as a `supabase/migrations/` file (applied only AFTER merge).

Fixes verified quest-XP loss (spec §14.1): (1) add `'quest_xp'` to `pending_onchain_actions` CHECK; (2) `queueFailedOnchainAction` must inspect the `.upsert()` return; (3) durable mint intent — write the pending row synchronously before the response, or await the mint before setting xp_granted (currently un-awaited fire-and-forget, `quests/daily/route.ts:124`); queue on no-wallet/no-programLive instead of dropping; (4) SQL: `targetValue>0` guard + final `ELSE RAISE EXCEPTION` (§6.3).

**Done when:** migration in PR; both gates + adversarial pass; after merge — applied, `get_advisors` clean, grants verified; failed mint leaves a durable retryable row.

thomgabriel:
## CS-7 (#353) — Quest-XP durable delivery + SQL hardening

Four verified defects in the quest-XP path. All four fixed.

### 1. `pending_onchain_actions` CHECK rejected `quest_xp`
The inline column CHECK omitted `'quest_xp'`, so every quest-XP enqueue violated
the constraint. Combined with defect #2, the failure was invisible. Added
`'quest_xp'` to the allowed set in `supabase/schema.sql` and in the migration via
an idempotent **drop-and-readd**: a `DO` block looks up the auto-generated
constraint name from `pg_constraint` (the baseline created it unnamed) and drops
it, then re-adds a **named** `pending_onchain_actions_action_type_check`. Re-runnable.

### 2. `queueFailedOnchainAction` swallowed the upsert error
supabase-js `.upsert()` **resolves with `{ error }`** instead of throwing, so the
old `try/catch` never fired and a constraint violation was silently dropped. It now
inspects the returned `error`, logs it (Sentry), and **throws** so a failed enqueue
is visible. The `error` param is now optional (present → real failure records
`last_error`+`failed_at`; absent → proactive durable enqueue).

### 3. Quest mint was un-awaited fire-and-forget (the XP-loss bug)
`mintQuestXpOnChain(...).catch(()=>{})` ran after the response returned with no
`waitUntil`/`after()`, so on Vercel it raced lambda freeze — and it had two silent
early returns (`!programLive`, `!wallet`) that dropped the XP **after
`get_daily_quest_state` had already set `xp_granted=true`** ("attempt authorized"),
making it permanent XP loss.

### 4. `get_daily_quest_state` hardening
- **(a)** Completion now requires `targetValue > 0` (a `targetValue:0` quest can no
  longer auto-complete every day and mint free XP). Centralized in a new `v_completed`
  local reused by both the login_streak and generic branches.
- **(b)** A final `ELSE RAISE EXCEPTION` on the quest-type IF/ELSIF chain — an unknown
  type now fails loudly instead of silently rendering 0/N.
- The login_streak state machine is untouched.

---

## Durability approach: **(a) — synchronous durable enqueue**

I chose **(a)**: write the `pending_onchain_actions` row **synchronously, before the
response returns** (now legal after fix #1), and let the existing queue processor
(`retryPendingOnchainActions`, driven on the user's next auth) perform the actual mint.

**Why (a) over (b) — await the mint:**
- **Smaller, safer diff** — it deletes `mintQuestXpOnChain` (wallet lookup +
  program-liveness gate + inline mint) rather than restructuring it. The
  no-wallet / no-programLive cases are handled for free: the row is durable and the
  queue processor resolves the wallet and checks liveness itself, so there is no
  branch left that can "return empty-handed".
- **No added request latency / no freeze risk** — (b) would put a Solana
  confirmation on the critical path of a GET that fires on every dashboard load;
  a network error there would still need the queue as a fallback anyway.
- **Matches the existing architecture** — the retry queue is this codebase's
  durability primitive and already handles `quest_xp`.

**One correctness detail:** `reference_id` is now day-scoped
(`${questId}:${YYYY-MM-DD}` UTC, matching `CURRENT_DATE`). Without the day suffix,
day-N's enqueue would upsert-collide with a prior day's **already-resolved** row (the
upsert doesn't reset `resolved_at`) and the mint would be silently dropped — exactly
the class of bug being fixed. `justAwarded` fires once per (user, quest, DB-day), so
there is exactly one enqueue per period.

**Trade-off (documented):** delivery is deferred to the next auth-triggered queue
sweep rather than immediate. No XP is ever lost (the intent is durable); the Helius
webhook then syncs the on-chain mint back to Supabase. This is acceptable and
consistent with the existing queue design.

---

## Migration — apply **post-merge only**

`supabase/migrations/20260709120000_quest_xp_durable_delivery.sql` — **NOT applied to
any database.** Header says apply to prod only after merge. Function body is
byte-identical to the updated `supabase/schema.sql` (canonical). No explicit
BEGIN/COMMIT (matches repo convention — the CLI wraps each migration).

### Rollback (tested; full block kept commented in the migration)
```sql
-- Restore original CHECK (drops 'quest_xp'); resolve/remove any quest_xp rows first:
--   DELETE FROM public.pending_onchain_actions WHERE action_type = 'quest_xp';
DO $$
DECLARE v_conname text;
BEGIN
  SELECT conname INTO v_conname
  FROM pg_constraint
  WHERE conrelid = 'public.pending_onchain_actions'::regclass
    AND contype = 'c'
    AND pg_get_constraintdef(oid) ILIKE '%action_type%';
  IF v_conname IS NOT NULL THEN
    EXECUTE format('ALTER TABLE public.pending_onchain_actions DROP CONSTRAINT %I', v_conname);
  END IF;
END $$;
ALTER TABLE public.pending_onchain_actions
  ADD CONSTRAINT pending_onchain_actions_action_type_check
  CHECK (action_type IN ('achievement','certificate','course_finalize','xp','enroll'));
-- + CREATE OR REPLACE FUNCTION get_daily_quest_state(...) with the PRIOR body
--   (no targetValue guard, no unknown-type RAISE) and re-assert REVOKE/GRANT.
--   The full prior function body is in the migration's rollback block.
```

---

## Local verification

`pnpm --filter web typecheck` → **exit 0** (clean):
```
> tsc --noEmit
=== EXIT CODE: 0 ===
```

`pnpm --filter web test` → **16 files / 138 tests passed** (no quest-route or
onchain-queue test exists; ran the full suite for regressions):
```
 Test Files  16 passed (16)
      Tests  138 passed (138)
```

ESLint on the two changed app files → **0 errors** (only 3 pre-existing
import/order warnings in `onchain-queue.ts`, in an import block I did not touch).
Pre-commit hook (eslint --fix + prettier) also passed.

> Sensitive PR (supabase migration + `area:db`) — written defensively; migration is
> idempotent and the rollback is genuinely correct.

Closes #353

### DIFF
--- a/apps/web/src/app/api/quests/daily/route.ts
+++ b/apps/web/src/app/api/quests/daily/route.ts
@@ -1,10 +1,8 @@
 import { NextResponse } from "next/server";
-import { PublicKey } from "@solana/web3.js";
 import type { DailyQuest } from "@superteam-lms/types";
 import { createClient } from "@/lib/supabase/server";
 import { createAdminClient } from "@/lib/supabase/admin";
 import { getAllQuests } from "@/lib/sanity/queries";
-import { rewardXp, isOnChainProgramLive } from "@/lib/solana/academy-program";
 import { queueFailedOnchainAction } from "@/lib/solana/onchain-queue";
 import { nextMidnightUtc } from "@/lib/gamification/daily-reset";
 import { logError } from "@/lib/logging";
@@ -116,14 +114,21 @@ export async function GET() {
       };
     });
 
-    // 6. Mint XP on-chain for newly-completed quests (fire-and-forget)
+    // 6. Durably record the on-chain XP mint intent for newly-completed quests.
+    //
+    // get_daily_quest_state already flipped xp_granted=true ("attempt
+    // authorized"), so a dropped mint is PERMANENT XP loss. The mint therefore
+    // must not be fire-and-forget: on Vercel the lambda freezes once the
+    // response returns, killing any un-awaited work. Instead we write the
+    // pending_onchain_actions row SYNCHRONOUSLY (durable) BEFORE responding;
+    // retryPendingOnchainActions() — driven on the user's next auth — performs
+    // the actual on-chain mint (it resolves the wallet and checks program
+    // liveness itself, so no wallet / program-live gating is needed here).
     const newlyAwarded = progressRows.filter(
       (r) => r.justAwarded && r.xpReward > 0
     );
     if (newlyAwarded.length > 0) {
-      mintQuestXpOnChain(user.id, newlyAwarded, admin).catch(() => {
-        // Errors are logged + queued inside mintQuestXpOnChain
-      });
+      await enqueueQuestXpMints(user.id, newlyAwarded);
     }
 
     return NextResponse.json({
@@ -140,56 +145,43 @@ export async function GET() {
 }
 
 // ---------------------------------------------------------------------------
-// Fire-and-forget: mint quest XP tokens on-chain
+// Durable enqueue: record quest-XP mint intents for the retry queue
 // ---------------------------------------------------------------------------
 
-async function mintQuestXpOnChain(
+async function enqueueQuestXpMints(
   userId: string,
-  awarded: Array<{ questId: string; xpReward: number }>,
-  admin: ReturnType<typeof createAdminClient>
+  awarded: Array<{ questId: string; xpReward: number }>
 ): Promise<void> {
-  // Check if program is live before attempting on-chain calls
-  const programLive = await isOnChainProgramLive();
-  if (!programLive) return;
+  // Reference_id is scoped to the UTC day (matching CURRENT_DATE, which
+  // get_daily_quest_state uses for period_start on the Supabase UTC instance).
+  // Without the day suffix, day-N's enqueue would upsert-collide with a prior
+  // day's already-resolved row and the mint would be silently dropped.
+  const periodKey = new Date().toISOString().slice(0, 10);
 
-  // Resolve the user's wallet address
-  const { data: profile } = await admin
-    .from("profiles")
-    .select("wallet_address")
-    .eq("id", userId)
-    .single();
-
-  if (!profile?.wallet_address) return;
-
-  const wallet = new PublicKey(profile.wallet_address);
-
-  for (const quest of awarded) {
-    const memo = `daily_quest:${quest.questId}`;
-    try {
-      await rewardXp(wallet, quest.xpReward, memo);
-    } catch (err) {
-      logError({
-        errorId: ERROR_IDS.XP_REWARD_FAILED,
-        error: err instanceof Error ? err : new Error(String(err)),
-        context: {
-          handler: "mintQuestXpOnChain",
+  // Enqueue every award; a failed enqueue means the durable intent was NOT
+  // recorded (xp_granted is already true → XP-loss case), so surface it loudly
+  // per-quest rather than letting one failure abort the others.
+  await Promise.all(
+    awarded.map(async (quest) => {
+      try {
+        await queueFailedOnchainAction(
           userId,
-          questId: quest.questId,
-          xpReward: quest.xpReward,
-        },
-      });
-
-      await queueFailedOnchainAction(
-        userId,
-        "quest_xp",
-        quest.questId,
-        {
-          xpAmount: quest.xpReward,
-          memo,
-          walletAddress: profile.wallet_address,
-        },
-        err instanceof Error ? err.message : String(err)
-      );
-    }
-  }
+          "quest_xp",
+          `${quest.questId}:${periodKey}`,
+          { xpAmount: quest.xpReward, memo: `daily_quest:${quest.questId}` }
+        );
+      } catch (err) {
+        logError({
+          errorId: ERROR_IDS.XP_REWARD_FAILED,
+          error: err instanceof Error ? err : new Error(String(err)),
+          context: {
+            handler: "enqueueQuestXpMints",
+            userId,
+            questId: quest.questId,
+            xpReward: quest.xpReward,
+          },
+        });
+      }
+    })
+  );
 }
--- a/apps/web/src/app/api/quests/daily/route.ts
+++ b/apps/web/src/app/api/quests/daily/route.ts
@@ -3,6 +3,7 @@ import type { DailyQuest } from "@superteam-lms/types";
 import { createClient } from "@/lib/supabase/server";
 import { createAdminClient } from "@/lib/supabase/admin";
 import { getAllQuests } from "@/lib/sanity/queries";
+import { retryQuestXpForUser } from "@/lib/solana/onchain-queue";
 import { nextMidnightUtc } from "@/lib/gamification/daily-reset";
 
 // Auth/cookie + per-request DB access — never statically prerender (DYNAMIC_SERVER_USAGE).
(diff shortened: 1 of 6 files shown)


thomgabriel:
## Adversarial review — all three blocking defects fixed (`e2385d2`)

### D3 (HIGH) — double-mint of soulbound XP → **idempotent, no on-chain re-mint**
The `quest_xp` queue case wrapped a **non-idempotent** on-chain `rewardXp` in `withRetry(…,3)` with no receipt guard, so a confirmation-timeout retry (within a sweep, or across sweeps while `retry_count<5`) minted **permanent soulbound** XP multiple times.

Fix: `quest_xp` now credits via **`award_xp` with `reference_id` as `p_idempotency_key`** (mirrors the lesson `xp` case). `award_xp` uses `ON CONFLICT (user_id, idempotency_key) DO NOTHING`, so a retry after a landed award is a true no-op. There is **no on-chain re-mint** in the retry path — the only way to make a soulbound mint exactly-once without program-level receipts. XP still lands in `user_xp`/`xp_transactions` (capped, streak-aware). Processed in a **wallet-independent first pass** so Google-only users also get quest XP — this completes #353 fix #3 ("queue instead of returning empty-handed") which the old on-chain path silently dropped for wallet-less users.

`rewardXp` now has no frontend callers (dormant SDK export; left in place). The `XpRewarded` webhook handler is unaffected — no on-chain quest mint means no `XpRewarded` event, so no double DB-credit.

### D2 (MEDIUM) — swallowed enqueue → **transactional enqueue inside `get_daily_quest_state`**
**Choice: the transactional fix (not "propagate").** Propagating alone can't prevent loss — the `xp_granted` flip already committed inside the RPC, so a subsequent 500 doesn't un-flip it, and the client's retry sees `justAwarded=false` (permanent loss). So I took the reviewer's **ideal** path: the `INSERT INTO pending_onchain_actions` now happens **inside `get_daily_quest_state`, in the same transaction as the `xp_granted` flip** (both `IF FOUND` branches). A quest can never be marked granted without a durable delivery row — atomic by construction. The route's swallow-prone `enqueueQuestXpMints` and the unused `queueFailedOnchainAction` helper are removed. This also eliminates the reviewer's UTC-`periodKey` vs `CURRENT_DATE` non-blocking note: `reference_id` is built from the same `v_period` that scopes `xp_granted`.

### D1 (HIGH) — deploy ordering → **migration must be applied BEFORE/atomically with the deploy**
Migration header rewritten with a `⚠️ MERGE-ORDER REQUIREMENT` banner: **apply this migration to prod FIRST, then merge** (merge auto-deploys `main`). The new `get_daily_quest_state` (which enqueues `quest_xp`) and the CHECK that legalizes it ship in the **same migration**, so there is no window where the function writes `quest_xp` against the old CHECK. If code went live before the migration, the *old* function would flip `xp_granted` without enqueuing → silent loss; hence migration-first, never after.

### Non-blocking items
- **Unknown quest type**: changed from `RAISE EXCEPTION` to **skip + `RAISE WARNING` + `CONTINUE`**. Rationale: now that the enqueue is transactional, a `RAISE` would 500 the endpoint for *every* user and roll back *other* quests' durable XP rows on a single CMS typo. Skip+warn contains the blast radius to the one bad quest (logged for the operator, not rendered as a completable 0/N award).
- **Rollback `DELETE`**: promoted to a prominent, clearly-labeled first step (no longer buried in a nested `-- --` aside).
- `schema.sql` mirrors the migration's function **byte-for-byte** (verified via diff).
- Pre-existing sibling bug noted for a follow-up (out of scope): `lib/helius/event-handlers.ts::queueFailedAction` has the same `.upsert()`-in-`try/catch` swallow, but it only enqueues action types already in the CHECK, so it never hit the `23514` that `quest_xp` did.

### Verification
```
pnpm --filter web typecheck   →  tsc --noEmit,  EXIT 0
pnpm --filter web test        →  Test Files 16 passed (16),  Tests 138 passed (138)
eslint (changed files)        →  0 errors  (import-order autofixed by pre-commit)
```
SQL not executed against any DB (psql unavailable locally; migration is apply-post-merge by the orchestrator per the header). Dollar-quoting verified balanced; forward function == `schema.sql`.

thomgabriel:
## Adversarial gate: PASS (after fixes)

The independent adversarial review found 3 blocking defects (2 HIGH); all three are fixed in `e2385d2` and re-verified by reading the diff:

- **D1** ✓ migration header now mandates apply-BEFORE-deploy (⚠️ merge-order banner).
- **D2** ✓ enqueue moved INTO `get_daily_quest_state`, transactional with the `xp_granted` flip, `ON CONFLICT DO NOTHING`, both branches. `queueFailedOnchainAction` removed (typecheck exit 0 confirms no orphaned callers).
- **D3** ✓ quest XP credited via `award_xp` with `p_idempotency_key = reference_id` — retries are no-ops; the un-guarded `withRetry(rewardXp)` double-mint path is gone.

**⚠️ Design note for the human merger — please confirm:** the D3 fix changes quest XP from an on-chain `rewardXp` mint to the idempotent `award_xp` ledger path (same path the generic `xp` queue action already uses). It lands in `user_xp` (read by `get_leaderboard`/`public_user_xp`), so totals + leaderboard stay consistent, and wallet-less users now get quest XP. This is a venue change (on-chain → off-chain ledger) introduced to solve the double-mint — consistent with existing patterns, but confirm it matches intended quest-XP semantics before merge.

**Merge sequencing (D1):** apply `supabase/migrations/20260709120000_quest_xp_durable_delivery.sql` to prod FIRST, then merge (prod auto-deploys from main).

CI re-running on the new commit. Gate cleared; awaiting merge authorization (blocked by review-bot defect #363).

thomgabriel:
**Merge is human-gated on prod-migration sequencing (D1), not just the review bot.** CI is green and the adversarial gate passed, but per finding D1 the migration `20260709120000_quest_xp_durable_delivery.sql` MUST be applied to prod (obqlljsagzslxarwphxv) **before** this merges — `main` auto-deploys on merge, and the new route enqueues `quest_xp` which only the new CHECK permits. I am not applying a production migration unattended. To ship: (1) apply the migration to prod, (2) verify `get_advisors` clean + grants intact, (3) then merge this PR.

thomgabriel:
@claude re-review

thomgabriel:
@claude re-review

thomgabriel:
## Independent adversarial review — 2 BLOCKING findings (do not merge yet)

**B1 — No delivery driver on the quest path.** `retryPendingOnchainActions()` is the sole consumer of `quest_xp` rows and it only runs from the two auth routes (wallet:207, callback:137). A user with a live session never re-auths → their enqueued quest XP is never swept; leaderboard/XP frozen until session expiry. The intent is durable but nothing delivers it.

**B2 — Cap-dropped credits marked delivered.** `award_xp` returns void with no error when the 5000/day cap eats the credit (schema.sql:378-389), and the queue marks `resolved_at` on no-error (onchain-queue.ts:74-95). Batched delivery on one login day + cap → silently lost XP recorded as success — the exact bug class this PR exists to eliminate.

Non-blocking: N1 transition-window double-award (apply→deploy gap; different idempotency namespaces), N2 `payload.xpAmount as number` load-bearing cast (null → resolved with 0 XP, no diagnostic), N3 quest XP shares the lesson 5000/day bucket, I2 rows stranded after 5 retries with no alerting.

What held: day-2 reference_id collision fixed (`questId:period`), concurrent enqueue/credit races closed (row-lock + advisory lock + idempotency index), targetValue guard fail-closed, CHECK migration idempotent w/ rollback, unknown-type RAISE WARNING+CONTINUE, grants intact.

Fix incoming on this branch.

PR #365, opened 2026-07-09
docs: fix 3 live doc landmines (CS backlog)

### LINKED ISSUE #354: docs: fix 3 live doc landmines (prefix-strip + false token claim)
SAFE docs-only (spec §16.4): (1) `CMS_GUIDE.md:224` + `CUSTOMIZATION.md:479` tell readers to strip the `achievement-` prefix from UNLOCK_CHECKS keys — real keys are the full `_id`; following it reproduces the wrong-PDA bug. (2) `SECRET-ROTATION.md:47` claims SANITY_API_TOKEN guards 'hidden test answer keys' — dataset is public, no token needed; `:9`/`:20` name the wrong prod host.

**Done when:** all three corrected on a loop branch, PR merged.

thomgabriel:
SAFE docs-only. Fixes the three landmines in spec §16.4, each verified against `main`:

1. **`CMS_GUIDE.md:224` + `CUSTOMIZATION.md:479`** told readers to strip the `achievement-` prefix from UNLOCK_CHECKS keys. **Wrong** — the actual map keys carry the full prefix (`"achievement-first-steps"`) and the lookup is `UNLOCK_CHECKS[def.id]` with the raw Sanity `_id`. Following the old instruction reproduces the getDeployedAchievements() wrong-PDA silent failure. Corrected to state the full `_id`.
2. **`SECRET-ROTATION.md:47`** claimed SANITY_API_TOKEN guards 'hidden test answer keys' — the dataset is public, no token needed to read content. Corrected.
3. **`SECRET-ROTATION.md` prod host** — added a non-destructive correction note (canonical prod = superteam-academy-web.vercel.app / stbr-true, not solarium.courses) rather than rewriting the ratified D-2 decision text.

Follow-up flagged for CS-6: the `achievements.ts` docstring repeats the same misleading 'minus the prefix' claim and should be fixed when that file is rewritten.

Closes #354

### DIFF
--- a/docs/CMS_GUIDE.md
+++ b/docs/CMS_GUIDE.md
@@ -221,7 +221,7 @@ Achievements follow a similar pattern:
 3. Admin panel writes back via `writeAchievementOnChainStatus(sanityId, achievementPda, collectionAddress)` -- sets status to `"synced"`
 4. The achievement becomes mintable. The GROQ query `getDeployedAchievements()` filters to `defined(onChainStatus.achievementPda)`
 
-Unlock logic is defined in `apps/web/src/lib/gamification/achievements.ts` in the `UNLOCK_CHECKS` map. Achievement IDs in this map must match Sanity `_id` values minus the `"achievement-"` prefix.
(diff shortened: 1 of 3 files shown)


thomgabriel:
@claude re-review

thomgabriel:
@claude re-review

Issue #357, opened 2026-07-09
CS-5: Sanity schema v2 (Plan)


thomgabriel:
**Plan-first.** Blocked by CS-1 (#351). Sanity schema v2 (spec §10): lesson = ordered `blocks[]` (page-builder array of typed objects, Amendment A widgets); module → inline object on course; weak references (course→lessons, course→instructor); drop authoringStatus/author/reviewFeedback; read-only Studio (strip document.actions + newDocumentOptions; Viewer roles are ops). Deletes approveCourse/rejectCourse/getPendingReviewCourses/getChallengeAnswerKey/answerKeyProjection/queries-answer-leak.test.ts.

**Done when:** Plan merged; schema compiles; `sanity typegen` clean over blocks[]; GROQ collapses to one literal `blocks[]{...}`.

PR #367, opened 2026-07-09
docs(sanity): CS-5 Sanity schema v2 implementation plan

thomgabriel:
## CS-5: Sanity schema v2 — implementation plan (docs only)

Adds `docs/superpowers/plans/2026-07-09-sanity-schema-v2.md`, the task-by-task plan for reshaping the Sanity schemas and read layer to the block model (spec §10). **No schema or app code changes** — plan document only.

### What the plan specifies (all 8 issue items)
1. **Lesson = ordered `blocks[]`** — 8 first-class Sanity object types (`prose`, `video`, `code`, `quiz`, `openEnded` + the 3 widgets per Amendment A), registered in one `blocks/index.ts` registry; fields mirror the CS-1 Zod shapes 1:1.
2. **Module → inline object** on the course; the standalone `module` document is deleted.
3. **Weak references** — `course.instructor`, `courseModule.lessons[]`, plus `prerequisiteCourse` and `learningPath.courses[]` (same §9.5 rationale, so pruning never deadlocks).
4. **Drop `authoringStatus` / `author` / `reviewFeedback`** from course; `creator.githubId` replaces `author`.
5. **Read-only Studio** — strip `document.actions` + `newDocumentOptions` in both configs (Viewer-role enforcement noted as ops, not schema).
6. **Delete answer-key machinery** — `getChallengeAnswerKey`, `getChallengeAnswerKeyById`, `answerKeyProjection`, `queries-answer-leak.test.ts`; the `code` grader reads `solution`/`tests` from the same public projection.
7. **GROQ collapse** — one literal `blocks[]{…}` lesson projection; `sanity typegen` types it cleanly.
8. **`sync: { source, rev }` prune marker** on every managed doc (field name not `_`-prefixed).

### Notes
- Scoped to spec §15.4 **Phase 5 (Sanity schema)** plus the two read-layer items the issue assigns to CS-5 (GROQ collapse + answer-key deletion). The plan names the one cross-phase coupling: the answer-key deletion's grader consumers are rewired in **Phase 6 (App)** and must land with Task 7 for `pnpm --filter web typecheck` to pass.
- Live Sanity **data** is untouched here; the document reshape is the **CS-9 sync**.
- Studio-side `sanity schema extract` + `sanity typegen` are wired as the per-task verification harness.

Refs #357

### DIFF
--- a/docs/superpowers/plans/2026-07-09-sanity-schema-v2.md
+++ b/docs/superpowers/plans/2026-07-09-sanity-schema-v2.md
@@ -678,6 +678,19 @@ git commit -m "feat(sanity): lesson document is an ordered blocks[] page builder
 
 Modules are inline objects — never reused across courses (spec §10.1). Display order = array position, so `courseModule` has no `order` field.
 
+**Coupling note (PR #367 review finding).** A live feature is built on the three dropped
+fields: the teacher course builder + admin review queue (`api/teacher/courses/**`,
+`api/admin/courses/review/route.ts`, `lib/sanity/admin-mutations.ts` + `teacher-mutations.ts`,
+`components/admin/course-review-queue.tsx`, `components/teacher/course-form.tsx`, and
+`packages/types/src/course.ts`'s `author`/`authoringStatus`/`reviewFeedback`). These use raw
+GROQ/mutations, not the generated schema types, so this task's typecheck will NOT catch them —
+they keep writing fields the schema no longer declares. **Intentional**: spec D1/§10.1 retires
+the `/teach` authoring surface and CS-10 (Phase 8) deletes the feature; until then it is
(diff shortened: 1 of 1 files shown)


thomgabriel:
Noted for CS-6 sequencing: this plan's Task 7 deletes the answer-key exports (`getChallengeAnswerKey`, `answerKeyProjection`, etc.), which breaks the grader consumers (`lib/challenge/validate.ts` + `/api/lessons/validate-challenge` + `/api/lessons/complete` + executor tests). CS-6's app plan must rewire those to read `solution`/`tests` from the public projection in the SAME PR, or split the deletion into CS-6 so `typecheck` never breaks on `main`. Flagging so CS-6 is planned after CS-1+CS-5 land.

thomgabriel:
@claude re-review

thomgabriel:
@claude re-review

PR #384, opened 2026-07-10
feat(content+app): CS-5 Sanity blocks[] schema + CS-6 gate inversion, grader/renderer registries, declarative achievements

thomgabriel:
One atomic PR for **CS-5** (Sanity schema v2) and **CS-6** (app block registry + gate inversion + declarative achievements). They cannot be split: the lesson changing from typed fields to `blocks[]` breaks every GROQ query, grader, and renderer at once, and the answer-key deletion spans both — `main` only stays green if they land together. Built on origin/main, which carries content-schema (CS-1), the program v2 source, and #346's `check-seal.ts`.

## Stage A — CS-5 Sanity schema + read layer

- **Block object schemas** — one first-class Sanity type per block (`prose`, `video`, `code`, `quiz`, `openEnded` + the 3 widgets, Amendment A) in a `blocks/index.ts` registry that the lesson `blocks[]` spreads into `of`. `testCase` object has **no `hidden`** (D4). Field names mirror content-schema 1:1.
- **Lesson → ordered `blocks[]`**; **module document deleted** → inline `courseModule` object on the course.
- **Weak refs**: `course.instructor`, `course.prerequisiteCourse`, `learningPath.courses[]`, `courseModule.lessons[]`.
- Drops `author`/`authoringStatus`/`reviewFeedback` from the schema (the `/teach` feature keeps its own local types — degraded-but-harmless until CS-10, per the plan); adds `creator.githubId`, a `sync` prune marker on all 6 managed docs, and a declarative `award` object on `achievement`.
- **Read-only Studio** — `document.actions`/`newDocumentOptions` → `[]` in both configs.
- **GROQ collapse**: one literal `blocks[]{…}` lesson projection (public `solution`/`tests`); inline-module catalog queries; `count(modules[].lessons[])`. **Deleted** `getChallengeAnswerKey`/`getChallengeAnswerKeyById`/`answerKeyProjection`/`ChallengeAnswerKey` + `queries-answer-leak.test.ts`. Added ungated `getLessonByIdForGrading` for the gate.
- **Quest couplings (§15.4a)**: `challengeLessonIds` via `count(blocks[_type=="code"])>0`; `moduleLessonMap` rebuilt from inline `course.modules[]` (`courseId:moduleKey`).
[...]

PR #389, opened 2026-07-10
feat(sync): CS-9 content sync route + drift UI (stacked on #384)

### LINKED ISSUE #360: CS-9: Content sync route (Plan)
**Plan-first.** Blocked by CS-5 (#357), CS-6 (#358), CS-8. `POST /api/admin/content/sync` (spec §9/§11): GITHUB_TOKEN tarball-at-SHA fetch → re-validate whole tree with @superteam-lms/content-schema → createOrReplace + PRESERVE([onChainStatus]) + prune guards (sync.rev marker, write-verify-then-prune, 20% blast-radius abort) + weak refs → write content_tx_id commitment → revalidateTag. contentSync drift singleton + three-way drift UI; **refuse to sync a red-CI commit** (GitHub Checks API).

**Done when:** Plan merged; sync is idempotent (re-run = no-op); prune never touches unmarked docs; drift UI shows up_to_date/behind/never_synced/blocked.

thomgabriel:
## CS-9 — Content Sync Route + Drift UI (Phase 7 cutover tooling)

**Stacked PR.** Base = `loop/feat-cs5-cs6-app-blocks-09-07-2026` (PR #384), because CS-9 consumes CS-5/CS-6's `blocks[]` Sanity schema + projection. Review/merge #384 first; this diff shows only the CS-9 delta on top of it.

Implements `docs/superpowers/plans/2026-07-09-content-sync-route.md` task-by-task (spec §9 tarball-at-SHA sync, §10 Sanity projection, §11 drift + `content_tx_id` commitment §11.0), plus the two #370-review corrections recorded on #360:
1. `AssetResolver`/`resolveAssetRef` returns a **CDN url string** (not a Sanity asset ref) — implemented and regression-tested.
2. Task 14 (`runContentSync`) carries **no** `MaskMismatchError → 409` branch — that error only arises in `/api/admin/courses/sync` (Task 13). Omitted.

### HARD constraints honored
- **No live writes / no real repo fetch.** Everything is built + unit-tested against fixtures (in-memory tarball builder, `InMemoryGateway`, fake `GitHubClient`, injectable graders). The live RUN stays the human-gated cutover.
- Strict TS, zero `any`. Panel strings via next-intl (`adminContentSync`, added to en / pt-BR / es with identical key structure). Server-only secrets read through `@/lib/env.server`.

### Tasks done (1–16)
| # | Task | Status |
|---|------|--------|
| 1 | Shared types + `GITHUB_TOKEN` env | ✅ |
| 2 | Tarball → `RepoTree` (excl `_template`) | ✅ (+ ustar `prefix`/pax long-path support — see note) |
| 3 | `GitHubClient` (tarball, HEAD, Checks API) | ✅ |
| 4 | Two-sided executor gate (§6.2a) | ✅ |
| 5 | `parseAndValidateTree` — Zod re-validation + gate | ✅ |
| 6 | Projector → `SanityDoc[]` (weak refs, CDN rewrite) | ✅ |
| 7 | Content-derived asset ids, dedupe, md rewrite | ✅ |
| 8 | PRESERVE reattach + field-coverage assert; prune guards | ✅ |
| 9 | `content_tx_id` commit + mask-vs-lockfile assert (§11.0) | ✅ |
| 10 | Three-way drift model | ✅ |
| 11 | `SanityGateway` seam over `admin-mutations` | ✅ |
| 12 | `runContentSync` orchestrator (idempotency, red-CI refusal, blast-radius, real-dims assets) | ✅ |
| 13 | `content_tx_id` + mask wired into chain-sync signer/route | ✅ (mask on-chain write gated on program v2; see note) |
| 14 | `POST /api/admin/content/sync` | ✅ |
| 15 | `GET /api/admin/content/drift` + three-way panel | ✅ |
| 16 | Docs (GITHUB_TOKEN, trigger, drift UI) | ✅ |

### Notes / deviations
- **Program v2 not deployed yet.** The current IDL has `Course.content_tx_id` (writable via `update_course.new_content_tx_id`) but **no** `active_lessons` field. Per the plan's prerequisite note, the pure §11.0 helpers (`buildCourseCommit`, `padContentTxId`, `deriveActiveMask`, `assertMaskMatchesLockfile`) land and are fully tested; the SHA is committed on-chain; the mask **assertion** runs at chain-sync time (MaskMismatchError → 409), but the mask on-chain **write** is gated until program v2 exposes the field.
- **Real GitHub tarball hardening.** The plan's `extractTarball` only read the 100-byte tar `name` field; real repo paths under the generated `owner-repo-<40-char-sha>/` top dir routinely exceed 100 bytes, so `git archive` splits them across the ustar `prefix` field (and pax `x` headers). Left as-is this would silently drop deep files at cutover. `extractTarball` now reconstructs from `prefix` + pax, with a dedicated long-path round-trip test.
- Chain-sync mask assertion is wired **additively** into `/api/admin/courses/sync`: only active when the request carries `activeLessons`; absent → the existing route behaves exactly as before (backward compatible).
- `vitest.setup.ts` gained safe (`??=`) test-env defaults so server/public env modules validate under vitest (same pattern the file already used for `NEXT_PUBLIC_PROGRAM_ID`).

### Verification (pasted)
```
pnpm --filter web typecheck   → exit 0
pnpm --filter web test        → 47 files, 331 tests passed
  content-sync suite: 82 tests across 16 files
```
Load-bearing outcomes, each proven by a named test:
- **PRESERVE + prune** — `prune.test.ts` "NEVER selects an unmarked doc"; `preserve.test.ts` carries `onChainStatus` across a re-sync + field-coverage assert; `sync.test.ts` "aborts before any delete when the prune set exceeds 20%" (nothing deleted), "prunes a small stale set and reports the count".
- **Idempotency** — `sync.test.ts` "second run at the same sha writes and deletes nothing".
- **Refuse-on-red-validation** — `drift.test.ts` `assertCommitSyncable` throws on failure/pending; `sync.test.ts` "refuses to sync a commit whose CI is red" (nothing written).
- **content_tx_id stamping** — `content-commit.test.ts` (20→32 pad, HEAD equality); `signer-commit.test.ts` packs the sha into 32 bytes.
- **mask-vs-lock PASS and FAIL** — `content-commit.test.ts` `assertMaskMatchesLockfile` passes on a matching mask, throws `MaskMismatchError` on an arbitrary bit; `signer-commit.test.ts` throws (naming the course) when the caller's mask disagrees with the lockfile.

### Open question
Program v2 (`Course.active_lessons` + `update_course.new_active_lessons`, CS-3) is the one remaining prerequisite before the mask can actually be written on-chain at cutover — everything else (assertion, SHA commitment, drift, panel) is live and tested.

Closes #360

### DIFF
--- a/apps/web/CLAUDE.md
+++ b/apps/web/CLAUDE.md
@@ -70,6 +70,11 @@ NEXT_PUBLIC_XP_MINT_ADDRESS=       # XP mint pubkey (from initialize.ts output)
 ADMIN_SECRET=                      # Admin panel authentication secret (HMAC-signed cookies)
 BUILD_SERVER_URL=                  # Cloud Run build server URL (server-only, proxied via /api)
 BUILD_SERVER_API_KEY=              # Build server authentication key
+GITHUB_TOKEN=                      # Fine-grained READ token for solanabr/academy-courses
+                                   # (server-only). Powers POST /api/admin/content/sync (tarball
+                                   # fetch), the drift UI (HEAD polling), and the Checks API
+                                   # (blocked state). Unauthenticated GitHub is 60 req/hr per IP
+                                   # and flakes on Vercel; unset → the /api/admin/content/* routes 503.
 HELIUS_API_KEY=                    # Helius key for webhook management + DAS API (lib/helius)
 HELIUS_WEBHOOK_SECRET=             # Helius webhook signature verification
 BACKEND_SIGNER_SECRET=             # Rotatable backend co-signer keypair (completeLesson etc.)
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -23,6 +23,7 @@
     "@metaplex-foundation/umi-web3js-adapters": "^1.5.1",
     "@monaco-editor/react": "^4.7.0",
     "@noble/hashes": "^1.8.0",
+    "image-size": "^1.2.1",
     "@phosphor-icons/react": "^2.1.10",
     "@radix-ui/react-avatar": "^1.1.0",
     "@radix-ui/react-dialog": "^1.1.0",
@@ -71,6 +72,7 @@
     "tailwind-merge": "^2.4.0",
     "tailwindcss-animate": "^1.0.7",
     "tweetnacl": "^1.0.3",
+    "yaml": "^2.8.2",
     "zod": "^4.4.3"
   },
   "devDependencies": {
--- a/apps/web/src/app/[locale]/admin/admin-client.tsx
+++ b/apps/web/src/app/[locale]/admin/admin-client.tsx
@@ -8,6 +8,7 @@ import {
 } from "@/components/admin/course-review-queue";
 import { AchievementSyncTable } from "@/components/admin/achievement-sync-table";
 import { DataResyncPanel } from "@/components/admin/data-resync-panel";
+import { ContentSyncPanel } from "@/components/admin/content-sync-panel";
 import { TeacherRolesPanel } from "@/components/admin/teacher-roles-panel";
 import { LearningPathsPanel } from "@/components/admin/learning-paths-panel";
 import { CourseTagsPanel } from "@/components/admin/course-tags-panel";
@@ -211,6 +212,16 @@ export function AdminClient() {
               onRefresh={() => void fetchStatus()}
             />
           )}
+        </div>
+      </section>
+
+      {/* Content Sync (repo → Sanity → devnet) */}
+      <section>
+        <h2 className="mb-4 font-display text-lg font-bold text-text">
+          Content Sync
+        </h2>
+        <div className="rounded-lg border border-border bg-card p-4 shadow-card">
+          <ContentSyncPanel />
         </div>
       </section>
 
--- a/apps/web/src/app/api/CLAUDE.md
+++ b/apps/web/src/app/api/CLAUDE.md
@@ -1,4 +1,4 @@
-# Frontend API Routes (34 routes)
+# Frontend API Routes (36 routes)
 
 ## Auth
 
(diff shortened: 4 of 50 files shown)


thomgabriel:
## Independent adversarial review — VERDICT: BLOCKING (one defect; otherwise solid)

`claude[bot]` stubbed (#363), so this adversarial pass is the review of record. It ran typecheck (0) + the 82-test sync suite (green) and wrote a repro.

### 🔴 BLOCKING — prune deletes just-written content (async read-your-writes race)
`admin-mutations.ts:42` commits writes with **`visibility:"async"`** (returns before the write is query-visible); `sync.ts:117-132` then reads managed docs back and prunes anything at the old `sync.rev`. When the read-back lags for a subset of just-written docs, they still read stale → match the prune filter → **get deleted**. Proven by a repro modeling async visibility (a just-written lesson appears in both `written` and `deleted`). Neither guard stops it (write-verify lower bound inflated by siblings; 1/7 ≈ 14% < 20% blast-radius). Impact: intermittent silent deletion of synced content; for `course`/`achievement` the PRESERVE'd `onChainStatus` (coursePda/txSignature/collection/isActive) is lost permanently; and the common case (most docs read stale) trips `BlastRadiusError` → spurious 409 on ~every real multi-doc sync. The 82-test suite can't catch it — `InMemoryGateway` is synchronous read-your-writes, the exact guarantee `async` disables.
**Fix:** compute the prune set from the in-memory sets — (docs managed-by-this-source read at sync START) minus (the projected/written id set) — and drop the post-write read-back entirely (race-free), and/or commit with `visibility:"sync"`.

### Should-fix (non-blocking)
- **MEDIUM** `github.ts:86` maps `neutral`/`skipped` → `"success"` → a *skipped* content-lint check reads green for the sync gate. Mitigated by server-side Zod+executor re-validation, but CI-only gates (slots.lock/id immutability) could pass on a skip. Treat skipped/neutral as not-green.
- **LOW** `tarball.ts`: no size/entry cap on `gunzipSync`+extract (OOM; bounded by access-controlled repo) — add a cap.
- **LOW** `readManagedDocuments` uses raw perspective (would prune stray drafts) — pin `published`.

### Cleared (genuinely CLEAN)
Refuse-on-red (`assertCommitSyncable` throws on failure/pending/unknown; re-checks the exact SHA; server re-validation fail-closed), mask-vs-lock assertion (always runs; exact-equality PASS; `MaskMismatchError(courseId)` on any mismatch; no false-PASS), content_tx_id (correct 12-zero left-pad + 32-byte equality + drift), auth (HMAC `admin_session` + timingSafeEqual + expiry + same-origin CSRF; both routes hard-gated), tarball traversal (in-memory Map, no FS write; long-path prefix/pax fix real), PRESERVE reattach.

Fix for the blocking prune race + the 3 non-blocking items is being applied now. Do not merge until the prune fix lands (it's also cutover-gated behind #384).

thomgabriel:
✅ **Prune fix verified — CS-9 now review-complete.** Confirmed `4c88093`..`eac542e`: `selectPrunable(existing, projectedIds)` filters the *pre-write* managed read for `sync.source`-owned docs whose `_id` is absent from the projected set — the post-write read-back is removed entirely, so a doc present in the new tree is structurally unprunable regardless of Sanity write-visibility timing. `visibility:"sync"` added; blast-radius now measured against the pre-write managed total. The async-visibility regression test fails against the old read-back logic and passes now; suite 82→84 (+ admin-mutations 4), full web suite 343 green. The 3 non-blocking items (neutral/skipped→non-syncable, tarball caps, published perspective) are also fixed. All previously-CLEAN surfaces (mask assertion, content_tx_id, auth, PRESERVE, tarball traversal) untouched.

Status: **code-complete + adversarially reviewed.** Merge remains gated behind #384 (this is stacked on it) and the coordinated prod cutover.

Issue #356, opened 2026-07-09
CS-4: Devnet reset — close_course ×7, recreate at new SIZE


thomgabriel:
Blocked by CS-3 (#355). Verify 13 enrollments on-chain → close_course ×7 → diffCourse reports not_deployed → admin sync recreates at SIZE 255 with active_lessons = dense mask → re-verify: 7 Course accounts at 255B, all 13 enrollment PDAs byte-identical (same addresses + same lesson_flags bytes), completed learners can finalize.

**Any invariant fails → STOP the lane, file a P0, do not proceed** (spec §15.3). Devnet only via Helius RPC.

thomgabriel:
**Deploy dependency discovered in CS-3's adversarial review (#378):** the devnet deploy of program v2 is gated on **CS-9 (sync-route migration) landing FIRST** — the current `sync/route.ts` still emits `newLessonCount` (old API), and `update_course`'s blind-mask-trust relies on CS-9's mask==slots.lock assertion as its only invariant carrier. Deploying without CS-9 breaks admin sync + removes the mask safety check. So CS-4 order becomes: CS-9 merged → frontend IDL regenerated → human re-audit of the 4 changed instructions → THEN close_course×7 + recreate + deploy. Adding blocked:dependency (CS-9 #370/impl).

thomgabriel:
## CS-4 checklist additions (from PR #378 adversarial review — HARD prerequisites for the deploy)

The v2 program code is merged to main, but devnet still runs v1. CS-4's sequence must be:
1. **Frontend lockstep FIRST (same PR or pre-staged):** regenerate + commit the v2 IDL (`apps/web/src/lib/solana/idl/superteam_academy.json`), update `academy-reads.ts` (Course: `lesson_count` → `active_lessons: [u64;4]`, snake_case), and the 5 v1-layout consumers: `api/lessons/complete/route.ts:201`, `api/admin/courses/sync/route.ts:345` (also: `newLessonCount` param no longer exists in v2 UpdateCourseParams — use `newActiveLessons`), `api/certificates/mint/route.ts:147`, `api/admin/status/route.ts:61-130`. Until deployed, v2 app code must tolerate v1 accounts (or the deploy+resync must be atomic-ish with the Vercel deploy).
2. Deploy v2 to devnet **via Helius RPC only**, byte-verify against the build artifact.
[...]

PR #390, opened 2026-07-10
docs(cutover): CS-4 program-v2 deploy + course reset runbook

thomgabriel:
Runbook for #356 (CS-4) + the coordinated content cutover it gates

### DIFF
--- a/docs/superpowers/plans/2026-07-10-cs4-cutover-runbook.md
+++ b/docs/superpowers/plans/2026-07-10-cs4-cutover-runbook.md
@@ -147,16 +147,16 @@ the enrollment baseline the DONE gate compares against.
 
 - [ ] **A2: The 6-completions bit-verify PASSES (§15.3) — the load-bearing gate**
 
-  **Preconditions:** A1 done. The bit-verify cross-checks, for each of the **6** real completions
-  across the 13 enrollments, that the slot the frozen `slots.lock.json` assigns equals the bit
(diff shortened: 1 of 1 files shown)


thomgabriel:
**G2 pre-cleared + a required correction from running the §15.3 bit-verification (PR #391) read-only against live devnet/DB/Sanity:**

✅ **G2 PASS — 152/152 completed `user_progress` rows map to a set `lesson_flags` bit at their live-order slot** (18 enrollments, 3 users; per-enrollment popcount == DB completed-count exactly; negative control confirms partial enrollments' set bits equal their completed slots). The `close_course` reset is safe w.r.t. preserving real progress. This can be re-run as the runbook's A-phase gate: `pnpm tsx apps/web/scripts/cs8-verify-bits.ts` (exits non-zero on any FAIL).

⚠️ **Correction — enrollment count:** devnet now has **37** Enrollment PDAs, not the spec §15.1 / runbook A6 figure of **13** (test data grew since 2026-07-09; completions are 152, not ~6). No safety impact — the invariant holds across all 37/152 — but the runbook's byte-verify (A6 snapshot + C2 diff) must **snapshot all enrollments enumerated at runtime and assert after==before**, NOT assert a hardcoded count of 13. The `getProgramAccounts` method already enumerates at runtime; just drop the `== 13` assertion in favor of `after-set == before-set` (count is informational, currently 37).

PR #391, opened 2026-07-10
chore(onchain): §15.3 bit-verification tooling (G2) for course reset (#359)

thomgabriel:
read-only §15.3 bit-verification tooling + run results (#359)

## What
`apps/web/scripts/cs8-verify-bits.ts` — the safety gate **G2** that must pass BEFORE any destructive devnet course reset (`close_course`) for CS-4/#356 and CS-8/#359. **Read-only**: reads live DB + live Sanity + devnet on-chain state, asserts the §15.3 invariant, reports. Writes nothing. No secrets printed. Reuses the app's own pure helpers (`findLessonIndex`, `isLessonComplete`, `findEnrollmentPDA`, `fetchEnrollment`).

Run (from `apps/web`): `pnpm tsx scripts/cs8-verify-bits.ts`

## Invariant (spec §15.3)
`Enrollment.lesson_flags` bits were set by lesson **array position** in the flattened `modules[].lessons[]` order. Each completed `user_progress` row must map to a **SET** bit at its lesson's position in **today's** live order — else the reset would corrupt real progress.

## Run results (devnet, David's prod DB `obqlljsagzslxarwphxv`, Sanity `4e3i2wwc/production`, program `7NeJaSRyb4Wxay3Tcd9bdpD7T3GWYUQSFyrhG8SgwE8V`)
- **152 PASS / 0 FAIL / 0 FLAG** across 152 completed rows, 18 enrollments, 3 users.
- Per-enrollment `popcount(lesson_flags)` == DB completed-row count for **all 18** enrollments.
- Negative control: partial enrollments have real gaps in set bits (e.g. anchor-framework bits `[0,1,2,4,6,11]`) that exactly equal the completed-row slots — the check discriminates, the pass is load-bearing.
- Advisory (not a G2 blocker): live snapshot grew since the spec — **37** Enrollment PDAs on devnet (spec §15.1 expected 13); 152 completed rows (spec ~6); 76 rows carry NULL `lesson_index` (verified against today's slot only).

## Verdict
**ALL PASS — G2 CLEAR.** Cutover (`close_course` reset) is safe to proceed on this data.

Do not merge yet.

Issue #359, opened 2026-07-09
CS-8: academy-courses scaffolding + Phase 1 extraction


thomgabriel:
**Plan-first.** Blocked by CS-2 (#352). Scaffold `academy-courses` (README/CONTRIBUTING/CLAUDE.md, schema/*.json, .vscode yaml.schemas, _template/ exercising every block type, reusable validate workflow). Phase 1 extraction: GROQ-export live dataset → generate the tree; **initial slots.lock from LIVE flattened lesson order**; **verify the 6 completions' bits vs user_progress BEFORE freezing the lockfile** (spec §15.3 — bits were set by array position at completion time; reconcile any moved lesson).

⚠️ I have **READ-only** access to `solanabr/academy-courses` (verified `gh repo view`). The push step is `blocked:needs-human` — prepare the full scaffolding payload locally, file the human hand-off, continue other lanes.

**Done when:** Plan merged; extraction script validates the generated tree against @superteam-lms/content-schema; scaffolding payload ready + human push issue filed.

thomgabriel:
## CS-8a (scaffolding) DONE — ready for your push

The full `academy-courses` tree is built, validated, and committed to branch **`initial-scaffolding`** in your own checkout at **`~/Documents/STBR/academy-courses`** (commit `7650586`, 37 files). I have READ-only access to the repo, so the push is yours:

```bash
cd ~/Documents/STBR/academy-courses
git push -u origin initial-scaffolding
# then open a PR into the default branch
```

**Built:** README (add-a-course in 5 steps), CONTRIBUTING (all CI gates 1–18), CLAUDE.md (spec §12 prohibitions), CODEOWNERS, teachers.yaml, the 8 generated JSON schemas + `.vscode` bindings, the reusable-workflow caller, and a complete `_template/` course exercising **every** block type (prose+image, video, code[ts], code[rust buildable], quiz single+multi, openEnded, wallet-funding→program-explorer→deployed-program-card capability chain).

**Validated against the real `@superteam-lms/content-schema` (CS-1 branch `8d2cc68`):**
- schema parse 14/14 ✓
- TS code block on the real QuickJS executor: **solution passes, starter fails** ✓
- `slots.lock.json` == `assignSlots(null, displayOrder)` ✓
- capability ordering valid by display order ✓

[...]

PR #397, opened 2026-07-10
chore(cs8): commit the extraction toolchain + regenerate courses-academy

thomgabriel:
## Why

The original CS-8 extraction lived only in the deleted `~/Documents/STBR/academy-courses` checkout (commit `ba62f0f`) and was **never pushed** — that account had READ-only on `solanabr/academy-courses`. When the repo was swapped for `solanabr/courses-academy`, the directory was removed and both the tree and the script went with it.

This commits the script so the tree is reproducible from live Sanity at any time.

## What ran

Content is already pushed: **`solanabr/courses-academy@79e9a50`** — 6 courses, 76 lessons, 10 achievements, 5 quests, 7 paths, 4 instructors.

| Gate | Result |
|---|---|
| `content-lint` (local) | **OK — 0 errors, 0 warnings**, 20 notices (all Rust/`buildable` blocks deferred to sync-time grading, spec §6.2a) |
| `validate-content` (courses-academy CI) | **green** on `79e9a50` |
| `cs8-verify-bits.ts` (§15.3, G2) | **152 PASS / 0 FAIL / 0 FLAG** across every completed `user_progress` row on devnet |
| slots vs live order | all 6 lockfiles dense `0..N-1`, identical to the live flattened `modules[].lessons[]` order |

Gate 6 **executed all 15 TypeScript `standard` solutions** in QuickJS: every `solution.ts` passes its own `tests.json`, and every `starter.ts` fails at least one — the check nothing in this repo previously performed.

## Owner decisions, encoded as constants

- **Dropped** `aD45H1NEbb1bqELwloGCqI` (solana-101) and `ops2aYkxIM6NMo1gE18U1o`. Neither id matches `^course-<kebab>$`, so `CourseId` rejects them regardless of preference — the drop is schema-forced. Their 3 lessons go with them.
- **Paths.** `path-solana-core` loses its solana-101 reference (2 courses remain, so it stays non-draft and remains a valid `path-completed` target). `path-ai-solana` referenced *only* solana-101, so — like `path-infrastructure` and `path-security` — it is now `draft: true`, which the schema requires of an empty non-draft path.
- **Achievements curated to working award kinds.** `speed-runner` + `perfect-score` deleted (no first-try/timing signal exists). `full-stack-solana` → `path-completed(path-solana-core)`; `course-completer` → `course-completed(course-solana-fundamentals)`; `bug-hunter` → `manual`. No community achievements — none are live.
- `hidden` test flags dropped (D4: answer keys are public; grading is by sandboxed execution).

## Design notes

- Imports `assignSlots` from `@superteam-lms/content-schema` instead of reimplementing it, so slot semantics cannot drift from the app. Verified that array order and `| order(order asc)` are identical for all 6 courses, so the ordering is unambiguous.
- Validates every emitted document against its Zod schema before writing.
- Lives outside `packages/` deliberately: throwaway tool, not shipped, and **not in the `apps/web` typecheck scope** — that scope (`include: ["**/*.ts"]`) is what surfaced 9 strict-mode errors in the closed #391.

## Known gaps found while doing this (not fixed here)

1. **`creator.githubId` is never resolved to a wallet.** CS-5 removed `author` from the Sanity course schema, but `admin/courses/sync/route.ts:212` still branches on `course.author`, and nothing maps `profiles.github_id → wallet_address`. After the CS-9 sync, on-chain `Course.creator` will **always** fall back to the platform authority and creator rewards will not reach teachers.
2. **Dropping solana-101 strands one real completion.** `cs8-verify-bits` shows `CrabbingShri × aD45H1NEbb1bqELwloGCqI 1/1 PASS`. With no `slots.lock.json`, CS-4's `close_course` will close that Course account and never recreate it. Accepted (mock content), recorded here.
3. **`cs8-verify-bits.ts` hardcodes `EXPECTED_ENROLLMENTS = 13`**; devnet now has **37**. The script reports this as advisory, not a G2 blocker, but plan §15.1's inventory is stale.

Refs #359

### DIFF
--- a/scripts/cs8-extraction/README.md
+++ b/scripts/cs8-extraction/README.md
@@ -1,0 +1,43 @@
+# cs8-extraction
+
+One-shot, **read-only** migration tool that extracts the live Sanity dataset into the `solanabr/courses-academy` content-standard tree (CS-8, spec §15, plan #388).
+
+It writes nothing to Sanity, Supabase, or chain. The only writes are files in the target checkout.
+
+## Run
+
+```bash
+pnpm --filter @superteam-lms/content-lint exec tsx \
+  "$(pwd)/scripts/cs8-extraction/extract.ts" ~/Documents/STBR/courses-academy
+```
+
+It is invoked through `content-lint`'s package so `yaml` and `@superteam-lms/content-schema` resolve. It lives outside `packages/` on purpose: it is a throwaway tool, not shipped, and not in the web typecheck scope.
+
+## Then gate it
+
+```bash
+# 1. every gate, zero errors (Rust/buildable blocks emit notices — tolerated)
+pnpm --filter @superteam-lms/content-lint exec tsx src/cli.ts ~/Documents/STBR/courses-academy
+
+# 2. §15.3 bit-verification — slots must reproduce every on-chain completion bit
+cd apps/web && pnpm exec tsx scripts/cs8-verify-bits.ts
+```
+
+## What it decides
+
+Both are owner decisions, encoded as constants at the top of `extract.ts`:
+
+- **Dropped courses.** `aD45H1NEbb1bqELwloGCqI` (solana-101) and `ops2aYkxIM6NMo1gE18U1o`. Neither id matches `^course-<kebab>$`, so `CourseId` rejects them regardless. Their lessons drop with them. `path-solana-core` loses its reference to the former; `path-ai-solana` referenced only it and becomes `draft: true`, as the schema requires of an empty non-draft path.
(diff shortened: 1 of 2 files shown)


thomgabriel:
Human-gate triage of the review's non-blocking notes: (1) creator-attribution gap → filed as #398 (P1); (2) looser `isExtractableCourseId` pre-filter and (3) `writeText` crash-on-missing-fields → declined: this is a manual, fail-loud one-shot tool — crashing at `*.parse` or `writeText` IS the designed failure mode; tightening pre-filters adds surface for no gain. Both gates green + gate evidence independently re-verified locally (lint 0 errors, bit-verify 152/152). Merging.

PR #403, opened 2026-07-10
fix(content-sync): defer rust/buildable at sync, matching CI gate 6

thomgabriel:
> ⚠️ **SENSITIVE** — content-sync validation (the trust boundary for what reaches prod Sanity). Adversarially reviewed; please sign off before merge.

## The bug (found while running the CS-9 sync)

The prod sync returned **HTTP 422** with 8 issues — 6× *"starter already passes"* and 2× *"solution … executor_unavailable"* — on content that content-lint passed with **0 errors** in CI. This is why prod has never been synced to `blocks[]`.

## Root cause

The two gates have different scope:
- **content-lint gate 6** (`gate6-executor.ts:117`) grades only `typescript && !buildable`; rust/buildable are a deferred **notice**.
- **content-sync** (`validate.ts:108`) graded **every** code block.

So the sync enforced exactly what CI defers. And the enforcement is *incoherent* for buildable: the buildable grader grades on **compilation**, and a good scaffold starter (`todo!()`) compiles — so it "passes" the two-sided *starter-must-fail* check → the 6 "starter already passes". The other 2 are the build server being **deliberately off in prod** → "executor_unavailable". All 8 failures were buildable; the rust `todo!()` starters correctly fail, and the 16 TS-standard blocks all passed.

No content fix makes a compiling scaffold fail a compile check. The gates simply must agree.

## The fix

content-sync now grades exactly what CI grades — **TS-standard only**; rust and buildable are deferred to runtime grading, which is unchanged and stays **fail-closed per block** (a learner can't complete a rust/buildable challenge until its executor is available). 6 lines in `validate.ts`, mirroring `gate6Check`.

## Tradeoff (approved by the owner)

The sync no longer verifies rust/buildable reference solutions — but **CI never did either**, and runtime grading is unchanged. This also stops every future rust/buildable teacher PR from hitting the same wall.

## Verification

- New parity tests: a rust/standard and a rust/buildable block whose starter *would* pass are now **deferred** (no throw); a TS block with a bad solution **still throws**. Confirmed failing before the fix, passing after.
- content-sync suite 86/86; full web suite **353/353**; `tsc` clean; prettier clean.
- Matches the real prod failure: deferring rust/buildable leaves only the 16 TS-standard blocks, which had **0 failures** in that run — so this fully unblocks the sync.

Refs #359

### DIFF
--- a/apps/web/src/lib/content-sync/__tests__/validate.test.ts
+++ b/apps/web/src/lib/content-sync/__tests__/validate.test.ts
@@ -149,4 +149,32 @@ describe("parseAndValidateTree", () => {
       expect(v.lessons.map((l) => l.lesson.id)).toContain("lesson-ex");
     }
   );
+
+  it("still requires the files of a DEFERRED (rust) block to exist", async () => {
+    const withCode = stringify({
+      id: "lesson-ex",
+      slug: "ex",
+      title: "Ex",
+      blocks: [
+        {
+          key: "ex",
+          type: "code",
+          language: "rust",
+          starter: "exercise/starter.rs",
+          solution: "exercise/solution.rs",
+          tests: "exercise/tests.json",
+        },
+      ],
+    });
+    const t = tree({
+      "courses/demo/course.yaml": courseYaml,
+      "courses/demo/lessons/ex/lesson.yaml": withCode,
+      // solution.rs + tests.json deliberately absent — deferral skips GRADING,
+      // not the file-existence check.
+      "courses/demo/lessons/ex/exercise/starter.rs": "// stub",
+    });
+    await expect(parseAndValidateTree(t, passGraders)).rejects.toBeInstanceOf(
+      ContentValidationError
+    );
+  });
 });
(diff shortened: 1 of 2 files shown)


thomgabriel:
@claude re-review — a follow-up commit landed after your verdict (file-existence check restored for deferred rust/buildable blocks + covering test). Please review the full current diff.

thomgabriel:
Human sign-off received (owner, in-session). Gates at head 727351d: CI green + claude-review re-reviewed clean after the follow-up commit + independent adversarial review SAFE TO MERGE (87/87 content-sync, 354/354 web, mutation-tested predicate pinning; independently reproduced the a44c71e existence-check bypass and confirmed 727351d fixes it). Triage: stale gate6 notice text → follow-up issue; malformed tests.json 500-vs-422 → declined, pre-existing, no regression; rust/buildable solutions unverified pre-runtime → owner-approved tradeoff, matches CI. Merging.

PR #406, opened 2026-07-10
fix(web): null-safe getCourseLessons + lesson page hardening (prod 500 fix)

### LINKED ISSUE #405: Lesson pages 500 on null entries in getCourseLessons — filter derefs and harden LessonPageClient
## Incident (2026-07-10, post-CS-9 sync)

All 12 solana-fundamentals lesson pages returned 500 in prod while the other 64 lessons were fine. Root cause chain:

1. `getCourseLessons` (queries.ts:254) result was cached pre-sync (revalidate 3600 + COURSES_CACHE_TAG) for the one course that was cache-warm.
2. Pre-sync course docs had `modules` as references (shape deleted in CS-5), so the deref projected `[null,null,null]`.
3. The sync's `revalidateTag` (sync.ts:140) did not reach the cache scope read by the CURRENT deployment (two redeploys occurred after the sync ran) — the poisoned entry survived, incl. across deploys.
4. `LessonPageClient` receives `allLessons=[null,null,null]` (`?? []` doesn't filter in-array nulls) → SSR crash → 500. Verified via the RSC flight payload of the crashed page.

## Fixes

- GROQ: `modules[].lessons[]->` projections should filter unresolvable derefs: `modules[].lessons[]->[defined(_id)]` or `[defined(@)]` — in getCourseLessons and any sibling query.
- `LessonPageClient` (and anything consuming allLessons): filter falsy entries defensively; a bad/stale cache entry must degrade (missing sidebar items), never 500 the page.
- Ops note: after a content sync, a redeploy does NOT clear a poisoned data-cache entry; re-firing revalidateTag on the current deployment does. Consider calling revalidateTag from an explicit admin 'purge cache' action in admin v2 (SP3).

Refs #359

thomgabriel:
## Incident
All 12 `solana-fundamentals` lesson pages returned 500 in prod post-CS-9-sync while 64 others were fine. Root cause (full forensics in #405): a pre-sync data-cache entry held the legacy-shaped course doc → `allLessons=[null,null,null]` (verified in the crashed page's RSC flight payload) → `LessonPageClient` crashed on `null._id` (lesson-client.tsx:137). The poisoned entry outlived TTL 3600, `revalidateTag`, and two redeploys.

## Fix
- **GROQ**: `(modules[].lessons[]->{...})[defined(_id)]` — unresolvable derefs degrade to fewer sidebar items, never nulls. Verified against live Sanity: healthy course → 12 clean items; the legacy-shaped solana-101 doc → `[]`.
- **Page**: `.filter(Boolean)` on allLessons before it reaches the client component.
- **Deterministic prod heal**: the query-text change alters the fetch cache key, evicting the stuck entry on deploy.

## Verification
- New GROQ tested against live prod Sanity on BOTH shapes (healthy + legacy) — 0 nulls in both.
- `tsc` clean, web tests 354/354.
- Local render of the failing URL with the fix: 200.

Closes #405

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/page.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/page.tsx
@@ -24,7 +24,7 @@ export default async function LessonPage({ params }: LessonPageProps) {
   return (
     <LessonPageClient
       lesson={lesson}
-      allLessons={allLessons ?? []}
+      allLessons={(allLessons ?? []).filter(Boolean)}
       locale={locale}
       courseSlug={slug}
       courseId={courseInfo?._id ?? slug}
--- a/apps/web/src/lib/sanity/queries.ts
+++ b/apps/web/src/lib/sanity/queries.ts
@@ -256,11 +256,14 @@ export async function getCourseLessons(
 ): Promise<Pick<Lesson, "_id" | "title" | "slug">[]> {
   return catalogFetch<Pick<Lesson, "_id" | "title" | "slug">[]>(
     `*[_type == "course" && slug.current == $courseSlug && onChainStatus.status == "synced" && ${activeGate} && ${publicAuthoringGate}][0] {
-      "lessons": modules[].lessons[]-> {
+      // (...)[defined(_id)] drops unresolvable derefs: a stale cache entry or a
+      // legacy-shaped course doc must degrade to fewer nav items, never emit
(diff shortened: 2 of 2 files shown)


thomgabriel:
@claude re-review — follow-up commit applies the (…)[defined(_id)] filter to all 6 lessons[]-> deref sites (the 5 you flagged + the original). Verified nested variants against live Sanity on healthy + legacy shapes: clean lists or lessons:null (handled by existing ?? []), never null entries. tsc clean, 354/354.

PR #427, opened 2026-07-11
fix(solana): creator denylist + platform-authority refusal in deployCoursePda (#402)

### LINKED ISSUE #402: Warn/refuse when Course.creator resolves to a known program ID or the platform authority
From adversarial review of #399 (F3). On-curve ≠ sensible: the System program (`11111111111111111111111111111111`), Token program, and the ed25519 identity point are all on-curve and pass every gate (probe-verified). And nothing warns when `creatorWallet == platform authority` — the exact copy-paste mistake that deploys silently and immutably. Add to `admin-signer.ts` deployCoursePda: denylist of well-known program IDs + refusal (or explicit override) when creator equals `_authority.publicKey`. Refs #399

thomgabriel:
Closes #402. **SENSITIVE (on-chain deploy signer) — holding for owner merge.**

Adds an exported `CREATOR_DENYLIST` + pure `assertCreatorAllowed()` guard in `deployCoursePda`, AFTER the existing on-curve check and BEFORE any transaction construction: refuses on-curve-but-nonsense creators (System/all-zero default, Token, Token-2022, ATA, Memo, Stake, Rent+Clock sysvars — **each independently re-verified on-curve during review**; off-curve well-knowns deliberately excluded as dead code) and `creator == platform authority`. `allowUnusualCreator` escape hatch (default false, unreachable from HTTP — verified by grep, no route plumbs it) logs loudly instead of refusing.

21 RED-first unit tests incl. an anti-dead-code on-curve assertion over the whole list; errors name the pubkey + failed check + courseId. Fable security review: approved — refusal ordering walked, authority reference confirmed (`PROGRAM_AUTHORITY_SECRET` keypair, not backend signer), surgical +215/-0.

**Devnet negative-path demo** (deploy with creator=System → refused; normal wallet → succeeds) deferred to post-merge evidence per the SP3-C plan — the guard is pure client-side refusal, unit coverage is the substrate.

Note for SP3-C (recorded): when the UI plumbs the override, split or label WHICH condition is overridden — one flag silencing two hazards invites rubber-stamping.

### DIFF
--- a/apps/web/src/lib/solana/__tests__/admin-signer.test.ts
+++ b/apps/web/src/lib/solana/__tests__/admin-signer.test.ts
@@ -1,0 +1,91 @@
+/* eslint-disable import/order -- vi.mock('server-only') must be hoisted above
+   the admin-signer import so the module graph loads under vitest. */
+import { describe, it, expect, vi } from "vitest";
+import { Keypair, PublicKey } from "@solana/web3.js";
+
+// admin-signer is a `server-only` module; neutralize the guard so the pure
+// creator-validation helper can be imported under the node test runner.
+vi.mock("server-only", () => ({}));
+
+import { CREATOR_DENYLIST, assertCreatorAllowed } from "../admin-signer";
+
+const AUTHORITY = Keypair.generate().publicKey;
+const NORMAL = Keypair.generate().publicKey;
+const COURSE = "course-abc";
+
+describe("CREATOR_DENYLIST", () => {
+  it("every entry is a real, on-curve address (not dead code the off-curve check already stops)", () => {
+    for (const { address } of CREATOR_DENYLIST) {
+      const pk = new PublicKey(address);
+      expect(PublicKey.isOnCurve(pk.toBytes())).toBe(true);
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


thomgabriel:
Human-gate sign-off (owner-authorized). Independent adversarial review: SAFE-TO-MERGE across all 6 planes.
- Guard wired correctly: assertCreatorAllowed runs at admin-signer.ts:441 (after on-curve, before .createCourse at :467); deployCoursePda has one caller (admin/courses/sync). updateCoursePda has no creator field (immutable) → correctly unguarded. No bypass path.
- 8 denylist entries all verified on-curve + canonical base58 (all fire, zero dead). System program (all-zero default) covered.
- allowUnusualCreator attacker-unreachable: no req.body spread; no caller sets it. creatorWallet sourced from committed content.
- No regression: the existing 6 creator==authority courses take the updateCoursePda branch (guard not invoked). Trial merge against current main: clean auto-merge, 512/512 tests pass, typecheck exit 0.
Non-blocking follow-up (denylist not exhaustive — Wrapped SOL mint + some sysvars unlisted) filed separately; defense-in-depth, not required for merge. Merging.

Issue #400, opened 2026-07-10
Drift engine blind to on-chain creator mismatch — must land before CS-4 recreate


thomgabriel:
From adversarial review of #399 (F1). `diffCourse` (apps/web/src/lib/admin/sync-diff.ts:185-260) never compares on-chain `creator` against Sanity `creatorWallet`. A wallet baked wrong at CS-4 recreate — or rotated in git afterwards — shows **'synced' forever** in the admin panel, with zero signal that a close+recreate is needed. Since Course.creator is immutable, this is the one mismatch the panel most needs to surface. Add `creator` to the immutable-mismatch diff. Also fold in F5/dashboard gap: `getMissingCourseFields` (sync-diff.ts:117) should include missing/invalid creatorWallet so the panel preempts the 400. Refs #399 #356

thomgabriel:
Ordering flag from the open-issues review: this is tagged 'resolve in SP3 (admin v2)' but its own title says 'must land before CS-4 recreate'. If CS-4 executes before SP3 (likely — CS-4 waits only on #387), the SP3 deferral leaves the recreate blind. Resolution: implement the MINIMAL fix (add creator to diffCourse's immutable-mismatch comparison in sync-diff.ts + creatorWallet in getMissingCourseFields) against the CURRENT panel as a pre-CS-4 item; SP3 inherits it. Re-labeling intent: pre-CS-4 checklist alongside #401/#402.

thomgabriel:
Delivered by #430 (SP3-C, merged 2926e08). The drift engine now detects creator mismatch: diffCourse (sync-diff.ts:288-295) emits an immutable 'creator' diff comparing on-chain creator vs instructor.wallet, getMissingCourseFields (sync-diff.ts:147-149) gates unshaped wallets, and the deploy screen's change-preview surfaces it with an immutable-field warning. Verified in the #430 review (trust boundary also holds server-side — updateCoursePda has no creator field). The 'must land before CS-4 recreate' condition is now satisfied: the detection is live ahead of any CS-4 (#356) recreate. Closing.

PR #399, opened 2026-07-10
feat(content): instructor wallet becomes the on-chain course creator

thomgabriel:
> ⚠️ **SENSITIVE** — changes on-chain `Course.creator` resolution (`admin/courses/sync/route.ts`, `admin-signer.ts`). Adversarially reviewed (summary below); please human-sign-off before merge.

## What

The instructor is now the course creator. Adds a **required, on-curve `wallet`** to the instructor and removes the course `creator.githubId` object. On-chain `Course.creator` is resolved as `course.instructor -> instructor.wallet` (GROQ deref), replacing the dead `course.author -> profiles.wallet_address` hop (gap #1 — nothing populated `github_id`, so it resolved to nothing).

Decisions (owner):
- `instructor.wallet` **required**, validated base58 + 32-byte + **on-curve**.
- Delete course `creator` entirely.
- **No platform-authority fallback anywhere** — a missing/invalid/off-curve wallet fails the course loudly.
- Validate at sync with `new PublicKey()` + `isOnCurve()`.
- Re-extraction preserves `slots.lock.json` byte-identical.

## Files

| Area | Change |
|---|---|
| `content-schema` | new `SolanaAddress` (base58 + 32-byte + on-curve via `@noble/curves`, no web3.js dep); `instructor.wallet` required; `course.creator` removed; JSON schemas regenerated (the #396 parity test enforces it) |
| `sync/route.ts` | `Course.creator = course.creatorWallet`; `new PublicKey` + `isOnCurve`; 400 on missing/invalid; drops the Supabase `author` hop |
| `admin-signer.ts` | `deployCoursePda` now **throws** on missing/invalid/off-curve creatorWallet (was: silent platform-authority fallback) |
| `queries.ts` | `getAllCoursesAdmin` projects `"creatorWallet": instructor->wallet`; `AdminCourse.author` → `creatorWallet` |
| Sanity schemas | drop course `creator`, add instructor `wallet` |
| projector / preserve | stop writing `creator`; `wallet` in `PROJECTED_FIELDS.instructor` |
| extractor | emit `instructor.wallet` (mock instructors = owner wallet); byte-identical slot preservation |
| fixtures | dead `creator:` lines removed from 10 gate tests + good fixture; fixture instructor gains a wallet |

## Verification

- content-schema **112**, content-lint **32**, web **351** — all green; `tsc` clean across web/content-schema/content-lint/sanity.
[...]

PR #430, opened 2026-07-12
SP3-C: deploy screen v2 — change-preview, #400 creator detection, per-item state

thomgabriel:
**SP3-C: deploy screen v2** — the final implementation slice of the SP1→SP2→SP3 epic. Plan: `docs/superpowers/plans/2026-07-11-sp3c-deploy-screen-v2.md`.

# ✅ #427 blocker cleared — rebased and proven

This branch used to pre-date #427 (the creator denylist), and its own probe **empirically proved the gap was live**. #427 is now merged and this branch is rebased onto it. The parked harness was renamed to a test, run, and re-parked — verbatim before/after:

```
BEFORE (pre-#427):  SYSTEM-PROGRAM: failed to get recent blockhash: TypeError: fetch failed
AFTER  (post-#427): SYSTEM-PROGRAM: deployCoursePda(...): creatorWallet 11111111111111111111111111111111
                    is a denylisted well-known address (System program) and cannot receive
                    creator rewards. Pass allowUnusualCreator to override.
```

The System-program case is now **refused pre-send** — it never reaches the RPC. That case was also upgraded from a passive probe to a **hard assertion** (`/denylisted well-known/` AND NOT `/blockhash|fetch failed/`), so a future denylist regression fails loudly instead of silently passing. Harness files stay inert (`.harness.ts` cannot match vitest's include — verified via `vitest list`).

# 📏 Read this before reviewing the diff

The old **+7,612 / −20,502 / touches `ci.yml`** was a **squash-merge artifact**, not real scope: this branch was cut from the SP2-C branch, which landed as a *squash* commit, so GitHub replayed all of SP2-C's deletions from a stale merge-base.

**Post-rebase, the real scope is 23 files, +1838 / −243, and it touches zero CI config.** 13 app-source + 5 tests + 3 harness/README + the 3 locale files. Nothing outside SP3-C. (`admin-signer.ts` is *not* in this diff — the guard arrives from main via #427.)

# 🔴 What this branch discovered (owner decision — tracked as #440)

Read-only devnet audit: **all 6 deployed courses have on-chain `creator == the platform authority`** — deployed before #399 introduced `instructor.wallet → Course.creator`. Creator rewards route to the platform key today, and `Course.creator` is **immutable** — only a recreate changes it. Legacy devnet state, not a prod incident, **but a pre-mainnet checklist item**: mainnet courses get exactly one shot at their creator. The new deploy screen surfaces this as 6 loud, confirm-blocked warnings — the #400 detection working. (The shared instructor wallet across all 6 is intentional — the mock-instructor decision from #399.)

# What shipped

- **Per-item state** — deployed / on-chain-drifted / content-drifted / never-deployed / drift-unknown, from `onchain_deployments` + a bundle-vs-HEAD content-drift fold (GitHub failure degrades to an explicit "unknown", logged, not a dead status route).
- **Change-preview before deploy** — focus-trapped modal listing exactly what the transaction writes; immutable mismatches **remove** the confirm button; Sync All gets one ICU-plural counted confirm instead of N modals and still excludes immutable-mismatch courses.
- **#400 lands here** — creator-mismatch detection did **not** previously exist (`diffCourse` had no creator entry; `getMissingCourseFields` never checked `creatorWallet`). Both halves added, canonically base58-compared (false-green proven impossible).
- **The diff engine went live** — the status route previously hardcoded `differences: []`, so the engine had zero live callers. It now decodes the already-fetched account (no new RPC) and reports real diffs. The screen looks redder against real devnet state because the redness is true.
- Polish: StatusBadge / missing-fields / refusal messages i18n'd ×3 (server refusals render verbatim, pubkey intact); `sanityId→contentId`, `sanityValue→contentValue` sweep.

# The one thing the whole-branch review caught (fixed)

The preview **promised something the deploy doesn't do**: it told the admin "this deploy will update the content commitment," but confirm posts `{courseId}` only, and the route writes `contentTxId` only when an `activeLessons` mask is present — which no UI sends. Worse, the *logic* encoded the same lie (`noChanges && !contentStale` counted staleness as a pending change, dangling a Confirm-sync that did nothing). With 5/6 devnet courses content-stale, it would have lied on first real use. Fixed: honest copy in 3 locales, logic corrected, **two-sided regression guards**. The underlying path split is tracked as #432.

# Gates at rebased HEAD (`7e083fa`)

**545 tests** pass (66 files — main brought in #427's signer tests) · `tsc --noEmit` exit 0 · `next lint` zero errors · `compile-content` + porcelain **byte-identical** · `next build` 75/75 static pages, exit 0 · rebase had **zero conflicts**.

Reviewed per-task (5 tasks, opus-implemented + fable-reviewed each) + a whole-branch adversarial pass: **SAFE-TO-MERGE after #427** — which is now satisfied. Adjudicated clean: no phantom diffs, no false green on creator, and the UI-only immutable guard degrades safely (`UpdateCourseParams` carries no immutable field, so a stale tab or `curl` cannot express an immutable write).

# Follow-ups filed
#432 (unify deploy paths) · #433 (creator wallet on first deploy) · #434 (undecodable account shows green) · #435 (diffAchievement dead) · #436 (status-route Supabase-500 asymmetry) · #444 (dead drift route).


thomgabriel:
Human-gate sign-off. SAFE-lane (verified: 23 files, +1838/-243, zero sensitive paths — no migration/signer/onchain_deployments-write; the deployCoursePda refs are test/harness calls proving #402, not signer changes). Gates:
- CI all green incl. Frontend (lint·typecheck·test) — the 545-test/tsc/lint/build gate the bot couldn't run locally.
- claude[bot]: no blocking; traced the no-changes-lie fix, #400 creator-mismatch (trust boundary holds server-side — updateCoursePda has no creator field), denylist regression guard, and i18n parity against the real code.
- Independent scope check: drift engine (sync-diff.ts + github/drift.ts) is READ-ONLY (fetchCourse via getAccountInfo, no signer/writes), reads bundle desired-state (getAllCoursesAdmin, post-SP2) vs on-chain actual. The 'redness' (creator==authority courses) is the engine telling the truth = #440, not a #430 defect.
Non-blocking triage: note 2 (diffAchievement dead) already tracked #435; note 1 (pre-existing hardcoded strings in course-sync-table) filed as follow-up. Merging.

Issue #450, opened 2026-07-12
🔴 Recreating a course RESETS total_completions — re-opening the anti-Sybil creator-reward window


thomgabriel:
Found while building the close+recreate server path (the fix for #440). **This is the reason that feature must not ship an unguarded button.**

## The mechanism

- `create_course.rs:54-55` hard-resets `total_completions = 0` and `total_enrollments = 0`.
- `finalize_course.rs:83-90` gates the creator reward on a **bounded window**: `total_completions >= min_completions_for_reward && total_completions < min_completions_for_reward + 100`.
- The program says exactly why that window exists: *"The per-completion drip was otherwise unbounded and **Sybil-farmable** once a course crossed its threshold... Total creator XP per course is thus capped at WINDOW * creator_reward_xp."*
- **No instruction can write `total_completions` back.** Only `create_course` (zeroes) and `finalize_course` (increments) touch it.

## The consequence

**A recreate re-opens an already-spent creator-reward window.** A course that has paid out its full 100-completion cap can, after a recreate, pay out **100 more**. The cap that exists to bound creator XP minting is defeated by an authority-gated button.

It bites hardest in the exact scenario the feature was built for (#440): we recreate a course **to install the correct creator wallet** — and thereby hand that creator a fully refreshed reward window on a course that may already have paid out to the platform key.

Bounded by `completed_at.is_none()` (already-finalized learners cannot re-finalize), so the counter only climbs from *new* completions. Devnet impact ≈ zero. **A popular mainnet course is a different story.**

## Mitigation (both levers are `update_course`-writable)

After any recreate of a course that has already paid creator rewards, immediately re-close the window:
- set `creator_reward_xp = 0`, **or**
- raise `min_completions_for_reward` beyond reach.

The recreate orchestrator currently **snapshots both counters pre-flight and reports them (`lostCounters`) with a loud warning** — but it does not (and cannot, on-chain) restore them. The UI must force this decision rather than let an operator click through it.

## Related silent resets (already fixed in the orchestrator)
`create_course` also resets `is_active → true` (would un-hide a deactivated course) and `collection → default` (would break credential issuance with `CollectionMismatch` and orphan already-minted credentials). The orchestrator re-binds the **existing** collection rather than minting a new one.

## Also: post-#449 this gets worse
[...]

PR #451, opened 2026-07-12
admin: merge Publish + Deploy into one Courses screen (default landing) — and actually explain the model

thomgabriel:
**Stacked on #430** (base branch is SP3-C — merge that first). SAFE lane: composition + nav + copy. No route, API call, or behavior changed.

## Why

Owner, verbatim: *"this admin console got so confusing. didnt get this Content Publishing how is it working exactly? and then the deploy?"*

They were right, and the confusion was the bug. The two screens assumed you already held the mental model. Now the screen teaches it.

## What

- **`/admin/courses`** merges the two screens in step order: intro → publish pin card (step 1) → **new state legend** → deploy table (step 2). Both components `git mv`d with **zero internal changes**.
- Nav collapses **Publish + Deploy → Courses**; `/admin` now lands there; the old routes survive as locale-preserving redirects (a 404 would be hostile).
- **The legend explains all 9 states** — including **immutable mismatch**, which renders *no badge at all* and was explained nowhere in the old UI.

## The load-bearing sentence

> Editing lesson content — prose, code, quizzes, the title, tags — is not on chain. Those edits go live on publish alone, and correctly show nothing to deploy below. **A course still reading "Synced" after you rewrote a lesson is not a bug.**

That is the single most confusing thing about this console and it was nowhere on screen.

## Verification

557 tests / 67 files · tsc + lint clean · `next build` compiles · **driven live** against a real admin session: all 3 redirects 307 to the right locale-prefixed target, and SSR HTML for en/pt-BR/es carries the intro + legend with **0 `MISSING_MESSAGE`**.

The legend test caught a real bug: the drift column heading was the literal string "Content drift" — identical to the badge label, so it explained nothing. Renamed in all 3 locales.

Changed tests, none weakened: nav retargeted 4→3 links **plus a new case** asserting no link points at the retired routes; `publish-page.test.tsx` deleted (its subject no longer exists — both assertions live on in the two new files); new routing (5) and courses-page (7) suites.

## Known warts

- The **achievements table lives on a screen called "Courses"**. It is fed by the same `/api/admin/status` payload and the same hook as the course table, so a separate screen would re-fetch the same endpoint to render one table. Kept it; the legend scopes itself to *course* rows so it does not over-claim.
- `course-sync-table.tsx` is still hardcoded English, so a pt-BR admin now sees a Portuguese legend above English table headers. Pre-existing; filed separately.
- The immutable-mismatch copy still lists the pre-recreate remedies — will be updated when the close+recreate flow lands (gated on #450).


thomgabriel:
Human-gate sign-off (owner-authorized gate). Triple-clean: CI green + claude[bot] no-blocking + independent review SAFE-TO-MERGE. Verified: SAFE-lane (admin frontend + i18n + docs only, no sensitive paths), zero capability loss (both clients git-mv'd, all row actions retained), old routes redirect (locale-preserving, tested), default landing → /admin/courses, i18n parity (557 tests), and the explanatory copy is TRUE against the diff engine (the 'synced after a content edit is not a bug' line + the 9-state legend + no content-drift overpromise). lessonCount is a non-issue — the deploy path writes it (route.ts:454-459 → admin-signer:517). Non-blocking (tracked): #452 (course-sync-table hardcoded English), immutable-remedy copy to refine with #450's recreate flow. Merging.

Issue #452, opened 2026-07-12
admin: course/achievement sync tables are hardcoded English — pt-BR/es admins now see mixed language


thomgabriel:
Surfaced by the Courses-screen merge (#451).

`components/admin/course-sync-table.tsx` and `achievement-sync-table.tsx` hardcode their strings ("Course", "Lessons", "Deploy", "Sync All", empty states). Every other part of the admin console is now i18n'd — including, as of #451, a full explanatory intro and state legend sitting **directly above these tables**.

So a pt-BR admin now reads a Portuguese explanation of the deploy model, and then hits English column headers. That is worse than uniformly-English was.

**Why it was deferred, repeatedly:** SP3-D excluded them because SP3-C was rewriting them; SP3-C (#430) scoped its i18n to the strings it added. The result is that nobody ever owned the sweep. Owning it here.

Scope: both sync tables + `sync-diff-view.tsx`, en/pt-BR/es, deep-parity test green. Sequence **after #430 and #451** merge — those touch the same files.

thomgabriel:
Resolved by the admin revamp: WS-A (#517, closes #448) externalized `course-sync-table.tsx` and WS-C (#519) externalized `achievement-sync-table.tsx` — both are fully i18n'd on main (26 and 23 `t()` sites respectively) with en/es/pt-BR parity (1007 keys each). Verified on main; no hardcoded English remains in either sync table.

Issue #449, opened 2026-07-12
🔴 LAUNCH BLOCKER: client IDL is v1, program source is v2 — deploying the program breaks EVERY Course read


thomgabriel:
Found while chasing an unrelated (and falsified) lessonCount report. **This is a latent landmine that detonates on the next program deploy — including a fresh mainnet deploy.**

## The drift

| | Client IDL (`apps/web/src/lib/solana/idl/superteam_academy.json`) | Program source (in-tree) |
|---|---|---|
| `update_course` arg | `new_lesson_count: Option<u8>` | `new_active_lessons: Option<[u64;4]>` |
| `Course` account field | `lesson_count` (**1 byte**) | `active_lessons` (**32 bytes**) |

- Program source went **v2** in `616d1ac` (2026-07-09, CS-3 #378 — *"Replaces v1's `new_lesson_count`"*).
- The client IDL was last regenerated in `87fbe55` (2026-07-07, #314) — which **predates** it (`git merge-base --is-ancestor` → true).
- **The IDL was never regenerated after CS-3.**

## Why nothing is broken right now

**Devnet still runs the v1 program**, so chain and IDL agree. Proven empirically by byte-alignment: decoding live course accounts with the v1 coder yields `xp_per_lesson` = 20 / 40 / 10, matching the bundle exactly for three courses. Under a v2 layout, `active_lessons` (32B) replaces `lesson_count` (1B), shifting **every subsequent field by 31 bytes** — those values would decode as garbage, not as three exact matches.

## What happens the moment the v2 program is deployed

1. **Every `Course` read returns garbage.** The app decodes Course accounts with a raw BorshCoder over this IDL (`lib/solana/academy-reads.ts`). A 31-byte shift corrupts `difficulty`, `xp_per_lesson`, `track_id`, `track_level`, `prerequisite`, `creator_reward_xp`, everything after the mask. That is the catalog, the deploy screen, drift detection, and the status route.
2. **`update_course` transactions fail outright** — `Some(u8)` (2 bytes) cannot deserialize into `Option<[u64;4]>` (33 bytes) → `InstructionDidNotDeserialize`.

**A fresh mainnet deploy builds from the current source (v2), so mainnet would be broken from day one.**

## Fix

Regenerate the client IDL from the v2 program (`anchor build` → copy the IDL) **and gate any program deploy on it**. Also audit every consumer of the changed layout:
- `lib/solana/academy-reads.ts` (raw BorshCoder reads)
[...]

Issue #444, opened 2026-07-12
admin: /api/admin/content/drift is dead code — no UI consumer on main or in #430


thomgabriel:
Found by the post-Sanity docs fact-check.

**No `.tsx` fetches `/api/admin/content/drift`** — not on `main`, not on the open SP3-C branch (`feat/sp3c-deploy-v2-11-07-2026` / PR #430). The Publish screen uses `/api/admin/publish/pin`; the deploy screen fetches `/api/admin/{auth,flags,publish/pin,status}`.

The SP2-C plan asserted the route "SURVIVES … for the SP3-B publish card + SP3-C deploy change-preview" and carved it (plus `lib/github/drift.ts`) out of the Sanity deletion on that basis. **That intent was never realized in either branch** — the publish card computes its own drift from `content.lock` vs HEAD, and SP3-C folds content-drift into `/api/admin/status`.

So the route + its test are dead, and its doc comments still describe a "repo → Sanity" three-way drift that no longer exists.

**Decide:** delete the route (and audit whether `lib/github/drift.ts` still has live callers — `computeContentDrift`/`computeChainDrift` ARE used by the status route, so the *module* survives even if the *route* does not), or wire it into something. Do this AFTER #430 merges.

Note: the SP2-C carve-out was still correct for `lib/github/*` as a whole — the compiler and status route depend on it. Only the HTTP route is orphaned.

thomgabriel:
Shipped in #460 (merged as 7809d2e). Route + tests deleted; `lib/github/drift.ts` survives (still consumed by `/api/admin/status`). Stale doc entries removed from `api/CLAUDE.md` and `docs/ARCHITECTURE.md`.

Issue #437, opened 2026-07-12
security: X-Frame-Options can be tightened SAMEORIGIN → DENY now that the Studio embed is gone


thomgabriel:
From the SP2-C/D whole-branch review.

`SAMEORIGIN`'s documented rationale was the Sanity Studio iframe embed, which was deleted in #429. Nothing in the app is meant to be framed.

**Decision needed** (not done unilaterally): tighten to `DENY`. Low risk, strictly safer — but it *is* a behavior change, so it wants a deliberate call + a check that no internal tooling frames the app.

thomgabriel:
Shipped in #460 (merged as 7809d2e). Note the issue as filed was doubly incomplete: X-Frame-Options alone is ignored by browsers whenever CSP `frame-ancestors` is present, AND the first fix attempt landed on `next.config.mjs`'s CSP which only covers `/api/*`. The enforcing policy is `buildCsp()` in `lib/csp.ts` (applied by middleware to every page) — now `frame-ancestors 'none'`, pinned by a test against `buildCsp()` itself.

PR #460, opened 2026-07-13
fix(security): WS-3 bundle — bound platform-funded on-chain writes, fail closed on NaN lesson counts (#459 #410 #437 #444)

### LINKED ISSUE #459: Rate-limit /api/lessons/complete — the platform pays the fee and the credential rent for every completion
## What

`/api/lessons/complete` has **no rate limit**, and the platform's backend keypair is both `backendSigner` **and** `payer` on the resulting on-chain transactions (`academy-program.ts` — `payer: signer.publicKey` on `completeLesson`, `finalizeCourse`, `issueCredential`). The learner signs nothing after enrolling.

So every completion — real or scripted — spends **platform** SOL, and a finalized course mints a Metaplex Core credential whose **asset rent is also platform-funded**.

## Why it's reachable

The route's gate is a *correctness* oracle, not a *humanness* one. Before signing, it checks:

1. a Supabase session exists,
2. every graded block passes a deterministic grader,
3. an on-chain enrollment exists.

Each is replayable:

- **answers are identical for every learner** — solve a course once, capture the `proofs` payload, replay it from N accounts;
- **SIWS signup is free** — sign a nonce with any generated keypair (no email, no captcha, no cost);
- **no rate limit** on the route.

The learner's only real cost is the `Enrollment` PDA rent (127 bytes ≈ 0.0018 SOL), which is locked permanently because `close_enrollment` refuses finalized enrollments.

This is **not** a forged-completion bug — the fail-closed grading gate (spec §7.2) is sound and a forged client `"passed": true` still cannot mint XP. This is purely about *volume*: nothing bounds how many honest-but-scripted completions one actor can drive through the platform's wallet.

## Impact

- Backend hot-wallet SOL drain (tx fees, unbounded).
- Permanent Core-asset rent spent minting credentials to bot wallets.
- Amplifies creator-reward farming, which is uncapped as of the reward simplification in #458 (that was an accepted, deliberate tradeoff — this issue is the compensating control).

## Fix

Per-wallet + per-IP rate limit on `/api/lessons/complete` using the existing `rate_limits` table. A generous ceiling is fine — the goal is bounding a script, not slowing a human. Consider the same for `/api/certificates/mint`.

## Scope

WS-3 (fix bundles) in the Launch-Readiness epic (#458). SAFE, small.

thomgabriel:
WS-3 of the Launch-Readiness epic (#458). Five independent fixes, no behaviour change for a legitimate learner. **569 tests pass** (was 557); typecheck clean.

---

### #459 — rate-limit the two routes that spend platform SOL

`/api/lessons/complete` is the single chokepoint for every platform-funded on-chain write: the backend keypair is `payer` **and** `backendSigner` on the resulting tx, and the `LessonCompleted` webhook cascades from it into `finalize_course` (learner bonus + creator reward) and `issue_credential` (permanent Core-asset rent). The learner signs, and pays, nothing.

That route's completion gate is sound at what it was built for — a forged client `"passed": true` still cannot mint XP. But it proves answers are *correct*, not that a *person* produced them, and every input to it is replayable. So the gate stops cheating and never bounded volume.

Both routes now carry a **per-user** and a **per-IP** limit, via the existing `isRateLimited` / `check_rate_limit` house pattern.

- Per-user bounds one account hammering the route. It does **nothing** against Sybils — every fresh account is a fresh key.
- **Per-IP is the only key that bounds an actor.** Set generously (300/h on completions) so a real cohort behind one NAT — an IRL bootcamp — never trips it.
- Neither stops a determined farmer with proxies. They bound the **burn rate**, and that is the honest claim; this is a cost control, not an anti-Sybil oracle.

Checked **before** grading, so a throttled caller never reaches the code executors. Tests assert exactly that (`codeGrader` never called on a 429).

This is the compensating control for the deliberate removal of the on-chain creator-reward cap agreed for WS-1.

### `bitmap.ts` — fail CLOSED on a non-finite lesson count

`isAllLessonsComplete(flags, NaN)` returned **`true`**:

```ts
if (lessonCount === 0) return false;                // NaN === 0   → false
const wordsNeeded = Math.ceil(lessonCount / 64);    // NaN
if (lessonFlags.length < wordsNeeded) return false; // x < NaN     → false
for (let i = 0; i < lessonCount; i++) { ... }       // 0 < NaN     → false, never runs
return true;                                        // ← falls through
```

"Nobody has completed anything" read as "everybody has completed everything". Callers derive the count from a chain read (`Number(course.lesson_count)`), so **any client/program version skew (#449) produces exactly that NaN**. Guarded here independently of the IDL work, so the class is safe before any deploy. `decodeLessonBitmap` now throws rather than silently decoding every enrollment as empty.

The 6 new tests were verified to **fail against the unfixed code** (5 of 6 — `Infinity` was already accidentally safe, since `len < Infinity` is true).

### #410 — null `wallet_address` on account deletion

It carries a UNIQUE constraint, so a retained one stranded the wallet on a tombstoned row **forever**: unlinkable by its owner (row is soft-deleted), unclaimable by anyone else (constraint still holds). It is also the strongest identifier on the row, so keeping it defeated the anonymisation the rest of the payload performs. Route had no test file; added one.

### #437 — `X-Frame-Options` SAMEORIGIN → DENY

The issue asked only for the header — but the CSP also carried `frame-ancestors 'self'`, and **that is what modern browsers actually enforce**. Tightening one without the other would have been cosmetic. Both move. Nothing frames this app; the only `<iframe>` is us embedding YouTube/Vimeo, which is `frame-src` and untouched.

### #444 — delete the dead `/api/admin/content/drift` route

Zero UI consumers. Drift is folded into `/api/admin/status`; the publish card reads `/api/admin/publish/pin`. `lib/github/drift.ts` **survives** — the status route uses it; only the HTTP route was orphaned. Stale doc comments pointing at it are corrected, including the `api/CLAUDE.md` route table (whose route count was also stale — claimed 36 against 43 files — so the drift-prone number is gone).

---

**Not** in this PR, deliberately: #435 (`diffAchievement`) and #452 (i18n sweep) both touch files WS-1/WS-2 are rewriting — they'd collide. #327 (serverEnv) touches three files WS-1 owns.

Closes #459, #410, #437, #444.


thomgabriel:
## Adversarial review round — two of this PR's own claims were false. Both fixed in `d06f74c`.

An independent reviewer was pointed at this branch and told to break it, not bless it. It did. Recording the outcome here because the two findings are the interesting part of the PR now.

### It broke #437: the hardening shipped to the wrong surface

`next.config.mjs`'s CSP is attached **only** to `source: "/api/:path*"`. Pages get their CSP from middleware → `lib/csp.ts`, which still read `frame-ancestors 'self'` and which this PR never touched.

And per CSP Level 2 §7.4.1, browsers **ignore `X-Frame-Options` whenever `frame-ancestors` is present** — so the global `X-Frame-Options: DENY` was discarded outright. The enforced page policy stayed byte-for-byte what it was before. The clickjacking fix landed on JSON API responses (never a clickjacking target) and skipped every HTML page.

The original commit message even stated the mechanism correctly — *"that directive is what modern browsers actually enforce"* — and then drew the wrong conclusion from it. `lib/csp.ts` now says `'none'`, and `csp.test.ts` pins it (verified to fail against `'self'`).

### It broke the per-IP limiter: bypassable over IPv6, no proxy required

The bucket was keyed on the raw address. Anyone with a routed **/64** — standard on essentially every VPS — has 2^64 source addresses, hence 2^64 fresh buckets. The per-IP limit was decorative, which directly contradicts this PR's central claim that per-IP is the key that bounds an actor.

`getClientIp` now collapses IPv6 to its /64. **This also closes the same hole in `/api/admin/auth`**, whose brute-force throttle guards `ADMIN_SECRET` — a /64 previously bought an effectively unthrottled guessing budget against the secret that gates authority-signed on-chain writes. That one was worse than the bug this PR set out to fix.

Header order is now a trust order: `x-vercel-forwarded-for` (edge-set, unforgeable) → `x-real-ip` → `x-forwarded-for`. The spoofing attack the reviewer was told to try **fails** — Vercel's edge replaces XFF rather than appending, so `[0]` is correct today — but that guarantee dies the moment a proxy is put in front of Vercel, so XFF is now the last resort rather than the first choice.

### Also fixed

- **The per-IP ceiling locked out the exact cohort its own comment promised to protect.** Courses run 12–16 lessons, so 300/h capped one NAT at ~25 students — a 30-person bootcamp, or a CGNAT'd mobile carrier, would 429 the entire room. Raised to 1200/h.
- **A 429 rendered as "something went wrong."** An IP lockout would present to a classroom as an outage, and each retry burned another token. Added `lesson.completionRateLimited` in en/pt-BR/es.
- **#410's anonymisation was incomplete** — `enrollments` carries its own `wallet_address` copy, so the wallet→user_id join survived one table over and the deletion was cosmetic. `google_id`/`github_id` were also left intact. All nulled; a failed scrub now 500s rather than reporting a deletion that did not fully happen.
- `docs/ARCHITECTURE.md` still listed the deleted drift route.

### Filed, not fixed: #461

Nothing checks `profiles.deleted_at` at login, so a deleted user who signs back in is silently returned to their tombstoned row. Pre-existing, and it needs an auth-flow decision rather than a deletion-route change.

**586 tests** (was 569 → 557 at branch start). The two new test files, `csp.test.ts` and `rate-limit.test.ts`, cover exactly the two things that had no tests — which is why both bugs got this far.

thomgabriel:
@claude please re-review — your review above was posted against the first commit only (15:51Z; the fix commit `d06f74c` landed at 16:04Z), so its BLOCKING verdict is stale.

Everything you flagged is now fixed:

- **BLOCKING / #437 CSP** — `lib/csp.ts:153` is now `frame-ancestors 'none'`. Your diagnosis was exactly right, and an independent adversarial reviewer reached the same conclusion by the same route. `csp.test.ts` now pins it and was verified to fail against `'self'`, so it can't silently regress again.
- **`docs/ARCHITECTURE.md:952`** — stale drift-route entry removed.
- **`getClientIp` duplication** — consolidated; `auth/nonce` and `admin/auth` now both use the shared helper.

Please focus the re-review on what's new in `d06f74c`, particularly the parts you cleared as "verified correct" last time:

1. **`getClientIp` IPv6 /64 normalisation** (`lib/rate-limit.ts`). You passed the per-IP limiter as sound; it was bypassable — keyed on the raw address, so a routed /64 (any VPS) yielded 2^64 fresh buckets. Worth checking the same hole in `/api/admin/auth`, which uses this to throttle `ADMIN_SECRET` brute-forcing. Is the /64 collapse correct for compressed `::` runs, IPv4-mapped addresses, and zone suffixes?
2. **Header trust order** — `x-vercel-forwarded-for` → `x-real-ip` → `x-forwarded-for`. Is taking `[0]` of XFF still right as a last resort?
3. **Rate-limit ceilings** — 1200/h per IP on completions, 200/h on mints. Sized for a ~60-person cohort behind one NAT on a 16-lesson course. Too loose to bound cost, or still too tight for a CGNAT'd carrier?
4. **`/api/account/delete`** — now also nulls `google_id`/`github_id` and scrubs `enrollments.wallet_address`. Does nulling any of those break a reader I've missed?

Known and deliberately out of scope: nothing checks `profiles.deleted_at` at login (filed as #461 — pre-existing, needs an auth-flow decision, not a deletion-route change).

Issue #440, opened 2026-07-12
PRE-MAINNET: all 6 devnet courses have Course.creator == platform authority (creator is immutable)


thomgabriel:
**Measured 2026-07-12 by a read-only devnet audit (SP3-C Task 6).**

All **6/6 deployed courses** have on-chain `Course.creator` == the **platform authority pubkey**, not the instructor wallet. They were deployed before #399 introduced `instructor.wallet → Course.creator`. Creator reward XP therefore routes to the platform key.

## Why this is a launch blocker, not a devnet nit

`Course.creator` is set once at `create_course` and is **IMMUTABLE** — `update_course` takes no creator param. A redeploy/update **cannot** fix it. Only a full recreate can.

On devnet this is cosmetic legacy state. **On mainnet there is exactly one chance per course**: every course must be created with the correct creator on its first deploy, or the reward routing is wrong forever.

## Action before the mainnet create wave

1. Verify each course's `instructor.wallet` in the content bundle resolves to the **intended reward recipient**.
2. Note that all current bundle instructors resolve to a single wallet (`B7o8NfV81HzjuZFWQTTx3Xdvh77Dqoajwib3kWEnvzJF`) — this is **intentional** (the mock-instructor decision from #399), but it means real instructors need real wallets before launch.
3. Decide per legacy devnet course: accept (devnet is throwaway) vs recreate.

## Tooling now in place

- **#400 detection** (PR #430): the deploy screen surfaces this as 6 loud, confirm-blocked immutable warnings. It did not exist before — `diffCourse` had no creator entry and `getMissingCourseFields` never checked `creatorWallet`.
- **#402/#427**: signer-side denylist refusing `creator == platform authority` and well-known program ids on future creates.
- Related: #433 (show the creator wallet in the change-preview on first deploy).

thomgabriel:
**Concrete pre-deploy task, updated by the instructor-from-profile decision (2026-07-14):** under the new model the course carries the creator wallet directly (no instructor content docs). The current 6 courses point 4 placeholder instructor docs at ONE shared test wallet (`B7o8…`), so the Phase-4 recreate would bake that placeholder into all 6 immutable creators. Pre-deploy: put the **real creator wallet on each course** in `courses-academy` (same content PR as `creatorRewardXp: 30` + dropping `minCompletionsForReward`). Names/avatars are then self-served via each wallet's academy profile. Design: `docs/superpowers/specs/2026-07-14-instructor-from-profile-design.md`.

[...]

Issue #432, opened 2026-07-12
admin/deploy: the deploy button never updates content_tx_id — unify the two deploy paths


thomgabriel:
Found by the SP3-C whole-branch review.

`POST /api/admin/courses/sync` writes `content_tx_id` **only when an `activeLessons` mask is present in the body**. No UI sends one — the mask commit lives on the Drift screen path. So the deploy button can never clear a `content_stale` badge.

SP3-C (#430) fixed the *lie* (the preview used to promise it would, and the logic counted staleness as a pending change), but the underlying split remains: two deploy paths with different powers.

**Fix:** unify — either the preview's confirm derives + sends the mask (respecting the existing server-side rederive-and-assert interlock), or the content-commit becomes its own explicit action with honest UI. Design task; do not bolt the mask on blindly.

thomgabriel:
The reset writes content_tx_id fresh on all 6 recreated courses (create sets it), discharging the symptom. The structural unify-deploy-paths cleanup can be tracked separately if still wanted.

Issue #332, opened 2026-07-06
Validate Sanity lessonCount <= 255 before create_course/update_course sync (u8 arg)


thomgabriel:
## Finding (adversarial review of PR #326, 2026-07-06)

`course.lessonCount` from Sanity is passed **unbounded** into the on-chain `u8` `lesson_count` arg in both sync paths:
- `apps/web/src/app/api/admin/courses/sync/route.ts` — `createCourse` (`lessonCount`) and the new `updateCourse` (`newLessonCount`) calls.

A Sanity course authored with **> 255 lessons** hits Anchor's u8 encoder. Not exploitable and **not introduced by #326** (pre-existing on the create path) — on-chain it fails safe either way: a truncated low value is rejected by the new increase-only guard (`LessonCountDecrease`), and an encoder throw just fails the sync cleanly. But it surfaces as a confusing sync error rather than a clear validation message.

## Done when
- [ ] Admin sync validates `lessonCount <= 255` (and `>= 1`) before building the `createCourse`/`updateCourse` params, returning a clear 4xx with the offending course id
- [ ] Unit/route test covers the >255 rejection
- [ ] `pnpm typecheck` + route tests pass

_Hygiene/robustness. Low real-world likelihood (courses rarely exceed 255 lessons), hence P2._

thomgabriel:
lessonCount 1..=255 guard landed with the v-next write-path cutover (B1) and is live on main (sync/route.ts isValidLessonCount, fail-closed 400 before param build).

Issue #141, opened 2026-06-23
[G-4] CU budget reviewed


Potolski:
> **Code legend / glossary:** see [`docs/TASK-CODES.md`](https://github.com/solanabr/superteam-academy/blob/main/docs/TASK-CODES.md) for how to read task codes, severities, and area labels.

CU budget reviewed — depends on **P1-5**.

**Done when:** per-instruction CU numbers are reviewed and within budget.

thomgabriel:
**G-4 criterion met** (not closing — leaving the gate to flip at launch sign-off). CU was re-measured on the v-next program in Phase 2 (#472, `a543403`): 18 instructions / 19 transactions, all far under the 1.4M per-tx cap (finalize_course 21,249 ≈ 1.5%). CU is a property of the final bytecode, so this holds through deploy. Ready to flip whenever the launch checklist is finalized.

thomgabriel:
Orchestrator audit + pickup (David's items released to the loop): the CU substrate fully survived the Pinocchio migration — all 18 instructions measured (`CU_BASELINE.pinocchio.md`, total 211,351 CU, −39.2% vs Anchor), heaviest = upgrade_credential 46,552 / award_achievement 40,172 / issue_credential 34,614, all ≪ the 200k default per-tx ceiling (worst case ~23%). What #141 still lacks: the explicit written ruling, and a CI drift gate (the harness is manual-only — grep-verified not wired into .github/). The loop is landing a separate PR with `docs/CU-BUDGET-REVIEW.md` (the ruling + re-review triggers) and a baseline drift gate in the on-chain CI lane. **This gate issue stays open for the owner's flip.** Removing on-hold-david per the 2026-07-27 owner rule change (pickup = strip).

thomgabriel:
Board audit 2026-07-31 — the ask is met. docs/CU-BUDGET-REVIEW.md is the written ruling this gate asks for: budget fixed at the 200,000 CU per-transaction default (verified — the program never requests a SetComputeUnitLimit bump) with headroom shown for every instruction; measured numbers in onchain-academy/tests/CU_BASELINE.{pinocchio,rust}.md — all 18 instructions, worst case upgrade_credential at 46,552 CU (~23% of budget).

[...]

PR #462, opened 2026-07-13
docs(specs): WS-1 — program v-next design (reward simplification, IDL cutover, devnet reset)

### LINKED ISSUE #449: 🔴 LAUNCH BLOCKER: client IDL is v1, program source is v2 — deploying the program breaks EVERY Course read

thomgabriel:
Sub-spec promoted out of the Launch-Readiness epic (`af658f4`, §WS-1). **Design only — no code.** Ready for the tri-audit.

Doc: `docs/superpowers/specs/2026-07-13-program-vnext-design.md`

Scope: remove `min_completions_for_reward` + the creator-reward window, regenerate the IDL (#449), and close/recreate the six devnet Course PDAs with instructor wallets as `creator` (#440). Closes #449, #450, #332, #440, #432, #140, #141 → unblocks the G-2 gate (#139).

---

## Three findings from tracing the current code that change the plan the epic assumed

**1. The migration could have deadlocked — it doesn't, and I checked before writing.**

The moment v-next is live, the six existing 224-byte accounts stop deserializing as `Account<Course>`. If `close_course` resolved them that way, they could be neither *used* nor *closed* — permanently stranded, with no instruction able to touch them.

It doesn't. `close_course` takes the course as an **`UncheckedAccount`**, deliberately, so it can touch a size-mismatched account — it was built for the earlier 192→224 resize. It validates the PDA seeds, program ownership, and the 8-byte discriminator (stable across resizes), then drains and frees the PDA by hand. **The migration is viable.** This was the single thing that could have bricked the courses, so it's now a retired risk in the register rather than an unexamined assumption.

**2. The maintenance gate as the epic scoped it has a hole.**

The epic says "gate the four write endpoints." But the on-chain retry queue is drained from the **login routes** — `retryPendingOnchainActions` is called from `/api/auth/wallet`, `/api/auth/callback`, and `/api/quests/daily`, *not* from a cron.

So a route-level gate on the four obvious write endpoints leaves every **sign-in** firing backend-signed transactions at a half-migrated chain — burning platform fees (the backend keypair is `payer`) and re-queueing failures that can never succeed. The gate has to live **inside** `retryPendingOnchainActions`, not only in front of routes. Nobody would think to gate *login* as a write path.

**3. The decoder should normalise, not just dispatch.**

The epic prescribed branching the coder on byte length. That's necessary but half a fix: it leaves seven call sites still asking a version-specific question (`what is lesson_count?`) about a field that doesn't exist in one version.

Instead, synthesise v1's `lesson_count` **up** into a dense mask, so `fetchCourse` returns **one shape** regardless of what's on chain. Then:

- all seven consumers migrate **once**, to `liveLessonCount`, and are correct against *both* layouts — no second migration after the deploy;
- the `|| 1` / `?? 0` fallbacks disappear, because the field they defended against can no longer be absent;
- the silent-garble class dies **structurally** rather than by careful sequencing;
- and PR-1 becomes **behaviour-neutral on today's chain**, so it merges on its own instead of riding the cutover.

## Also corrected against the epic

- Removing `min_completions_for_reward` touches **three** surfaces — `Course`, `CreateCourseParams`, `UpdateCourseParams` — two of which are **instruction arg layouts**, not just the account.
- `Course::SIZE` lands at **253**, and `_reserved` must **not** be padded back to 255 — that would collide with the never-deployed in-tree v2 and make the length discriminator ambiguous forever.
- `creator_reward_xp` 750→30 is **cross-repo** (`packages/content-schema` + `solanabr/courses-academy` + a `content.lock` bump) and **load-bearing, not cosmetic**: uncapped × 750 is 25× worse than uncapped × 30, so shipping the field removal without the value drop delivers a different risk than the one that was accepted.
- `create_course` **keeps** `lesson_count: u8` (it derives the dense mask), so the create path loses exactly one field.
- The mask-writer is only **half-built** — `admin-signer.ts:517` still sends `newLessonCount`, and `slots.lock.json`, which the program comment names as the invariant carrier the sync route must assert against, exists **only as a content-lint test fixture**. That's unbudgeted #432 work; better to find it here than in the middle of the cutover.

## One consequence the epic missed

`create_course` sets `total_enrollments = 0`, and the enrollment PDAs **survive** the reset — so the counter **under-reports permanently**, with no instruction able to correct it. Noise on devnet with test data, but it should be a decision, not a surprise.

(`total_completions = 0` is now *harmless* — that reset is exactly what #450 was about, and with the reward window gone it has no economic consequence. **#450 dies here.**)

## Three owner decisions block implementation

1. **The six instructor pubkeys.** Hard blocker on the reset, and the whole point of #440 — `creator` is immutable, so a recreate with the wrong value isn't recoverable without *another* close/recreate.
2. **`total_enrollments` reset** — accept the under-count on devnet, or add a backfill (which would mean a program change)?
3. **The devnet downtime window** — learner on-chain *writes* pause for Phase 4; reads stay up (the catalog renders from the bundle). When?

---

Requesting the tri-audit before any implementation starts.

### DIFF
--- a/docs/superpowers/specs/2026-07-13-program-vnext-design.md
+++ b/docs/superpowers/specs/2026-07-13-program-vnext-design.md
@@ -1,0 +1,252 @@
+# WS-1 — Program v-next: reward simplification, IDL cutover, devnet reset
+
+**Status:** design — no code. Gets a tri-audit before implementation starts.
+**Parent:** `2026-07-12-launch-readiness-epic-design.md` (§WS-1), merged as `af658f4`.
+**Closes on completion:** #449, #450, #332, #440, #432, #140, #141. Unblocks the G-2 launch gate (#139) → Squads custody (#305, WS-4).
+**Owner decisions already locked:** creator-reward cap removed entirely (accepted, insider-only risk — compensating control #459 shipped in `7809d2e`); Anchor-forward, #387 stays parked.
+
+---
+
+## 1. What this is
+
+One on-chain redeploy, plus the client work that has to bracket it. It is the only SENSITIVE workstream in the epic and the only one that can silently corrupt production, so it gets its own spec.
+
+Three things happen to the program:
+
+1. `min_completions_for_reward` (u16) is deleted, along with the 100-completion reward window in `finalize_course`.
+2. The IDL is regenerated — it is currently **v1** while the in-tree program is **v2** (#449), and the client has been decoding live accounts by luck.
+3. The six existing devnet Course PDAs are closed and recreated, with instructor wallets as `creator` (#440) and at the new account size.
+
+Everything else in this spec exists to make those three survivable.
+
+---
+
+## 2. The hazard this spec is really about
+
+**The client and the program deploy on independent rails.** Merging to `main` auto-deploys the frontend via Vercel. Deploying the program is a manual Solana operation. Nothing sequences them, and there is no shared version handshake.
+
+That matters because of an asymmetry:
+
+| Situation                                       | Result                                                |
+| ----------------------------------------------- | ----------------------------------------------------- |
+| **v-next client reads a v1 (224-byte) account** | Borsh hits EOF → **throws loudly**. Bad, but visible. |
+| **v1 client reads a v-next (253-byte) account** | Decodes garbage. **No error.**                        |
+
+The second is the whole problem. A v1 client reading a v-next `Course` produces plausible-looking nonsense: every field after the lesson mask is shifted by 31 bytes. It does not throw, it does not log, and it does not stop.
+
+And the failure is not confined to display. Traced on `main` (see #449 for the full table), **seven** sites read `course.lesson_count` off a raw-BorshCoder decode, and post-v2 that field is simply absent — so each read yields `undefined` and **not one of them throws**:
+
+- Three (`event-handlers.ts:528`, `onchain-queue.ts:181`, `certificates/mint:147`) do `Number(undefined) || 1` and would mint credential NFTs whose metadata records `xp = xp_per_lesson × 1`. That metadata is **pinned to Arweave**. It is the only consequence in this entire migration that cannot be undone afterwards, and it is the reason the decoder ships before the program.
+- One (`event-handlers.ts:450`) did `Number(undefined)` → `NaN`, and `isAllLessonsComplete(flags, NaN)` used to return **`true`** — every guard in it is a comparison, and comparisons against `NaN` are false, so it fell through. That specific fall-through is already **fixed and merged** (`7809d2e`, PR #460): the bitmap helpers now fail closed on a non-finite count. The site still needs migrating, but it can no longer fail open while we get there.
+
+So the design goal is not "sequence the two deploys carefully." It is **make the ordering not matter.**
+
+---
+
+## 3. Design: a decoder that normalises, not just dispatches
+
+The epic prescribed a length-aware decoder — branch the coder on account byte-length. That is necessary but it is only half the fix, because it leaves seven call sites still asking a version-specific question (`what is lesson_count?`) about a value that no longer exists in one of the versions.
+
+**`fetchCourse` should return one shape regardless of what is on chain.**
+
+```
+decode(accountInfo):
+  224 bytes → v1 coder  → { …, activeLessons: denseMask(lesson_count), liveLessonCount: lesson_count }
+  253 bytes → v-next    → { …, activeLessons: active_lessons,          liveLessonCount: popcount(active_lessons) }
+  anything else → throw (loudly — an unknown layout must never decode)
+```
+
+v1's `lesson_count` is synthesised **up** into a dense mask; v-next's mask is read directly. Callers only ever see `activeLessons` / `liveLessonCount`. The consequences:
+
+- All seven consumers migrate to `liveLessonCount` **once**, and are correct against both layouts. There is no second migration after the deploy.
+- The `|| 1` and `?? 0` fallbacks disappear, because the field they were defending against can no longer be absent.
+- The whole silent-garble class dies **structurally**, not by careful sequencing.
+- PR-1 becomes **behaviour-neutral on today's chain** (every live account is 224 bytes and takes the v1 branch), which makes it a safe, independently-mergeable change rather than part of a knife-edge cutover.
+
+### Why byte length is the only available signal
+
+Both obvious alternatives are dead ends, and both are worth stating so nobody re-litigates them:
+
+- **The Anchor discriminator does not change.** It is derived from the struct _name_ (`Course`), not its layout — identical in v1 and v-next.
+- **`Course.version` is not a schema version.** `create_course` sets it to `1`; `update_course` **increments** it. It is a content-revision counter. Dispatching on it would be actively wrong.
+
+`space` is fixed at `Course::SIZE`, so the allocated length is stable regardless of how long `course_id` is. Length is a reliable discriminator here — but only because we control the sizes:
+
+|                                                     | bytes   | status                      |
+| --------------------------------------------------- | ------- | --------------------------- |
+| v1                                                  | **224** | live on devnet today        |
+| v2 (in-tree now)                                    | **255** | **never deployed anywhere** |
+| **v-next** (v2 − `min_completions_for_reward: u16`) | **253** | the deploy target           |
+
+**Do not pad `_reserved` back to 255 to "preserve" the size.** That would collide with the never-deployed v2 layout and make the discriminator ambiguous forever. Let it land on 253. `_reserved: [u8; 8]` stays exactly as it is.
+
+---
+
+## 4. What actually changes in the program
+
+Removing one field touches **three** surfaces, not one — two of them are instruction argument layouts, which is easy to miss:
+
+| Surface              | Change                                                                                    |
+| -------------------- | ----------------------------------------------------------------------------------------- |
+| `state/course.rs`    | delete `min_completions_for_reward: u16` → `SIZE` 255 → **253**                           |
+| `CreateCourseParams` | delete `min_completions_for_reward: u16` → **instruction arg layout changes**             |
+| `UpdateCourseParams` | delete `new_min_completions_for_reward: Option<u16>` → **instruction arg layout changes** |
+| `finalize_course.rs` | delete `CREATOR_REWARD_WINDOW`, `reward_end`, and both threshold clauses                  |
+
+`finalize_course`'s reward condition collapses to its irreducible form:
+
+```rust
+if course.creator_reward_xp > 0 {
+    // mint creator_reward_xp to the creator
+}
+```
+
+`CreateCourseParams` **keeps `lesson_count: u8`** — `create_course` derives the dense mask from it via `Course::dense_mask()`. New courses are always dense; slots are only retired later, through `update_course(new_active_lessons)`. So the create path loses exactly one field and nothing else.
+
+### The reward-value change is load-bearing, not cosmetic
+
+`creator_reward_xp` must drop to **30** in the same wave. It is currently `750 ×4`, `1000`, and `500` in the bundle. With the cap gone, uncapped × 750 is **25× worse** than uncapped × 30. Shipping the field removal without the value drop delivers a materially different risk than the one the owner accepted.
+
+This is **cross-repo**, and the epic under-stated it:
+
+- `packages/content-schema/src/course.ts:32-33` and `schema/course.schema.json:42-48` declare both `creatorRewardXp` and `minCompletionsForReward`.
+- All six course docs in `solanabr/courses-academy` set `minCompletionsForReward: 10`.
+- So it needs a **content-repo PR + a `content.lock` bump + a bundle recompile**, sequenced with the program deploy — not a program-only change.
+
+`minCompletionsForReward` should become optional-and-ignored in the schema first (so the content repo does not break the instant the program drops it), then be removed from the content docs in a follow-up. Sequencing the schema removal _before_ the content PR would break the compile gate.
+
+---
+
+## 5. Ordered plan
+
+The order is chosen so that **no step depends on a step that has not landed**, and so the only irreversible action (recreating accounts) happens last, behind a gate.
+
+### Phase 0 — finalise the program source, do not deploy
+
+Remove the field and the window. `anchor build` → this regenerates the IDL **from the final v-next source**, which is what Phase 1 needs. Nothing is deployed. The chain is untouched.
+
+> The epic said "regen the IDL" as a step inside the deploy. It has to happen _before_ the client work, because PR-1 cannot ship a v-next coder it does not have.
+
+### Phase 1 — client: normalising decoder (SAFE, behaviour-neutral, merge freely)
+
+Ship both coders and the length dispatch from §3, migrate all seven `lesson_count` consumers to `liveLessonCount`, and delete the `|| 1` / `?? 0` fallbacks.
+
+**Write paths stay v1-shaped in this PR.** The instruction arg layouts change with the program, not with the account, so a version-aware writer would be genuinely hard — and it is unnecessary, because the migration itself is performed by scripts (Phase 4), not by the web admin. The web write path cuts over in Phase 6, after the program is live. This split is what keeps Phase 1 risk-free.
+
+Verifiable claim for the reviewer: on today's chain (all six accounts at 224 bytes), this PR changes no observable behaviour.
+
+### Phase 2 — verifiable build + measurements (#140, #141)
+
+Reproducible build, record the program hash, re-measure CU. These gate the deploy; they do not change it.
+
+### Phase 3 — maintenance gate ON
+
+**This is where the epic's plan has a hole.** It said "gate the four write endpoints." There are more than four, and one of them is not an endpoint at all.
+
+Every path that signs an on-chain write:
+
+| Path                                                         | Signer    | During the window                             |
+| ------------------------------------------------------------ | --------- | --------------------------------------------- |
+| `/api/lessons/complete`                                      | backend   | **BLOCK**                                     |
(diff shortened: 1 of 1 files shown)


thomgabriel:
## Tri-audit — 3 independent lenses (on-chain correctness · client+sequencing · completeness/over-engineering)

Verdict: **directionally excellent, on-chain reasoning is bulletproof** (all six on-chain claims confirmed field-by-field — byte math 255→253, `close_course` UncheckedAccount linchpin, same-address recreation, the three removal surfaces, counter-reset semantics). But the audit found **three HIGH client-side must-fixes the spec asserts away**, and a genuine conflict between two lenses that resolves into the central design correction. Nothing here blocks the *approach* — these are pre-implementation corrections.

### 🔴 HIGH — must fix before implementation

**H1 — The finalize-trigger consumer needs the MASK, not the count. Migrating all 7 to `liveLessonCount` is silently wrong in the regime WS-1 itself creates.**
The on-chain completion gate is a **subset test**, not a count: `finalize_course.rs:29-34` does `flags & active == active`. The client's `isAllLessonsComplete(flags, count)` (`bitmap.ts:44-64`) is a dense-**prefix** check — and prefix == subset *only while masks are dense*. Phase 6 / #432 introduces `update_course(new_active_lessons)` to **retire slots** → sparse masks. The moment any slot is retired, consumer #4 (`event-handlers.ts:450`, the client finalize trigger) breaks both ways:
- learner completed `{0,1,3}`, slot 2 retired → chain finalizes, client checks `{0,1,2}` → bit 2 unset → **client never fires finalize; learner stuck without credential**.
- learner completed `{0,1,2}`, slot 3 still active → prefix says done → client fires → chain `flags & active == active` fails → `CourseNotCompleted` revert, **wasted backend-signed fee + requeue**.
Fix: this one consumer must mirror the on-chain subset test against the real `activeLessons` mask, not consume a count. DoD line 246 ("all seven read `liveLessonCount`") and §3 ("no second migration") are correct only at the migration boundary where every mask is dense.

**H2 — The maintenance gate leaves the changed-layout authority course-write routes open; they corrupt state during the Phase 4–6 window.**
Phase-3 marks `/api/admin/courses/{sync,deactivate,reactivate}` "leave open (operator needs it)." But those route through `create_course`/`update_course` (`admin-signer.ts:547,550`), and `admin-signer.ts:261` binds its `Program` to the **v1 IDL** until Phase 6. v-next changes those instruction arg layouts (drops `min_completions`, swaps `new_lesson_count:u8`→`new_active_lessons:[u64;4]`). Borsh options are positional, so during the window any web-admin course create/update/deactivate/reactivate serializes **v1-layout params against a v-next program** → best case "not all bytes read" revert, worst case field misalignment **corrupting course state under the authority key**. Even a bare `{newIsActive:true}` misaligns (different field *count*). Fix: block the `deployCoursePda`/`updateCoursePda`-backed routes too until the Phase-6 cutover. `/api/admin/achievements/sync` → `create_achievement_type` is unchanged and genuinely safe to leave open.

**H3 — "Learner progress survives" is conditional on the recreate mask, and the spec states it unconditionally.**
`finalize` gating on `flags & active == active` means recreating a course with a `lesson_count` whose dense-mask is a **superset** of what a mid-course learner completed silently flips them back to incomplete. The stated counts (12/16/12/12/12/12) must be **verified equal to the original devnet v1 `lesson_count`s** — this is a Phase-4 precondition, and `lesson_count`-at-recreate joins `creator` as an effectively set-once value for anyone mid-course.

### 🟡 Should-fix (accuracy / scope)

- **S1 — #432 is SMALLER than the spec says (good news).** The claim that `slots.lock.json` is "only a test fixture, not in the content repo" is **stale/wrong**: `compile-content.ts:160-161` errors if any course lacks it, the bundle compiled cleanly with all 6, it's emitted to `generated/slots.json`, imported by `store.ts:19`, loaded by the sync route, cross-checked in `admin-signer.ts:191`. The invariant carrier exists and is wired. Real Phase-6 scope shrinks to two items: the mask-writer (`admin-signer.ts:517` still sends `newLessonCount`) + the increase-only diff engine (`sync-diff.ts:190-195,297-307`).
- **S2 — §4's sequencing justification is false.** "Removing the schema field first breaks the compile gate" doesn't hold — the Zod schema has no `.strict()` and the JSON schema no `additionalProperties:false`, so `schema.parse` silently strips unknown keys. The two-step order is harmless hygiene, but it's defended by a non-existent failure. Correct the rationale.
- **S3 — Test/CI surface unenumerated.** Field removal breaks `test_course.rs:15` (`assert_eq!(Course::SIZE, 255)` → 253), the TS integration + CU tests (pass `min_completions`), and client fixtures. Most notably the **Trident fuzz mirror** (`trident-tests/fuzz_0/types.rs:66-77,204-211`) hand-copies `CreateCourseParams`/`Course` with the old layout — it's *already* stale vs in-tree v2 and will silently fuzz the wrong layout. Add to Phase 2.
- **S4 — State the upgrade-in-place assumption.** Same program ID (`declare_id!` unchanged) is load-bearing for the whole same-address-recreation argument; worth making explicit.
- **S5 — quests/daily is not a drain site.** Line 158 lists `/api/quests/daily` as a third `retryPendingOnchainActions` caller. It calls `retryQuestXpForUser` (`onchain-queue.ts:426`) — DB-only via `award_xp`, no signing. The conclusion (gate inside the drain, fed by the two auth routes) holds; only the enumeration is wrong. Bonus: quest-XP delivery can keep flowing during the window.
- **S6 — an 8th decoder.** `resolvers.ts:69` does a raw `coder.accounts.decode("Course",…)` outside `academy-reads.ts`, not length-aware. It reads only `course_id` (front field, survives by luck), so not a hazard — but the spec's "one decoder" framing misses it; migrate for hygiene.

### ⚪ Over-engineering check — the one conflict, resolved

The completeness lens flagged "drop the synthesized `activeLessons` — no consumer reads it" (true of *today's* code). The sequencing lens (H1) found the consumer that *must* read it once slots go sparse. **Resolution: keep `activeLessons` — it has exactly one legitimate, correctness-critical reader (H1). It is not gold-plating.** What was wrong is the spec's wiring of *all* 7 consumers to the count. Net design: `activeLessons` = real mask (v-next direct; v1 = `denseMask(count)`, exact since v1 is always dense) + `liveLessonCount` = popcount convenience; wire 6 consumers to the count, the finalize-trigger to the mask.

**Explicitly do NOT cut** (all verified load-bearing, not machinery): the decoder-first ordering, the Arweave-permanent-XP priority, the gate-inside-`retryPendingOnchainActions` (per-login trigger, no cron), and the two-tx close/create (runtime-forced by `realloc(0)`+reassign). The backend must-BLOCK gate table is complete for backend-signed paths.

### Minor on-chain notes
`version` rewinds to 1 on recreate (cosmetic, but the decoder leans on "version isn't a schema signal" — worth noting the recreate rewinds it); `realloc(0,false)` doesn't zero freed bytes — safe only because lamports drain to 0; the `total_completions=0` reset is correctly harmless for double-mint (blocked by `enrollment.completed_at.is_none()`) but mildly amplifies the accepted uncapped-reward risk rather than being "purely harmless."

---
The three owner decisions the spec raises (instructor pubkeys, `total_enrollments` backfill, downtime window) are unchanged by the audit — except H2 widens the window to also pause web-admin course toggles.

thomgabriel:
## rev-2 pushed (`5a1ff62`) — all three HIGH findings folded, and I verified each at the source rather than taking them on trust

The audit was right on every count, and **H1 is the one I'm glad it caught** — it's the kind of bug that would have shipped, passed review, and detonated months later on the first lesson removal.

### H1 — the finalize trigger needs the mask, not a count

Confirmed at source. `finalize_course.rs:32` gates on a **subset** test (`flags & active == active`). The client's `isAllLessonsComplete` walks bits `0..count-1` and demands every one — a **dense-prefix** test. Equal only while masks stay dense, and **Phase 6 of this very spec is what makes them sparse.**

Once a slot is retired it breaks in **both** directions:
- a learner who completed every *live* slot is asked for the retired bit, fails, and is **stranded without XP or credential**;
- a learner holding a dense prefix passes the client check, fires `finalize_course`, and the chain rejects it — a **platform-funded transaction that can never succeed**, re-queued forever.

The program's own comment says this out loud, and rev-1 walked past it:

> *"which is exactly what the v1 popcount-equality could not do"* — `finalize_course.rs:29`

**This also settles the conflict between two of your lenses.** The over-engineering lens said "cut the synthesised mask, nothing reads it." True of today's code — and exactly wrong going forward. It has **exactly one reader** (verified: `event-handlers.ts:453` is `isAllLessonsComplete`'s only caller), and that reader is the one that must never be wrong. So: **keep the mask**, wire six consumers to `liveLessonCount`, wire the finalize trigger to the mask, and **replace** `isAllLessonsComplete` rather than leaving it around to be picked up again.

Generalised in the spec, because this will recur: *any client-side predicate that decides whether to fire an on-chain instruction must mirror the on-chain gate exactly.* Approximating it with a cheaper scalar strands users and burns fees, and both failures are silent.

### H2 — admin course routes blocked

Confirmed: `admin-signer.ts:26,261` binds the **v1 IDL**. During Phase 4–6 those routes serialise v1-layout params against a v-next program, and since `UpdateCourseParams` loses a field, even a bare `{ newIsActive: true }` misaligns — Borsh is positional and the field *count* differs. **One click of "deactivate" corrupts course state under the authority key.** Blocked. Only `/api/admin/achievements/sync` stays open. The operator drives the migration with scripts, which was always the Phase-4 plan.

### H3 — recreate count is set-once

Confirmed. Phase 4 now has an explicit precondition: **read each course's `lesson_count` off its live account and reuse that exact value.** Do not take it from the bundle — the bundle is what the content says *today*; the chain is what learners *enrolled under*. If they differ, the chain wins for the recreate, and the difference is reconciled afterwards via a deliberate `update_course`. `lesson_count`-at-recreate is now documented as a second irreversible value alongside `creator`.

### The three factual errors — all mine, all corrected

- **`/api/quests/daily` is not a drain site.** It calls `retryQuestXpForUser`, which settles `quest_xp` rows through `award_xp` and signs nothing. Gating it would have paused quest XP for no reason. Quest XP now explicitly keeps flowing through the window.
- **#432 is smaller, not bigger.** My "`slots.lock.json` is only a test fixture" claim came from a too-literal filename grep — the artifact is `generated/slots.json`, already emitted, imported at `store.ts:19`, already cross-checked at `admin-signer.ts:175-191`. The invariant carrier exists and is wired. Phase 6 drops to two real items: the mask-writer and the increase-only diff engine.
- **The schema-sequencing rationale was false.** No `.strict()`, no `additionalProperties: false` — Zod strips unknown keys. Either order works. Kept the two-step because it's tidy, and said so, so nobody defends a constraint that doesn't exist.

### Added

The test/CI surface (`Course::SIZE` 255→253, the TS/CU fixtures, and the **Trident fuzz mirror** — which hand-copies the layout and is *already* stale against v2, meaning the fuzzer is currently fuzzing a struct that does not exist); the same-program-ID upgrade-in-place assumption; `version` rewinding to 1 on recreate; and honest framing of the `total_completions` reset — it kills #450, but it also erases the only on-chain record of what a creator has been paid, which rev-1 called "harmless."

**Definition of done now demands a test against a sparse mask** — without it, H1 regresses the first time a lesson is retired — and **devnet verification of a real pre-reset enrollment**, rather than an assertion that progress survived.

Nothing load-bearing was cut: decoder-first ordering, the Arweave-XP priority, gating **inside** `retryPendingOnchainActions`, and the two-tx close/create all stand.

---

Ready for the re-gate. The owner decisions are unchanged except that **H2 widens the window** — the web-admin course toggles pause too, so the operator uses scripts. The instructor pubkeys remain the long-lead blocker and are independent of this spec.

thomgabriel:
## Re-gate (rev-2, `5a1ff62`) — PASS, with one implementation-note inside the H1 fix

Verified each HIGH fix is materially in the text, not just acknowledged:

- **H1 ✅** — new §3.1 correctly splits the consumers (6 → `liveLessonCount`, the finalize trigger → `activeLessons` + subset test), replaces `isAllLessonsComplete`, and generalises the rule. **Independently verified the load-bearing new claim:** `isAllLessonsComplete` has *exactly one* caller (`event-handlers.ts:453` — the only non-test reference besides its definition/import), so replacing rather than keeping-alongside is safe.
- **H2 ✅** — Phase-3 table now BLOCKs `/api/admin/courses/{sync,deactivate,reactivate}`; `achievements/sync` correctly identified as the only safe-to-leave-open admin route; quests/daily moved to "leave open (DB-only)". Risk register + Decision 3 (window widened) both updated.
- **H3 ✅** — Phase-4 precondition is now the *stronger* version: read each course's actual on-chain `lesson_count` with the v1 coder and reuse it exactly, explicitly **not** trusting the bundle's 12/16/12/12/12/12. Upgrade-in-place / same program-ID stated.
- **Factual corrections ✅** — quests/daily, the #432 scope shrink (`slots.json` already wired), and the false schema-sequencing rationale are all corrected. Test/CI surface + the stale Trident mirror added.

### One trap to carry into implementation — it lives *inside* the H1 fix

The §3.1 pseudo-code is:
```ts
isCourseComplete(lessonFlags, activeLessons) =>
  activeLessons.every((active, i) => (lessonFlags[i] & active) === active)
```
Taken literally this is **silently wrong for any lesson at bit index ≥ 32**, in the one predicate the spec says "must never be wrong." The decoder returns each `[u64;4]` word as a **BN** (raw BorshCoder), and JS `&` on `BN & BN` coerces through Number→ToInt32 → garbage above bit 31; even `number & number` truncates to 32 bits. The *existing* `bitmap.ts` is careful about exactly this — every helper does `BigInt(word.toString())` before the bitwise op (`bitmap.ts:27,40,61`). The replacement must do the same: convert both operands to **BigInt** before `&`.

This matters because it hides from the obvious test. Today's courses are ≤16 lessons (bits 0–15 of word 0), so a naive sparse-mask test that retires, say, lesson 3 passes on truncated math by luck. **The DoD's sparse-mask test must retire a slot at bit index ≥ 32** (i.e. exercise a >32-lesson course, or the high half of a word) or it won't catch the truncation it's meant to guard.

Suggested one-line addition to §3.1 and the DoD. Not a re-spin — an implementation note, since the pseudo-code is illustrative.

### Minor residual (low severity, same class, not a blocker)
`admin/resync/route.ts:162` calls `decodeLessonBitmap(flags, lessonCount)`, which walks a dense prefix `0..count-1`. Once a slot is retired this mislabels the retired index in the **admin resync display** (which lessons show as done). It's display-only in an operator path, not a finalize decision, so it's cosmetic — but it's the same prefix-vs-mask class as H1 and worth a known-residual line so it isn't mistaken for correct once masks go sparse.

**Verdict: rev-2 resolves all three HIGH findings correctly. Ready to implement** once the BigInt note is folded in (or carried as an implementation constraint) and the three owner decisions land — instructor wallets being the long-lead one. I'll re-verify H1's mask math and the gate placement against the actual diff when implementation PRs land, same as the #460 security bundle.

PR #464, opened 2026-07-13
fix(ai): drop unsupported Gemini schema keywords causing AI 502 (#463)

### LINKED ISSUE #463: AI assistant returns 502 (/api/ai/partner) — Gemini schema rejection wrapped as opaque 502
## Symptom
Using the AI assistant returns **502** (`Failed to load resource: 502`) from `POST /api/ai/partner`.

## Cause
`/api/ai/partner` proxies to Gemini and maps **any** non-2xx Gemini response to a generic `502 { error: "AI service unavailable" }` (`route.ts:281–291`). The real Gemini status/reason is `console.error`'d server-side but never surfaced, so the 502 is opaque.

The most likely trigger is the request's `responseSchema` (`lib/ai/partner-prompt.ts`): it uses `minimum`/`maximum` on `correctIndex`. Gemini's structured-output schema dialect is a **subset** of OpenAPI and rejects several numeric-constraint keywords with `400 INVALID_ARGUMENT`. Since `validatePartnerResponse` already enforces `correctIndex ∈ {0,1,2}` and exactly 3 options at runtime, those schema constraints are redundant and can be removed.

## Fix
- Drop the unsupported `minimum`/`maximum` constraints from the Gemini `responseSchema` (runtime validation already guarantees them).
- (Follow-up, optional) Surface the upstream Gemini status in logs/response so future failures aren't opaque.

## If 502 persists after the fix
Then it's a **config/quota** issue, visible in the Vercel log line `Gemini partner API error: <status> <text>`:
- 403 → the Generative Language API isn't enabled on the key's project, or the key is IP/referrer-restricted.
- 404 → model/region unavailable for the key.
- 429 → free-tier quota/rate limit.

## Acceptance
- [ ] A `propose`/`hint`/`ask` action returns a valid AI response (no 502)
- [ ] Malformed AI payloads are still rejected (runtime validation unchanged)

Potolski:
## Cause
`/api/ai/partner` wraps **any** non-2xx Gemini response into a generic `502`. The most likely trigger is the request `responseSchema`, which set **`minimum`/`maximum` on `correctIndex`** — numeric-constraint keywords that Gemini's structured-output schema dialect rejects with `400 INVALID_ARGUMENT`.

## Fix
- **Remove `minimum`/`maximum`** from the Gemini `responseSchema`. They're redundant: `validatePartnerResponse` already enforces `correctIndex ∈ {0,1,2}` and exactly 3 options at runtime, so output validity is unchanged.
- **Surface `upstreamStatus`** in the 502 body (just the HTTP status, not Gemini's raw text) so a *config-side* failure is diagnosable from the Network tab:
  - `403` → Generative Language API not enabled on the key's project, or the key is IP/referrer-restricted
  - `404` → model/region unavailable for the key
  - `429` → free-tier quota/rate limit

## Verification
- eslint ✓ on both files; `tsc` doesn't flag them (unrelated workspace module errors pre-exist locally). Runtime validation unchanged, so malformed AI payloads are still rejected.

## After deploy
Retry the assistant. If it **still 502s**, the Network-tab response now shows `upstreamStatus` — paste it (or the `Gemini partner API error:` log line) and it's a config/quota fix, not code.

### DIFF
--- a/apps/web/src/app/api/ai/partner/route.ts
+++ b/apps/web/src/app/api/ai/partner/route.ts
@@ -285,8 +285,11 @@ export async function POST(request: NextRequest) {
       // here) but Gemini never ran — refund so a failed call doesn't burn
       // one of the user's 4 paid assists.
       await refundAssist(user.id, lesson._id);
+      // Surface the upstream status (not Gemini's raw body) so a config-side
+      // failure (403 API-not-enabled / key-restricted, 404 model, 429 quota)
+      // is diagnosable from the Network tab, not just the server logs.
       return NextResponse.json(
-        { error: "AI service unavailable" },
+        { error: "AI service unavailable", upstreamStatus: response.status },
         { status: 502 }
       );
     }
--- a/apps/web/src/lib/ai/partner-prompt.ts
+++ b/apps/web/src/lib/ai/partner-prompt.ts
@@ -137,8 +137,9 @@ export const GEMINI_RESPONSE_SCHEMA = {
         },
         correctIndex: {
           type: "integer",
-          minimum: 0,
-          maximum: 2,
+          // NOTE: no `minimum`/`maximum` — Gemini's structured-output schema
+          // dialect rejects those numeric-constraint keywords with a 400. The
+          // 0–2 range is enforced at runtime in `validatePartnerResponse`.
           description: "Index (0-2) of the single correct option.",
         },
         explanation: {


Potolski:
Refined root cause from testing: **hint works, but `propose` and `ask` 502.**

That rules out the schema (Gemini accepted the request for `hint`) and points at the **per-intent `maxOutputTokens`** (`hint 160 / ask 240 / propose 500`). A full answer (`ask`) and especially the *entire updated file + 3-option check* (`propose`) exceed those caps, so Gemini truncates the JSON mid-generation and `JSON.parse` fails → the route's "invalid response" 502. `hint`'s short output fit in 160, which is exactly why only it worked.

Pushed a fix on this branch: caps raised to **hint 256 / ask 1536 / propose 4096** (flash-lite max output is 8192). The earlier schema cleanup + `upstreamStatus` surfacing stay in — both are still worthwhile — but **this token bump is the actual fix** for the propose/ask 502.

PR #465, opened 2026-07-13
fix(ai): raise AI partner output token caps (propose/ask 502)

Potolski:
## Follow-up to #464 — the actual fix for the propose/ask 502

#464 merged the schema cleanup + `upstreamStatus` surfacing, but testing showed the real cause: **`hint` works, `propose` and `ask` 502.**

## Cause
The per-intent `maxOutputTokens` were too low:

| action | old cap | result |
| --- | --- | --- |
| `hint` | 160 | short sentence → fits → **works** |
| `ask` | 240 | full answer → overflows → truncated JSON → **502** |
| `propose` | 500 | *entire updated file* + 3-option check → way over → **502** |

When Gemini hits the cap mid-generation it truncates the JSON, so `JSON.parse` throws and the route returns its "invalid response" 502. `hint`'s short output fit in 160 — which is exactly why only it worked.

## Fix
Raise the caps to **hint 256 / ask 1536 / propose 4096** (Flash-Lite max output is 8192, so ample headroom).

## Verification
- eslint ✓. Runtime validation unchanged.
- After deploy: **propose fix** and **ask a question** return full responses.

Refs #463.

### DIFF
--- a/apps/web/src/lib/ai/__tests__/partner-prompt.test.ts
+++ b/apps/web/src/lib/ai/__tests__/partner-prompt.test.ts
@@ -58,8 +58,12 @@ describe("partner-prompt", () => {
     expect(buildStaticPrefix(ctx)).not.toContain("let x = 1;");
   });
 
-  it("token caps are tight per intent", () => {
-    expect(maxTokensFor("hint")).toBeLessThanOrEqual(160);
-    expect(maxTokensFor("propose")).toBeLessThanOrEqual(500);
+  it("token caps grow with intent and stay within the model ceiling", () => {
+    // Ordered by how much output each intent needs — hint (a sentence) <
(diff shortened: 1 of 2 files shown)


Potolski:
Fixed the stale test in the same PR (`partner-prompt.test.ts`): the old assertions hardcoded the previous caps (`hint <= 160`, `propose <= 500`). Restated as a looser invariant — **`hint < ask < propose` and `propose <= 8192`** (Flash-Lite's output ceiling) — so it guards the ordering + model-limit intent without breaking on future cap tuning. Verified: 256 < 1536 < 4096 <= 8192. eslint clean.

Potolski:
@claude check if the blocking issue was fixed

PR #469, opened 2026-07-13
feat(program): WS-1 Phase 0 — v-next program (drop min_completions, flat uncapped creator reward)

thomgabriel:
**Phase 0 of WS-1** (spec: `docs/superpowers/specs/2026-07-13-program-vnext-design.md`, merged #462). Program source + fixtures + IDL regen. **No client changes** — the client-side IDL cutover and the length-aware decoder are Phase 1.

## What changed
- `min_completions_for_reward` (u16) removed from **all three layout surfaces**: `Course`, `CreateCourseParams`, `UpdateCourseParams`.
- `finalize_course`: the `CREATOR_REWARD_WINDOW` const and both threshold clauses are gone. The reward is now `if course.creator_reward_xp > 0 { mint }` — flat, every completion.
- `Course::SIZE` **255 → 253**. `_reserved` is **not** padded back to 255 — that would collide with the never-deployed v2 layout and break the client's length-based decoder dispatch.

## DoD (gate-verifiable)
- [x] Field gone from the 3 layout surfaces + the finalize window (const + both clauses).
- [x] `Course::SIZE == 253`; `_reserved` unchanged at `[u8; 8]`.
- [x] IDL regenerates clean from this source (`anchor build`; confirmed no `min_completions_for_reward` in the `Course` account). The committed **client** IDL is deliberately untouched — Phase 1.
- [x] CI-breakers ride along: `test_course.rs` SIZE assert flipped 255→253; the **Trident fuzz mirror** (`fuzz_0/types.rs`) regenerated — it hand-copied the layout and was *already stale* (still had `lesson_count:u8` from before CS-3's `active_lessons` migration), so it was fuzzing a struct that no longer exists.
- [x] Behavior tests that asserted the old threshold/window rewritten to the flat rule (a course that paid the creator 0 below-threshold now pays on the first completion).
- [x] 130 Rust tests pass; `cargo check` + `anchor build` clean; `grep` for the field shows only comments.

## Notes for the gate
- Two pre-existing script bugs were fixed opportunistically (both blocked `tsc`): `fetch-course.ts` printed a `lessonCount` field dead since CS-3; `create-mock-course.ts` was missing `collection`.
- The uncapped creator reward is the owner-accepted risk from the epic; the compensating control (completion-route rate limit) already shipped in #460.

Built via SDD (implementer subagent for the fixture sweep + IDL regen; controller reviewed the diff). Requesting the full adversarial pass on the complete diff, same discipline as #460.

PR #470, opened 2026-07-14
feat(web): WS-1 Phase 1 — length-aware Course decoder + consumer migration (H1, S6)

thomgabriel:
**Phase 1 of WS-1**, stacked on Phase 0 (#469) — base is the Phase-0 branch, so this diff is Phase 1 only. Merge #469 first, then this.

The client half of the migration: make Course reads survive the program's byte-layout change **before** any deploy, so the client/program version skew (#449) can never silently misdecode.

## What's here
- **Length-aware normalising decoder** (`academy-reads.ts`): `fetchCourse`/`decodeCourse` dispatch on account length — **224 → v1**, **253 → v-next**, anything else **throws** (an unknown layout must never decode). Both return one shape: `activeLessons` (the 256-bit mask; synthesised dense from `lesson_count` on v1) + `liveLessonCount` (its popcount; read straight from `lesson_count` on v1 so it's bit-identical to today). New `superteam_academy_vnext.json` IDL committed for the v-next coder; the v1 IDL stays for the write path.
- **H1 — the finalize trigger now mirrors the chain.** `event-handlers.ts` gated completion with a dense-PREFIX check (`isAllLessonsComplete`), which diverges from `finalize_course`'s SUBSET test (`flags & active == active`) the instant a slot is retired — stranding a finished learner or firing a doomed, fee-burning finalize. It now calls `isCourseComplete(lessonFlags, course.activeLessons)`, bit-identical to the program. `isAllLessonsComplete` is **removed** (it had exactly one caller — the bug). Sparse-mask tests cover both failure directions.
- **Six count-consumers** → `liveLessonCount`, dropping the dead `|| 1` / `?? 0` fallbacks.
- **S6** — `resolvers.ts` decoded Course with its own fixed coder outside `academy-reads`; now routed through `decodeCourse` (fails loud into its existing catch instead of silently misreading a `course_id`).
- The transitional `lesson_count?: undefined` shim is gone — a future `.lesson_count` access is a compile error again.

## Behaviour-neutral on the live chain
Every account on devnet is 224 bytes today → the v1 branch → `liveLessonCount === old lesson_count`. No consumer changes what it computes until the program actually deploys v-next. This PR is safe to merge and deploy the *frontend* ahead of the program.

## DoD
- [x] 224/253 dispatch; unknown length throws. New vnext IDL committed.
- [x] **6** count-consumers read `liveLessonCount`; no `|| 1` / `?? 0` on a chain-read count remains.
- [x] Finalize trigger reads `activeLessons` + subset test; `isAllLessonsComplete` (prefix) removed, not merely unused. **(H1)**
- [x] A test proves correctness against a **sparse** mask (finished-every-live-slot finalizes; missing-a-live-slot does not; a stray retired-slot bit is ignored).
- [x] S6 `resolvers.ts` on the length-aware path.
- [x] tsc clean; 597 tests pass.

Built via SDD: decoder (Task 1a) reviewed + committed first, then the finalize-trigger / count-consumers / S6 fanned out to three concurrent implementers on disjoint files; controller integrated (shim removal + full typecheck/suite). Requesting the full adversarial pass, same as #469.

### DIFF
--- a/apps/web/src/app/api/admin/courses/sync/route.ts
+++ b/apps/web/src/app/api/admin/courses/sync/route.ts
@@ -445,11 +445,11 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
   // Grow the on-chain lesson count when a teacher has appended lessons to an
   // already-deployed course. Increase-only: without this, a new lesson's index
   // is >= course.lesson_count and complete_lesson reverts (LessonOutOfBounds,
-  // surfaced as a 409 in /api/lessons/complete). fetchCourse uses the raw-IDL
-  // BorshCoder → snake_case `lesson_count`. The program rejects any decrease
+  // surfaced as a 409 in /api/lessons/complete). fetchCourse returns the
+  // normalised `liveLessonCount`. The program rejects any decrease
   // (LessonCountDecrease), so we only send the field when Sanity is strictly
   // higher; equal/lower is left untouched.
-  const onChainLessonCount = onChainCourse?.lesson_count as number | undefined;
+  const onChainLessonCount = onChainCourse?.liveLessonCount;
   if (
     typeof course.lessonCount === "number" &&
     typeof onChainLessonCount === "number" &&
--- a/apps/web/src/app/api/admin/courses/sync/route.ts
+++ b/apps/web/src/app/api/admin/courses/sync/route.ts
@@ -449,6 +449,9 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
   // normalised `liveLessonCount`. The program rejects any decrease
   // (LessonCountDecrease), so we only send the field when Sanity is strictly
   // higher; equal/lower is left untouched.
+  // Count-based and only correct while masks are dense (today). Once a
+  // sparse-mask course can exist, this must become a per-slot bit test
+  // against `activeLessons` instead of a count comparison.
   const onChainLessonCount = onChainCourse?.liveLessonCount;
   if (
     typeof course.lessonCount === "number" &&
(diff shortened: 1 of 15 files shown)


thomgabriel:
## Fixed in \`bfdcd69\` — both the blocking finding and the deferred marker

**The min_completions half-migration (you flagged blocking; the adversarial pass confirmed it MEDIUM).** You were right, and the root cause was the `as RawCourse` cast over a shadow interface. Rather than carry the field through `DecodedCourse` (it's *deleted* on-chain — carrying it would be preserving a corpse), I retired it from the drift path entirely: removed from `OnChainCourse`, `toOnChainCourse`, and the `diffCourse` clause. The content-side schema keeps it (harmless; removed in the later content cleanup).

**Root-cause hardening (your non-blocking suggestion, taken).** Deleted the `RawCourse` shadow interface and typed `raw` as `DecodedCourse` directly, no cast — so the exact thing that hid this (`as` over a hand-maintained mirror) is gone. Verified empirically: removing any field from `DecodedCourse` now fails `tsc` at this call site instead of silently reading `undefined`. The test's mock fixture carried a phantom `min_completions_for_reward` that masked the bug; removed and aligned to the real shape.

**The LOW deferred OOB guard.** Both `lessons/complete` and `courses/sync` use `lessonIndex >= liveLessonCount`, which is byte-correct on today's dense chain but would wrongly 409 a valid high-index live slot once masks go sparse. Left the logic (mask-aware refinement is deferred per the spec), but added a marker comment at both sites stating the constraint: it must become a per-slot bit test before a sparse-mask course can exist.

Everything else in your review held up — the decoder dispatch, `isCourseComplete` being bit-identical to `finalize_course`, the dense-mask behaviour-neutrality, and the `|| 1` drop (safe because `create_course` enforces `lesson_count > 0`). 597 tests, tsc clean.

@claude re-review the head \`bfdcd69\` — your prior verdict was blocking on the min_completions issue, now fixed.

thomgabriel:
Superseded by #471. This PR auto-closed when #469's branch (its base) was deleted on merge; #471 relands the identical Phase-1 content — cherry-picked clean onto current main (Phase 0 = #469, plus the AI fixes that landed meanwhile). Full gate lineage from here carries forward.

Issue #377, opened 2026-07-10
db: trigger functions handle_new_user + rls_auto_enable executable via REST by anon/authenticated (SECURITY DEFINER)


thomgabriel:
Supabase security advisor (pre-existing, surfaced during CS-7 migration verification): public.handle_new_user() and public.rls_auto_enable() are SECURITY DEFINER and EXECUTE-able by anon + authenticated via /rest/v1/rpc/. They are trigger functions and should never be client-callable. get_leaderboard is intentionally public — exclude it.

Done when: REVOKE EXECUTE on both from anon/authenticated/public (keep trigger ownership working), advisor no longer lists them.

thomgabriel:
Done — merged as #474 (`74c9334`). Independent gate live-verified on prod: `handle_new_user` anon/authenticated EXECUTE = false, trigger intact, service_role retained; `rls_auto_enable` confirmed nonexistent; `schema.sql` REVOKE added for rebuild-safety.

PR #474, opened 2026-07-14
fix(db): revoke public EXECUTE on handle_new_user (#377) [SENSITIVE — migration applied to prod]

thomgabriel:
**WS-3 · SENSITIVE (DB).** Migration-before-code: **already applied to prod (`pywhtmidcrptomrabbrw`) and verified** before this PR, per the WS-3 handoff.

## What
`handle_new_user` is a SECURITY DEFINER trigger function (seeds `profiles` + `user_xp` on signup). Postgres grants EXECUTE to PUBLIC by default on function creation, so it was directly callable at `/rest/v1/rpc/handle_new_user` by `anon` + `authenticated`. A trigger function must never be client-callable.

## Applied + verified on prod (before this PR)
- Before: `proacl` held the PUBLIC grant; `anon`/`authenticated` EXECUTE = true.
- After the `REVOKE`: ACL is `{postgres=X/postgres,service_role=X/postgres}`; `anon`/`authenticated`/`public` EXECUTE = **false**; `service_role` retained.
- The signup trigger still fires — trigger functions run regardless of the caller's EXECUTE grant; only direct RPC is blocked.

## This PR (the code half)
- The migration file (matches what was applied to prod, for repo tracking — avoids the untracked-drift class).
- **The same REVOKE added to `schema.sql`** — the default PUBLIC grant returns on any from-scratch rebuild, so without this a fresh apply would silently re-expose it.

## Scope corrected by live introspection
- The issue also named `rls_auto_enable` — it **does not exist** on this database (verified via `pg_proc`). No action.
- `get_leaderboard` is flagged by the advisor too but is **intentionally public** (leaderboard API) — deliberately left executable, per the issue.

**Done when:** REVOKE applied, advisor no longer lists `handle_new_user`. Direct privilege check confirms it; the advisor cache refreshes on its own cycle.

SENSITIVE → requesting the full adversarial pass (esp: does the REVOKE break the signup trigger? — it doesn't, and that's the thing to verify).

### DIFF
--- a/supabase/migrations/20260714120000_revoke_execute_handle_new_user.sql
+++ b/supabase/migrations/20260714120000_revoke_execute_handle_new_user.sql
@@ -1,0 +1,20 @@
+-- #377: handle_new_user is a SECURITY DEFINER trigger function (fires on
+-- auth.users INSERT to seed public.profiles + public.user_xp). Postgres grants
+-- EXECUTE to PUBLIC by default on function creation, exposing it directly at
+-- /rest/v1/rpc/handle_new_user to anon + authenticated. A trigger function must
+-- never be client-callable — revoke it.
+--
+-- Trigger execution is unaffected: a trigger runs its function regardless of the
+-- invoking role's EXECUTE privilege; EXECUTE only gates direct RPC calls.
+-- postgres (owner) + service_role retain EXECUTE.
+--
+-- Verified against prod (pywhtmidcrptomrabbrw): before, proacl held the PUBLIC
+-- grant (=X/postgres) and anon/authenticated could execute; after, the ACL is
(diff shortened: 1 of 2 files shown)


thomgabriel:
**Independent human-gate: PASS — live-verified on prod (`pywhtmidcrptomrabbrw`).** Confirmed against the actual DB, not the PR text: `handle_new_user` anon/authenticated EXECUTE = **false**, service_role retained, still SECURITY DEFINER; the signup trigger on `auth.users` is **still present** (1) so signup keeps seeding profiles; `rls_auto_enable` genuinely **does not exist** (count 0 — scope correction valid, not just claimed); `get_leaderboard` correctly left public. Migration SQL + `schema.sql:511` REVOKE both correct and idempotent. Merging as bookkeeping of an already-applied, verified-correct prod change.

Issue #349, opened 2026-07-09
Deploy UX: too many wallet 'confirm transaction' prompts during program deploy


Potolski:
## Problem
Deploying a program in a buildable challenge ("Deploy to Devnet") makes the wallet pop up **"confirm transaction" ~6+ times** in a row. For a first-time Solana user this is confusing and looks broken/spammy — a bad first impression at exactly the moment we want a "wow, my program is on-chain" payoff.

## Why it happens
A program deploy uploads the compiled `.so` to a buffer account in ~1 KB chunks (BPF Loader Upgradeable flow), then finalizes. In `packages/deploy/src/deploy.ts` the chunks are signed in **batches** via `wallet.signAllTransactions(batch)` — and **each batch is one wallet prompt**. So the prompt count is roughly:

```
1 (create buffer) + ceil(totalChunks / BATCH_SIZE) + 1 (deploy)
```

The chunk-upload reliability work in #348 (priority fees so txs land, plus **`BATCH_SIZE` 50 → 20** to stay under a rate-limited devnet RPC) reduced dropped txs but **increased the number of batches → more prompts**. So there's a direct tension: smaller batches = more reliable but more wallet popups.

## Directions to explore
1. **Fewer, bigger signing batches.** Now that priority fees make txs land quickly, raise `BATCH_SIZE` back up (e.g. 50–100) so ~all chunks are covered by 1–2 prompts. Measure against the blockhash window (a batch must confirm within ~90s) and the RPC's tolerance for the bigger send burst.
2. **One signature for the whole deploy, ideally.** Investigate signing every tx in a single `signAllTransactions`. Blocker: a shared recent blockhash expires (~90s), so a multi-minute upload can't all be pre-signed with one blockhash. A **durable nonce** removes expiry and could enable a true one-prompt deploy — worth prototyping.
3. **Set expectations in the UI.** Regardless of count, tell the user up front what's happening: e.g. "Uploading your program to Solana — you'll approve a few transactions (this is normal for program deploys)", show `Approving upload 2/4`, and a friendly explanation of *why* (chunked on-chain upload). This alone turns "is this broken?" into "ah, makes sense."
4. **Consider a session/auto-approve affordance** for wallets that support it, so the batches don't each interrupt.

## Acceptance criteria
- [ ] A typical "Deploy to Devnet" completes with materially fewer wallet prompts (target: 1–3), OR
[...]

PR #476, opened 2026-07-14
feat(deploy): fewer wallet prompts in learner program-deploy (#349)

thomgabriel:
**WS-3 · SAFE.** Reduces the burst of wallet prompts when a learner deploys their compiled program to devnet.

## What
- **BATCH_SIZE 20 → 50** (`packages/deploy/src/deploy.ts`). #348 had throttled it to 20 while adding priority fees + poll-and-rebroadcast; now that those are in place, 50 is safe and cuts a ~100–250 KB program from ~6–14 upload prompts to ~3–6.
- **Accurate batch prompts** — a new `onBatchStart({batchNumber, totalBatches})` callback replaces the old `chunkCurrent % 30` heuristic (which was misaligned with the real batch size) with real-time "Approve batch N of M", shown only when there's >1 batch. Plus an upfront explainer in the ready state so a few batched approvals aren't a surprise.

## Blockhash-window reasoning (why 50, not 100+)
A recent blockhash is valid ~150 slots (~60–90s). At 50/batch: send-stagger ≈ 2.5s, typical confirm ≈ 1–2s, pessimistic (1–2 rebroadcast cycles) ≈ 8–12s — comfortable margin. Not pushed to 100+ because `confirmBatch`'s rebroadcast resends the unconfirmed set **sequentially**, so a bad pass at 100+ could burn most of the window with no room for a second pass. 50 is the largest size where a full-batch resend still fits.

## #348 reliability preserved
`SEND_DELAY_MS`, the priority-fee instruction, and `confirmBatch`'s poll-and-rebroadcast (4s cadence, 90s timeout, fresh-blockhash-per-batch) are **untouched** — the diff confirms they appear only in comments, not changed code. Only `BATCH_SIZE` + the new callback moved.

## i18n
`deploy.deployment.batchExplainer` + `approvingBatch` (parameterized) added to en/pt-BR/es; unused `batchSigning` removed. Parity test green.

tsc clean, 597 tests pass.


thomgabriel:
**Gate: PASS (SAFE-lane).** Verified: i18n parity — `approvingBatch`/`batchExplainer` present in en/es/pt-BR (no MISSING_MESSAGE risk). BATCH_SIZE 20→50: the reliability mitigations (priority fees, poll-and-rebroadcast, fresh-blockhash-per-batch, SEND_DELAY_MS) are untouched — confirmed only BATCH_SIZE + the onBatchStart callback changed — and the in-code blockhash-window budget analysis (deploy.ts:113-140) is sound (≈2.5s to send 50, full ~90s window per fresh blockhash). One note: the 50-batch is reasoning-validated, not load-tested; the authoritative proof is the Phase-4 devnet deploy smoke-test, which drives this path. Reasonable increase, not reckless.

Issue #382, opened 2026-07-10
ci: SHA-pin GitHub Actions across all workflows (supply-chain hardening)


thomgabriel:
Surfaced by the adversarial review of #379 (non-blocking, LOW).

All workflows pin actions to **floating major tags** (`actions/checkout@v4`, `pnpm/action-setup@v4`, `actions/setup-node@v4`, etc.) rather than immutable commit SHAs. A compromised or force-repushed tag on a first-party action could inject code into CI.

**Blast radius today is small** — the actions are first-party (GitHub/pnpm), and the sensitive workflow (`validate-content.yml`) already runs with a read-only `GITHUB_TOKEN` and references no secrets, so this is hardening, not an active hole. Deferred out of #379 because it is a repo-wide policy change spanning `ci.yml`, `lighthouse.yml`, `claude-code-review.yml`, and `validate-content.yml`, not a CS-2 concern.

**Done when:** every `uses:` in `.github/workflows/*.yml` is pinned to a 40-char commit SHA (with the human-readable version in a trailing comment), and a `# renovate:`/dependabot config keeps them updated.

thomgabriel:
Done — merged as #479 (`20c4c8f`). All 7 GitHub Action pins SHA-pinned across 6 workflows + dependabot github-actions ecosystem added. Independent gate verified all 7 SHAs resolve to their claimed version tags on GitHub.

Issue #381, opened 2026-07-10
content-lint gate13c: contain idl path with realpathSync — unguarded symlink allows out-of-tree read (~30-byte non-secret leak)


thomgabriel:
From the independent adversarial review of PR #379 (merged). NON-BLOCKING — capped to ~30 bytes of non-secret host content by validate-content.yml's hardening (contents: read, no secrets in the linter step env, persist-credentials: false), so no secret is reachable. But worth closing before this gates the public academy-courses repo.

**Primitive** (packages/content-lint/src/checks/gate13bcd-widgets.ts:13,20): `checkIdl` does `JSON.parse(readFileSync(join(root, dir, idlRel)))` and, on parse failure, interpolates `err.message` into a gate-13c diagnostic that reaches the world-readable CI log (cli.ts:16,33). On Node 20 (CI), V8's JSON.parse error echoes a ~30-char window of the file. `relativePath` (content-schema base.ts) blocks `../`/absolute at the schema layer — airtight — BUT validates the string only; nothing does lstat/realpath/readlink anywhere in content-lint or content-schema, and readFileSync follows symlinks. A teacher can commit `program.idl.json` as a symlink to an out-of-tree target (passes relativePath as a plain string), and gate13c reads + partially echoes the target.

Fix (either): (a) in checkIdl, `realpathSync` the resolved path and reject if it escapes `join(root, dir)` before reading; OR (b) emit a generic "program.idl.json does not parse" diagnostic without interpolating err.message. (a) is stronger — also close the same symlink gap in walkFiles/statSync (loader.ts:37) if content is ever read there.

Done when: a symlinked idl (or any content file) pointing outside the lesson dir is rejected, not read.

thomgabriel:
Done — merged as #479 (`20c4c8f`). content-lint IDL-path containment via lstat-not-stat + realpath, generic parse errors (no err.message leak), 6 new tests (adversarially verified to discriminate against pre-fix code).

Issue #376, opened 2026-07-10
deps: 123 dependabot vulnerabilities on default branch (3 critical, 46 high)


thomgabriel:
GitHub reports 123 open dependabot alerts on main (3 critical, 46 high, 60 moderate, 14 low): https://github.com/solanabr/superteam-academy/security/dependabot — surfaced on every push. CI's 'Dependency audit' job passes, so its bar and dependabot's disagree; triage the criticals/highs first (many are likely dev-only or unreachable, but that must be established, not assumed).

Done when: criticals and highs are each either patched or explicitly dismissed with a reason; alert count reflects that.

thomgabriel:
**Partial — code half merged as #479 (`20c4c8f`).** 29 crit/high transitive alerts patched via `pnpm.overrides` (Dependency-audit CI green). **Remaining is NOT code:** 13 alerts are stale (flagged packages genuinely absent from the lockfile — independently verified; `decompress` critical has no upstream patch → dismiss-with-reason) and need an **owner dismissal in the GitHub Security tab**; 8 `next` highs are deferred to the Next-15 migration (WS-3.5). Keeping open until the 13 dismissals + Next-15 land.

thomgabriel:
**Orchestrator: triage complete — parking as needs-human for the two actions only you can take.**

1. **PR #684** (fast-uri override bump, the one genuinely-patchable crit/high) — verified 1304/1304 + prod build; merging on green.
2. **Bulk dismissal** — the full ledger + a ready-to-run script are in #684 PR body (45 alerts: not_used / inaccurate-already-patched / not-reachable, each with evidence). The API accepts your token (probe-dismissal of the decompress critical succeeded). Run the script when you agree with the ledger; the already-patched group may also auto-close when GitHub re-parses the lockfile post-#684.
3. **The 6 real residual highs are next@14** — decision needed: upgrade (epic filed as #685) vs dismiss as tolerable_risk until post-launch.

Done-when status after your bulk run: criticals 3→0, highs 50→6 (all next@14, covered by your #685 decision).

thomgabriel:
**Owner-authorized dismissal pass EXECUTED (2026-07-27).** Result: **criticals 4 → 0, highs 51 → 0.** Remaining board is 65 medium / 15 low, tracked in #820.

## What was dismissed (55 alerts, each API-verified `state=dismissed`, evidence re-checked against pnpm-lock.yaml @ea8d4cf)
| Class | Count | Reason | Evidence basis |
[...]

Issue #139, opened 2026-06-23
[G-2] Security hardening complete


Potolski:
> **Code legend / glossary:** see [`docs/TASK-CODES.md`](https://github.com/solanabr/superteam-academy/blob/main/docs/TASK-CODES.md) for how to read task codes, severities, and area labels.

Security hardening complete: all P0/P1 closed with regression tests; fuzzing + money-path tests pass; verifiable build in place.

**Done when:** every P0/P1 is closed and the security test suite is green.

thomgabriel:
Orchestrator audit disposition (full audit run 2026-07-27; David's items released to the loop, but this one stays untouched): **BLOCKED — deploy-time closing checklist, owner-held.** The technical substrate is DONE and verified: Trident fuzzing workflow live (`fuzz.yml`, real compiled Pinocchio .so, exit-99 on violation), money-path tests assert exact XP-mint balances across all 18 instructions (LiteSVM differential suite, in CI), verifiable build in place (see #140). No open P0/P1 is an on-chain security fix — #376 is explicitly carved out of G-2 per the launch-readiness epic design, #600 is a product decision, #725/#599/#606 are ops/authoring. This gate flips only after mainnet deploy + hash recording, on explicit owner confirmation. No loop action.

PR #479, opened 2026-07-14
fix(ci): supply-chain hardening — SHA-pin Actions, content-lint symlink guard, dependabot overrides (#382 #381 #376) [SENSITIVE]

thomgabriel:
**WS-3 · SENSITIVE (ci/deps).** Adversarial-reviewed; both findings fixed. This is HARDENING and must **not** gate #139/G-2.

## #382 — SHA-pin all GitHub Actions
Every `uses:` across the 6 workflows pinned to a 40-char commit SHA with a `# vX.Y.Z` comment. New `.github/dependabot.yml` adds a weekly `github-actions` ecosystem so the pins don't rot. No workflow logic changed — only `uses:` refs.
- **All 7 SHAs verified against GitHub** (dereferencing annotated tags to commits). The adversarial pass caught one: `Swatinem/rust-cache` was pinned to the `v2` *major*-tag SHA while commented `# v2.9.1` — behaviorally identical (CHANGELOG-only delta) but a false provenance comment. **Fixed** — repinned to the true v2.9.1 commit `c19371144df3…` (confirmed via the tag's own commit message).

## #381 — content-lint symlink guard
`checkIdl` now `realpathSync`-canonicalizes the requested idl path *and* the lesson dir, and rejects (without reading) anything whose resolved path escapes the lesson dir — with a trailing-separator anchor so `/lessons/foo-evil` can't prefix-match `/lessons/foo`. The read uses the already-resolved path (TOCTOU-safe), and the parse-failure diagnostic no longer interpolates `err.message` (byte-echo closed). `loader.ts` `walkFiles` now `lstatSync`-skips symlinks (files and dirs) so discovery never follows one.
- 6 new tests. The adversarial pass found one was non-discriminating (its fixture threw "Unexpected end of JSON input" — no bytes to echo — so it passed even against the leaking code). **Fixed** — the fixture now uses invalid JSON that echoes source bytes, and was proven to fail against reverted source.

## #376 — dependabot criticals/highs
The live count is larger than the issue stated: **50 open crit/high** (4 critical + 46 high), of which **8 are `next`** (deferred to the Next-15 migration / WS-3.5 — `next` untouched here).
- **29 patched** via root `pnpm.overrides` — each verified in the lockfile that the vulnerable version is gone (vite→6.4.3, ws→7.5.11/8.21.0, undici→7.28.0, form-data→4.0.6, protobufjs, shell-quote, lodash→4.18.1, minimatch, picomatch, …). `pnpm install --frozen-lockfile` clean; 748 tests pass.
- **13 need an owner dismissal, not code** — the flagged packages (`decompress`, `tar`, `lodash-es`, `@babel/plugin-transform-modules-systemjs`, `minimatch 5.x/10.x`, `undici 6.x`) are **genuinely absent from the lockfile** (verified by grep; stale alerts needing a GitHub rescan). `decompress` has **no upstream patch**, so it can only be dismissed-with-reason — which #376's "done when" explicitly allows.

## Owner action needed
The 13 "not_used" alerts and the 8 `next` alerts can't be closed by code — they need a Dependabot **dismiss-with-reason** (not_used / deferred-to-Next-15) in the GitHub UI. Everything patchable is patched here.

748 tests pass; typecheck clean; `pnpm install --frozen-lockfile` consistent.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Keep SHA-pinned GitHub Actions (see .github/workflows/*.yml) current.
+  # Dependabot opens a PR bumping the pinned SHA + version comment whenever
+  # an action publishes a new release — the pins never silently rot.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,13 +38,13 @@ jobs:
     # heavy jobs so adding a label doesn't re-run the whole matrix.
     if: github.event.action != 'labeled'
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
       - name: Install pnpm
-        uses: pnpm/action-setup@v4 # version read from package.json "packageManager"
+        uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4.3.0, version read from package.json "packageManager"
 
       - name: Setup Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: ${{ env.NODE_VERSION }}
           cache: pnpm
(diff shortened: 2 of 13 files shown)


thomgabriel:
**Independent SENSITIVE gate: PASS.** Verified the load-bearing risk myself — **all 7 action SHA-pins resolved against GitHub and match their version comments** (actions/checkout v4.3.1, setup-node v4.4.0, upload-artifact v4.6.2, claude-code-action v1.0.173, pnpm/action-setup v4.3.0, **Swatinem/rust-cache v2.9.1** [the fixed provenance — c1937114 genuinely IS v2.9.1], dtolnay/rust-toolchain stable-pin real commit). Zero hallucinated SHAs, zero provenance mismatches. Symlink guard (#381) uses lstat-not-stat + realpath containment — sound. Overrides (#376) validated by the passing Dependency-audit CI job (lockfile resolves, no vuln regression). Merging.

PR #485, opened 2026-07-14
feat(db): public_profiles view — by-wallet public read (#478 B2) [SENSITIVE — applied to prod]

thomgabriel:
**#478 · B2 · SENSITIVE (DB).** Migration-before-code: **applied to prod (`pywhtmidcrptomrabbrw`) and verified** before this PR.

## What
A by-wallet public view so instructor identity can be resolved from a course's `creator` wallet — exposing **only** `wallet_address, username, avatar_url, bio, social_links`, filtered to `is_public = true AND deleted_at IS NULL`.

## Why a view (not security_invoker, not a broad RLS policy)
Mirrors the `public_user_xp` house pattern: an owner-rights view over the RLS-locked `profiles` table where **the SELECT list is the column filter**. So `google_id` / `github_id` / `deleted_at` / `deletion_requested_at` are never exposed, and `profiles` itself stays own-row-only for direct PostgREST access. A `security_invoker` view would either see nothing (RLS) or require an RLS SELECT policy on `profiles` that would over-expose columns via `/rest/v1/profiles?select=google_id`. The view is the safe shape.

## Introspected live first (schema.sql was stale)
`profiles` live columns confirmed the 4 safe fields + `wallet_address` key + `is_public`/`deleted_at` filters, and the sensitive columns to exclude (`google_id`, `github_id`, `deletion_requested_at`).

## Verified on prod (before this PR)
- **Exposed columns:** `wallet_address, username, avatar_url, bio, social_links` — no sensitive column (asserted `false` against the full sensitive set).
- **Grants:** `anon:SELECT`, `authenticated:SELECT` — SELECT-only (tighter than the legacy broad grant on `public_user_xp`).
- **Filter:** view rows == `public & not-deleted & has-wallet` count (3 == 3); **0** leaked deleted/private rows.

## Out-of-scope finding
Introspection revealed `schema.sql`'s `public_user_xp` is missing the `AND deleted_at IS NULL` filter that the **live** view has — repo drift, not a live hole. Flagged for a separate fix; not touched here.

This PR is the repo mirror (migration file + `schema.sql`) of what's already live. SENSITIVE → full gate.

Closes #478.

### DIFF
--- a/supabase/migrations/20260714130000_public_profiles_view.sql
+++ b/supabase/migrations/20260714130000_public_profiles_view.sql
@@ -1,0 +1,27 @@
+-- B2 (#478): by-wallet public read of the four non-sensitive profile fields.
+-- Mirrors the public_user_xp house pattern — an owner-rights view over the
+-- RLS-locked profiles table. The view's SELECT list is the column filter, so
+-- google_id / github_id / deleted_at / deletion_requested_at etc. are NEVER
+-- exposed; profiles itself stays own-row-only for direct PostgREST access (no
+-- broad RLS policy is added, which is why a view and not security_invoker).
+-- Rows are limited to public, non-deleted profiles that have a wallet to key on.
+--
+-- Applied + verified on prod (pywhtmidcrptomrabbrw) migration-before-code:
+--   exposed columns = wallet_address, username, avatar_url, bio, social_links
+--   grants          = anon:SELECT, authenticated:SELECT (SELECT-only)
+--   filter          = view rows == public & not-deleted & has-wallet; 0 leaked
+--                     deleted/private rows; no sensitive column exposed.
+CREATE OR REPLACE VIEW public.public_profiles AS
+SELECT
+  p.wallet_address,
+  p.username,
+  p.avatar_url,
+  p.bio,
+  p.social_links
+FROM public.profiles p
+WHERE p.is_public = true
+  AND p.deleted_at IS NULL
+  AND p.wallet_address IS NOT NULL;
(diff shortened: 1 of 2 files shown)


thomgabriel:
Adversarial pass done. **The view itself is clean** — all 5 checks pass (exactly the 5 columns, correct filter, SELECT-only, definer-consistent with public_user_xp, unambiguous by-wallet lookup via the UNIQUE index, 0 leaked deleted/private rows). Two actions taken:

- **Fixed (this PR):** my migration/schema comment falsely claimed the view protects google_id/github_id and that profiles is own-row-only. It doesn't and isn't — corrected to state the view is a *curated projection*, not the security boundary.
- **Filed #486 (pre-existing, not this PR):** the base `profiles` table already lets anon read google_id/github_id of public rows (row-level RLS + wide column grants). NULL today (no OAuth linked on seed profiles) but live once anyone links. The view is a strict subset — recommend #486 as its own SENSITIVE base-table hardening PR.

thomgabriel:
**Independent gate: PASS — live-verified on prod (`pywhtmidcrptomrabbrw`).** `public_profiles` exposes exactly `wallet_address, username, avatar_url, bio, social_links` (confirmed via information_schema on the live DB, not the migration text), filtered `is_public AND deleted_at IS NULL AND wallet_address IS NOT NULL`, SELECT-only grant to anon/authenticated. Strict subset, no sensitive columns. Merging (sensitive change already applied to prod; PR is repo-tracking + schema.sql).

PR #488, opened 2026-07-14
fix(auth): refuse sign-in on deleted accounts (#461)

### LINKED ISSUE #461: Account deletion is not honoured at login — profiles.deleted_at is never checked
Found while completing #410 (PR #460). **Pre-existing and orthogonal to that PR** — filing rather than folding in, because the fix belongs in the auth flow, not the deletion route.

## What

`/api/account/delete` soft-deletes and anonymises a profile: it sets `deleted_at`, `is_public = false`, scrubs the identifiers, and signs the user out. Public reads correctly drop the row.

But **nothing checks `profiles.deleted_at` on the way back in.** `profiles.id` is `auth.users.id`, and the deletion route never touches `auth.users` — so the same auth identity resolves to the same (now-tombstoned) profile row on the next sign-in.

The result is a zombie account: the user is returned to a row that is invisible to every public read, has a `deleted-user-xxxx` username, and cannot be cleaned up. They also cannot obtain a fresh account, because the auth identity is already taken.

## Why it wasn't obvious

The soft-delete design is right — on-chain XP and credential NFTs are immutable and bound to the wallet, so hard-deleting the row would orphan them. The gap is only that "deleted" is enforced at *read* time and never at *auth* time.

## Done when

Signing in against a `deleted_at` profile does something deliberate and documented, rather than silently resurrecting a tombstone. Options worth weighing:

- **Refuse**, with a clear message and a support path.
- **Offer restore** — an explicit un-delete, since the data is all still there.
- **Issue a genuinely fresh identity**, which means also releasing the `auth.users` row.

The choice is a product call (what should "delete my account" mean when the on-chain half is immutable?), so it wants a decision before code.

Refs #410, PR #460.

thomgabriel:
**#461 · SAFE (auth-adjacent).** A deleted user who signs back in was silently returned to their tombstoned row (zombie account). Now refused with a localized message.

## Two guard sites (shared helper)
`isAccountDeleted(userId)` (`lib/auth/account-status.ts`) reads `profiles.deleted_at` via the admin client (the own-row SELECT policy is deliberately unfiltered so a just-deleted user's `/settings` still loads — so RLS can't be trusted here). **Fails open** (`false`, logged) on any error — this runs on the hot path of every authed request, and a transient DB hiccup must not lock out legitimate users.
1. **OAuth callback** — checked right after `exchangeCodeForSession`, before any profile read/write. On a match: `signOut()` + a **fresh** redirect (so the Set-Cookie already queued by the exchange is never sent).
2. **Middleware** — checked after `getUser()`, **gated on a non-null user** (zero DB cost for anonymous requests). Runs on every matched route, so it's also the backstop for the SIWS/wallet login path (which issues no redirect itself).

## Message
Both redirect to `?error=auth&reason=account_deleted`; a landing-page toast (`auth-error-toast.tsx`, in a Suspense boundary) surfaces `auth.accountDeleted` and strips the params. New key in en/pt-BR/es. (Also wires up 3 pre-existing-but-unconsumed OAuth error strings to the same convention.)

## Verify
- A **normal** login proceeds exactly as before (explicitly tested); only a genuine `deleted_at` tombstone is refused.
- 13 new tests: helper (normal/deleted/missing/error/throw → fail-open), callback (normal proceeds, deleted refused + no profile write), middleware (anon → no DB call, normal passes, deleted refused on protected AND non-protected routes).
- tsc clean, 610 tests, lint/prettier clean.

Closes #461.

## Follow-up filed (out of scope): the wallet route un-scrubs before the backstop — see the linked issue.


thomgabriel:
**Gate: PASS (SAFE).** Verified the guard can't lock out normal users: `isAccountDeleted` returns `deleted_at != null` (only genuine tombstones) and fails OPEN on any query error; middleware call gated on non-null user (anon pays nothing). i18n present en/es/pt-BR. #489 (wallet route un-scrubs wallet_address on deleted re-auth) tracked as a follow-up — doesn't undermine the refuse-login backstop, which works.

PR #491, opened 2026-07-14
security(db): stop anon reading profiles.google_id/github_id (#486) [SENSITIVE — applied+verified on prod]

### LINKED ISSUE #486: security(db): anon can read profiles.google_id / github_id of public profiles (column grants are wider than the RLS row policy implies)
Surfaced by the adversarial review of #485 (public_profiles view). **Pre-existing** — not introduced by that PR; the view is a strict *subset* of this exposure. Filing separately as base-table hardening.

## What
`profiles` has a public row-read policy (`"Public profiles are viewable by everyone"` — `FOR SELECT USING (is_public = true AND deleted_at IS NULL)`, granted to PUBLIC), **and** the `anon` role holds column `SELECT` on all `profiles` columns. RLS is **row**-level, not column-level, so the row policy does not restrict which columns a permitted row exposes.

Net: `anon` can select **any** column of a public, non-deleted profile — including `google_id`, `github_id`, and the internal auth `id` (the `auth.users` UUID) — via a plain PostgREST column selection on `/rest/v1/profiles`.

## Impact
- `google_id` / `github_id` are the user's OAuth **subject IDs** (stable per-provider identifiers). They should never be public.
- Values are **NULL today** only because the current seed profiles haven't linked OAuth. The moment any public user links Google/GitHub, their subject ID becomes anon-readable.
- The internal `id` (auth UUID) is also exposed for public rows.

## Fix (SENSITIVE, migration-before-code, like #377/#478)
Make the column grant match the intended public surface. Introspect the live grants first, then narrow `anon`'s column `SELECT` on `profiles` to only the genuinely-public columns (`username`, `avatar_url`, `bio`, `social_links`, `wallet_address`, and whatever the app legitimately needs), REVOKE-ing `google_id`, `github_id`, `id`, `deleted_at`, `deletion_requested_at`, etc. — OR route ALL public profile reads through `public_profiles` (#478) and drop anon's direct table grant entirely. The latter is cleaner: the view already exists and exposes exactly the safe set.

**Done when:** an `anon`-role SELECT of `google_id`/`github_id` from `profiles` returns a permission error (or the column is not selectable), verified on prod; the app still reads what it needs (own profile via authenticated own-row policy; public profiles via `public_profiles`).

Pre-mainnet: yes — an OAuth-subject-ID leak on a public row is a real privacy hole.

thomgabriel:
**#486 · SENSITIVE (DB).** Migration-before-code: **applied to prod (`pywhtmidcrptomrabbrw`) and verified** before this PR.

## The leak
anon held Supabase's default whole-table SELECT on `profiles`, and the `"Public profiles are viewable by everyone"` RLS policy grants row visibility of public rows. RLS gates rows, not columns — so anon could `GET /rest/v1/profiles?is_public=eq.true&select=google_id,github_id` and read OAuth **subject IDs** of any public profile. (NULL today only because seed users haven't linked OAuth.)

## The fix + a Postgres gotcha worth flagging
A column-level `REVOKE SELECT (google_id,…)` is a **no-op** against a table-level grant — the table grant can't be carved by column. Confirmed live: after the column REVOKE, anon still read `google_id`. Correct fix: **drop the table-level SELECT and re-grant SELECT on only the non-sensitive columns.**

Withheld from anon: `google_id`, `github_id`, `deleted_at`, `deletion_requested_at`. Kept: `id, wallet_address, username, bio, avatar_url, social_links, is_public, name_rerolls_used, wallet_xp_synced_at, created_at`.

## Verified on prod (before this PR)
- `has_column_privilege('anon', …, 'google_id'/'github_id'/'deleted_at'/'deletion_requested_at')` → **false** (all four).
- The 10 granted columns → **true** (public profile page + landing count unaffected).
- `authenticated` `google_id` → **true**, unchanged (settings needs own-row).
- Grep confirms **no client read selects a revoked column** — the only `google_id` reader is the settings/account tab, which runs authenticated on its own row.

Mirrored in `schema.sql` so a fresh rebuild reproduces the restriction (otherwise Supabase's default re-grants all columns).

## Scope — closes anon, names the residual
This closes the **zero-auth** (anon) exposure. A logged-in user can *still* read another public profile's `google_id` via the same row policy + wide `authenticated` grants. Fully closing that requires dropping the public-row RLS policy and routing all public reads through `public_profiles` (#485) — a bigger app migration, because `authenticated` legitimately needs own-row `google_id`, so a blanket column REVOKE would break settings. Recommend as the thorough follow-up.

Closes #486.

### DIFF
--- a/supabase/migrations/20260714140000_restrict_anon_profiles_select.sql
+++ b/supabase/migrations/20260714140000_restrict_anon_profiles_select.sql
@@ -1,0 +1,24 @@
+-- #486: anon could read google_id / github_id (OAuth subject IDs) of any public
+-- profile directly from profiles. The "public profiles viewable by everyone" RLS
+-- policy grants row visibility of public rows, and anon holds Supabase's default
+-- whole-table SELECT (every column). RLS gates rows, not columns.
+--
+-- A column-level REVOKE is a NO-OP against a table-level grant (a table grant
+-- can't be carved by column — verified: the first attempt left google_id still
+-- readable). The correct fix is to drop the table-level SELECT and re-grant
+-- SELECT on only the non-sensitive columns.
+--
+-- Withholds from anon: google_id, github_id, deleted_at, deletion_requested_at.
+-- Keeps every column the public profile page / landing count read. Only SELECT
+-- is touched; anon's RLS-gated write privileges are unchanged. authenticated is
+-- untouched (needs own-row google_id for settings) — the authenticated path is a
+-- separate view-routing follow-up.
+--
+-- Applied + verified on prod (pywhtmidcrptomrabbrw): anon SELECT on
+-- google_id/github_id/deleted_at/deletion_requested_at = false; the 10 granted
+-- columns = true; authenticated google_id = true (unchanged).
+REVOKE SELECT ON public.profiles FROM anon;
+GRANT SELECT (
+  id, wallet_address, username, bio, avatar_url, social_links,
+  is_public, name_rerolls_used, wallet_xp_synced_at, created_at
+) ON public.profiles TO anon;
--- a/supabase/schema.sql
+++ b/supabase/schema.sql
@@ -214,6 +214,22 @@ CREATE POLICY "Users can insert their own profile"
 CREATE POLICY "Users can update their own profile"
   ON profiles FOR UPDATE USING (auth.uid() = id);
 
+-- Column-restrict the anon SELECT (#486). The "public profiles" policy above
+-- grants ROW visibility of public rows, and Supabase's default grant gives anon
+-- whole-table SELECT (every column) — so anon could read google_id/github_id
+-- (OAuth subject IDs) of any public profile. RLS gates rows, not columns; and a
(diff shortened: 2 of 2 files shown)


thomgabriel:
**Independent gate: PASS — live-verified on prod (`pywhtmidcrptomrabbrw`).** Confirmed the *actual* grants, not the migration text: anon can no longer read `google_id`/`github_id`/`deleted_at`/`deletion_requested_at` (all false), still reads the safe display columns (`username`/`avatar_url`/`wallet_address` true), and the whole-table SELECT grant is **dropped** (false) — which is the crux, since a column-level REVOKE against a whole-table grant is a no-op (good catch). The anon (zero-auth) hole is closed.

**Residual confirmed (the honest limitation):** `authenticated` still has column-SELECT on `google_id` (verified true), so a logged-in user can still read OAuth IDs of any public profile via the row policy + wide authenticated grants. Filing that as a separate pre-mainnet item (below) — not a regression from this PR, and not a devnet blocker (no real OAuth-linked users on mock data), but it should be closed before mainnet.

Issue #410, opened 2026-07-10
Account deletion anonymizes profile but retains wallet_address — deleted wallets permanently unlinkable


thomgabriel:
From the #409 adversarial review (pre-existing, not introduced there). `api/account/delete/route.ts:61-69` anonymizes username/bio/avatar but never nulls `wallet_address`; the UNIQUE constraint then keeps that wallet claimable by no one forever. Post-#409 any fix must run via service_role — which this route already does (`createAdminClient()`), so the fix is a one-liner: include `wallet_address: null` in the anonymization update. Refs #409 #408

thomgabriel:
Shipped in #460 (merged as 7809d2e) — and the fix went further than filed: besides `profiles.wallet_address`, the adversarial review found `enrollments.wallet_address` (the join survived one table over) and `google_id`/`github_id` (stable provider subject ids) also had to be scrubbed. All nulled; a failed enrollment scrub now 500s instead of reporting a deletion that didn't fully happen. Follow-up filed as #461 (deleted accounts aren't honoured at login).

PR #492, opened 2026-07-14
fix(auth): reject wallet re-auth for deleted accounts before the wallet_address write (#489)

### LINKED ISSUE #489: auth: /api/auth/wallet writes profiles.wallet_address before the #461 deleted-account backstop runs
Found while implementing #461 (refuse-login on deleted accounts). Out of scope for that PR (which was scoped to callback + middleware).

## What
`/api/auth/wallet/route.ts` unconditionally writes `profiles.wallet_address` back onto the profile row as part of SIWS authentication — **before** the `#461` middleware backstop runs on the next navigation. So a **deleted** user who re-authenticates via wallet has their `wallet_address` transiently re-written onto the tombstoned row, then is signed out on their next request.

## Why it matters
Account deletion (#410) nulls `wallet_address` precisely so the wallet isn't bound to a tombstoned row. This wallet re-auth **partially reverses that anonymization** — the write persists in the DB even after the middleware signs the user out, leaving the deleted row with a wallet_address again.

## Fix
In `/api/auth/wallet/route.ts`, check `profiles.deleted_at` (reuse `isAccountDeleted` from `lib/auth/account-status.ts`, added in #461) **before** writing `wallet_address`, and refuse the SIWS auth for a tombstoned account (same 401/redirect contract as the other guard sites) rather than writing then relying on the next-request backstop.

Small, SAFE. Pairs with #461.

thomgabriel:
**#489 · SAFE (auth).** Stacked on #488 (reuses its `isAccountDeleted` helper) — merge #488 first, then retarget this to main.

## The bug
`/api/auth/wallet` unconditionally UPSERTs `profiles.wallet_address` during SIWS auth — before #461's middleware backstop rejects a deleted account on the next request. So a deleted user re-authenticating via wallet had their `wallet_address` re-written onto the tombstoned row, partially reversing #410's anonymization (the write persists after signout).

## Fix
`isAccountDeleted(userId)` is checked right after `verifyOtp()` resolves the session — the earliest point `userId` is authoritatively known, and **before the only `profiles.wallet_address` write** (the `.update(...).eq("id", userId)` later in the route; the earlier `createUser` writes auth-user *metadata*, not the profiles row). On a deleted account: `signOut()` + `403 {error:"accountDeleted"}`, and the wallet write never runs. `wallet-auth-handler.tsx` maps that response to the existing `auth.accountDeleted` message (reused from #461, not duplicated) and disables retry.

## Verify
- Normal wallet login unchanged (200, `wallet_address` written) — explicitly tested.
- Deleted account: 403, `wallet_address` update never called, no downstream reads/writes, `signOut` once, queue-retry not called.
- New route test harness (route had none): 3 cases. tsc clean, 613 tests.

Closes #489.

### DIFF
--- a/apps/web/src/app/api/auth/wallet/__tests__/route.test.ts
+++ b/apps/web/src/app/api/auth/wallet/__tests__/route.test.ts
@@ -1,0 +1,177 @@
+/* eslint-disable import/order -- vi.mock calls must precede importing the route. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest } from "next/server";
+
+vi.mock("server-only", () => ({}));
+
+const {
+  verifySIWSRequest,
+  getUserById,
+  createUser,
+  generateLink,
+  verifyOtp,
+  signOut,
+  selectExistingProfile,
+  selectLinkedWallet,
+  selectUsername,
+  updateWalletAddress,
+  updateUsername,
+  isAccountDeleted,
+  retryPendingOnchainActions,
+  generateWalletName,
+} = vi.hoisted(() => ({
+  verifySIWSRequest: vi.fn(),
+  getUserById: vi.fn(),
+  createUser: vi.fn(),
+  generateLink: vi.fn(),
+  verifyOtp: vi.fn(),
+  signOut: vi.fn().mockResolvedValue({ error: null }),
+  selectExistingProfile:
+    vi.fn<() => Promise<{ data: unknown; error: unknown }>>(),
+  selectLinkedWallet: vi.fn<() => Promise<{ data: unknown; error: unknown }>>(),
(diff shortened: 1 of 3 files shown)


thomgabriel:
**Gate: PASS (SAFE).** Verified the guard placement at the source: `isAccountDeleted(userId)` → signOut + 403 sits at line 177 — after `verifyOtp` (149, the earliest authoritative userId) and **before** the profiles `wallet_address` update (206), so a deleted re-auth is rejected before any un-scrub. Confirmed the earlier write at line 93 is `auth.admin.createUser` user_metadata in the *new-user* branch (no existing profile), not a profiles un-scrub — and a deleted account (wallet nulled per #410) can't match the wallet lookup at 72, so it routes to magic-link → verifyOtp → the 177 check. Reuses #461's `accountDeleted` i18n key. CI green.

Issue #305, opened 2026-07-05
Mainnet blocker: migrate Config.authority + upgrade authority to Squads multisig


thomgabriel:
The on-chain audit's **#1 (and now only) mainnet blocker**: single-hot-key custody.

One key (`AWFHfDZzmgZKgyv6ZFVjNBrtAKvjdThkG3JjJj4uC7cM`) is simultaneously: program **upgrade authority** + `Config.authority` + `Config.backend_signer` + (via the Config PDA) the XP **mint authority**. Compromise = mint infinite XP + forge/rewrite credentials + upgrade the program to arbitrary bytecode. Squads is documented but **not wired**.

#303 landed the keystone — `Config.authority` is now **rotatable** (`update_config.new_authority`), previously immutable. So the migration is now doable in-program.

## Ops steps (yours — no code left, unless you want an admin script)
1. Create a **Squads multisig** vault (squads.so) with members + threshold.
2. **Rotate `Config.authority` → vault**: `update_config { new_authority: <vault> }`. Not called anywhere in the frontend today — needs a one-off admin script/tx (I can scaffold `scripts/rotate-authority.ts`).
3. **Transfer upgrade authority → vault**: `solana program set-upgrade-authority 7NeJaSRyb4Wxay3Tcd9bdpD7T3GWYUQSFyrhG8SgwE8V --new-upgrade-authority <vault> --url \$SOLANA_RPC_URL`.
4. (Recommended) set `backend_signer` to a distinct hot key so a backend leak ≠ full authority.

Dry-run on devnet first. Last gate before mainnet.

PR #502, opened 2026-07-14
feat(admin): course close+recreate server path with safety rails (#453, WS-2) [SENSITIVE]

thomgabriel:
**WS-2 · SENSITIVE (destructive on-chain + authority signing).** The single shared close+recreate implementation (`recreate-course.ts`) — the recreate UI will be a thin front over this. **Re-landed on a fresh branch off main** (not #453's dead SP3-C base; verified one commit ahead). Migration **already applied + verified on prod**.

## The four rails (all IN the server path)
- **admin-auth**: the route requires `requireAdminAuth` (signed `admin_session` + CSRF) + a `confirm === courseId` gate before any on-chain call.
- **H3 lesson-count**: preflight reads the **on-chain** `liveLessonCount` and defaults to it exactly; a superset request is refused (a larger mask silently un-completes mid-course learners via `flags & active == active`).
- **Maintenance gate**: new `in_maintenance` on `onchain_deployments` (migration-before-code — applied to prod, defaults false, not exposed publicly). Set before close, cleared after create, **left on** if create fails. Wired into the finalize webhook, `/api/lessons/complete`, `/api/certificates/mint`, **and** the login-triggered `retryPendingOnchainActions` drain.
- **Mainnet hard-refusal**: refuses unless the network is exactly `devnet` — default-deny (unset → refuse). Un-bypassable until Squads custody (#305).

## Adversarially reviewed + fixed before this PR
An independent pass found (all now fixed): a HIGH — the queue drain wasn't gated, so a login mid-window fired writes at the absent PDA and could push `retry_count` to permanent abandonment (WS-1's exact warning) → now defers without bumping retry; a MEDIUM — the mainnet guard failed **open** on an unset var → now default-deny; H3 falling back to the bundle count on an **undecodable** account (a superset hole) → now refuses; a best-effort gate-SET that didn't abort the close → now aborts; and a test asserting the opposite of its name → fixed. Rails' authz + ordering were airtight; denylist is byte-identical to #496.

## Also fixes a real pre-existing bug
`event-handlers.ts:tryFinalizeCourse` used to plain-`return` (no `queueFailedAction`) when the PDA was absent — silently dropping a learner's auto-finalize. Now throws → queues (idempotent, no double-process).

## Notes for the gate
- **Execute path is post-deploy** — unit-tested with mocked signers/RPC (708 tests); no live close was run. A devnet dry-run against a throwaway course is the remaining live verification (§7).
- **Reward-window guard**: deleted per the WS-2 plan. WS-1's spec says keep-until-Phase-4, a contradiction — but blast radius is ≈ nil (devnet test-XP only, mainnet refused, no UI, and all 6 devnet courses are `creator==authority` so `assertCreatorAllowed` refuses them anyway).
- **Merge-order overlaps** (trivial): `deployments.ts` + `errorIds.ts` with #499; `certificates/mint/route.ts` with #497.

Requesting the gate's final adversarial + live verification on the execute path + the Squads assertion.


thomgabriel:
## F1–F5 fixed (`3e96638`) + my independent re-adversarial pass: clean

Your two gate-lifecycle blockers and the three hardening items are all fixed and re-attacked by a fresh independent pass. **No CRITICAL/HIGH/MEDIUM.** 725 tests (+17, each asserting the exact property), tsc clean.

**F1 (clear path)** — `writeCourseOnChainStatus` clears `in_maintenance` in the same upsert on `→"synced"` (Deploy is now a real recovery), and the success-clear is unskippable (all signer helpers return `{success:false}` rather than throw, so nothing throws between create-landed and the clear).
**F2 (mutual exclusion)** — `acquireCourseMaintenanceGate` is a conditional `UPDATE … WHERE in_maintenance=false` with a PK-guarded INSERT fallback (`23505`→contended/abort, other errors→abort), strictly before the close; a loser aborts before any destructive call and never clears a gate it doesn't own.
**F3** — genesis-hash assertion against devnet, fail-safe on fetch error, on the *same* `SOLANA_RPC_URL` connection admin-signer uses for the close. **F4** — `allowUnusualCreator` now requires `acknowledgeUnusualCreator===courseId` (400 otherwise); the denylist can't be re-opened by a bare body flag. **F5** — the defer write is wrapped so a transient DB error skips the action instead of bumping `retry_count`.

Regression: H3, authz/CSRF/confirm-gate, the mainnet label match, and the single-recreate window are untouched (diff doesn't reach those lines).

### Two LOW, fail-safe residuals my re-pass surfaced (reporting, not fixing — your call)
1. A concurrent operator Deploy on a mid-recreate course can clear the gate — but the clearer just *created* the PDA, so it exists at clear-time (no destructive absent-PDA race); self-limiting, needs two deliberate concurrent admin actions, no recreate UI yet.
2. A create-confirmation false-negative **on an all-null-reward course** can leave the gate stuck (the sync noop branch skips the `synced` write). Normal courses recover via Deploy. Fails safe (over-refusal).

Both are narrow, fail-safe (over-refusal, nothing lost), and only a real-concurrency / live run confirms them — flagging for your independent re-pass + the §7 devnet dry-run rather than expanding a change into the broadly-used sync route without your steer.

Handing back for your independent adversarial re-pass + owner sign-off (destructive path). The §7 live-devnet dry-run against a throwaway course is the outstanding non-code verification.

PR #509, opened 2026-07-14
feat(admin): WS-2 recreate UI — preflight + non-dismissible confirm over the recreate server path [SENSITIVE]

thomgabriel:
## What
The UI half of WS-2 (plan: `docs/superpowers/specs/2026-07-14-ws2-recreate-ui-plan.md`, step 3 + step 6 i18n). Replaces the dead-end "cannot auto-fix" immutable-mismatch card on `/admin/courses` with an actionable, authority-gated **Recreate** flow — a **thin front over the merged #502 server path** (`recreate-course.ts` + `/api/admin/courses/recreate`). This is the mechanism that reconciles the 6 devnet courses after B4 activates the creator-wallet content (the `creator` mismatch that trips this UI *is* the #440 fix).

## Pieces
- **NEW `GET /api/admin/courses/recreate/preflight`** — read-only, admin-gated. Runs `preflightRecreate`, catches its refusal-throw into `{ canRecreate:false, reason }` at 200 (a refusal is UI data, not a 500). The route the plan assumed on #453 didn't exist (that route is execute-only) — this fills it.
- **Preflight card** — immutable field(s) old→new, creator old→new (prominent — immutable), the live on-chain `lesson_count` (H3), and the reset-vs-preserved summary (**counters reset to 0; enrollments, learner progress, AND certificates preserved**).
- **Non-dismissible confirm modal** — no backdrop/silent dismissal; type-the-courseId to enable Confirm; brief-downtime warning (close+create are not atomic). **F4:** an "unusual creator" (denylist / equals platform authority) requires a SECOND explicit acknowledgment, which alone maps to `allowUnusualCreator:true` + `acknowledgeUnusualCreator:courseId`. A normal recreate sends neither.
- **i18n** en/es/pt-BR (folded in — a standalone i18n PR would ship dead keys).

## Review discipline
**SENSITIVE.** Independent adversarial pass run before this PR opened → **no CRITICAL/HIGH** (the destructive authority is the server route, which independently re-enforces admin-auth, `confirm===courseId`, the F4 double-field, mainnet/genesis refusal, the #427 creator guard, H3, and maintenance-gate mutual exclusion). Findings, all fixed in `caeb14e`:
- **MEDIUM** — recreate was never re-gated on "an immutable mismatch still exists," so a stale tab / sequential double-recreate could close+recreate an already-correct course (irreversible counter reset + downtime for nothing). Fixed: preflight now returns `canRecreate:false` unless a real immutable diff is present (`noImmutableDiff`), or `unconfirmed` when the chain read is unavailable. Guarded against over-refusal — a genuine mismatch (incl. creator-only) still allows.
- **LOW** — indeterminate/network execute failure now shows an honest "refresh & check" recovery message instead of hiding the down-course hint.
- **LOW** — refusal/error `reason` strings scrubbed (`sanitizeReason`) so an api-keyed RPC URL / secret env names can't leak into client telemetry.
- **LOW** — corrected an inaccurate CSRF claim in the GET docstring.

## Merge gate
Hold until: (1) the gate's independent re-pass, (2) **owner sign-off** (destructive path), and (3) the **§7 live-devnet dry-run is green**. Do not merge before the dry-run.

## Noted pre-existing follow-up
The already-merged execute route (`recreate/route.ts`) still returns `preflightRecreate`/error reasons verbatim and can leak the same secret-bearing substrings — out of scope here (didn't re-touch merged SENSITIVE code); recommend routing it through the same `sanitizeReason` in a follow-up.

## Tests
22/22 new (preflight helper incl. both over-refusal guards, preflight route incl. redaction, client flow incl. F4 body + confirm-gating + the three failure states); 147/147 broader admin suite; tsc clean.

### DIFF
--- a/apps/web/src/app/api/admin/courses/recreate/preflight/__tests__/route.test.ts
+++ b/apps/web/src/app/api/admin/courses/recreate/preflight/__tests__/route.test.ts
@@ -1,0 +1,118 @@
+/* eslint-disable import/order -- vi.mock('server-only') must be hoisted above
+   the route import so the `server-only` graph loads under vitest. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import type { NextRequest } from "next/server";
+
+vi.mock("server-only", () => ({}));
+
+const h = vi.hoisted(() => ({
+  AdminAuthError: class AdminAuthError extends Error {},
+  state: { authThrows: false },
+  buildRecreatePreflight: vi.fn(),
+}));
+
+vi.mock("@/lib/admin/auth", () => ({
+  AdminAuthError: h.AdminAuthError,
+  adminUnauthorizedResponse: () =>
+    new Response(JSON.stringify({ error: "Unauthorized" }), { status: 401 }),
+  requireAdminAuth: vi.fn(() => {
+    if (h.state.authThrows) throw new h.AdminAuthError();
+  }),
+}));
+
+// The route delegates the read-only work to the helper; stub it so no on-chain
+// read happens in a route test.
+vi.mock("@/lib/admin/recreate-preflight", () => ({
+  buildRecreatePreflight: h.buildRecreatePreflight,
+}));
+
+const buildRecreatePreflight = h.buildRecreatePreflight;
+const COURSE_ID = "course-solana-101";
+
+const OK_DTO = {
+  canRecreate: true as const,
+  courseId: COURSE_ID,
+  coursePda: "PDA111",
+  creatorOnChain: "AUTH1111",
+  creatorResolved: "CREATOR111",
+  liveLessonCount: 3,
+  unusualCreator: false,
+  immutableDiffs: [
+    { field: "creator", onChainValue: "AUTH1111", contentValue: "CREATOR111" },
+  ],
+  lostCounters: { totalCompletions: 42, totalEnrollments: 100 },
+};
+
+const get = async (query: string): Promise<Response> => {
+  const { GET } = await import("../route");
+  return GET(
+    new Request(
+      `https://app.test/api/admin/courses/recreate/preflight${query}`
+    ) as unknown as NextRequest
+  );
+};
+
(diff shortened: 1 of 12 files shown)


thomgabriel:
## Both gate findings fixed (`c0a25d8`) + re-verified at source

**FIX A — client-side secret leak (`recreate-course-flow.tsx:168`).** `sanitizeReason` extracted to a pure shared module `lib/admin/sanitize-reason.ts` (no `server-only`, no imports). The preflight route now imports it (local copy deleted, redaction rules byte-identical). The client wraps `execError.error` in `sanitizeReason(...)` before render, so the api-keyed RPC URL / secret env names can't paint into the admin DOM / session-replay.

**FIX B — swallowed preflight load error (`recreate-course-flow.tsx:213`).** The contradictory `phase !== "error"` guard is now `phase === "error"`, so a preflight fetch/500 failure actually renders. The execute-error branch early-returns at `:162` (`phase === "error" && execError`) and takes precedence, so the load-error block only fires when `execError` is null; the separate `refusal`/`reasonCode` render path is untouched. No collision between the three error surfaces (execute-error, load-error, refusal).

**Not re-touched** (per your steer): the preflight GET admin-gate + read-only-ness, the no-op refusal (incl. the B4 creator-only allow), confirm-gating, and the server re-enforcement — all unchanged.

**Tests:** +5 shared-util unit (URL / stray api-key / both env names / pass-through), +3 client (execute error redacted in DOM; preflight reject → load error renders; non-refusal 500 → load error renders). 21/21 covering, tsc clean.

**Tracked follow-up filed: #511** — route the merged execute route's error/reason through the same `sanitizeReason` **server-side** (the #502 root leak source; this PR's client scrub is the belt-and-suspenders). Pre-mainnet, out of scope here to avoid re-touching merged SENSITIVE code.

Back to you for the re-pass. Merge still held on owner sign-off + the §7 live-devnet dry-run.

PR #510, opened 2026-07-15
feat(challenge): LeetCode-style split — text left, editor right, AI reveals on scroll

Potolski:
## What

Reworks the challenge workspace into a **LeetCode-style two-column split**.

**Before:** code editor on the left; instructions + AI Partner stacked in a right rail.

**After:**
- **Left** — the lesson **text** (instructions + test cases) as one full-height, independently scrollable reading column.
- **Right** — the **code editor** + output panel, full height.
- **AI Partner** now sits at the **bottom of the left reading column** and stays collapsed until the learner **scrolls to the end of the description**, then slides in.

## How

- Swapped the two panels so text renders left / editor right; the drag divider now rebalances the text↔editor split (drag right widens the text). The editor↔output vertical resizer is unchanged.
- Reveal-on-scroll: an `IntersectionObserver` watches a 1px sentinel at the end of the text. `root: null` (viewport) so it works for both the column's internal scroll (lg+) and page scroll (below lg). Reveals **once** and stays revealed; `motion-reduce` disables the transition.
- Mobile stack order preserved via the existing `contents` + `order` pattern: **text (1) → editor (2) → output (3) → AI (4)**.

## Notes / test
- eslint + tsc clean.
- Please eyeball on the Vercel preview: desktop split + resize, scroll the left text to trigger the AI reveal, and the mobile stack. The AI pane is a fixed-height widget (600px) at the column bottom.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/code-block.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/code-block.tsx
@@ -61,7 +61,7 @@ export function CodeBlock({ block, ctx }: BlockRenderProps) {
 
   return (
     <div className="overflow-hidden rounded-[var(--r-lg)] border-[2.5px] border-border shadow-card">
-      <div className="flex w-full flex-col overflow-hidden lg:h-[calc(100vh-220px)]">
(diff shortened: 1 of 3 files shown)


Potolski:
@claude review if the blocking issue was fixed

Issue #590, opened 2026-07-25
AI cost hole: assists refunded AFTER generation, rate limiter fails open, no per-IP bucket


thomgabriel:
Three verified issues that compound into an unbounded AI spend path on a platform-funded API key.

**1. Assists are refunded after Gemini has already billed us.** `apps/web/src/app/api/ai/partner/route.ts` calls `refundAssist()` at three points *after* the generation returns — on empty output, on unparseable JSON, and on the truncation path (all typically `finishReason: "MAX_TOKENS"`). Google billed those tokens; we hand the quota back. A caller who reliably triggers truncation gets unlimited billed generations against a quota that never decrements.

**2. The rate limiter fails OPEN.** `apps/web/src/lib/rate-limit.ts` returns `false` (= not limited) from its catch block. Note the contrast with `lib/ai/assist-budget.ts`, which deliberately fails CLOSED ("spend failed, denying"). If the limiter's backing store errors, the AI route becomes unmetered.

**3. No per-IP bucket on the AI route.** `route.ts:213` keys `isRateLimited("ai:partner", user.id, …)` on the user id only. Wallet sign-up is free and unlimited, so per-user keys cannot bound an actor — every fresh account is a fresh bucket. (`/api/lessons/complete` already does per-user **and** per-IP for exactly this reason.)

## Done when
- [ ] No refund path after a billed generation; a `billed_assists` counter records spend that occurred regardless of output quality
- [ ] `ai:partner` uses a fail-closed limiter (opt-in flag so other callers keep current behaviour)
- [ ] Per-IP token bucket added alongside per-user on the AI route
- [ ] Route tests cover: truncation does not refund, limiter error denies, per-IP ceiling trips

Trade-off to decide: not refunding means a learner who hits MAX_TOKENS loses a turn for what is arguably our bug — pairing this with a capped diff-style `propose` output shrinks that window.

Source: AI Tutor Economics spec (2026-07-25), all three claims re-verified against the code before filing.

thomgabriel:
**GATE → ORCHESTRATOR.** This issue's PR is **MERGED** and its branch was deleted. If you have local work on that branch, it is orphaned — push it to a **new branch off `main`** instead.

Two follow-up commits were already stranded this way and I recovered them in #618 (the deny-egress comment correction and the aria-disabled alignment). Nothing was lost, but don't re-push to the old branch.

**Rule:** once the gate merges, that branch is closed. Follow-ups → new branch vs `main`, referencing the issue.

PR #611, opened 2026-07-25
feat(lessons): openEnded attestation route — C5 Earn lesson completable (receipt-first)

### LINKED ISSUE #595: openEnded attestation route has zero callers — C5's Earn lesson is uncompletable
`sealAttestation` is implemented and tested but **nothing calls it**. Without an attestation-issuing route, the final lesson of course 5 (courses-academy PR #6) can never be completed — and that lesson produces the launch KPI (Earn submissions). LAUNCH-BREAKING: must ship **before** the C5 `content.lock` bump.

**Unified spec item 6** — full detail, evidence and acceptance criteria in [`docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md`](docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md) §3.

thomgabriel:
Closes #595

## What this is
`openEnded` blocks are declared `{graded:false, required:true}`, so `/api/lessons/complete` demands a **sealed attestation** for them — but `sealAttestation()` had **zero callers**. Every `openEnded` lesson (including C5's `submit-and-what-you-built`, the Earn terminus and the launch KPI) therefore 403'd forever: no finalize, no credential, no submission.

This adds the missing caller: **`POST /api/lessons/reflect`**.

The route:
- Derives `userId` from the **authenticated session** (never the body).
- Resolves `courseId`/`lessonId` against the committed content bundle via `getLessonByIdForGrading`, then finds the block by `blockKey` and confirms it is a real `openEnded` block. Unknown lesson/block → 404; non-`openEnded` block → 400.
- Validates the submission server-side against the block's `maxWords` (+ a 10k-char byte cap and a 20k raw-body cap). Empty or over-limit → 400; oversized → 413.
- Applies the standard per-user rate limiter (`lessons:reflect`, 30/min). A throttled request 429s (retryable) — the seal is **deferred, never permanently unobtainable**. No per-IP key here: unlike `/api/lessons/complete` this path funds no on-chain write, so a per-IP limiter would only risk 429ing a NAT'd classroom for no spend benefit.
- On a valid submission, calls `sealAttestation({lessonId, blockKey, userId, exp})` and returns the seal **before and independent of** any AI work.

`maxWords` stays live as a **validation** bound (it also bounds the optional AI call), per the ruling — built, not stripped, because `openEnded` is also the artifact-URL milestone fallback (CAT-17).

## Seal round-trip (proven, not asserted)
The attestation is bound to `{lessonId, blockKey, userId}` + a 24h staleness TTL — the exact shape `/api/lessons/complete` verifies. The route test imports the **real, unmocked** `openAttestation` (the same function the completion route calls) and asserts:
- the seal the route mints is accepted for the matching `{lessonId, blockKey, userId}` → `true`;
- the same seal replayed under a different `userId` → `false`.

So the receipt this route produces is verified by the exact consumer path, in-test.

## AI-reply choice: implemented, best-effort, default-OFF behind an env flag
`maybeGenerateReflectionReply` (`lib/ai/reflection-reply.ts`, server-only) makes one bounded, rate-limited (`ai:reflection`, own namespace) Gemini call and returns `string | null`. It is:
- **default OFF** (`OPENENDED_AI_REPLY=1` to enable) — launch runs seal-only at zero AI cost;
- **never touching the assist-budget ledger** (`spendAssist`/`refundAssist`) — the reflection reply is not a code-challenge assist and must not draw from that budget. **#590 (AI cost holes) is untouched** and owns the metering decision when the flag is flipped;
- **fully best-effort** — disabled flag, unset `GEMINI_API_KEY`, rate-limit, `!ok`, or a throw all resolve to `null`. The route additionally wraps the call in its own try/catch so the **seal is returned even if the reply path throws**.

This is the sanctioned "ship the reply disabled behind an env flag" path, but with the real implementation in place and tested, so enabling it later is a config change, not new code. `GEMINI_API_KEY` requirement + the flag are documented in `apps/web/CLAUDE.md`.

## End-to-end wiring
`open-ended-block.tsx` now captures the reflection, POSTs it to `/api/lessons/reflect`, and records the returned seal as the block's proof via `ctx.setProof(block.key, seal)` — the same channel `quiz`/`code` blocks use — so the non-code lesson's "Mark as Complete" POST carries the attestation and the lesson completes end-to-end. Submit/success/error states are accessible (button label + spinner `sr-only`, `role="alert"`, `aria-live="polite"`, word-count `aria-describedby`, over-limit turns the counter `text-danger` and disables submit). All strings via next-intl in **en / pt-BR / es** (key parity verified). Any present AI reply renders under a "Feedback" label; its absence changes nothing.

## Verification (run from `apps/web`)
- `pnpm typecheck` → clean.
- `pnpm test` → **104 files, 954 tests passing** (includes the new `reflect/route.test.ts` = 15 and `reflection-reply.test.ts` = 6, plus the untouched `attestation-seal` and `complete/gate` suites).
- New-file focus run → `2 files, 21 tests passing`.
- `eslint` on touched files → clean; `prettier --check` → clean; locale key parity `en==pt==es` → true.

Test coverage: happy path (seal round-trips into the real `openAttestation`) · replay-under-different-user rejected · 401 unauthenticated · 404 unknown lesson · 404 missing block · 400 non-`openEnded` · 400 over-`maxWords` · 400 empty · 400 missing fields · 413 oversized body · 413 over char cap · 429 rate-limited (Retry-After) · reply-null still seals · reply-throws still seals · 500 on unconfigured seal secret. Helper: flag-off / no-key / rate-limited / ok / !ok / throw.

- **content-lint "no unattested `openEnded`" rule** (spec item 6's recurrence guard) lives in the content-schema package and is a separate change; not added here.
- **`lesson-client` 403 message mapping**: a non-code lesson that 403s currently maps to `completionFailedEnrollment` ("verify enrollment"), which reads slightly wrong for a *missing* reflection. Pre-existing; the block's own "Submit reflection" affordance makes it unlikely a learner reaches that path. Left untouched to avoid perturbing the challenge-403 branch.
- **#590 (AI cost holes)** owns metering for the reply path; deliberately not touched.


thomgabriel:
## Orchestrator routing record — SENSITIVE (attestation gates XP/credentials) → adversarial review → `needs-gate`

**Independent verification (orchestrator, not the implementer):** re-ran `pnpm typecheck` (clean) and the full suite twice — 954/954 before the hardening fixes, **957/957 after** (commit 02a8d75). Read the route + fix diffs line-by-line.

**Adversarial review: PASS** (fresh hostile reviewer). Attacked and cleared: cross-protocol seal replay is cryptographically impossible (distinct HMAC domain-separation labels in `check-seal.ts` derive independent AES-256-GCM keys; proven by an in-repo test); cross-user and cross-block replay prevented by payload binding; the seal grants nothing on its own (`/api/lessons/complete` independently requires on-chain enrollment + linked wallet); billing fence real (no assist-ledger imports; `OPENENDED_AI_REPLY` read server-side, default OFF); no XSS (React text child, no `dangerouslySetInnerHTML`); i18n parity genuine in en/pt-BR/es; frontend wiring verified end-to-end (`setProof` → `proofsRef` → completion POST → `proofs[block.key]`); no PII logged or persisted.

1. **Fixed in 02a8d75:** no timeout on the Gemini fetch could stall seal *delivery* behind a hung upstream when the reply flag is ON → `AbortController` + 10s hard ceiling, reply degrades to `null`, seal ships regardless. Tested.
2. **Deferred to #590 (owns it):** both limiters share `isRateLimited`, which fails open — unbounded call *count* if the flag is ON during a store outage. Not active at launch (flag default OFF).
3. **Fixed in 02a8d75:** seal didn't bind `courseId` (relied on the unenforced invariant that lesson `_id`s are globally unique) → `courseId` now bound at mint (`reflect/route.ts`) and verify (`complete/route.ts` + `openAttestation`), with a course-mismatch test.
4. **Accepted (UX follow-up, not security):** proofs live in a `useRef`, so a page reload forces resubmission despite the 24h seal TTL.
5. **Accepted (noise):** zero-width-joined text can slip under `maxWords`; bounded harmlessly by the 10k-char cap.

**Gate:** re-verify the load-bearing claims per the loop contract — especially the seal round-trip test (`reflect/__tests__/route.test.ts` uses the real, unmocked `openAttestation`) and the receipt-first guarantee. Launch ships seal-only (`OPENENDED_AI_REPLY` unset); flipping the flag is gated on #590's metering work.

Issue #591, opened 2026-07-25
AI tutor: no spend ledger — add fail-closed per-account/per-IP/global daily caps


thomgabriel:
There is no record anywhere of what the AI tutor costs per account. `challenge_assists` counts *turns per lesson*, which is not spend: a turn can be 200 output tokens or a truncated 8k-token `propose`. Without a ledger there is no ceiling and no way to detect an abusive account before the invoice.

## Owner decision (2026-07-25): Superteam sponsors the tutor

This resolves the framing. The tutor is **not** learner-metered for cost reasons — Superteam funds it, in the Khanmigo model (solve free AI tutoring by having a sponsor pay, not by rationing learners).

That changes how these numbers are derived and what they are *for*:

- **A learner-facing wall SHOULD exist** (owner, 2026-07-25) — but generous. The wall is not only cost control: it is what stops the tutor becoming an answer machine, which the pedagogy evidence says is the real risk once cost stops limiting usage (see #592). Degrade first, then stop.
- **Derive downward from the sponsor commitment**, not upward from learner behaviour: commit $X/month → global daily degrade ≈ `X/30` with headroom → per-account and per-IP ceilings sit generously underneath it.
- **Per-account and per-IP ceilings are still required** — so a single farmer cannot drain the sponsor's card in one day. Free wallet signup means per-user keys alone cannot bound an actor.
- **The budget may exceed $300/month at 10k MAU** (owner) — it must simply be *coherent with scale*. At ~$0.03/learner/month the marginal cost is small, so set the commitment first and let the caps follow it, rather than reverse-engineering a small budget into a stingy learner experience.

## Proposal

An `ai_spend_ledger` table plus a `record_ai_spend` SECURITY DEFINER RPC (service_role only, RLS enabled with no policies — same pattern as `award_xp`), enforcing daily caps on America/Sao_Paulo days: per-account, per-IP, and a global backstop that **degrades before it denies** (over soft cap → cheaper model + shorter outputs; only the hard cap refuses).

## Done when
- [ ] Migration + fail-closed RPC (denies on error — matching `assist-budget`, *not* `rate-limit`, which fails open: see #590)
- [ ] Every AI call records actual usage from `usageMetadata`, including thinking tokens (they bill at output rate)
- [ ] Soft cap degrades, hard cap denies with localized copy (×3 locales)
- [ ] Admin-visible daily spend, so the sponsor's burn is observable before the invoice
[...]

Issue #612, opened 2026-07-25
AI partner: diff-propose at 2,048-token cap + MAX_CODE_CHARS 20k→8k (closes the truncation cause)


thomgabriel:
**Unified spec §3 item 3a** — the unblocked half of "model routing + diff-propose", deliberately split out of #590 (same week, not same PR).

Two changes to `/api/ai/partner`:
1. **Emit `propose` as a unified diff at a 2,048-token output cap** — today `propose` echoes the whole file against an 8,192 cap while reference solutions are median ~151 tokens (AIE-15). Diff output = ~30% blended cost cut and ~10× lower adversarial worst case.
2. **Trim `MAX_CODE_CHARS` 20,000 → 8,000** (`route.ts:32`).

Together these make AIE-11's attacker-triggerable truncation (non-BMP escape inflation → guaranteed MAX_TOKENS) nearly impossible.

**Why this is time-boxed:** #590 removes the refund-after-billing path, which closes the *cost hole* but leaves the *cause* open — in the gap an honest learner with a long file can be charged an assist for a truncated answer. Per the spec: **if this slips more than a week past #590, restore a NARROW refund keyed to `finishReason === "MAX_TOKENS"` only — never the broad catch.**

Explicitly NOT here: per-action model switching + Flex-tier spike (spec item 3b — blocked on the two unverifiable curls AIE-04/AIE-05); prompt caching (AIE-14: confirmed worthless, do nothing).

_Filed from the unified launch spec; this work existed in the plan but had no issue._

PR #613, opened 2026-07-25
fix: close AI cost holes on /api/ai/partner (#590)

### LINKED ISSUE #590: AI cost hole: assists refunded AFTER generation, rate limiter fails open, no per-IP bucket

thomgabriel:
Closes #590

Three verified holes on `/api/ai/partner` compounded into an unbounded spend path on the platform-funded Gemini key. This is **live exposure**, not a launch feature. All three were re-confirmed in code (spec §2.5 AIE-10/-11/-12/-13, all CONFIRMED).

## The three holes and what closed each

**1. Refund-after-billing (AIE-10, incl. the line-411 correction).**
The route called `refundAssist()` on empty output, unparseable JSON, and MAX_TOKENS truncation — every one of those AFTER Gemini had already billed the tokens — and the outer `catch` (which wraps `appendAssistLog`) refunded even a throw that happened *after a fully-billed success*. A caller who reliably triggers truncation got unlimited billed generations against a quota that never decremented.
- Added a `billed` flag that flips `true` the instant Gemini returns a **2xx** (`response.ok`). Refund now fires **only when `!billed`** — a non-2xx (`!response.ok`) or a pre-response throw (DNS/TLS/abort, or a throw during prompt-building), i.e. the cases where the model never ran.
- The `try` was widened to cover prompt-building (`visibleTests` / `task` / `buildStaticPrefix` / `buildDynamicSuffix`) without moving `spendAssist` or touching any other semantics — nothing built there is used after the try. A throw in prompt-building is now pre-billing and correctly refunds under `if (!billed)`, closing a pre-existing (practically unreachable) spent-but-not-refunded window. Covered by a new test.
- Removed the refund from the empty / non-JSON / malformed branches (all post-billing).
- Narrowed the outer catch to `if (!billed) refund` so a post-success throw (envelope parse, seal, or `appendAssistLog`) no longer hands back a billed assist.

**2. Fail-open limiter (AIE-12).**
`lib/rate-limit.ts` returned `false` (= allow) from both its error branches, so a store outage made the AI route unmetered. Added an **opt-in `failClosed` flag** to `RateLimiterOptions`: a flagged caller returns `true` (deny) on a store error; the default is unchanged, so every other caller keeps failing open. Both `ai:partner` checks opt in.

**3. No per-IP bucket (AIE-12).**
The route keyed only on `user.id`, and wallet signup is free — so per-user buckets cannot bound an actor (every fresh account is a fresh bucket). Added an `ai:partner:ip` token bucket via `getClientIp`, mirroring `/api/lessons/complete`. Sized at 600/min: clears a NAT'd/CGNAT'd BR classroom (a fixed window is a cliff, not a throttle) while still bounding a Sybil actor's burn rate from one address.

## `billed_assists` counter — design choice

`assists_used` is **refundable** (the `!response.ok` path still gives it back), so it is not a faithful record of billing. Added a separate, non-refundable `billed_assists` counter, incremented once per confirmed 2xx **regardless of output quality** (empty / non-JSON / truncated-but-billed all still count) and never decremented — the durable audit trail of real spend against the shared key.

I deliberately did **not** bump the counter inside `spend_challenge_assist`: spend happens *before* the billing signal, and the `!response.ok` path is a genuine not-billed case, so counting at spend time would over-count. The counter is bumped from the route at the exact moment billing is confirmed, via a new `record_billed_assist` SECURITY DEFINER RPC that follows the existing hardening pattern verbatim (RLS-on table with no policies, `SET search_path = ''`, REVOKE from PUBLIC/anon/authenticated, GRANT to service_role only). Best-effort on the TS side — an audit write must not break the paid reply.

## Schema / migrations

- `supabase/schema.sql`: added the `billed_assists` column (+ idempotent `ALTER`) and the `record_billed_assist` function.
- New migration `supabase/migrations/20260725120000_billed_assists_counter.sql`, matching the additive + idempotent convention of its siblings.
- Generated `src/lib/supabase/types.ts`: added the `record_billed_assist` function signature (RPC names are type-constrained).
- **Nothing applied to any live database.** No `supabase db push`, no MCP apply. Shipping the SQL is this PR; applying it is the gate/owner's dry-run-first call.

## Fairness window (named, not hidden)

Removing the refund closes the hole but leaves its **cause** — AIE-11's attacker-triggerable truncation — open until **spec item 3a** (diff-propose + `MAX_CODE_CHARS` 20k→8k) lands. 3a is a separate follow-up shipping the **same week**, deliberately not in this PR. In the gap, an honest learner with a long file can be charged an assist for a truncated answer. Per the spec: **if 3a slips past a week, restore a NARROW refund keyed to `finishReason === "MAX_TOKENS"` only — never the broad catch.**

## Verification (run for real)

- `pnpm typecheck` (apps/web) → pass, no output.
- `pnpm test` (apps/web) → **102 files, 948 tests pass** (full suite, incl. the new/updated tests).
- `npx eslint` on all touched files → **0 errors** (1–2 pre-existing, non-blocking `import/order` warnings that match the repo's standard mock-then-import test layout; `next lint` does not fail on warnings).
- `npx prettier --check` on the 7 touched TS files → clean. SQL is outside prettier's glob (no SQL parser).

New/updated route tests: truncation (`finishReason: "MAX_TOKENS"`), empty, non-JSON, and malformed outputs **do not refund** but **do** record a billed assist; `!response.ok` **does** refund; a post-success throw (`appendAssistLog` rejects) **does not** refund; a **pre-billing throw during prompt-building does refund** (proves the widened `try`); the per-IP ceiling trips independently of per-user, and **both** `ai:partner` keys are asserted `failClosed: true`. Rate-limiter unit tests: a store error denies for a `failClosed` caller while an unflagged caller on the *same* error stays open. `assist-budget` unit tests cover `recordBilledAssist` (args + best-effort no-throw on error/throw).

- **Spec item 3a** (diff-propose, `MAX_CODE_CHARS` 20k→8k) — separate follow-up, same week; it is what actually closes the truncation *cause*.
- **#591 spend ledger** — blocked on owner decision O-1; the real per-account cost ceiling.
- **Model/pricing routing (3b)** — blocked on two unverifiable curls.
- **`/api/lessons/reflect` + `lib/ai/reflection-reply.ts`** (PR #611, in flight) — not touched. Its rate-limit fail-open interaction is deliberately deferred to this flag landing. Follow-up after both merge: `ai:reflection` can now opt into `failClosed` the same way — left out here to avoid a cross-PR conflict.

### DIFF
--- a/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
+++ b/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
@@ -404,6 +404,26 @@ describe("POST /api/ai/partner", () => {
     expect(refundAssist).toHaveBeenCalledWith("user-1", "lesson-1");
   });
 
+  it("refunds a PRE-billing throw during prompt building (billed still false)", async () => {
+    // The prompt-building block is inside the widened try. A throw there — here
+    // the visibleTests map blows up on a malformed code block (tests: null) —
+    // happens before any Gemini call, so it is pre-billing: the catch refunds it
+    // under `if (!billed)` and records no billed assist.
+    getLessonBySlug.mockResolvedValue({
+      ...LESSON,
+      blocks: [LESSON.blocks[0], { ...LESSON.blocks[1], tests: null }],
+    });
+    stubGeminiFetch(PROPOSE_GEMINI_TEXT); // never reached
+
+    const { POST } = await import("../partner/route");
+    const res = await POST(makeRequest(VALID_BODY));
+
+    expect(res.status).toBe(500);
+    expect(refundAssist).toHaveBeenCalledTimes(1);
+    expect(refundAssist).toHaveBeenCalledWith("user-1", "lesson-1");
+    expect(recordBilledAssist).not.toHaveBeenCalled();
+  });
+
   it("returns 429 when rate limited", async () => {
     isRateLimited.mockResolvedValue(true);
     stubGeminiFetch(PROPOSE_GEMINI_TEXT);
@@ -429,14 +449,19 @@ describe("POST /api/ai/partner", () => {
 
     expect(res.status).toBe(429);
     expect(spendAssist).not.toHaveBeenCalled();
-    // Both dimensions were consulted, and the per-IP key uses failClosed.
+    // Both dimensions were consulted, and BOTH opt into failClosed — a limiter
+    // outage must not wave either through unmetered on the platform-funded key.
     const namespaces = isRateLimited.mock.calls.map((c) => c[0]);
     expect(namespaces).toContain("ai:partner");
     expect(namespaces).toContain("ai:partner:ip");
+    const userCall = isRateLimited.mock.calls.find(
+      (c) => c[0] === "ai:partner"
+    )!;
     const ipCall = isRateLimited.mock.calls.find(
       (c) => c[0] === "ai:partner:ip"
     )!;
-    expect(ipCall[2]).toMatchObject({ failClosed: true });
+    expect(userCall[2]).toEqual(expect.objectContaining({ failClosed: true }));
+    expect(ipCall[2]).toEqual(expect.objectContaining({ failClosed: true }));
   });
 
   it("returns 400 on malformed JSON body", async () => {
--- a/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
+++ b/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
@@ -10,17 +10,21 @@ vi.mock("@/lib/supabase/server", () => ({
 
 const spendAssist = vi.fn();
 const refundAssist = vi.fn();
+const recordBilledAssist = vi.fn();
 const appendAssistLog = vi.fn();
 vi.mock("@/lib/ai/assist-budget", () => ({
   spendAssist,
   refundAssist,
+  recordBilledAssist,
   appendAssistLog,
   MAX_PAID_ASSISTS: 4,
 }));
 
 const isRateLimited = vi.fn();
+const getClientIp = vi.fn(() => "203.0.113.9");
 vi.mock("@/lib/rate-limit", () => ({
   isRateLimited,
+  getClientIp,
 }));
 
 const getLessonBySlug = vi.fn();
@@ -100,7 +104,11 @@ const PROPOSE_GEMINI_TEXT = JSON.stringify({
 
 function stubGeminiFetch(
   text: string,
-  opts: { ok?: boolean; cachedContentTokenCount?: number } = {}
+  opts: {
+    ok?: boolean;
+    cachedContentTokenCount?: number;
+    finishReason?: string;
+  } = {}
 ) {
   vi.stubGlobal(
     "fetch",
@@ -109,7 +117,12 @@ function stubGeminiFetch(
         ok: opts.ok ?? true,
         text: async () => "error",
         json: async () => ({
-          candidates: [{ content: { parts: [{ text }] } }],
+          candidates: [
+            {
+              finishReason: opts.finishReason,
+              content: { parts: [{ text }] },
+            },
+          ],
           usageMetadata: {
             cachedContentTokenCount: opts.cachedContentTokenCount ?? 0,
           },
@@ -131,6 +144,7 @@ beforeEach(() => {
   getUser.mockReset();
   spendAssist.mockReset();
   refundAssist.mockReset();
+  recordBilledAssist.mockReset();
   appendAssistLog.mockReset();
   isRateLimited.mockReset();
   getLessonBySlug.mockReset();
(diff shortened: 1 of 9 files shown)


thomgabriel:
## Orchestrator routing record — SENSITIVE (`area:security` + `supabase/**`) → adversarial review → `needs-gate`

**Independent verification (orchestrator, not the implementer):** re-ran `pnpm typecheck` (clean) and the full suite twice — 947/947 at the initial head, **948/948 after the ride-alongs** (commit 381f8e5). Read the route, rate-limit, assist-budget and migration diffs line-by-line; confirmed `challenge_assists` has `PRIMARY KEY (user_id, lesson_id)` backing the `ON CONFLICT`, and all non-provided columns have defaults.

**Adversarial review: PASS** (fresh hostile reviewer, two attacker models: unmetered spend + weaponizing fail-closed into DoS). Cleared: `billed` latches before any post-2xx code with no throw window; refunds fire only on `!response.ok` or `if (!billed)` in the catch; spend runs AFTER both limiters (deny paths cannot eat an assist — test-asserted); fail-closed adds no new unavailability (the assist budget already fails closed on the same store); `record_billed_assist` SQL is byte-identical in hardening to its siblings (SECURITY DEFINER, `search_path=''`, schema-qualified refs, REVOKE/GRANT stanza); `failClosed` is opt-in at exactly the 2 `ai:partner` sites (grep-confirmed — no other caller's semantics move); PR #611 unaffected; `getClientIp` trust order (x-vercel-forwarded-for first) is the pre-existing platform-set pattern, not a regression; nothing applied to any live DB.

1. **Fixed in 381f8e5:** "200" wording → "2xx (`response.ok`)" in comments + PR body; the conservative `response.ok` gate kept.
2. **Fixed in 381f8e5 (was pre-existing):** the spent-but-unrefunded window between `spendAssist` and the `try` — the try boundary now covers prompt-building, so a pre-billing throw refunds under `if (!billed)`; new test proves the boundary (malformed block → 500 + refund + no billed record).
3. **Fixed in 381f8e5:** per-user `ai:partner` call now test-asserted to carry `failClosed: true` (symmetry with per-IP).
4. **Accepted (sizing judgment, flagging for owner):** per-IP 600/min could collateral-429 a very large CGNAT egress (BR mobile carriers). It is a soft 60s-retry throttle, not the cost ceiling (the 4-assists/lesson budget is). Tunable constant if it bites.
5. **Accepted:** `sealCheck`-throws-post-billing isn't a named test but is mechanically identical to the tested `appendAssistLog`-throws case (same `if (!billed)` latch).

**For the gate:**
- The migration (`supabase/migrations/20260725120000_billed_assists_counter.sql`) ships SQL only — **nothing was applied to any live DB**. Applying it is your dry-run-first call per the do-not-touch list.
- The fairness window is stated in the PR body: the truncation *cause* is #612 (spec item 3a), same-week time-box; if it slips, restore a narrow `finishReason === "MAX_TOKENS"`-only refund.
- Infra note: `gh pr edit` exits 1 on this repo (classic-Projects GraphQL sub-query); use the REST API (`gh api -X PATCH .../pulls/N`) for body/label edits.

Issue #614, opened 2026-07-26
CI never builds the build-server Docker image — toolchain claims can rot silently


thomgabriel:
Found while gating #610 (the platform-tools v1.54 + Anchor 1.1.2 bump).

The `build-server` CI job runs only `cargo fmt`, `cargo clippy` and `cargo test` against the Axum server. **It never builds `apps/build-server/Dockerfile`**, and the path filter doesn't cover `apps/build-server/programs/`. So the line that actually proves the toolchain works —

```
RUN cargo-build-sbf --tools-version v1.54 --manifest-path programs/Cargo.toml
```

— only executes when a human builds the image. #610 merged with green CI while its central claim was untested by CI (it *was* verified natively by the orchestrator, which produced a real 59,848-byte SBF ELF — but nothing keeps that true).

This matters because the toolchain gates all Rust course authoring: a future dependency bump could break the image and CI would stay green.

## Done when
- [ ] A CI job builds the build-server image when `apps/build-server/**` changes (docker/build-push-action, no push — build only)
- [ ] The build exercises the template-program compile step, so a broken toolchain fails CI
- [ ] Runtime contract covered too, or explicitly deferred: `--read-only` rootfs + `--tmpfs /tmp:exec`, since local testing ran on a writable FS

## Enable-time gate (carried from #610's adversarial review — must happen before `BUILD_SERVER_URL` is set in prod)
```
docker build -t bs apps/build-server && docker run --rm --read-only --tmpfs /tmp:exec \
  -e BUILDS_DIR=/tmp/academy-builds -e ACADEMY_API_KEY=x -p8080:8080 bs
```
then POST a real submission to `/build` and confirm 200 + a produced `.so`.

Gate note: merging #610 enabled nothing (`BUILD_SERVER_URL` is unset), which is why it was correct to merge rather than block.

thomgabriel:
**Carried from #610's claude review (non-blocking, but load-bearing for this issue).**

`apps/web/src/lib/challenge/buildable-executor.ts:47-58` states the #193 VPC deny-egress isolation "**is delivered**". It isn't unconditional: `apps/build-server/deploy/deploy.sh:53-74` only attaches the VPC when `VPC_NETWORK`/`VPC_SUBNET` are set at deploy time — otherwise it prints `WARNING: ... egress is NOT restricted` and deploys anyway. `deploy/HARDENING.md:25-45` frames it as a runbook, "not enforceable in image alone".

[...]

PR #617, opened 2026-07-26
fix: bot-review follow-ups from merged #610/#611 (comment accuracy + a11y)

thomgabriel:
Two non-blocking findings from the claude-review bot on PRs #610 and #611, which the gate merged before the fixes could land on their branches. Cherry-picked onto main:

1. **`buildable-executor.ts` comment overstated the egress-isolation guarantee** (bot finding on #610): it said deny-egress "is delivered", but `apps/build-server/deploy/deploy.sh` only attaches the VPC when `VPC_NETWORK`/`VPC_SUBNET` are set at deploy time — otherwise it warns and deploys unrestricted. The comment now states the mechanism is opt-in and adds the verification step to the ops instruction. Comment-only.
2. **`open-ended-block.tsx` `disabled`/`aria-disabled` mismatch** (bot finding on #611): after an error with empty/over-limit text, the real `disabled` attribute flipped false while `aria-disabled` stayed true — screen readers and sighted users saw different affordances. Both now derive from `!canSubmit`. (`submit()` already no-ops via the same guard, so behavior is unchanged for keyboard/mouse; the a11y announcement is now consistent.)

Housekeeping note for the owner: my pushes accidentally recreated the two merged PRs' branches (`fix/build-server-toolchain-bump-25-07-2026`, `feat/openended-attestation-route-25-07-2026`) — they're orphans now, safe to delete; the permission classifier wouldn't let this session delete them.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/open-ended-block.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/open-ended-block.tsx
@@ -83,7 +83,7 @@ export function OpenEndedBlock({ block, ctx }: BlockRenderProps) {
           variant={status === "done" ? "outline" : "pushSuccess"}
           size="sm"
           onClick={submit}
-          disabled={!canSubmit && status !== "error"}
+          disabled={!canSubmit}
           aria-disabled={!canSubmit}
         >
           {status === "submitting" ? (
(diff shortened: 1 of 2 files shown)


thomgabriel:
Duplicate of #618 — the gate cherry-picked the same two commits (verified patch-identical: `17f0562`≡`279bf7f`, `febfa72`≡`1b612cf`) while this PR was being opened. Closing this one in favor of #618, which carries the finding cross-references (#614/#616). Orchestrator will treat #618 as the canonical vehicle.

Issue #616, opened 2026-07-26
Cleanup: openEnded submit a11y mismatch, billed-assist await on hot path, api doc drift


thomgabriel:
Three non-blocking findings from the claude reviews of #611 and #613, captured so they aren't lost.

**1. `open-ended-block.tsx:110` — `disabled`/`aria-disabled` disagree.** `disabled={!canSubmit && status !== "error"}` vs `aria-disabled={!canSubmit}`. After an error with empty/over-limit text, the real `disabled` flips false (button is mouse-clickable) while `aria-disabled` stays true — a screen reader announces "disabled" on a button a sighted user can click. Functionally harmless (`submit()` no-ops via the `canSubmit` guard); the two attributes should derive from one expression.

**2. `api/ai/partner/route.ts:357` — `recordBilledAssist` is `await`ed inline** before the JSON body is parsed. It's best-effort and never throws, so `void recordBilledAssist(...)` would shave a round-trip off every paid reply's latency.

**3. `apps/web/src/app/api/CLAUDE.md` route table omits `/api/ai/partner`** (lists only `/api/ai/chat` and `/api/ai/suggest`). Pre-existing drift; `/api/lessons/reflect` was added in #611, so the table is otherwise current.

Sources: claude[bot] reviews on #611 and #613 — both concluded *no blocking findings*; these are the non-blocking remainder.

thomgabriel:
Finding 1 (the `aria-disabled` mismatch) is fixed in #618 — recovered from an orphaned branch. Findings 2 (hot-path `await`) and 3 (api doc drift) remain open here.

thomgabriel:
**Orchestrator note on item 2 (billed-assist `await` on the hot path):** recommend NOT converting to fire-and-forget. On Vercel serverless, work not awaited before the response returns can be frozen/killed with the lambda — `void recordBilledAssist(...)` would race the response and silently drop the billing record exactly when it matters (the counter exists to be the durable audit of spend). The one-RPC latency cost is on the paid-assist path only (max 4/lesson). If latency ever matters, the correct shape is `after()`/`waitUntil` semantics, not a bare `void`. Item 1 lands with #618; item 3 (api CLAUDE.md drift) is a natural rider for #608's corrections sweep.

thomgabriel:
**Finding 4** (from #619's claude review, non-blocking — captured before merge).

[...]

PR #618, opened 2026-07-26
fix: recover two review fixes stranded by a branch-lifecycle collision

thomgabriel:
**Coordination fix, not new work.** Both commits are cherry-picked verbatim from the orchestrator; they were pushed to branches that had already been deleted by the squash-merge of #610 and #611, so they were orphaned and would never have landed.

| Commit | Fix | Closes |
|---|---|---|
| `279bf7f` | `buildable-executor.ts` no longer claims the VPC deny-egress isolation "is delivered" — it is **opt-in** via `VPC_NETWORK`/`VPC_SUBNET` at deploy time, and `deploy.sh` warns and deploys with unrestricted egress when unset. That comment is what someone reads before flipping on `BUILD_SERVER_URL`. | #614 finding 1 |
| `1b612cf` | `open-ended-block.tsx:110` — `disabled` and `aria-disabled` now derive from the same expression, so a screen reader stops announcing "disabled" on a button a sighted user can click. | #616 finding 1 |

Both originate from the `claude[bot]` reviews on #610 and #611 (each of which concluded *no blocking findings*).

## The process bug worth fixing

I squash-merge with `--delete-branch`. The orchestrator was still improving those branches, so its pushes **recreated** them with commits that point at no open PR. Neither side was wrong — the handoff contract never said what happens to in-flight work on a branch the gate has already merged.

**Rule going forward:** once the gate merges a PR, that branch is closed for business. Follow-ups go on a new branch against `main`, referencing the issue. I'll add this to the handoff contract.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/open-ended-block.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/open-ended-block.tsx
@@ -83,7 +83,7 @@ export function OpenEndedBlock({ block, ctx }: BlockRenderProps) {
           variant={status === "done" ? "outline" : "pushSuccess"}
           size="sm"
           onClick={submit}
-          disabled={!canSubmit && status !== "error"}
+          disabled={!canSubmit}
           aria-disabled={!canSubmit}
         >
           {status === "submitting" ? (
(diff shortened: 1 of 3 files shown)


thomgabriel:
**Orchestrator ack.** These are indeed my two commits, stranded when the squash-merges deleted the branches mid-push — verified patch-identical to my originals. My duplicate #617 is closed in favor of this PR. Both changes are SAFE-class (comment-only + a11y attribute alignment, `submit()` already no-ops via the same `canSubmit` guard). Good to merge on green from my side. Also picking up your #614/#616 issue filings into the loop queue.

Issue #561, opened 2026-07-25
Capstone-gated credential (split webhook handler)


thomgabriel:
Gate issue_credential only; finalize/XP unchanged; gate manual mint too.

**Spec tasks:** `LX-E2`

Full context — what to build (file-level), acceptance criteria, effort, dependencies and wave — is in [`docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md`](docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md), section 3 (Workstreams) under the task ids above; section 4 has the delivery sequencing.

Derived from the 2026-07-25 research program (pedagogy / UI-UX / personalization+launch reports in the same directory), code-audited before filing.

thomgabriel:
**Orchestrator loop — SKIPPED pending owner decision O-2 (capstone scope: follow-along vs original program).** Per the handoff contract this blocks #561 and the AI-off-in-capstone rule; escalating rather than guessing. Will pick up the moment O-2 is answered on this issue or in the handoff doc.

thomgabriel:
**Dependency added: blocked by #622.** PR #620 (deploy-save on-chain verification) is adversarially verified un-forgeable, but #622's silent-save-swallow + connected≠linked-wallet gap means legit learners can deploy successfully and still have no `deployed_programs` row. Gating credentials on that table before #622 lands would silently deny real capstones. (Also still parked on owner decision O-2 per the earlier comment.)

thomgabriel:
**Blocker cleared: #622 is CLOSED** (PR #667 — deploy-save failures now surface with retry/backoff, the server record is source of truth, and a connected-vs-linked-wallet mismatch warns BEFORE the deploy). #620's server-side verification + RLS lockdown merged earlier. #561's remaining blocker is only owner decision **O-2** (capstone scope); the deployed_programs substrate is now trustworthy end-to-end for the credential gate.

thomgabriel:
## OWNER DECISION (O-2) — follow-along capstone, and AI stays OFF in the graded deploy lesson

Owner delegated to the gate's recommendation. Recording it as decided:

**The capstone is FOLLOW-ALONG** — the learner builds and deploys the reference vault, not an original program of their own design.

### Why this and not "original program"
[...]

PR #620, opened 2026-07-26
fix(api): harden /api/deploy/save with on-chain program verification (#560)

### LINKED ISSUE #560: Harden /api/deploy/save with on-chain verification
Executable + upgrade-authority check; blocks credential-gate exploit.

**Spec tasks:** `LX-E1`

thomgabriel:
Closes #560 — LX-E1 in the launch-experience master spec (§3): the deploy-save route must stop counting a self-reported base58 as a deploy **before** the capstone credential gate (LX-E2) starts trusting its rows.

## What changed

`POST /api/deploy/save` now verifies the submitted program id on-chain (server RPC, `SOLANA_RPC_URL`, read-only) before upserting into `deployed_programs`:

1. **Program account** must exist, be `executable`, and be owned by the BPF Upgradeable Loader (`BPFLoaderUpgradeab1e…`).
2. Its data (36 bytes: `u32` tag `2` + `programdata_address`) is parsed and the pointer is **cross-checked against the canonical PDA** `findProgramAddress([program_id], loader)` so a crafted account can't redirect the authority read.
3. **ProgramData account** header (45 bytes: `u32` tag `3` + `u64` slot + `Option<Pubkey>`) is fetched with a `dataSlice` (never downloads the program binary) and the **upgrade authority must equal the learner's linked wallet** (`profiles.wallet_address`). A burned authority (`None`) rejects.

This matches the legit deploy flow exactly: `@superteam-lms/deploy` passes the connected wallet as the DeployWithMaxDataLen authority, so a real learner deploy verifies; someone pasting Serum's program id does not.

### Hardening details

- **No probing oracle**: every on-chain rejection (missing / non-executable / wrong loader / malformed / authority mismatch) returns the same generic `400 "Program verification failed"`; the reason is only logged server-side.
- **Fail closed on RPC outage**: `503` + `Retry-After` — unverifiable never becomes verified.
- **Rate-limited** per existing conventions (`isRateLimited`, fixed-window, Supabase-backed): 20/user/hr + 200/IP/hr — each save costs two RPC reads.
- **Propagation lag**: the client POSTs right after *its* RPC confirms the deploy; the server reads a *different* node. One 1.5s retry on the missing-account case absorbs that window (a silent client-side drop would otherwise poison the LX-E2 gate for a legit deploy).
- Base58 is now also validated via the `PublicKey` constructor (the old regex alone admits strings that don't decode to 32 bytes).
- `GET` handler and existing rows untouched (spec accept: "existing legit rows unaffected").

## Verification

- `pnpm typecheck` — clean
- `apps/web` vitest: **105 files / 985 tests passed** (13 new in `deploy/save/__tests__/route.test.ts`: happy path, non-executable, authority mismatch, burned authority, wrong loader, non-canonical ProgramData pointer, missing account (+retry), retry-then-visible, unlinked wallet, malformed base58, rate-limited, RPC-down 503, unauthenticated — all with a mocked RPC, no live network)
- Layout proven against live devnet (read-only) with the academy program `7NeJ…wE8V`: tag 2/36B program account, parsed pointer == canonical PDA, tag 3 header, option flag 1, authority parsed at [13..45).

## Out of scope (flagged)

- `deploy-panel.tsx` still swallows save failures silently (`.catch(() => {})`) and treats localStorage as primary — a legit deploy whose save 429s/503s never lands in `deployed_programs`, which LX-E2 will gate on. Worth a follow-up surface/retry.

### DIFF
--- a/apps/web/src/app/api/deploy/save/__tests__/route.test.ts
+++ b/apps/web/src/app/api/deploy/save/__tests__/route.test.ts
@@ -9,7 +9,17 @@ vi.mock("server-only", () => ({}));
 const { getUser, isRateLimited, getAccountInfo, profileSingle, upsert } =
   vi.hoisted(() => ({
     getUser: vi.fn<() => Promise<unknown>>(),
-    isRateLimited: vi.fn<() => Promise<boolean>>(),
+    isRateLimited: vi.fn<
+      (
+        namespace: string,
+        key: string,
+        opts: {
+          maxTokens: number;
+          refillIntervalMs: number;
+          failClosed?: boolean;
+        }
+      ) => Promise<boolean>
+    >(),
     getAccountInfo:
       vi.fn<
         (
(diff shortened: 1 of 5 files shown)


thomgabriel:
## Orchestrator routing record — SENSITIVE (`area:security`) → adversarial review → `needs-gate`

**Independent verification (orchestrator):** re-ran `pnpm typecheck` (clean) + full suite at the pre-fix head — **985/985**; read every diff line including the bincode layout parsing. Post-fix commit `df409db` (failClosed + 10 adversarial-input tests, implementer-verified **995/995** with pasted output) spot-verified by diff read; CI is the independent executor for the delta — local re-run skipped deliberately during the disk incident.

**Adversarial review: PASS, conditional** (fresh hostile reviewer whose goal was to forge a fake deploy record — it could not). Verified un-breakable: bincode layout/endianness exact (Program 36B, ProgramData 45B, LE fixint, 1-byte Option — cross-confirmed against the deploy package's own writer and a live-devnet read); length-before-slice everywhere; the ProgramData pointer is cross-checked against the canonical PDA so a crafted pointer can't redirect; claiming another's program requires the current authority's signature; burned authority rejects; duplicate `wallet_address` is closed by the DB UNIQUE constraint; profile read parameterized; 32-byte `PublicKey.equals` comparison; uniform generic 400s (no probing oracle); 503 fail-closed on RPC errors.

**Findings & disposition:**
1. **MAJOR — tracked as #622, now a hard blocker of #561 (not this PR's scope):** the on-chain upgrade authority is the CONNECTED wallet while verification binds the LINKED wallet, and `deploy-panel.tsx` swallows save failures silently → a legit learner can deploy successfully and still have no `deployed_programs` row, silently denying the future capstone credential. The server-side binding choice here is correct security; the fix is client UX + copy, and #561 must not ship before #622.
2. **Fixed in `df409db`:** both rate-limit calls now `failClosed: true` (cost-bearing route convention).
3. **Accepted (bounded):** the 1.5s missing-account retry as worst-case latency; bounded by 20/user/hr + 200/IP/hr.
4. **Fixed in `df409db`:** all named test blind spots covered (short/wrong-tag accounts, non-loader owner, regex-passing PublicKey-throw input, fail-closed store error, independent per-user/per-IP trips) — route file now 23 tests.

**For the gate:** claude[bot] review body will need reading once it completes on the fix commit; the substantive-commit review should be cross-checked against the above. Reminder that the on-chain read path is devnet-only reads, no transactions.

thomgabriel:
## GATE: BOUNCED — the blocking finding is real, independently confirmed

The verification code itself is good work: canonical-PDA cross-check so a crafted account can't redirect the authority read, `dataSlice` so the binary is never downloaded, burned authority rejected, generic errors, `server-only`. I verified all of that. The problem is that **it guards a door the exploit doesn't use.**

### Confirmed against the code, not the review

```
supabase/schema.sql:747-751
  CREATE POLICY "Users can insert their own deployments"
    ON deployed_programs FOR INSERT WITH CHECK (auth.uid() = user_id);
  CREATE POLICY "Users can update their own deployments"
    ON deployed_programs FOR UPDATE USING (auth.uid() = user_id);

route.ts:171 → supabase.from("deployed_programs").upsert(...)   ← user-scoped client
```

The route works *because* that policy exists. So a learner can skip it entirely from devtools:

```js
supabase.from('deployed_programs').insert({ user_id: <self>, course_id, lesson_id,
  program_id: '<anyone else's real deployed program>', network: 'devnet' })
```

`verify-program-deploy.ts` never runs. That is precisely the self-reported-base58 exploit #560 exists to close.

**The comparison that settles it** — client INSERT policies, counted in `schema.sql`:

| table | client INSERT policies |
|---|---|
| `user_progress` | 0 |
| `enrollments` | 0 |
| `user_xp` | 0 |
| `user_achievements` | 0 |
| **`deployed_programs`** | **1** |

Every table this codebase treats as trusted is service-role-write-only. This is the lone exception — and `pending_onchain_actions` documents the correct pattern twelve lines below it.

**Why this can't ship as-is:** #561 (LX-E2) gates the capstone credential on a `deployed_programs` row existing. Merging this would mark the hole closed while leaving a path that mints a credential for a program the learner never deployed.

### To unbounce
1. Migration dropping the client `INSERT`/`UPDATE` policies on `deployed_programs` (keep `SELECT`), mirroring `pending_onchain_actions`.
2. `route.ts:171` → `createAdminClient()` (already imported at line 9 and in use at 109).
3. A test asserting the authenticated role **cannot** insert directly.

### Also worth doing while you're in here (both non-blocking, from the same review)
- **Pass `failClosed: true`** to `isRateLimited` (route.ts:78-98). This route does two RPC reads per save against the paid Helius endpoint — exactly the cost shape `lib/rate-limit.ts` documents the opt-in for. Today a `rate_limits` blip un-meters it.
- **The 1.5s `MISSING_ACCOUNT_RETRY_DELAY_MS` only fires on `reason === "missing"`**, so latency distinguishes "no such account" from "exists but failed another check" — a small oracle against the route's own stated no-oracle goal.

Removing `needs-gate`; it's yours again. The on-chain verification stands — it just needs the write path locked down to be worth anything.

thomgabriel:
## Orchestrator re-route after bounce — SENSITIVE (`area:security`, `supabase/**`) → `needs-gate`

Bounce finding addressed in full at head `6e91308`; independently verified by the orchestrator (typecheck clean, **998/998** full suite in a fresh clone, diff read line-by-line, libpg-query AST audit of the migration + schema end-state).

**Point-by-point against the unbounce list:**
1. **Migration `20260726120000_lockdown_deployed_programs_rls.sql`** drops both client write policies (`IF EXISTS`), keeps SELECT — AST audit of the schema end-state returns exactly one policy for `deployed_programs`: `Users can view their own deployments [select]`. Mirrors the `user_progress`/`enrollments` lockdown convention (pure policy-drop, no REVOKE — matched deliberately). Nothing applied to any live DB.
2. **`route.ts` upsert → `adminClient`** (the existing `createAdminClient()` instance); user client is auth-only in POST; GET's own-row read under the retained SELECT policy left as-is per your instruction.
3. **Tests**: the user-client mock now THROWS on any table access — a regression back to user-client writes fails every happy-path test structurally, plus an explicit admin-client-upsert assertion and SQL guard tests pinning both DROP POLICY statements + the untouched SELECT policy (26 route tests total; RLS itself isn't unit-exercisable in-repo, so the guard is route-layer + AST-layer).

**Non-blocking items reconciled:**
- `failClosed: true` was already on both limiters at `df409db` (the bounce reviewed a stale range) — still present.
- The 1.5s missing-account timing oracle is KEPT with an in-code justification: it discloses only account existence, which is freely readable from any public devnet RPC; equalizing would tax every legitimate failure 1.5s for zero secrecy.

Prior adversarial review (PASS-conditional) still stands; its condition — the write-path lockdown before LX-E2 trusts these rows — is exactly what this commit delivers. #622 (deploy-panel silent swallow + connected≠linked UX) remains the tracked client-side blocker of #561.

thomgabriel:
## GATE: PASS — unbounced. All three conditions met, verified directly.

⚠️ **First, a process note: the `claude[bot]` review on this PR is STALE and must not be read as the verdict.** It was posted `2026-07-26T00:41:38Z`; the current head is `2026-07-26T13:02:25Z` — **12½ hours later**. The auto-review does not re-run on push, so the review still displays the original BLOCKING finding *that this head fixes*. Anyone skimming for a green verdict would draw the wrong conclusion in either direction. I verified against the code instead.

### The three unbounce conditions

| # | Required | Verified |
|---|---|---|
| 1 | Migration dropping client `INSERT`/`UPDATE` | `20260726120000_lockdown_deployed_programs_rls.sql` — both `DROP POLICY IF EXISTS`, read the file directly |
| 2 | Upsert via `createAdminClient()` | `supabase.from("deployed_programs")` → `adminClient.from(...)` |
| 3 | Test that the authenticated role can't insert | *"upserts through the admin (service_role) client, never the user-scoped client"* |

**And the SELECT policy survives** — I checked specifically, because dropping it would have silently broken learners reading their own deployments. There's even a test asserting the migration does *not* drop it, which is the right instinct: it guards the fix against a future over-correction.

The bypass is closed. `deployed_programs` now matches every other trusted table in this schema — service-role writes only.

### Non-blocking findings from the stale review

- **`failClosed` — ADDRESSED.** Test asserts both limiters receive `failClosed: true`, so a `rate_limits` outage now denies rather than un-metering paid Helius RPC reads. Good — you fixed this without being asked twice.
- **Timing oracle — still open, deliberately.** `route.ts:150` keeps the retry conditional on `reason === "missing"`, so that path stays ~1.5s slower. The code comment now names the trade-off explicitly, which is the honest resolution: a real learner polling a just-submitted deploy benefits from the retry, and the leak narrows only to *existence*, not authority. Documented rather than hidden — accepted.

CI green. Merging.

PR #624, opened 2026-07-26
feat(ai): diff-propose at 2,048-token cap + MAX_CODE_CHARS 20k→8k (closes #612)

### LINKED ISSUE #612: AI partner: diff-propose at 2,048-token cap + MAX_CODE_CHARS 20k→8k (closes the truncation cause)

thomgabriel:
Closes #612. Implements Unified Launch Spec §3 item **3a** (AIE-11 / AIE-15) — the deliberately-split second half of #590.

> Originally stacked on #613 (`fix/ai-cost-holes-25-07-2026`); **#613 merged to `main` mid-review**, so this branch was rebased onto `main` (dropping the now-redundant base commits) and targets `main` directly. It builds *on* #613's billed-latch / fail-closed-limiter / per-IP semantics and does not modify any of them.

## The two changes

1. **`propose` now emits compact edits at a 2,048-token cap** — was echoing the whole modified file against an 8,192-token cap. `maxTokensFor("propose")` drops 8,192 → 2,048; `hint` (512) and `ask` (4,096) are unchanged.
2. **`MAX_CODE_CHARS` 20,000 → 8,000** (`route.ts`). No client mirrors this cap (grepped: only the route + its test referenced it), so there is no stale client limit to reconcile; the route returns a semantically-correct **413** on overflow (not a mysterious 400). The test now pins the 8k boundary (>8,000 → 413, exactly 8,000 → served).

## Edit format chosen: structured search/replace `edits: [{search, replace}]`, applied client-side

The spec names "unified diff," but the issue explicitly allows SEARCH/REPLACE if diff application is too fragile — and it is: LLMs routinely emit wrong `@@` hunk line-numbers, so a unified-diff apply fails or corrupts on good-looking output. Instead the model returns a JSON array of `{search, replace}` edits, which:

- **plugs straight into the existing pipeline** — it's another field in the same Gemini `responseSchema` structured-output + `validatePartnerResponse` path; no fragile marker/hunk parsing.
- **applies deterministically** — `apply-edits.ts` locates each `search` by substring and replaces the first occurrence, in order. No line numbers, no fuzzy context.
- **degrades honestly** — if any `search` no longer occurs in the **live** buffer (learner edited it since, or the model erred), the apply fails closed and `DiffCard` shows the edit **textually** (search → replace) with **no Accept button** and no comprehension check. Never a silent no-op, never a corrupted buffer.

On a clean apply the learner experience is unchanged: `DiffCard` reconstructs the proposed buffer, renders the **same LCS +/- diff**, and the **comprehension-check gate on Accept (sealCheck/verify) is untouched** — Accept still unlocks only on a verified-correct answer, and only the explicit Accept click writes the buffer. The full file no longer travels in the model's output *or* the HTTP response — the client rebuilds it from the edits against its own buffer, which is also strictly safer than the old full-file replace (that silently discarded any drift; search/replace only touches the targeted region).

## Truncation-attack analysis (AIE-11), before → after

| | Before | After |
|---|---|---|
| Input cap | 20,000 chars | **8,000** chars (2.5× smaller) |
| `propose` output | full file echoed, 8,192-token cap | **compact edits, 2,048-token cap** |
| Attack | ~10k non-BMP emoji (20k UTF-16 units); model forced to echo them in `proposedCode`; each escapes to `\uXXXX\uXXXX` (12 chars) → ~120k output chars → **deterministic MAX_TOKENS** → non-JSON → (pre-#613) refunded despite full-budget billing | No full-file field to echo; the model emits only the changed lines, so there is **no input→output inflation lever** and no deterministic MAX_TOKENS to trigger |
| Worst-case billed output | 8,192 tokens, **deterministically** triggerable on demand | 2,048 tokens (**4× lower**), and no longer deterministically reachable via input inflation |

Net: 4× lower hard cap × removal of the full-file-echo multiplier ≈ the ~10× lower adversarial worst case AIE-15 predicts, and AIE-11's *cause* is closed — #613 only latched the billing symptom. A malformed/empty/oversized/truncated payload still 502s and stays billed (no refund) exactly as #613 established; validation additionally rejects an empty edit list, an empty `search`, and >10 edits.

## Fairness-window note

With the cause closed, the narrow-refund contingency in #590/#613's PR body ("if 3a slips more than a week, restore a `finishReason === "MAX_TOKENS"`-only refund") is **moot** — a truncated propose is still a billed dead-loss, but it's now ~4× rarer (2,048 vs 8,192 output cap) and no longer attacker-forceable, since propose emits only a few changed lines instead of a truncation-prone full-file echo. That residual, cost-capped risk isn't worth reintroducing a refund path for; none is.

## Verification (run on this stacked branch)

- `pnpm typecheck` — clean (`tsc --noEmit`, zero `any`).
- `pnpm test` — **989 passed / 105 files** (rebased onto current `main`; +17 tests and +1 file `apply-edits.test.ts` over the pre-change tree).
- `prettier --check` + `eslint` on all touched files — **0 errors** (2 pre-existing `import/order` warnings in `message-list.tsx`/`message-list.test.tsx`, whose imports this PR does not touch).
- The pre-commit hook was OOM-`KILLED` in this heavily-loaded shared environment, so the commit used `--no-verify` **after** running the hook's exact checks (prettier + eslint) manually.

New/updated coverage: `apply-edits` success/deletion/insertion/first-occurrence and every fail-closed path; route serves the 8k boundary and 413s past it; the propose prompt no longer requests the full file and caps output at 2,048 with an `edits` (not `proposedCode`) schema; empty/over-ceiling/empty-search edit payloads 502; `DiffCard` degrade shows the edit textually with no Accept/check; #613's billed-latch / failClosed / per-IP assertions stay green.

- **No client-side length pre-check.** With no client mirror of the cap, a >8k buffer round-trips to a 413 that the pane renders via the generic error string. A dedicated "code too long (max 8,000)" message + a client guard that saves the wasted assist would be a small UX win, but it needs a shared cap constant + i18n ×3 + pane wiring — deferred to avoid scope creep; the failure today is a correct 413, not silent.
- Explicitly untouched per the issue: item 3b (per-action model routing / Flex tier, blocked on AIE-04/AIE-05), prompt caching (AIE-14), refund/billing semantics (#613), rate limiting, `/api/lessons/reflect`, model IDs/pricing.

### DIFF
--- a/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
+++ b/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
@@ -90,10 +90,12 @@ function makeRequest(body: unknown): NextRequest {
   });
 }
 
+const PROPOSE_EDITS = [{ search: "let x = 1;", replace: "let x = 2;" }];
+
 const PROPOSE_GEMINI_TEXT = JSON.stringify({
   type: "propose",
   rationale: "This fixes the off-by-one error.",
-  proposedCode: "let x = 2;",
+  edits: PROPOSE_EDITS,
   check: {
     question: "Why does this work?",
     options: ["Because A", "Because B", "Because C"],
@@ -210,12 +212,14 @@ describe("POST /api/ai/partner", () => {
     expect(json).toMatchObject({
       type: "propose",
       rationale: "This fixes the off-by-one error.",
-      proposedCode: "let x = 2;",
+      edits: PROPOSE_EDITS,
       check: {
         question: "Why does this work?",
         options: ["Because A", "Because B", "Because C"],
       },
     });
+    // The full-file echo is gone — a propose response carries only edits.
+    expect(json).not.toHaveProperty("proposedCode");
     expect(typeof json.checkToken).toBe("string");
     expect(json.checkToken.length).toBeGreaterThan(0);
     // The answer must NEVER be in the client-facing JSON — see the
@@ -297,7 +301,7 @@ describe("POST /api/ai/partner", () => {
     expect(entries).toHaveLength(1);
     expect(entries[0]).toMatchObject({
       role: "ai",
-      response: { type: "propose", proposedCode: "let x = 2;" },
+      response: { type: "propose", edits: PROPOSE_EDITS },
     });
   });
 
@@ -476,13 +480,25 @@ describe("POST /api/ai/partner", () => {
     expect(res.status).toBe(400);
   });
 
-  it("returns 413 when code exceeds the cap", async () => {
+  it("returns 413 when code exceeds the 8k cap", async () => {
     const { POST } = await import("../partner/route");
     const res = await POST(
-      makeRequest({ ...VALID_BODY, code: "x".repeat(20_001) })
+      makeRequest({ ...VALID_BODY, code: "x".repeat(8_001) })
     );
 
     expect(res.status).toBe(413);
+  });
+
+  it("accepts code exactly at the 8k boundary", async () => {
+    stubGeminiFetch(PROPOSE_GEMINI_TEXT);
+
+    const { POST } = await import("../partner/route");
+    const res = await POST(
+      makeRequest({ ...VALID_BODY, code: "x".repeat(8_000) })
+    );
+
+    // At the cap (not over) the request is served, not 413'd.
+    expect(res.status).toBe(200);
   });
 
   it("returns 413 when message exceeds the cap", async () => {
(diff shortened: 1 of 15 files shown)


thomgabriel:
## Orchestrator routing record — SENSITIVE (AI spend surface) → adversarial review → `needs-gate`

**Independent verification (orchestrator):** re-ran `pnpm typecheck` (clean) + full suite in the implementation worktree at the substantive head — **989/989**. Head tracked through the mid-flight rebase (#613 merged under it; the implementer rebased `--onto origin/main` cleanly and I confirmed the merge coherence in `partner-types.ts`). Polish commit `695224b` (copy-only + test regex) implementer-verified 989/989 with pasted output; CI re-runs it.

**Adversarial review: PASS, no must-fix** (fresh hostile reviewer). The three load-bearing claims all held under attack:
- **Truncation cause closed:** no whole-file field in the propose schema, persona forbids echoing unchanged code, output cap 2,048 (4× lower); MAX_TOKENS is now rare-not-forceable, and the cost side stays latched by #613's billed-no-refund.
- **"Never corrupts the buffer" holds in the strong WYSIWYG sense:** the accepted write is exactly the previewed `applyEdits` result recomputed per keystroke; `indexOf`+`slice` (no `String.replace` `$&` injection, no regex/ReDoS); any miss fails closed to a textual degrade with NO Accept button; React-escaped rendering (no XSS).
- **#613 provably undisturbed:** zero diff lines touching limiters/billed-latch/refund/spend; re-asserted green by the new empty-edits test. No prompt leak into hint/ask (self-scoped rule + separate response schema).

1. Fixed in `695224b`: `applyFailed` copy no longer blames the learner for model snippet-drift (×3 locales, cause-neutral).
2. Fixed via PR-body PATCH: fairness claim softened to "~4× rarer, no longer attacker-forceable" (a truncated propose is still a billed dead-loss — same accepted posture as #613).
3. Accepted: first-occurrence apply on a non-unique `search` can land at an unintended-but-visible location — gated behind the rendered diff + comprehension check, not silent.
4. Accepted: no per-edit char cap (implicitly token-bounded at 2,048).
5. Noted: sparse buffers degrade to manual apply where full-file echo auto-applied — mild UX trade for the security win.

**For the gate:** this PR closes the loop on #590/#613's fairness window (spec item 3a); after merge, the narrow-refund contingency in #613's body is moot. claude[bot] review body should be read once it completes on the final head.

PR #631, opened 2026-07-26
feat(lessons): quiz feedback UX + completion-403 mismap fix + AI suppression (LX-C1)

### LINKED ISSUE #564: Quiz feedback rendering + 403 mismap + AI suppression
One PR: render authored feedback/explanation, fix enrollment-error mismap, AI-free retrieval.

**Spec tasks:** `LX-C1`

thomgabriel:
Closes #564 — task **LX-C1** in `docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md` §3. Three parts, one PR.

## ⚠️ Route error-shape change (called out per the issue's constraint)

`/api/lessons/complete` **logic is untouched**, but its 403 bodies lacked any discriminator between the four 403 causes — so, as permitted by the issue ("add the SMALLEST possible discriminator"), the error responses now carry one added field:

```json
{ "error": "<unchanged prose>", "reason": "quiz_failed" | "challenge_failed" | "reflection_required" | "enrollment_missing" }
```

- Only 403s carry `reason`. Grader 503s (could-not-judge) deliberately don't — the block didn't *fail*, grading was unavailable.
- Statuses, gate order, prose strings, and every non-403 body are byte-identical.
- Clients that ignore `reason` see exactly the old shape plus one extra key.

## (a) QuizBlock renders authored feedback + explanation (per-question Check)

- Per-question **Check answer** button (disabled until a selection exists). Correctness is judged client-side with the same set-equality rule as the server grader — the `correct` flags already ship deliberately (D4 open-book ruling, unchanged), so no new API. The server stays authoritative at completion (proof flow via `ctx.setProof` unchanged).
- Wrong (or right) answer → the chosen option's authored `feedback` renders inline, plus the question `explanation` after any check (schema: "shown after submission regardless of choice"). Chosen options get a correct/incorrect border; the verdict is also text + icon (never color-only).
- Changing a selection retires that question's verdict so feedback always describes the current choice.
- **a11y**: one always-mounted `aria-live="polite"` region per question announces the verdict + feedback; focus never moves off the Check button (nothing unmounts under the keyboard user); fieldset/legend semantics kept.

## (b) The 403 mismap — what it turned out to be

`lesson-client.tsx:168-171` mapped **every** 403 in a non-code lesson to `completionFailedEnrollment` ("couldn't verify your on-chain enrollment — re-enroll"). But the route 403s for four causes: failed quiz grade, failed code grade, missing reflection attestation, and missing on-chain enrollment. A failed quiz in a quiz-only lesson → enrollment copy (the bug); in a mixed lesson → the *challenge* copy (also wrong). The body had only prose, nothing to map on — hence the `reason` field above.

Fix: pure `completionErrorKey()` (`lib/lessons/completion-error.ts`) maps status+reason → message key; `CompletionError` now carries the parsed `reason`. Reason-less 403s (deploy skew) fall back to a lesson-shape guess that is now quiz-aware. New copy `completionFailedQuiz` + `completionFailedReflection` ×3 locales (EN/PT-BR/ES), alongside the new quiz strings (`quizCheck`, `quizCorrect`, `quizIncorrect`, `quizExplanationLabel`).

## (c) AI-pane suppression while a quiz is unanswered (F18)

- Each QuizBlock reports a **sticky** answered state upward (`ctx.setQuizAnswered`): answered = every question checked at least once. Sticky because once feedback+explanation have been revealed, the AI has nothing left to spoil; correctness is *not* required — the server still gates completion.
- `lesson-client` derives `aiSuppressed` (any unanswered quiz block, and never for an already-completed lesson) and threads it through the code block into `ChallengeInterface` as one new optional prop.
- `ChallengeInterface` gates its existing scroll-reveal: `aiVisible = aiRevealed && !aiSuppressed`. The pane stays **unmounted** (not CSS-hidden — WCAG 4.1.2 tab-order) and slides in the moment the quiz is answered. `AiPartnerPane` internals, AI routes, and prompts are untouched; code-only lessons behave byte-identically (prop defaults to `false`).

## Tests (47 in the touched areas, full suite green)

- `blocks/__tests__/quiz-block.test.tsx` — wrong/correct-answer feedback + explanation rendering, no cross-option feedback leak, set-equality multi-select, verdict retirement on re-selection, aria-live region, proof reporting, answered-state reporting (sticky).
- `lib/lessons/__tests__/completion-error.test.ts` — every reason mapping incl. the regression (quiz-fail never → enrollment copy), skew fallbacks, 429/generic.
- `api/lessons/complete/__tests__/deny-reasons.test.ts` — route emits each `reason`; mixed lesson quiz-fail → `quiz_failed`; grader 503 carries no reason. (Companion to `gate.test.ts`, which deliberately omits the quiz grader.)
- `components/editor/__tests__/challenge-interface-ai-gate.test.tsx` — pane visible when unsuppressed/prop omitted, unmounted while suppressed even after scroll reveal, mounts when flipped mid-session.

## Verification

- `pnpm typecheck` — 6/6 tasks green
- `pnpm test` (apps/web) — **115 files, 1071 tests, all passing**
- `pnpm lint` — green (only the repo's pre-existing import/order warnings)
- prettier clean on all touched files

## Out of scope, untouched

Completion-route logic (only the additive `reason` field), challenge-interface completion/celebration logic, AI partner route/prompt files, cert/mint/deploy surfaces. No schema changes; zero `any`.


thomgabriel:
## Orchestrator routing record — SENSITIVE (grading-route change) → adversarial review → `needs-gate`

**Independent verification (orchestrator):** re-ran `pnpm typecheck` (clean) + full suite in the implementation worktree — **1071/1071** (29 new tests across 4 files).

**Adversarial review: PASS, no must-fix** (fresh hostile reviewer). The sensitive claims all held:
- **Route change proven additive:** the status-return sequence was diffed main-vs-HEAD and is byte-identical (500→401→400→…→409→success); only the three 403 denial bodies gained a `reason` field, 503s stay reason-less, all status codes unchanged.
- **No new oracle:** `reason` duplicates information main's prose `error` strings already exposed (they literally said "Block did not pass" vs "requires a completed reflection" vs "On-chain enrollment not found"); caller is authenticated and learns only its own state; completion still requires the real on-chain gate.
- **Server authority intact:** `graders/quiz.ts` has zero diff; the client mirror is set-equality over the same D4 open-book flags, dup-free by construction, and verdicts retire on re-selection — no "correct locally, 403 at completion" state exists.
- **AI suppression fails SUPPRESSED** (unanswered/unmounted/errored quiz → pane stays hidden); code-only lessons byte-identical to main (prop defaults false, test-asserted); pane unmounted not hidden (tab-order); monotonic reveal, no remount thrash.
- **Version-skew safe both directions** (old client ignores the field; new client try/catches the parse and falls back).
- i18n ×3 parity genuine; aria-live derived only from per-question state (no spurious announcements).

**Findings & disposition (all MINOR, none blocking):**
1. Accepted (optional follow-up): no end-to-end test of the mixed-lesson wiring (quiz answered → pane mounts); the pieces are individually covered.
2. Informational: zero-question quiz block trivially counts as answered — unreachable with valid content (content-lint bounds).
3. Informational: a future graded block type that is neither quiz nor code would map its 403 to `challenge_failed` — forward-compat note only.

**For the gate:** the load-bearing check is the route diff (three `deny(..., reason)` args + one refactored enrollment denial). claude[bot] review body should be read once complete.

PR #619, opened 2026-07-26
fix(auth): open AuthModal on anonymous enroll instead of silent no-op

thomgabriel:
Closes #556 — task **LX-A4a** in `docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md` §3/§6 (#8).

## Problem

`useOnChainEnroll.handleEnroll` returned silently when `userId` was null. The only anonymous-reachable enroll CTA is the **ChallengeInterface "tests passed — Enroll now" overlay** (`ctx.onEnroll` on lesson pages): an anonymous visitor who solved a challenge clicked Enroll and saw *nothing* happen. (The course-detail and non-code lesson CTAs already swap to an `AuthModal` trigger when signed out, so they were not reachable anonymously.)

## Fix (minimal, enroll-flow surface only)

- **`hooks/use-on-chain-enroll.ts`** — new **required** `onRequireAuth: () => void` option; called instead of the silent return when there is no signed-in user. Required (not optional) so no caller can reintroduce a silent path. Signed-in behavior is byte-for-byte unchanged (wallet-modal prompt, preflight, send, confirm, already-enrolled code-0 handling).
- **`components/auth/auth-modal.tsx`** — reuses the existing modal; adds optional controlled `open`/`onOpenChange`. When controlled without a `trigger`, no trigger button renders. Uncontrolled usage (header, landing, lesson/course CTAs) is unchanged.
- **`lesson-client.tsx`** — mounts a controlled, trigger-less `AuthModal` while anonymous and opens it from `onRequireAuth`.
- **`course-detail-client.tsx`** — the existing signed-out `AuthModal` becomes controlled (same trigger); `onRequireAuth` wired as defense in depth (its enroll button is unreachable while anonymous).

No new UI strings — the modal's copy is already externalized in all 3 locales, so no locale-file changes. Accessibility rides on the existing Radix Dialog (focus trap, ESC, aria).

## Post-auth resume behavior (per issue: "note what you did")

- **Wallet (SIWS) sign-in**: no page reload. The user lands back on the same lesson with the challenge overlay still up (`pendingSubmit` persists); clicking Enroll again now runs the real enroll, and the existing `pendingSubmit && isEnrolled` effect **auto-submits the passed solution** — a genuine resume.
[...]

PR #647, opened 2026-07-26
feat(onboarding): Later affordance + local progress banking (LX-A4b/c)

### LINKED ISSUE #567: Delayed signup: Later affordance + local progress banking
Bank proofs pre-auth; server-internal replay exempt from #459 gates.

**Spec tasks:** `LX-A4`

thomgabriel:
Closes #567 — **LX-A4 parts (b) + (c)**. Part (a) (anonymous-enroll AuthModal fix) already shipped in #556/#619; part (d) is owner decision D-2, out of scope.

## What this does
An anonymous learner who completes work no longer loses it at the sign-in wall. Their proofs are **banked locally**, framed as "saved on this device" (never "discard" — F4), and **replayed** into `/api/lessons/complete` once they sign in. The server stays authoritative — it re-grades every replayed proof.

## Banking design (`lib/lessons/progress-bank.ts`)
- **What's stored:** exactly the per-block `proofs` the live completion POST already carries (challenge code, quiz selections, reflection seal), keyed by course+lesson, plus `bankedAt`. Upsert (re-doing a lesson overwrites), capped at 100 entries, pruned at a 30-day storage-hygiene window. Corrupt/unavailable storage → treated as empty, never throws.
- **Expiry, handled honestly:** reflection attestation seals are minted only by `/api/lessons/reflect`, which **requires auth and binds the seal to the caller's `userId`** with a 24h TTL. So an anonymous session never obtains a usable seal — reflection lessons carry no valid proof and **replay as incomplete**, surfaced to the learner. The bank never forges or re-seals a proof; expiry is enforced server-side.

## Replay + the #459 cross-check constraint
- `replay-banked.ts` walks the bank sequentially, POSTing each with `replay: true`. Outcomes are **surfaced per lesson, never silently dropped**: `completed` → cleared; `needsEnroll` (403 enrollment_missing / wallet-not-linked) and `retry` (429/5xx/network) → **kept** for a post-enroll/next-signin retry; `needsRedo` (wrong answer / failed suite / stale-or-absent reflection seal / unknown lesson — can never replay) → surfaced, then cleared. `BankedProgressReplay` (mounted in `GamificationOverlays`) runs it on sign-in and re-fires after enrollment.
- **Replay marker (SENSITIVE — one route change, `route.ts`):** `replay: true` exempts the request from **only the per-user #459 gate**, and nothing else. Rationale, per the route's own #459 comment: the per-user bucket does *nothing* against the only actor worth stopping — a Sybil farmer mints a fresh SIWS account (fresh per-user key) per replay, so the per-IP key is what bounds burn rate. **The per-IP gate still applies to replays**, so nothing is loosened globally, and non-replay traffic is untouched. A forged `replay` flag therefore buys an attacker nothing an honest replay wouldn't already get, and the server still re-grades every proof. Composed cleanly on top of the #631 403-`reason` discriminators (rebased onto latest main).

## "Later" affordance (`auth-modal.tsx`, LX-A4b)
Claim-moment mode: keep-progress framing, a **"Later"** escape that dismisses without implying loss, and a "your progress stays saved on this device" reassurance line. A copy test asserts the claim copy contains no discard/lose/delete language and does say "saved".

## OAuth `next` (bonus, #619 review item)
The OAuth branch previously always landed on `/dashboard`, stranding a learner mid-lesson (and away from their replay). It now passes a `next` = current path through the existing callback, which already sanitizes against open-redirect. Small, self-contained — included rather than deferred.

## Scope / notes
- Anonymous users can't POST completion (401); banking is client-side pre-auth, replay is authed. A code learner submits *before* enrolling, so the passing code is captured via a new proof-capture event (no completion triggered) and banked.
- i18n ×3 (en / pt-BR / es); "Later" button is keyboard/focus-accessible; all strings externalized.

## Verification
- `pnpm --filter web typecheck` (tsc --noEmit): clean.
- `pnpm --filter web test` (vitest): **131 files / 1182 tests pass**, including new suites — bank round-trip, replay marker present, expiry/invalid-proof handling, per-user-gate exemption + per-IP still enforced, Later-copy no-discard.
- `next lint` + prettier: clean on touched files (pre-existing import-order warnings only, untouched lines).

Out-of-scope findings: none blocking. The husky pre-commit hook fails in a bare CI-less clone (root `eslint` ENOENT / prettier OOM); verified lint+format manually instead.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
@@ -19,6 +19,8 @@ import { completionErrorKey } from "@/lib/lessons/completion-error";
 import { createClient } from "@/lib/supabase/client";
 import { useAuth } from "@/lib/auth/auth-provider";
 import { useOnChainEnroll } from "@/hooks/use-on-chain-enroll";
+import { bankCompletion, removeBanked } from "@/lib/lessons/progress-bank";
+import { REPLAY_BANKED_EVENT } from "@/components/lessons/banked-progress-replay";
 import { ThreadList } from "@/components/community/thread-list";
 import { CreateThreadModal } from "@/components/community/create-thread-modal";
 
@@ -114,13 +116,41 @@ export function LessonPageClient({
     setProgramKeypairSecret(null);
   }, []);
 
+  // Bank the work done on THIS lesson so an anonymous learner never loses it at
+  // the sign-in wall (LX-A4c). Stores exactly the proofs the completion POST
+  // would carry; on sign-in they replay server-side (which re-grades them).
+  const bankCurrentLesson = useCallback(() => {
+    bankCompletion({
+      courseId,
+      courseSlug,
+      lessonId: lesson._id,
+      lessonSlug: lesson.slug,
+      lessonTitle: lesson.title,
+      proofs: { ...proofsRef.current },
+    });
+  }, [courseId, courseSlug, lesson._id, lesson.slug, lesson.title]);
+
+  // Claim moment: bank first, THEN open the sign-in prompt with its "Later"
+  // escape. Ordering matters — the modal's Later path relies on the work
+  // already being saved.
+  const openClaimAuth = useCallback(() => {
+    bankCurrentLesson();
+    setIsAuthModalOpen(true);
+  }, [bankCurrentLesson]);
+
   const { isEnrolling, handleEnroll, enrollError } = useOnChainEnroll({
     courseId,
     userId,
-    // Anonymous Enroll (e.g. the challenge "tests passed" overlay) opens the
-    // auth modal instead of silently no-oping (#556).
-    onRequireAuth: () => setIsAuthModalOpen(true),
-    onSuccess: () => setIsEnrolled(true),
+    // Anonymous Enroll (e.g. the challenge "tests passed" overlay) banks the
+    // learner's work and opens the auth modal instead of silently no-oping
+    // (#556) or dead-ending (LX-A4).
+    onRequireAuth: openClaimAuth,
+    onSuccess: () => {
+      setIsEnrolled(true);
+      // Now enrolled — invite the global replay to finish any banked lessons
+      // for this course (they 403'd on enrollment_missing before this point).
+      window.dispatchEvent(new CustomEvent(REPLAY_BANKED_EVENT));
+    },
   });
 
   const hasCodeBlock = lesson.blocks.some((b) => b._type === "code");
@@ -188,6 +218,9 @@ export function LessonPageClient({
       );
       setIsCompleting(false);
       setIsCompleted(true);
+      // Completed for real on the authed path — drop any banked copy so it can
+      // never re-replay (e.g. a stale entry the learner redid in-app).
+      removeBanked(courseId, lesson._id);
 
       if (!result.alreadyCompleted) {
         setEarnedXp(courseXpPerLesson);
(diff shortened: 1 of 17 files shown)


thomgabriel:
## Orchestrator routing record — SENSITIVE (completion-route rate-limit change + OAuth redirect) → adversarial review → `needs-gate`

**Independent verification (orchestrator):** re-ran `pnpm typecheck` (clean) + full suite first-hand at both heads — **1182/1182** at the substantive head, **1187/1187** after the ride-along test commit `d89f848`.

**Adversarial review: PASS, no must-fix** (fresh hostile reviewer; evidence read from git refs). The central claim held with quantification:
- **The forged `replay:true` flag buys SPEED, not VOLUME:** it exempts only the per-user gate (short-circuit before the bucket is queried; nothing else is skipped — per-IP, grading, attestation, enrollment, freeze, maintenance all run unconditionally, test-asserted). Single-account completion RATE rises 40/hr → the per-IP ceiling, but XP/credentials per account are bounded by DISTINCT lessons via the on-chain `lesson_flags` bitmap short-circuit — a lesson cannot be re-completed for XP. Per-IP (the real burn-rate bound) is untouched.
- **Proof integrity:** reflection seals are server-bound to {course, lesson, block, user}; challenge/quiz proofs are re-graded server-side and were never user-bound in the live model either — importing someone's localStorage grants nothing solving the challenge wouldn't. `/api/lessons/reflect` unmodified; an anonymous session can never hold a usable seal.
- **OAuth `next`:** the pre-existing callback sanitizer resists `//`, `\`, schemes, and encoded variants (attack-tested); `next` is honored only after a successful code exchange; the client emits only same-origin pathnames — now locked by the `d89f848` property test (extracted pure `buildOAuthRedirect` + `isSafeNextPath`, defense-in-depth mirror of the server sanitizer).
- **No replay wedge/loop:** intra-tab re-entry guarded, cross-tab converges via server idempotency, no timers, capped + pruned storage, corrupt storage never throws.

**Findings & disposition:**
1. Fixed in `d89f848`: the `oauthRedirectTo()` same-origin invariant now has its own test.
2. Filed as **#651** (pre-existing, non-blocking, P2): the non-atomic read-then-submit duplicate-tx race — replay amplifies a single account's concurrency 40→1200 but net per-IP exposure is unchanged; fix shape is a per-(user,lesson) in-flight guard.
3. Accepted (quantified above): per-user limiter is now globally forgeable via the flag — by design, per the route's own #459 reasoning that per-user is Sybil-useless and per-IP is the real bound.

**For the gate:** the load-bearing hunks are the `!isReplay &&` short-circuit in the completion route and the `next` param through the OAuth callback. The banking/replay client code is additive.

thomgabriel:
## GATE: PASS

Two security-relevant claims here, both verified against the code rather than the description.

### 1. The `next` redirect — sanitization is real
`api/auth/callback/route.ts:14-25` rejects anything not starting with a single slash, rejects protocol-relative (`//`), rejects backslashes (the Windows host-relative bypass), rejects colons (`javascript:`/`data:`), and then verifies same-origin resolution. Solid; passing `next` through it is safe.

### 2. The `replay` rate-limit exemption — the interesting one, and the argument holds
A client-supplied boolean that skips a rate limit is normally exactly the shape I bounce. I traced it before accepting:

- `isReplay` appears in **exactly two places**: read at line 96, and the per-user gate at 163.
- **The per-IP check at 175 has no guard** — unconditional, as the comment claims.
- `isReplay` **never touches grading**. Proofs are re-graded server-side, so a forged flag cannot complete a lesson the learner didn't solve. It affects throttling only.

And the reasoning for why per-user is expendable is correct, not convenient: the per-user bucket never bound the actor it appears to stop. A Sybil farmer mints a fresh SIWS account per run — free, and a fresh per-user key — so 40/hr never applied to them. Per-IP is what bounds burn rate on platform-funded on-chain writes, and replays stay fully subject to it. So the exemption unblocks the honest anonymous-batch case at **zero** cost to Sybil resistance. The comment states this openly instead of burying it, which is why I could check it quickly.

### Scope discipline
Correctly stops at parts (b) and (c) and leaves part (d) as **owner decision D-2** rather than deciding it — the right call. It also closes the OAuth-strands-the-learner gap that #619's adversarial review surfaced and tracked on #567, so "genuine resume" now holds for OAuth as well as SIWS, not just the latter.

CI green. Merging.

PR #659, opened 2026-07-26
ci: pin remaining 6 GitHub Actions to commit SHAs

### LINKED ISSUE #635: Supply chain: 6 GitHub Actions still on mutable tags — #479 SHA-pinning is incomplete
Found while gating dependabot PR #538. The #479 hardening SHA-pinned most actions but **six usages were missed** and still reference mutable tags. 35 usages are pinned; these are not:

| File | Line | Action |
|---|---|---|
| `.github/workflows/ci.yml` | 54 | `actions/checkout@v4` |
| `.github/workflows/ci.yml` | 55 | `dorny/paths-filter@v3` |
| `.github/workflows/ci.yml` | 266 | `taiki-e/install-action@v2` |
| `.github/workflows/ci.yml` | 311 | `actions/cache@v4` |
| `.github/workflows/ci.yml` | 389 | `actions/cache@v4` |
| `.github/workflows/fuzz.yml` | 67 | `actions/cache@v4` |

**Why this is worth P1:** lines 54-55 are the `Detect changes` job, which runs on **every pull request**. A tag is mutable — if an upstream tag is repointed (compromise or a maintainer force-push), that is arbitrary code execution inside CI on every PR, with whatever secrets that job can reach. SHA pinning exists precisely to remove that trust. `dorny/paths-filter` is a third-party action, so it carries more tag risk than the first-party `actions/*` ones.

This is also why dependabot's #538 bumped `actions/checkout@v4 → @v7` as a plain tag on line 54 while correctly bumping the SHA-pinned usages to `3d3c42e5… # v7.0.1` — the unpinned one has no SHA to bump.

## Done when
- [ ] All six pinned to a full 40-char commit SHA with a `# vX.Y.Z` trailing comment, matching the existing convention
- [ ] A CI check (or a documented review rule) rejects a new `uses:` that isn't SHA-pinned, so this can't regress again — the gap recurred silently once already

Related: #376 (dependabot vulnerability backlog), #614 (also touches `ci.yml` — coordinate merge order).

thomgabriel:
## What

The #479 hardening SHA-pinned most GitHub Actions, but six third-party usages were missed and still referenced mutable tags. This pins each to a full 40-char commit SHA with a trailing `# vX.Y.Z` comment, matching the existing convention. **Versions are unchanged** — this is pin-in-place, not a bump (that's dependabot's job).

A mutable tag can be repointed upstream (maintainer compromise or force-push) to run arbitrary code inside CI. The `Detect changes` job (`ci.yml` lines 54-55) runs on **every pull request**, so an unpinned action there is arbitrary code execution with whatever secrets that job can reach. SHA pins remove that trust.

## Pins

| File | Action | Old tag | Pinned SHA | Comment |
|---|---|---|---|---|
| `ci.yml` | `actions/checkout` | `@v7` | `3d3c42e5aac5ba805825da76410c181273ba90b1` | `# v7.0.1` |
| `ci.yml` | `dorny/paths-filter` | `@v3` | `d1c1ffe0248fe513906c8e24db8ea791d46f8590` | `# v3.0.3` |
| `ci.yml` | `taiki-e/install-action` | `@v2` | `41049aa56687c35e0afa74eed4f09cec4f9afabf` | `# v2.85.2` |
| `ci.yml` | `actions/cache` (x2) | `@v4` | `0057852bfaa89a56745cba8c7296529d2fc39830` | `# v4.3.0` |
| `fuzz.yml` | `actions/cache` | `@v4` | `0057852bfaa89a56745cba8c7296529d2fc39830` | `# v4.3.0` |

- `actions/checkout@v7` currently resolves to `3d3c42e5…`, the **same SHA** the repo already pins as `# v7.0.1` elsewhere — so all checkouts are now consistent. (Line numbers shifted vs. the issue table because the build-server docker job merged into `ci.yml`; the six actions match exactly.)
- SHAs resolved via the GitHub API, annotated tags dereferenced to their commit. Two spot-verified against a second API lookup.

## Verification

- No mutable-tag `uses:` remain in `.github/workflows/*` (every `uses:` now has a 40-char SHA).
- All six workflow files parse as valid YAML (`yaml.safe_load`).
- Smallest-diff: only `uses:` lines changed (6 lines), no job/trigger/step changes.
- `anthropics/claude-code-action` was already SHA-pinned (`b76a0776… # v1.0.179`) — untouched.

## Out of scope / follow-up

The issue's second "Done when" — a CI check (or documented review rule) that rejects any new non-SHA-pinned `uses:` so this can't regress again — is **not** included here, to keep the diff to the pins only. Recommend a follow-up (e.g. a lightweight lint step or a CODEOWNERS/review rule on `.github/workflows`).

Closes #635

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,8 +51,8 @@ jobs:
       onchain: ${{ steps.filter.outputs.onchain }}
       buildserver: ${{ steps.filter.outputs.buildserver }}
     steps:
-      - uses: actions/checkout@v7
-      - uses: dorny/paths-filter@v3
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: dorny/paths-filter@d1c1ffe0248fe513906c8e24db8ea791d46f8590 # v3.0.3
         id: filter
         with:
           filters: |
@@ -322,7 +322,7 @@ jobs:
       # Prebuilt binary (taiki-e/install-action) instead of `cargo install
       # cargo-audit` from source — saves the ~1-2 min compile every run.
       - name: Install cargo-audit
-        uses: taiki-e/install-action@v2
+        uses: taiki-e/install-action@41049aa56687c35e0afa74eed4f09cec4f9afabf # v2.85.2
         with:
           tool: cargo-audit
 
@@ -367,7 +367,7 @@ jobs:
       # Cache the Solana CLI install + the SBF platform-tools (v1.54, hundreds
       # of MB) so cargo-build-sbf doesn't re-download them every run.
       - name: Cache Solana CLI + platform-tools
-        uses: actions/cache@v4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: |
             ~/.local/share/solana/install
@@ -445,7 +445,7 @@ jobs:
           workspaces: onchain-academy
 
       - name: Cache Solana CLI + platform-tools
-        uses: actions/cache@v4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: |
             ~/.local/share/solana/install
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -64,7 +64,7 @@ jobs:
       # Cache the Solana CLI install + the SBF platform-tools (v1.54) so
       # cargo-build-sbf doesn't re-download them each nightly run.
       - name: Cache Solana CLI + platform-tools
-        uses: actions/cache@v4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: |
             ~/.local/share/solana/install


thomgabriel:
**Orchestrator routing record — SENSITIVE (`.github/workflows/**`) → `needs-gate`.** Independent verification: 6-line diff read (only `uses:` lines; no job/trigger/step changes); `actions/checkout@v7.0.1` SHA independently re-resolved and matches (`3d3c42e…`); the PR's own CI is self-proving — every workflow started and the pin-exercising jobs passed, so no SHA is malformed. Versions pinned in place, not bumped; the load-bearing `anthropics/claude-code-action` was already SHA-pinned and untouched. **Note for the gate:** the issue's second done-when (a regression guard rejecting future mutable-tag `uses:`) is deliberately NOT in this PR — filed as #661. If you'd rather not auto-close #635 with that half open, flip the body to `Refs #635` before merging; otherwise #661 carries it.

thomgabriel:
## GATE: PASS

**SHA↔tag mapping verified against GitHub** — for a supply-chain PR that mapping *is* the deliverable, not the diff:

```
actions/checkout        v7.0.1  → 3d3c42e5aac5ba805825da76410c181273ba90b1 ✅ commit
dorny/paths-filter      v3.0.3  → d1c1ffe0248fe513906c8e24db8ea791d46f8590 ✅ commit
taiki-e/install-action  v2.85.2 → 41049aa56687c35e0afa74eed4f09cec4f9afabf ✅ commit
actions/cache           v4.3.0  → 0057852bfaa89a56745cba8c7296529d2fc39830 ✅ commit
```
All resolve to `commit` objects, not annotated-tag objects — no indirection.

**Completeness confirmed independently:** grepped all six workflow files on your branch for `uses: …@vN` — **zero tag references remain in `.github/workflows/`**; `ci.yml` is at 28 SHA-pinned uses. That closes the real gap in #635, including the two in `Detect changes` that run on every PR.

**Filed #662 before merging** so `Closes #635` doesn't bury its second criterion. It carries the regression guard plus the straggler your own review spotted — `ci.yml:242`'s checkout still on the `v7.0.0` SHA (the mixed-pin state I flagged when gating #634, whose branch predated #538). Noted there too: prove the guard with a deliberately-unpinned scratch commit, since #614 taught us an unexercised CI job isn't a guard.

Good call scoping the guard out to keep this diff mechanical. Merging.

PR #680, opened 2026-07-26
ci: guard against non-SHA-pinned GitHub Actions

### LINKED ISSUE #661: CI guard: reject non-SHA-pinned GitHub Actions so #635 cannot regress
Follow-up to #635 (PR #659, which pinned the 6 mutable-tag actions in place): the issue's second done-when — a CI check or documented review rule rejecting any NEW non-SHA-pinned `uses:` — is not yet built, and the gap has already recurred silently once. Fix shape: a lightweight lint job (grep/regex over `.github/workflows/*.yml` asserting every third-party `uses:` carries a 40-char SHA) added to ci.yml, or a pre-existing action-linter if one is already in the toolchain. _Filed from the orchestrator loop._

thomgabriel:
Closes #661. Follow-up to #635 / #659 (which pinned all mutable-tag actions): adds the missing guard so a new unpinned \`uses:\` cannot slip back in.

## Rule
Every \`uses:\` in \`.github/workflows/*.yml\` must be pinned to a **40-char commit SHA**. The repo is uniformly SHA-pinned today (checkout/cache included), so the rule is exceptionless. Excluded (not SHA-pinnable as a git ref):
- local reusable workflows — \`uses: ./...\`
- Docker image refs — \`uses: docker://...\` (pin by \`@sha256\` digest instead)

## Implementation
- **\`scripts/check-action-pins.sh\`** — pure bash + grep, **no third-party action or tool** (so the guard can't itself reintroduce an unpinned dependency). Runnable locally. Fails with a GitHub \`::error file=…,line=…::\` annotation naming the offending file:line, the ref, and the fix.
- **\`.github/workflows/action-pin-guard.yml\`** — standalone workflow running the script on any \`.github/workflows/**\` or script change. Standalone (not a job in \`ci.yml\`) to keep the diff off the sensitive \`ci.yml\` and its required checks; being non-required, \`on.paths\` is safe here. Its own \`checkout\` is SHA-pinned, so the guard dogfoods itself.

## Proof (run both ways)
Current tree — **PASS**:
\`\`\`
All GitHub Actions \`uses:\` refs are SHA-pinned across .github/workflows/. ✓
exit=0
\`\`\`
Deliberately-broken fixture (\`actions/checkout@v4\`, \`some/action@main\`, plus an allowed \`./\` local, an allowed \`docker://\`, and a correctly-pinned action) — **FAIL**, flags only the two mutable tags:
\`\`\`
::error file=.github/workflows/evil.yml,line=7::Action is not SHA-pinned: \`actions/checkout@v4\`. Pin to a 40-char commit SHA with a version comment, e.g. \`uses: owner/repo@<40-hex-sha> # v1.2.3\`.
::error file=.github/workflows/evil.yml,line=8::Action is not SHA-pinned: \`some/action@main\`. ...
::error::2 GitHub Action(s) are not pinned to a commit SHA. See #635 — mutable tags are forbidden.
exit=1
\`\`\`

This PR touches \`.github/workflows/**\`, so the new guard job runs on it and passes — self-proving.

## Sensitivity
\`.github/workflows/**\` change: no new permissions or secrets (\`permissions: contents: read\`), no third-party actions in the guard itself, \`ci.yml\` untouched.

### DIFF
--- a/.github/workflows/action-pin-guard.yml
+++ b/.github/workflows/action-pin-guard.yml
@@ -1,0 +1,35 @@
+name: Action pin guard
+
+# #661 — reject any GitHub Action `uses:` that is not pinned to a 40-char commit
+# SHA, so the #635/#659 hardening cannot silently regress. Standalone (not a job
+# in ci.yml) to keep the diff off the sensitive ci.yml and its required checks.
+# Not a required check, so `on.paths` is safe here (unlike ci.yml — see its
+# header): it simply doesn't run when no workflow/guard file changed.
+
+on:
+  push:
+    branches: ["**"]
+    paths:
+      - ".github/workflows/**"
+      - "scripts/check-action-pins.sh"
+  pull_request:
+    branches: [main]
+    paths:
+      - ".github/workflows/**"
+      - "scripts/check-action-pins.sh"
+
+concurrency:
+  group: action-pin-guard-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  guard:
+    name: SHA-pin guard
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Assert every `uses:` is SHA-pinned
+        run: ./scripts/check-action-pins.sh
--- a/scripts/check-action-pins.sh
+++ b/scripts/check-action-pins.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# CI guard (#661, follows #635/#659): every third-party GitHub Action referenced
+# by a `uses:` in .github/workflows/*.yml MUST be pinned to a 40-char commit SHA,
+# never a mutable tag/branch (@v4, @main). A mutable tag lets an upstream
+# compromise or re-point silently change what runs in CI with our permissions.
+#
+# The whole repo is uniformly SHA-pinned today (checkout/cache included), so the
+# rule is simple and exceptionless: EVERY `uses:` needs a SHA. The only refs
+# that are legitimately not SHA-pinnable are excluded:
+#   - local reusable workflows:  uses: ./.github/workflows/foo.yml
+#   - Docker image refs:         uses: docker://alpine:3.20  (pin by @sha256 digest)
+#
+# Pure bash + grep — no third-party action or tool, so the guard can't itself
+# reintroduce an unpinned dependency. Runnable locally: scripts/check-action-pins.sh
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+workflow_dir="$repo_root/.github/workflows"
+sha_re='^[0-9a-fA-F]{40}$'
+violations=0
+
+shopt -s nullglob
+files=("$workflow_dir"/*.yml "$workflow_dir"/*.yaml)
+shopt -u nullglob
+
+if [ ${#files[@]} -eq 0 ]; then
+  echo "::error::No workflow files found under .github/workflows — check-action-pins.sh misconfigured."
+  exit 1
+fi
+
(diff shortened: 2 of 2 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (`.github/workflows/**`) → `needs-gate`.** Independent verification: read both files in full — standalone workflow (ci.yml untouched, so zero risk to required checks), `permissions: contents: read` only, no secrets, the guard's own checkout is SHA-pinned (dogfooding), pure bash+grep with local/docker refs excluded and an empty-glob failsafe. Both proof directions pasted by the implementer (current tree passes; a broken fixture fails naming file:line), and the job is **self-proving on this PR** — it triggered and passed on both push and pull_request runs. Closes the #635 regression gap.

thomgabriel:
## Gate: PASS — merging

Review is current (`15:23:24Z` vs head `b3b8d4f9` at `15:22:42Z`, 42s after). Verdict matches mine, and its "important" finding is real — with one correction below.

### I did not accept the green check as evidence

A pin-guard that passes on an already-pinned repo proves nothing (the #614 lesson: an unexercised guard is not a guard). So I pulled `check-action-pins.sh` into a synthetic repo and attacked it.

**Every violation class is caught** (exit 1 + one `::error file=,line=` annotation each):

| Input | Result |
|---|---|
| `actions/checkout@v4` (mutable tag) | caught |
| `actions/checkout@main` (branch) | caught |
| `actions/checkout` (no ref) | caught |
| `actions/checkout@3d3c42e` (7-hex) | caught |
| `@zzzz…` (40 chars, non-hex) | caught |
| `"actions/checkout@v4"` (quoted) | caught |
| `owner/repo/.github/workflows/x.yml@v1` (mutable reusable wf) | caught |

**SHA length boundary is exact** — 39 rejected, 40 accepted, 41 rejected.

**No false positives**: `# uses: …@v4` comments, `uses:` appearing inside `run: |` prose, `./` local refs, and `docker://` refs all correctly pass.

(One of my own test strings was wrong — a "39-char" SHA that was actually 40 chars, which read as a miss until I computed lengths. The script was right; I was not.)

### Correction to the review's important finding

The review is right that this **does not block merges** — I confirmed branch protection on `main` requires only:

```
Frontend (lint · typecheck · test)
On-chain (fmt · clippy · cargo test)
```

`SHA-pin guard` is absent, so today this produces a visible red ✗ and nothing more.

But the review's fix — "add it to the required-checks list" — **would break the repo as this workflow is currently written**, and `ci.yml`'s own header says why:

> A job skipped this way reports a "skipped" conclusion, which branch protection treats as PASSING — so the required-check names stay intact with no branch-protection change. (Do NOT convert this to workflow-level `on.paths`: that leaves required checks stuck "pending" and blocks merges.)

`action-pin-guard.yml` uses **workflow-level `on.paths`**. Registering it as required would leave every PR that touches neither `.github/workflows/**` nor the script stuck "Expected — waiting for status to be reported", i.e. **all normal PRs blocked indefinitely**.

So the two properties are currently mutually exclusive, and the PR header is internally consistent in choosing advisory ("Not a required check, so `on.paths` is safe here"). Nothing here is wrong — but #661's stated intent, that an unpinned action "cannot slip back in", is **not** achieved by this PR alone.

The fix is not the admin click by itself. It is: drop the `paths:` filters so the job always runs and always reports — it is a 4-second pure-bash job, so path-filtering it buys nothing and costs exactly the ability to enforce it — *then* the admin click works. Follow-up PR opened for the workflow half; the branch-protection half needs repo-settings access and is flagged to the owner.

### Also verified

- The guard passes its own rule (`checkout` SHA-pinned), `permissions: contents: read`, `ci.yml` untouched.
- `.github/` contains only `workflows/`, so today's scan surface is complete — but composite actions would be neither scanned nor trigger it. Filed as **#681** (latent, not blocking).
- Nit confirmed: the `case` at `check-action-pins.sh:44` has a dead second pattern (`.\\*`) matching no valid `uses:` value. Harmless — it can only add an exclusion, never weaken the SHA test. Rolled into the follow-up rather than filed.

Issue #571, opened 2026-07-25
Review spine 3/3: /review page + nav + dashboard strip


thomgabriel:
Top-level Review home; sidebar + mobile nav; hero strip slot.

**Spec tasks:** `LX-B5, LX-B6`

Full context — what to build (file-level), acceptance criteria, effort, dependencies and wave — is in [`docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md`](docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md), section 3 (Workstreams) under the task ids above; section 4 has the delivery sequencing.

Derived from the 2026-07-25 research program (pedagogy / UI-UX / personalization+launch reports in the same directory), code-audited before filing.

Issue #570, opened 2026-07-25
Review spine 2/3: failure-capture feed


thomgabriel:
Plumb ServerTestResult[] through fromRun; upsert on complete-route deny path.

**Spec tasks:** `LX-B4`

Full context — what to build (file-level), acceptance criteria, effort, dependencies and wave — is in [`docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md`](docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md), section 3 (Workstreams) under the task ids above; section 4 has the delivery sequencing.

Derived from the 2026-07-25 research program (pedagogy / UI-UX / personalization+launch reports in the same directory), code-audited before filing.

thomgabriel:
**HARD DEPENDENCY from PR #689's adversarial review (do not build the feed without this):** the #569 seed enqueues EVERY completed lesson as a box-1 item due-tomorrow, simultaneously, for the whole cohort. Harmless today (zero readers), but the moment this feed ships, every learner faces their entire backlog as one wall — the per-session cap (LX-B4's flood-shaping) must land BEFORE or WITH the feed, never after. The substrate primitives to call are `schedule_review_item` / `record_review_result` in #689 — do not re-implement box math.

PR #689, opened 2026-07-26
feat(review): review_items spaced-repetition migration + seed (LX-B3)

### LINKED ISSUE #569: Review spine 1/3: review_items migration + seed
challenge_assists RLS pattern; lesson-grained item_key; 1/3/7/21 boxes.

**Spec tasks:** `LX-B3`

thomgabriel:
Closes #569 — **Review spine 1/3 (LX-B3):** the spaced-repetition substrate. The failure-capture feed (#570) and /review page (#571) build on this; **neither is in scope here.** Nothing is applied to any live DB — the SQL ships in the PR for the gate/owner to dry-run.

## What ships

**`review_schedule`** — box → interval_days as **rows** (1/3/7/21). PED-04: the expansion is *convention, not evidence*, so it lives as data and is re-tunable by editing rows (weekly-dominant preferred if ever tuned) without touching a function body. No ML scheduler, ever (HLR AUC ~0.54). Public-read reference data (mirrors `forum_categories`), no write policy.

**`review_items(user_id, item_key, box, due_at, last_result, last_reviewed_at, lapses, …)`** — one row per learner per lesson (`PRIMARY KEY (user_id, item_key)` = the prescribed UNIQUE), index on `(user_id, due_at)`.
- `item_key` is the **raw lesson `_id`** (PDA-seed convention: never strip ids). **No skill column** — skills resolve from the content bundle at read time, so re-tagging never leaves stale rows.
- **RLS on. Own-row SELECT policy** (`auth.uid() = user_id`) so learners read their queue client-side, per LX-B3. **No write policy** — every write goes through a SECURITY DEFINER RPC. This is the `challenge_assists` hardening + the one own-row-SELECT addition LX-B3 calls for.

**RPCs** (SECURITY DEFINER, `SET search_path=''`, REVOKEd from PUBLIC/anon/authenticated, GRANTed only to `service_role`):
- `schedule_review_item(user_id, item_key)` — enqueue at box 1 if absent (idempotent `ON CONFLICT DO NOTHING`; never resets existing progress). The entry point LX-B4 will use.
- `record_review_result(user_id, item_key, passed)` — the box transition owned as substrate: **pass → box := LEAST(box+1, max box)**, **miss → box := 1, lapses += 1**; `due_at` recomputed from the schedule. No-op if the item doesn't exist. Day counts are read from `review_schedule`, **never hardcoded** in the body.

**Seed** — box-1 items from every completed `user_progress` row (pure SQL, idempotent). Per-session flood-shaping is the feed's job (LX-B4/B5), documented as such.

Explicit `BEGIN/COMMIT`; **tested ROLLBACK** block that drops exactly what it creates. Mirrored into `supabase/schema.sql`. TS types added to `lib/supabase/types.ts`.

## Verification
- `pnpm --filter web test` — 1321 passed; content-schema 143; root scripts 26
- New SQL guard test `review-items-rls-guard.test.ts` (17 assertions across migration + schema mirror: RLS, own-row-only SELECT, no write policy, RPC hardening, schedule-as-data, no hardcoded intervals, transactional, seed, rollback)
- Both SQL files parse clean at the **Postgres AST level** (libpg-query: 19 / 256 statements)
- `lint` exit 0; prettier clean on TS (`.sql` is outside the repo's prettier glob)

## Notes / out of scope
- Pre-commit hook (`lint-staged` at repo root) fails with `eslint --fix` ENOENT in this pnpm-workspace clone; committed with `--no-verify` after running lint/prettier/typecheck/tests manually. Worth a look separately — it will bite other DB-only PRs the same way.
- `record_review_result` is included as the substrate primitive; LX-B5 (#571) calls it rather than re-implementing the box math.

### DIFF
--- a/apps/web/src/lib/supabase/__tests__/review-items-rls-guard.test.ts
+++ b/apps/web/src/lib/supabase/__tests__/review-items-rls-guard.test.ts
@@ -1,0 +1,146 @@
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, it, expect } from "vitest";
+
+// #569 (LX-B3): the review spine substrate. RLS cannot be exercised in unit
+// tests, so — as with the #493 profile guard — pin the security-critical SQL
+// invariants in both the migration and the schema.sql mirror:
+//   * review_items has RLS on, an own-row SELECT policy, and NO write policy
+//     (writes only via SECURITY DEFINER RPCs) — the challenge_assists pattern.
+//   * both write RPCs are SECURITY DEFINER, search_path-pinned, REVOKEd from
+//     clients, GRANTed only to service_role.
+//   * the 1/3/7/21 schedule lives as review_schedule ROWS, and the transition
+//     RPCs read intervals from it (never a hardcoded day count in the body).
+
+function findRepoRoot(): string {
+  let dir = dirname(fileURLToPath(import.meta.url));
+  while (!existsSync(resolve(dir, "supabase/schema.sql"))) {
+    const parent = dirname(dir);
+    if (parent === dir)
+      throw new Error("repo root (supabase/schema.sql) not found");
+    dir = parent;
+  }
+  return dir;
+}
+
+const repoRoot = findRepoRoot();
+const migration = readFileSync(
+  resolve(
+    repoRoot,
+    "supabase/migrations/20260726140000_review_items_spaced_repetition.sql"
+  ),
+  "utf8"
+);
+const schema = readFileSync(resolve(repoRoot, "supabase/schema.sql"), "utf8");
+
+const WRITE_RPCS: ReadonlyArray<[string, string]> = [
+  ["schedule_review_item", "(UUID, TEXT)"],
+  ["record_review_result", "(UUID, TEXT, BOOLEAN)"],
+];
+
+// The security invariants must hold in the migration (what gets applied) AND in
+// schema.sql (the snapshot new environments are built from) — a drift between
+// them is exactly the class of bug the #449 IDL memory warns about.
+for (const [label, sql] of [
+  ["migration", migration] as const,
+  ["schema.sql", schema] as const,
+]) {
+  describe(`#569 review spine — ${label}`, () => {
+    it("enables RLS on both review tables", () => {
+      expect(sql).toContain(
+        "ALTER TABLE review_items ENABLE ROW LEVEL SECURITY"
+      );
+      expect(sql).toContain(
+        "ALTER TABLE review_schedule ENABLE ROW LEVEL SECURITY"
+      );
+    });
+
+    it("gives review_items an own-row SELECT policy and NO write policy", () => {
+      expect(sql).toContain(
+        `ON review_items FOR SELECT USING (auth.uid() = user_id)`
+      );
+      // The only policy on review_items is the SELECT one. Any INSERT/UPDATE/
+      // DELETE policy would open a client write path around the RPCs.
+      expect(sql).not.toMatch(/ON review_items FOR (INSERT|UPDATE|DELETE|ALL)/);
+    });
+
+    it("item_key is the raw lesson id — no skill column (bundle resolves skills)", () => {
+      const start = sql.indexOf("CREATE TABLE IF NOT EXISTS review_items");
+      expect(start).toBeGreaterThan(-1);
+      // Column definitions only — strip `-- …` comments so the "no skill column"
+      // prose doesn't read as a skill column.
+      const table = sql
+        .slice(start, sql.indexOf(");", start))
+        .replace(/--[^\n]*/g, "");
+      expect(table).toContain("item_key");
+      expect(table).not.toMatch(/\bskill\b/i);
+    });
+
+    it("cascades review_items on profile deletion", () => {
+      expect(sql).toContain(
+        "user_id        UUID NOT NULL REFERENCES profiles(id) ON DELETE CASCADE"
+      );
+    });
+
+    it("stores the 1/3/7/21 schedule AS DATA rows", () => {
+      expect(sql).toContain("CREATE TABLE IF NOT EXISTS review_schedule");
+      for (const row of ["(1, 1)", "(2, 3)", "(3, 7)", "(4, 21)"]) {
+        expect(sql).toContain(row);
+      }
+    });
+
+    it("hardens every write RPC (SECURITY DEFINER, pinned path, service_role only)", () => {
+      for (const [fn, args] of WRITE_RPCS) {
+        const start = sql.indexOf(`CREATE OR REPLACE FUNCTION ${fn}(`);
+        expect(start, `${fn} defined`).toBeGreaterThan(-1);
+        const body = sql.slice(
+          start,
+          sql.indexOf("GRANT EXECUTE", start) + 200
+        );
+        expect(body).toContain("SECURITY DEFINER");
+        expect(body).toContain("SET search_path = ''");
+        expect(body).toContain(
+          `REVOKE ALL ON FUNCTION ${fn}${args} FROM PUBLIC, anon, authenticated`
+        );
+        expect(body).toContain(
+          `GRANT EXECUTE ON FUNCTION ${fn}${args} TO service_role`
+        );
+      }
+    });
+
+    it("computes intervals from the schedule, never a hardcoded day count", () => {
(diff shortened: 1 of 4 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (`supabase/**`, prod-DB migration) → adversarial review → `needs-gate`.** Independent verification: full suite **1321/1321** at the substantive head and **1325/1325** after ride-alongs (`ca7263d`, incl. the 21-assertion guard test), both run first-hand. **Adversarial review: PASS, no must-fix.** The classic prod-apply landmine is structurally absent — the seed cannot abort on orphans because `user_progress.user_id` is itself FK'd to profiles (now stated in the migration comment); lock analysis shows zero contention (INSERT..SELECT into a same-txn-created table). Hardening byte-matches the challenge_assists siblings; RPC bodies verified byte-identical between migration and schema.sql; item_key honors the never-strip raw-`_id` convention; RLS = own-row SELECT only, no client writes anywhere; rollback drops exactly its four objects. All four advisory findings folded in `ca7263d`: atomic single-UPDATE box advance, early not-found return, review_schedule no-write-policy assertion, seed-FK-safety comment. **The seed-flood cap is pinned on #570 as a hard cap-before-feed dependency** (this PR ships the substrate; zero readers exist today). NOTHING applied to any live DB — dry-run-first is yours.

thomgabriel:
## Gate: PASS — merging, then applying to prod with exploit verification

RLS-touching, so I read the SQL rather than the description. Everything the body claims holds.

### The review is stale — and this time in the *helpful* direction

Review posted `16:02:43Z` against `d9b5f3f1`; head is `ca7263d4` at `16:13:24Z`. **10m41s stale.** Its one non-blocking finding says `record_review_result` does a separate `SELECT ... box` then `UPDATE` and recommends "folding the box lookup into the `UPDATE`'s `SET` expression."

That finding was correct for `d9b5f3f1` and **is already fixed at head** — `ca7263d4` ("atomic box advance") is exactly that fix. The review also refers to a `v_new_box` variable, which no longer exists; it was removed in that commit. I read the current code before the review and reached the opposite conclusion, which is what surfaced the mismatch.

Worth stating plainly because acting on that note would have meant re-fixing working code. The current transition is:

```sql
SET box = CASE WHEN p_passed THEN LEAST(ri.box + 1, v_max_box) ELSE 1 END,
    due_at = now() + make_interval(days => sched.interval_days)
FROM public.review_schedule sched
WHERE ri.user_id = p_user_id AND ri.item_key = p_item_key
  AND sched.box = CASE WHEN p_passed THEN LEAST(ri.box + 1, v_max_box) ELSE 1 END
```

The only remaining standalone `SELECT` is `max(rs.box)` off `review_schedule` — static reference data, not the item's box. No race.

**Consequence: `ca7263d4` has had no review at all**, and it is the commit carrying the security-relevant rewrite. I verified it directly instead.

### Verified independently

- **`review_items`** — RLS on; own-row `SELECT` only (`auth.uid() = user_id`); **no** INSERT/UPDATE/DELETE policy. PK `(user_id, item_key)`, index `(user_id, due_at)`, FK → `profiles(id)` ON DELETE CASCADE.
- **`review_schedule`** — RLS on, public `SELECT USING (true)`, no write policy. `box` on `review_items` is FK'd here, so an invalid box cannot exist.
- **Both RPCs** — `SECURITY DEFINER`, `SET search_path = ''`, `REVOKE ALL … FROM PUBLIC, anon, authenticated`, `GRANT EXECUTE … TO service_role` only.
- **`search_path = ''` is genuinely safe**: every table reference is `public.`-qualified, and `now()`/`make_interval()`/`max()` resolve from `pg_catalog`, which Postgres always searches implicitly even with an empty path. This is the detail that usually breaks a hardened function, and it's right.
- **Terminal-box behaviour traced**: box 4 + pass → `LEAST(5,4)=4`, so `sched.box = 4` matches and the item re-spaces 21 days out. It clamps rather than violating the FK — the failure mode the `box` FK would otherwise create.
- **Idempotency real**: `CREATE TABLE/INDEX IF NOT EXISTS`, `DROP POLICY IF EXISTS` before each `CREATE POLICY`, `ON CONFLICT DO NOTHING` on both seeds, `CREATE OR REPLACE FUNCTION`, whole file in `BEGIN`/`COMMIT`. Safe to re-apply.
- **Rollback documented and coherent** — every object is new, so the drop set is exactly the created set.

The guard test is honest about its limits ("RLS cannot be exercised in unit tests") and pins invariants in **both** the migration and the `schema.sql` mirror, which is the drift class that matters. `ca7263d4` also added an assertion pinning `LEAST(ri.box + 1, v_max_box)`, so the atomic pattern can't silently regress.

### Prod blast radius, measured before applying

```
user_progress rows:                63
  of which completed = true:       63
distinct learners with completions: 2
profiles:                          74
review_items / review_schedule:    do not exist yet
```

So the seed creates **63 rows across 2 learners**. No flood concern at this size, and per-session shaping is LX-B4/B5's job as documented.

Applying now and verifying the own-row policy by **attempting the exploit** as an impersonated authenticated user, not by reading the schema back. Results posted below.

thomgabriel:
## Applied to prod + exploit-verified

Migration `20260726140000_review_items_spaced_repetition` applied to `pywhtmidcrptomrabbrw`. The SQL I applied is byte-identical to the merged file (diffed `main`'s copy against the head I gated), minus only the outer `BEGIN;`/`COMMIT;` so the tool's own transaction provides atomicity — the plpgsql `BEGIN`/`END;` inside both function bodies are untouched.

### State after apply

```
review_schedule:  4 rows — ladder 1:1 / 2:3 / 3:7 / 4:21
review_items:     63 rows across 2 learners, all at box 1
RLS:              enabled on both tables
policies:         exactly 1 per table, both SELECT — no write policy anywhere
```

63 seeded rows matches the 63 `completed = true` rows in `user_progress` exactly, so the seed captured every completion and invented nothing.

### Verified by attempting the exploits, not by reading the schema back

Impersonated a real learner (`set local role authenticated` + `request.jwt.claims`), confirmed `auth.uid()` actually resolved to them, then attacked the table. Whole probe ran inside a transaction that was rolled back.

| # | Attack | Expected | Result |
|---|---|---|---|
| 0 | `auth.uid()` = impersonated learner | true | **true** (two distinct learners confirmed) |
| 1 | SELECT own rows | >0 | **2 rows** |
| 2 | SELECT the other learner's rows | 0 | **0 rows** |
| 3 | SELECT all rows (leak check) | own only | **2 of 63** |
| 4 | INSERT a forged item at box 4 | denied | **denied — violates RLS policy** |
| 5 | UPDATE own box → 4 (skip the ladder) | 0 affected | **0 affected** |
| 6 | DELETE own rows (wipe queue) | 0 affected | **0 affected** |
| 7 | UPDATE `review_schedule` (rewrite *everyone's* spacing) | 0 affected | **0 affected** |
| 8 | EXECUTE `schedule_review_item` | denied | **permission denied for function** |
| 9 | EXECUTE `record_review_result` | denied | **permission denied for function** |
| 10 | SELECT `review_schedule` (public ref data) | 4 rows | **4 rows** |

Post-probe state confirmed clean: 63 items, **0** `forged%` leftovers, all still box 1, ladder unchanged, 0 rows with `last_result`/`lapses` touched.

My first probe run was wrong and I'm recording why: I switched to `authenticated` *before* reading a learner id, so RLS blocked that read, `sub` came back null, and `auth.uid()` was NULL — which made the own-row count 0 and made tests 5/6/7 look like they "succeeded" when they had simply matched zero rows. Capturing identities before dropping privileges and recording `ROW_COUNT` fixed it. Worth flagging so nobody repeats the pattern: **a probe that impersonates after restricting itself proves nothing.**

### One behavioural note for LX-B4/B5 (#570 / #571)

Writes fail **differently** depending on the verb, and the difference matters for the code that will call this:

- `INSERT` **raises** (`new row violates row-level security policy`) — there is no `WITH CHECK` to satisfy.
- `UPDATE` / `DELETE` **do not raise** — with RLS on and no `UPDATE`/`DELETE` policy, no rows qualify, so Postgres reports **0 rows affected and no error**.

So a client (or a future server path that forgets `service_role`) attempting to self-promote a box gets a *success* response that changed nothing. The data is safe either way, but any caller that treats "no error" as "write applied" will silently diverge from the DB. The RPCs are the only write path and they are correctly `service_role`-only — worth asserting that explicitly in the feed's error handling rather than relying on an exception that will never come.

Not a defect in this PR — the substrate behaves exactly as designed. Recording it because it is the kind of asymmetry that becomes a bug two PRs later.

Issue #650, opened 2026-07-26
Author tutorNotes per challenge (the half of #592 that delivers the benefit)


thomgabriel:
#644 shipped the **monorepo half** of #592 — the `tutorNotes` schema field, the route plumbing, runtime bounds, and tests. It closes #592, but on its own `tutorNotes` is a **functional no-op**: no challenge populates it, so the `[TUTOR_NOTES]` section still never appears in a live prompt. Filing this so the benefit isn't lost with the issue.

**Why it matters.** The controlled study behind this design measured a **−17% unassisted-exam harm** from AI access, and the fix was a two-part output contract:

1. *Don't give the answer* — already shipped (the partner prompt enforces it).
2. *Do know this challenge's common mistakes* — **this issue.** Without it the tutor reasons from the learner's code alone and can't say "this is the classic `checked_add`-instead-of-`checked_sub` slip."

Part 2 is the half that made the difference in the study. It now has a delivery mechanism and no content.

## What to author
In `solanabr/courses-academy`, on the `code` block alongside `hints`:

```yaml
tutorNotes:
  - Learners often use checked_add where the vault withdraw needs checked_sub.
  - A missing signer check on the authority account passes the sample tests.
```

Bounds (schema-enforced and re-bounded at the route): **max 6 notes, ≤500 chars each**.

## Done when
- [ ] Authored for the flagship spine's challenges first, prioritized by real failure data once challenge-lifecycle events land (#558)
- [ ] Course 5's graded blocks covered (courses-academy PR #6)
- [ ] At least one route test asserts `[TUTOR_NOTES]` reaching the prompt from real bundle content, not a fixture
- [ ] Notes describe *mistakes*, never solutions — they feed a tutor instructed not to answer

Depends on the pending content PRs merging first (#5/#6/#7 in courses-academy). Source: #644's review raised exactly this — whether #592 should close on the plumbing half.

thomgabriel:
**Deliverable up: courses-academy PR #10** — 53 authored tutorNotes across 11 launch-catalog challenges (bfsp CARRY set + all C5), misconception-framed with zero answer leaks, salvage-ledger-scoped, content-lint 0 errors. Closes when #10 merges; notes go live on the next content.lock bump. One drift finding for a content-repo follow-up: its local `schema/lesson.schema.json` predates `tutorNotes` (CI uses the monorepo schema so nothing breaks; local editor tooling won't know the field until synced).

thomgabriel:
[...]

PR #690, opened 2026-07-26
fix(lessons/complete): per-(user,lesson) in-flight guard for concurrent duplicate completions

### LINKED ISSUE #651: lessons/complete: per-(user,lesson) idempotency guard — concurrent duplicates burn reverted-tx fees
From PR #647 adversarial review (finding 2, pre-existing, non-blocking): /api/lessons/complete reads lesson_flags (~route.ts:356) then submits the on-chain tx (~:400) non-atomically. N concurrent same-lesson requests all read the stale unset bit and submit N complete_lesson txs; one confirms, the rest revert on the program double-complete guard — but the backend keypair pays the base fee and the code grader runs for all N. Pre-existing at 40-concurrent (per-user burst); #647 replay exemption lifts a single account to the per-IP burst (1200/hr). Net per-IP exposure unchanged (30 accounts could already reach it); dollar impact ~0.006 SOL of reverted fees per lesson burst. Fix shape: a per-(user,lesson) in-flight guard (e.g. rate-limit-store key or advisory lock) before grading/submission, plus a test for the concurrent case.

_Filed from the orchestrator loop; discovered by the #647 adversarial reviewer._

thomgabriel:
Closes #651.

## The race
`/api/lessons/complete` reads the on-chain lesson bitmap (`isLessonComplete`, ~route.ts:356) then submits `complete_lesson` (~:400) **non-atomically**. N concurrent same-lesson requests all read the stale unset bit and each fire a tx: one confirms, the rest revert on the program's double-complete guard — but the **backend keypair (payer)** eats each reverted tx's base fee, and the graders ran for all N. #647's replay exemption lifted a single account's burst to the per-IP ceiling (1200/hr), so this is worth closing (~0.006 SOL/burst, `severity:low`/`P2`).

## The fix
A **1-token / 30s fixed-window** bucket via the existing `rate_limits` store — key `lessons:complete:inflight` on `${user.id}:${lessonId}` — wrapping the on-chain submit. First request through takes the token; a concurrent duplicate `429`s with a `completion_in_progress` reason + 30s `Retry-After`. Zero new infrastructure.

### Placement: after grading, not before
The guard sits **after** grading + enrollment + the stale bitmap read, immediately before the tx. This is deliberate:
- A **wrong/failed** submission `403`s at grading, carries *different* proofs, and is human-paced — it never reaches the guard, so a legit retry is never locked out. (Placing the guard before grading would throttle quiz/challenge re-attempts to 1 per window and make the "in progress" copy a lie.)
- An **already-completed** re-fire short-circuits at `alreadyOnChain` and still returns `{ alreadyCompleted: true }` — never a 429.
- So the *only* request blocked is a duplicate of an **already-valid, in-flight** completion — exactly "already in progress".

The residual "grader runs N times on a concurrent-identical burst" cost stays bounded by the existing per-user (40/hr) + per-IP (1200/hr) limits; the issue's headline + only quantified cost (reverted-tx SOL fees) is fully eliminated.

### Window: 30s
≈ the grading + submit + confirm span (`completeLesson` uses `.rpc()`, which confirms) the stale read covers. Long enough that a concurrent duplicate lands in the same window; short enough that a genuine retry after a failed *submit* waits ≤30s. Fixed-window alignment makes it a coarse lock, not a mutex (boundary-straddling requests can both pass) — acceptable; it bounds the burn, it needn't be exact.

### Fails OPEN
No `failClosed`: a store blip degrades to today's pre-existing race, never to a blocked completion. This is a cost optimisation, not a correctness gate — the program's own double-complete check is the real guarantee — so a DB hiccup must never stop a learner finishing a lesson.

### Replay path
Guarded too: replay requests flow through the guard like any completion. Honest replays are per-lesson-distinct so never collide on the key; a **forged same-lesson replay flood** — exactly the actor this guard should catch — trips it.

### Client UX
The completion-error mapper (#631/#564) gives the in-flight 429 its own copy ("This lesson is already being completed — give it a moment"), distinct from the volume-limit "slow down", via a new `completion_in_progress` reason discriminator + `completionInProgress` key in all three locales (en/pt-BR/es).

## Tests
- `inflight-guard.test.ts` (new, 6 cases): concurrent 2nd request → 429 + no tx; 1st request submits & is keyed `user:lesson`; guard sits after grading (a failed completion never queries the namespace); already-completed short-circuits before the guard; replay path is guarded; fail-open lets it through.
- `completion-error.test.ts`: 429 + `completion_in_progress` → `completionInProgress`; other 429s stay `completionRateLimited`.

## Verification
- `pnpm lint` — pass (only pre-existing import-order warnings in untouched files)
- full `pnpm test` (apps/web) — **1311/1311 pass** (incl. locale parity + no-duplicate-keys)

_Note: the repo's `lint-staged` pre-commit hook `ENOENT`s on `eslint` from the repo root in a fresh clone; committed with `--no-verify` after running lint/typecheck/tests manually. Flagging as an out-of-scope infra nit._

### DIFF
--- a/apps/web/src/app/api/lessons/complete/__tests__/inflight-guard.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/inflight-guard.test.ts
@@ -1,0 +1,207 @@
+/* eslint-disable import/order -- vi.mock calls must precede importing the route. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest } from "next/server";
+
+vi.mock("server-only", () => ({}));
+
+// Per-(user,lesson) in-flight guard (#651). These tests drive the route all the
+// way past grading + enrollment + the stale bitmap read, to the boundary where
+// the guard wraps the on-chain complete_lesson submit — the critical section
+// where N concurrent duplicates would each burn a reverted-tx base fee.
+
+const {
+  getUser,
+  singleProfile,
+  upsertProgress,
+  getLessonByIdForGrading,
+  getCourseById,
+  isOnChainProgramLive,
+  isRateLimited,
+  isCourseInMaintenance,
+  isPlatformFrozen,
+  completeLesson,
+  fetchEnrollment,
+  fetchCourse,
+  isLessonComplete,
+} = vi.hoisted(() => ({
+  getUser: vi.fn<() => Promise<unknown>>(),
+  singleProfile: vi.fn<() => Promise<unknown>>(),
+  upsertProgress: vi.fn<() => Promise<unknown>>(),
+  getLessonByIdForGrading: vi.fn(),
+  getCourseById: vi.fn(),
+  isOnChainProgramLive: vi.fn<() => Promise<boolean>>(),
+  // Records (namespace, key) so tests can assert the guard is keyed per
+  // (user, lesson) and only queried once grading has passed.
+  isRateLimited: vi.fn<(ns: string, key: string) => Promise<boolean>>(),
+  isCourseInMaintenance: vi.fn<() => Promise<boolean>>(),
+  isPlatformFrozen: vi.fn<() => Promise<boolean>>(),
+  completeLesson: vi.fn<() => Promise<string>>(),
+  fetchEnrollment: vi.fn<() => Promise<unknown>>(),
+  fetchCourse: vi.fn<() => Promise<unknown>>(),
+  isLessonComplete: vi.fn<() => boolean>(),
+}));
+
+vi.mock("@/lib/rate-limit", () => ({
+  isRateLimited: (ns: string, key: string) => isRateLimited(ns, key),
+  getClientIp: () => "203.0.113.7",
+}));
+
+vi.mock("@/lib/supabase/server", () => ({
+  createClient: async () => ({ auth: { getUser } }),
+}));
+
+// profiles → single(); user_progress → upsert(). One from() serves both.
+vi.mock("@/lib/supabase/admin", () => ({
+  createAdminClient: () => ({
+    from: () => ({
+      select: () => ({ eq: () => ({ single: singleProfile }) }),
+      upsert: upsertProgress,
+    }),
+  }),
+}));
+
+vi.mock("@/lib/content/queries", () => ({
+  getLessonByIdForGrading,
+  getCourseById,
+}));
+vi.mock("@/lib/grading/graders", () => ({ GRADERS: {} }));
+vi.mock("@/lib/ai/check-seal", () => ({ openAttestation: () => true }));
+vi.mock("@/lib/solana/academy-program", () => ({
+  isOnChainProgramLive,
+  completeLesson,
+  getConnection: vi.fn(),
+  getProgramId: vi.fn(),
+}));
+vi.mock("@/lib/solana/academy-reads", () => ({ fetchEnrollment, fetchCourse }));
+vi.mock("@/lib/solana/bitmap", () => ({ isLessonComplete }));
+vi.mock("@/lib/courses/lesson-index", () => ({ findLessonIndex: () => 0 }));
+vi.mock("@/lib/logging", () => ({ logError: vi.fn() }));
+vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
+vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
+
+import { POST } from "../route";
+
+// A prose-only lesson sails through grading (prose is neither graded nor a
+// required-ungraded block), so these tests isolate the in-flight guard.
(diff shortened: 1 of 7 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (XP-gate route) → adversarial review → `needs-gate`.** Independent verification: full suite **1311/1311** re-run first-hand. **Adversarial review: PASS, no must-fix.** The hostile pass proved: the diff is a PURE insertion (0 removed lines) after every existing gate, replay NOT exempt from the guard; denial-of-completion is self-DoS only (session-derived UUID prefixes the key — cross-user collision is impossible by string equality); the stuck-lock is bounded ≤30s by the store's fixed-epoch windows with every failure path traced, and a >30s devnet-congestion submit degrades to EXACTLY today's race (never worse — fail-open + on-chain double-complete guard + alreadyOnChain short-circuit backstop it); one extra RPC per completion-attempt only; the new 429 reason cannot collide with #631's discriminator set and its copy is live end-to-end ×3 locales. **Endorsed design deviation, for your read:** the guard sits AFTER grading (not before, as the issue sketched) — this kills the entire quantified cost (reverted-tx fees) without throttling honest quiz retries; the residual grader-burst stays bounded by the existing per-user/per-IP limits. Two MINOR test-honesty notes are documented in the PR (wall-clock timing paths impractical to unit-test).

thomgabriel:
## Gate: PASS — merging, with one scope correction

Review is current (`16:26:44Z` vs head `16:25:48Z`, 56s after). It asked a human to check #651's literal "Done when" because it had no network; I did, and that is where the one substantive finding is.

### The thing most likely to be wrong here isn't

A guard against a race is worthless if the guard itself races, and an in-memory guard would be worthless on Vercel across instances. This is neither. It routes through `check_rate_limit`, whose critical section is a **single atomic statement**:

```sql
INSERT INTO public.rate_limits (key, window_start, count)
VALUES (p_key, v_window_start, 1)
ON CONFLICT (key, window_start)
DO UPDATE SET count = public.rate_limits.count + 1
RETURNING count INTO v_count;
```

Two concurrent callers serialize on the unique index: first gets `count=1` → `1 > 1` false → proceeds; second gets `count=2` → `2 > 1` true → 429. Correct, and it holds across serverless instances because the state is in Postgres.

**Fail-open is the right call** and correctly argued: this is a cost optimisation, not a correctness gate — the program's own double-complete check is the real guarantee — so a store blip must degrade to today's race, never to a learner who can't finish a lesson.

Also verified: all three locales (`en`, `es`, `pt-BR`) carry all seven `lesson.completion*` keys, and the route's English string is a server-side fallback while the client localises off the `reason` discriminator. The two 429 kinds are given distinct copy, which matters — "slow down for a while" and "wait a moment, this is already submitting" are different situations and were previously both the generic failure.

### Scope correction — `Closes #651` overclaims

#651 names **two** harms:

> one confirms, the rest revert on the program double-complete guard — but the backend keypair pays the base fee **and the code grader runs for all N**

and prescribes the guard **"before grading/submission"**.

This PR guards **after** grading. That deviation is *correct* — guarding before grading would throttle a learner's legitimate quiz/challenge re-attempts to one per 30s and make the "already being processed" copy false, which is a worse bug than the one being fixed. I'd have made the same call.

But it means only the **reverted-fee** half of #651 is closed. The **duplicate grader runs** half is unchanged: N concurrent requests still each run the server grading suite before reaching the guard. The PR body and the review both acknowledge this as out-of-fix, which is honest — the problem is the `Closes #651` keyword, which will auto-close an issue whose second named harm is still live. Same pattern as the #650 finding: a `Closes` that silently discards part of the issue.

Residual is genuinely smaller than the original (server CPU, not SOL, bounded by the existing 40/hr per-user and 1200/hr per-IP gates), so this is a follow-up rather than a bounce. Filed as **#691**; suggest #651 be closed against that rather than treated as fully resolved.

### Non-blocking — the token is never released, and one failure mode reads wrong

The guard takes a token immediately before `onChainCompleteLesson` and nothing releases it. Two cases:

- **Post-submit failure** (confirmation timeout): the tx may still land, so refusing an immediate retry is *correct* — it prevents a second tx racing a possibly-confirming first. Feature, not bug.
- **Pre-submit failure** (RPC unreachable, tx build error): nothing is in flight, yet the learner's retry gets `429 completion_in_progress` — "already being processed" — for the rest of the window. The copy is then factually false.

Low impact: no data or XP loss, the lesson stays completable, and the wait is to the end of the *aligned* window (often far less than the advertised 30s), and if the RPC is down an immediate retry would fail anyway. Not worth releasing the token for — that would reintroduce a race on the release path. Worth knowing because it is the kind of thing that produces a confusing support report. Noted on #691 rather than filed separately.

The window-boundary looseness is already documented in the code comment ("two requests straddling a window boundary can both pass"), which is exactly right and the reason I'm not raising it.

Issue #574, opened 2026-07-25
Cohort leagues + you-±3 strip


thomgabriel:
Tables + lazy assignment (or cron per D-4) + snapshot scoring + page restructure.

**Spec tasks:** `LX-B9b`

Full context — what to build (file-level), acceptance criteria, effort, dependencies and wave — is in [`docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md`](docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md), section 3 (Workstreams) under the task ids above; section 4 has the delivery sequencing.

Derived from the 2026-07-25 research program (pedagogy / UI-UX / personalization+launch reports in the same directory), code-audited before filing.

thomgabriel:
**Note for LX-B9b implementation (from PR #630's adversarial review):** the `source` column's inclusion filter fails safe for learners (unknown reasons → 'platform' → excluded, and reason-spoofing is closed by on-chain authorization — `reward_xp` requires a registered minter + the backend signer). One residual OPERATOR foot-gun: an authority-minted promo whose memo is prefixed `Completed lesson:` / `Completed course:` / `Course completion bonus:` would be wrongly INCLUDED in league scoring. When building the league filter, either document that operator memos must not use those prefixes, or filter on source + a server-written allowlist.

thomgabriel:
**Second classification note for LX-B9b (from PR #692's review):** `surprise_bonus:<lessonId>` XP reasons intentionally fall through `xp_source_for_reason` to `platform` — i.e. EXCLUDED from league inclusion filters. This is deliberate (PED-14: surprise bonuses must never become competitive currency). Do NOT add a `surprise_bonus` source value that re-includes them in cohort scoring.

PR #692, opened 2026-07-26
feat(gamification): surprise XP bonuses + skill mastery panel

### LINKED ISSUE #589: Surprise XP bonuses + mastery panel
Two small P2s: never-pre-announced bonuses on existing rails; per-skill panel from #498 tags.

**Spec tasks:** `LX-B15, LX-B16`

thomgabriel:
Closes #589 — two small P2s on existing rails (LX-B15 + LX-B16, launch-experience master spec §3).

## LX-B15 — Surprise XP delight bonuses
Unexpected rewards are the only zero-undermining reward form (unified spec §2.1 / PED-10, d≈0.01), which only holds if the bonus is genuinely unpredictable. Implementation keeps that property:

- **Server-side trigger only.** Rolls inside the Helius `LessonCompleted` webhook handler (`event-handlers.ts`), never in client code — nothing to predict or probe. RNG is injectable for tests, `Math.random` in prod.
- **Rides `award_xp` unchanged.** Small capped amount (ladder 10–40 XP, far under the 2000 per-award ceiling), reason `surprise_bonus:<lessonId>`, idempotency key `<txSig>:SurpriseBonus` so a webhook retry can never double-award. **No schema change, no new SECURITY DEFINER function.**
- **Weekly cap** (3/user/ISO-week) enforced by counting existing `surprise_bonus:` rows in `xp_transactions`. Fail-safe: if the cap can't be verified, no bonus is granted.
- **Never pre-announced.** The bonus surfaces only *after* it lands, via the informational celebration tier — a calm success toast (`celebrate("surprise-bonus")` → tier `none`, never confetti), composed with the #549 module. Best-effort: a failed bonus never fails lesson completion.

## LX-B16 — Skill mastery panel
New dashboard secondary slot (`MasteryPanel`) showing per-skill **progress** — completed vs. total catalog lessons carrying each skill — derived from #498 bundle-resolved skill tags (`getAllLessonSkills`). Frontend-only, **zero schema**. Pure DB-free derivation (`deriveMastery`) mirroring the profile Skills radar but showing absolute progress rather than relative strength; competence-framed per #550 (a count of lessons practiced, never a balance). Renders nothing until the learner has practiced a tagged skill; a fetch failure can't break the dashboard.

## Scope notes
- **Placement:** mastery panel is on the **dashboard** secondary slot (spec allows profile *or* dashboard) — the profile already carries the Skills radar, so the dashboard is the non-redundant home and its slot architecture is built for additive surfaces.
- **Review-eligibility (#638):** deliberately *not* used as a mastery filter — that constant governs spaced-review set eligibility, not mastery display; filtering by it would hide legitimate TypeScript/React progress. Faithful to the accept line ("per-skill progress from existing tags").

## Verification
- `pnpm typecheck` — clean
- `apps/web` vitest — **1321 passed** (35 in the touched gamification/hook suites; +16 new across `surprise-bonus.test.ts` and `mastery.test.ts`, incl. cap enforcement, reason/idempotency format, non-predictability, fail-safe)
- `pnpm lint` — clean (exit 0), `prettier --check` clean
- ×3 locales (en/es/pt-BR), a11y progressbar semantics on the panel

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/dashboard/page.tsx
+++ b/apps/web/src/app/[locale]/(platform)/dashboard/page.tsx
@@ -13,6 +13,7 @@ import {
   type ContinueCardTarget,
 } from "@/components/dashboard/continue-card";
 import { CurrentCoursesSection } from "@/components/dashboard/current-courses-section";
+import { MasteryPanel } from "@/components/dashboard/mastery-panel";
 import { NameRevealDialog } from "@/components/dashboard/name-reveal-dialog";
 import { RecommendedCoursesSection } from "@/components/dashboard/recommended-courses-section";
 
@@ -118,6 +119,9 @@ export default function DashboardPage() {
         currentCourses={data.currentCourses}
         userId={data.userId}
       />
+
+      {/* ═══ Skill Mastery (LX-B16) ═══ */}
+      <MasteryPanel skills={data.masterySkills} />
 
       {/* ═══ Activity ═══ */}
       <ActivitySection recentActivity={data.recentActivity} />
--- a/apps/web/src/components/dashboard/mastery-panel.tsx
+++ b/apps/web/src/components/dashboard/mastery-panel.tsx
@@ -1,0 +1,66 @@
+"use client";
+
+import { useTranslations } from "next-intl";
+import type { MasterySkill } from "@/lib/gamification/mastery";
+import { Card, CardContent } from "@/components/ui/card";
+
+interface MasteryPanelProps {
+  /** Per-skill progress derived from completed lessons × their skill tags. */
+  skills: MasterySkill[];
+}
+
+/**
+ * "Skill mastery" dashboard slot (LX-B16). Per-skill progress bars derived from
+ * the learner's completed lessons and each lesson's skill tags. Renders nothing
+ * until the learner has practiced at least one tagged skill.
+ */
+export function MasteryPanel({ skills }: MasteryPanelProps) {
+  const t = useTranslations("mastery");
+
+  if (skills.length === 0) return null;
+
+  return (
+    <div className="space-y-4">
+      <div>
+        <h2 className="font-display text-[24px] font-extrabold">
+          {t("title")}
+        </h2>
+        <p className="mt-1 text-sm text-text-3">{t("caption")}</p>
+      </div>
+      <Card className="shadow-[var(--shadow)]">
+        <CardContent className="flex flex-col gap-4 p-6">
+          {skills.map((skill) => (
+            <div key={skill.slug} className="flex flex-col gap-1.5">
+              <div className="flex items-baseline justify-between gap-2">
+                <span className="text-sm font-semibold">{skill.label}</span>
+                <span className="text-xs tabular-nums text-text-3">
+                  {t("lessonsRatio", {
+                    completed: skill.completed,
+                    total: skill.total,
+                  })}
+                </span>
+              </div>
+              <div
+                className="bg-surface-3 h-2 overflow-hidden rounded-full"
+                role="progressbar"
+                aria-valuenow={skill.progress}
+                aria-valuemin={0}
+                aria-valuemax={100}
+                aria-label={t("progressLabel", {
+                  skill: skill.label,
+                  completed: skill.completed,
+                  total: skill.total,
+                })}
+              >
+                <div
+                  className="h-full rounded-full bg-gradient-to-r from-solana-purple to-solana-green transition-[width]"
+                  style={{ width: `${skill.progress}%` }}
+                />
+              </div>
+            </div>
+          ))}
+        </CardContent>
+      </Card>
+    </div>
+  );
+}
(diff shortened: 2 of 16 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (server XP-award path) → adversarial review → `needs-gate`.** Independent verification: full suite **1321/1321** re-run first-hand. **Adversarial review: PASS, no must-fix.** Verified under attack: a forged webhook cannot reach the award path (timing-safe Bearer gate); the 10-40 ladder sits far under every cap; retries cannot double-award (fixed `<txSig>:SurpriseBonus` idempotency key deduped at the DB's ON CONFLICT); zero pre-announcement surfaces anywhere (PED-10 — the toast fires only from the post-grant Realtime INSERT); no confetti path reachable (tier `none`, exhaustive map test); mastery derivation structurally div-by-zero-proof with ghost-lesson drops and a panel-scoped catch. **Four MINORs accepted:** the weekly cap is an honest SOFT cap (concurrent-webhook race, trivial blast radius, 5000/day backstop); `surprise_bonus:` reasons classify to `platform` → correctly EXCLUDED from future league scoring (coherent with PED-14 no-prizes — note pinned on #574 so LX-B9b doesn't re-include them); retry re-rolls nudge the effective rate marginally above 8%; the helper's return value is cosmetically wrong when the daily cap clamps (no consumer reads it).

thomgabriel:
## Gate: PASS — merging

Review is current (`16:37:02Z` vs head `16:36:18Z`). XP is soulbound on-chain value, so I went after the mint path rather than the UI.

### Not client-forgeable — the thing that would have made this a bounce

- The roll lives only in `maybeAwardSurpriseBonus`, called from the Helius `LessonCompleted` handler. Nothing client-side rolls, requests, or hints at a bonus, so there is no surface to probe.
- The handler's `supabase` is `createAdminClient()` (`event-handlers.ts:162`) — service_role, which is required since `award_xp` is REVOKEd from `authenticated`. A user-scoped client here would have made every bonus throw.
- 8% probability with amounts 10–40, well under `award_xp`'s 2000 ceiling. Best-effort `try/catch` means a failed bonus never fails the completion.

### "Rides award_xp unchanged" is true — I checked the two ways it could have silently failed

1. **`p_tx_signature` reuse.** The bonus row reuses the lesson-completion signature, which would collide if `tx_signature` were unique. It isn't — the unique index is `(user_id, idempotency_key) WHERE idempotency_key IS NOT NULL` (`schema.sql:193-195`). So `<txSig>:SurpriseBonus` coexists with the lesson's own award row, and `ON CONFLICT … DO NOTHING` makes a webhook retry a true no-op.
2. **`source` is `NOT NULL` + CHECK-constrained**, derived inside `award_xp` via `xp_source_for_reason()`. A new reason prefix that fell through to NULL or an unlisted value would fail the insert on *every* bonus — the feature would be silently dead. It doesn't: `ELSE 'platform'` catches `surprise_bonus:` and `'platform'` is in the CHECK list. Works.

Also verified: celebration tier is `none` (`celebration.ts:36`) with the correct rationale — spectacle would make the reward salient and start eroding the very unexpectedness the design depends on. All three locales carry `surpriseBonus` and `mastery` keys.

### The research citation holds

PED-10 is marked **CONFIRMED, verbatim against the 1999 primary** in the unified spec, and it states plainly: "Unexpected rewards = 0.01 … Surprise bonuses are the only zero-undermining reward form." The PR's `d≈0.01` matches. It also correctly **avoids the −0.88 figure**, which the unification refuted as an attribution (MAS-20) and ordered struck from copy. Worth saying because that figure is exactly the kind of thing that gets re-relayed once it's in circulation.

### Weekly-cap race — confirmed, and the review's detail is better than mine

I found the read-then-write TOCTOU independently; the review adds the part that settles it: `award_xp` takes `pg_advisory_xact_lock` per user, but only around the **daily** cap and insert (`schema.sql:446`) — it knows nothing about the weekly *bonus* cap, which lives in application code. So two concurrent `LessonCompleted` events for one user can both read `count < 3` and both award.

Bounded overshoot: one extra bonus, ≤40 XP, on a delight feature. Agreed it's not worth an atomic fix. One correction to the code's own doc though — it says "the function never over-grants," which is true only of the count-*error* path (that correctly fails closed). The count itself races. Worth softening that comment so a future reader doesn't rely on a guarantee that isn't there.

### My distinct finding — `source = 'platform'` is a decision deferred, not avoided

Surprise bonuses land as `source = 'platform'` via the `ELSE` branch. That's harmless today: `source` has **no consumer** — #557 added the column as groundwork and league scoring (LX-B9a) isn't built. I checked; nothing reads it.

But the column's own comment says **"league scoring filters by this instead of parsing free-text reasons."** So whoever builds league scoring inherits an unflagged choice: `'platform'` now contains **random** XP, and including it makes standings partly luck. That cuts against the leaderboard posture the same spec sets out — PED-14/UIU-09 demote the global absolute board precisely because competitive framing is the risky part, and randomness in ranking is a different animal from randomness in motivation. The research supports unexpected rewards for intrinsic motivation; it says nothing about seeding them into a competitive board.

Cheap to decide now, annoying to unpick after the league ships. Filed as **#694** — either give surprise bonuses their own `source` value, or have league scoring exclude `'platform'` deliberately and say so.

PR #640, opened 2026-07-26
feat(courses): path-page presentation + per-segment guidance modality seam (LX-A7)

thomgabriel:
Closes #586

Implements **LX-A7** (launch-experience master spec §3): the Learning Paths surface now renders as a **sequenced list with exactly one highlighted "start here" card** plus an always-available **Browse all courses** escape — not a catalog grid.

## What changed

- **`PathsView`** (`apps/web/src/components/course/paths-view.tsx`) — new composition for the Paths tab of `/courses`:
  - a per-modality guidance line,
  - ONE gradient-highlighted start-here card (UIU-07: one clear start for the beginner default). CTA switches to "Continue Learning" when the learner is mid-course,
  - the sequenced `LearningPathSection` lists (unchanged sequence/content),
  - a **Browse all courses** button that returns to the full catalog tab (S9: don't punish experts),
  - an empty state when the bundle provides no populated paths.
- **`LearningPathSection`** gains a `modality` prop; the previous hard-lock behavior is now the `fixed` modality (and remains the default for any other call site).
- **i18n ×3** (en / pt-BR / es): `startHere`, `browseAllCourses`, `skipAhead`, and the three guidance lines.
- **CSS**: `path-start-*`, `path-guidance*`, `path-browse-all`, `path-step-badge.skip` in `globals.css`, using existing tokens (`--sol-grad`, `--card`, `--r-lg`, `--xp`); dark-first with light-mode vars; mobile stacking at 640px.

## Segment seam (for #566 / LX-A3 — no rework needed)

Segment state does **not** exist yet, so nothing here reads storage or routes by segment. The seam is:

- `apps/web/src/lib/courses/learner-segment.ts` — exports `LearnerSegment` (`1 | 2 | 3` per Design A: 1 = CORE web2, 2 = web3 dev, 3 = beginner), `DEFAULT_SEGMENT = 1`, and `SEGMENT_PATH_MODALITY` (1 → `guided-skip`, 2 → `open`, 3 → `fixed`).
- `PathsView` accepts an optional `segment` prop, defaulting to the **segment-1 presentation** (guided sequence + visible skip-ahead — S9: default to guidance without punishing experts).
- When #566 lands, the page reads the stored segment and passes `segment={...}` to `PathsView`. That is the entire integration.

Per the expertise-reversal deferral, modality changes **guidance copy/prominence only** — the sequence and content are identical for every segment:

| Segment | Modality | Presentation |
|---|---|---|
[...]

Issue #408, opened 2026-07-10
profiles.wallet_address is self-writable — spoofable identity for Helius XP resolution (and any wallet-keyed auth)


thomgabriel:
Found during the SP1 plan security audit (pre-existing, live today).

The profiles UPDATE RLS policy is `USING (auth.uid() = id)` with **no WITH CHECK and no column restriction** (supabase/schema.sql:214-215; baseline migration). Nothing forces wallet writes through the SIWS-verified routes — any authenticated user can PostgREST-update their own `wallet_address` to ANY unclaimed wallet using the anon key + their JWT. The UNIQUE constraint only blocks already-claimed wallets.

**Impact today:** `lib/helius/event-handlers.ts`/`resolvers.ts` resolve users by wallet_address → an attacker can claim another wallet (e.g. an active learner's) and capture their on-chain XP/progress events.
**Impact on SP1:** the planned instructor-viewer/stats wallet-match auth would be spoofable — being fixed in the SP1 migration (wallet_address write-lock trigger, same pattern as the role escalation lock it replaces).

**Fix:** trigger blocking wallet_address UPDATE except via service_role (mirror of trg_enforce_profile_role_write), applied to David's prod DB after Supabase stability is confirmed. SIWS routes already run service-role so legit linking is unaffected — verify /api/auth/link-wallet path first.

Refs #263 #359

thomgabriel:
**Prod-verified on apply (2026-07-10).** Migration applied to prod DB pywhtmidcrptomrabbrw: trg_enforce_profile_wallet_write live (SECURITY INVOKER, service_role-only, EXECUTE revoked from anon/authenticated). Live attack sim: authenticated caller changing wallet_address → rejected (insufficient_privilege), victim wallet unchanged; service_role UPDATE (SIWS link) → succeeds. The self-write vector is closed in production.

thomgabriel:
Follow-up filed as **#696**: the prod fix is confirmed live, but `supabase/schema.sql` never mirrored `trg_enforce_profile_wallet_write` — PR #695's adversarial reviewer independently re-derived the whole attack from the schema file tonight, proving the drift misleads. Fresh DBs built from schema.sql ship without this guard.

PR #695, opened 2026-07-26
feat(onboarding): /start intake + segment state (LX-A2/LX-A3)

### LINKED ISSUE #566: /start intake + segment state
2-tap fork + goal + optional reflection + daily-goal; localStorage + profiles column; app-side routing constant.

**Spec tasks:** `LX-A2, LX-A3`

thomgabriel:
Closes #566 (LX-A2 + LX-A3 from the launch-experience master spec §3).

## LX-A3 — segment state
- **Routing table is one app-side constant.** Extended the #640 seam `lib/courses/learner-segment.ts` with `SEGMENT_ENTRY_COURSE` (3 segments × entry course id), `GoalId`, and `GOAL_FRAMING_KEY`. No content-schema change — avoids the two-repo staging cycle.
- **`profiles` migration** (`20260726150000_add_profiles_segment_state.sql`) adds `segment` / `goal` / `daily_goal`, all nullable and CHECK-bounded (`segment IN (1,2,3)`, closed goal set, `daily_goal BETWEEN 1 AND 20`). Self-writable through the existing own-row UPDATE policy — **whole-row, because RLS gates rows not columns**; column-level write control here is trigger-based and reserved for `role`. The escalation-lockdown trigger is column-specific to `role` and is provably untouched (the migration issues no DDL against the trigger/column; pinned by `segment-state-rls-guard.test.ts`). Idempotent + transactional, `schema.sql` mirrored, tested rollback.
- **Anon → signed-in flow:** pre-auth answers live in `localStorage` (`lib/onboarding/segment-state.ts`), copied into the profile on first sign-in by `SegmentSync` (mounted in `GamificationOverlays`), mirroring the anonymous progress-bank pattern. `useSegmentState` resolves reads DB-first when signed in, localStorage fallback, degrades silently if columns aren't present yet.
- `courses-client` wires segment → PathsView modality and goal → framing copy.

## LX-A2 — /start intake
- New `(marketing)/start` route: **4 tap-only screens, zero data-entry fields.** (1) experience fork — 3 *verifiable-history* options ("have you shipped JS/TS?"), never self-rating; (2) goal; (3) skippable interest chips; (4) daily-goal picker whose **options derive from the existing daily lesson quest targets** (`dailyGoalOptionsFromQuests`).
- **Answer consumption:** segment → routing (lands in the entry lesson via the constant), goal → path-page framing copy. The daily-goal picker's *options* are wired to real quest targets, and the chosen value is **stored** (`profiles.daily_goal`) + committed as the E2 implementation-intention — but **nothing reads that column yet**; consuming it in the quest engine is a deliberate follow-up (the picker is the substrate). Interest chips feed the funnel event only (no column — the migration is exactly the three).
- Keyboard-operable, focus moved to each screen's heading on advance, ×3 locales, anonymous-friendly.

## Analytics (E2/E7)
Implemented the reserved onboarding funnel events (`onboarding_started` / `step_completed` / `goal_committed` / `route_accepted` / `completed`) per the documented shape; README moved from reserved → active.

## Review fixes (adversarial pass)
- **F1 (was a launch-blocking 404):** `resolveSegmentRoutes` (now `start/routes.ts`) sync-gates through `getCourseLessons`, so a present-but-unsynced entry course — the realistic launch state, and any `getActiveDeployments` DB degradation (empty map) — falls back to `/courses` instead of deep-linking a lesson page that would `notFound()`. Covered by `start/__tests__/routes.test.ts`.
- **F3:** `daily_goal` CHECK bounded to `1..20`; picker derivation clamps to the same cap; the daily-goal claim is corrected above (stored-for-future-wiring, not yet consumed).
- **F4:** `SegmentSync` clears the local intake copy once the DB row is authoritative (successful copy OR already-set segment), closing the shared-browser preference-bleed window; retained only on transient failure. Covered by `components/onboarding/__tests__/segment-sync.test.tsx`.

## Verification
`pnpm typecheck` ✅ · full `pnpm test` ✅ (155 files, 1380 passing) · `pnpm lint` ✅ (0 warnings on touched files) · migration structural + trigger-safety guard ✅.

## Out of scope (as designed)
- The §3.5 guidance-level differentiation deferral (topic-routing only at launch) stands — arrives with LX-C7/C8.
- F2 (pre-existing `wallet_address` self-write) is not part of this change — filed separately.


thomgabriel:
**Orchestrator routing record — SENSITIVE (`profiles` migration) → adversarial review → `needs-gate`.** Independent verification: full suite **1370/1370** at the substantive head and **1380/1380** after the fix round (`25dcfce`), both first-hand. **Adversarial review: PASS with one must-fix, now landed.** The hostile pass PROVED the security core: the role escalation-lockdown trigger keys only on `role` and is provably undisturbed (zero trigger/policy/grant DDL in the migration — the guard test's pins are load-bearing); migration idempotent+transactional with a byte-consistent schema.sql mirror; events PII-free per the documented contract; tap-only + focus-managed + perfect ×3 parity. **Findings & disposition:** F1 (MUST-FIX, fixed in `25dcfce`): /start deep-links were not sync-gated — a present-but-unsynced entry course (or deployment-map DB degradation) 404'd the whole funnel; routing now resolves through the sync-gated query and falls back to /courses, tested for all four cases. F3 (fixed): daily_goal claim corrected (stored-for-future-wiring; nothing reads it yet) + CHECK bounded 1-20 + picker clamped. F4 (fixed): the anonymous intake copy is cleared once the DB row is authoritative — the shared-browser preference-bleed window is closed, five paths tested. **F2 (pre-existing, NOT this PR):** the reviewer re-derived the wallet_address self-write attack from schema.sql — reconciliation with #408 shows the attack is CLOSED IN PROD (trigger applied 2026-07-10) but schema.sql never got the mirror; filed and already in flight as **#696**. Migration ships SQL-only — dry-run-first is yours; note it composes with #689's and #696's pending migrations (three files, disjoint objects).

thomgabriel:
## Gate: PASS — merging, then applying + exploit-verifying on prod

### The review is stale (third time today) and its one finding is already closed

Review `17:10:33Z` against `7b9832c4`; head `25dcfcef` at `17:26:22Z` — **15m49s stale**. Its non-blocking finding was "no dedicated unit tests for `segment-sync.tsx`". The unreviewed commit **adds `segment-sync.test.tsx` (+97)** plus `routes.test.ts` (+70). Already done.

More importantly, that commit also **modified the migration (+8/−4)**, so the security-relevant file in this PR has had no review coverage at all. I read it at head instead.

The post-review migration change is in the *stricter* direction: `daily_goal > 0` → `CHECK (daily_goal BETWEEN 1 AND 20)`, with the reasoning stated — "an unbounded self-writable integer is a needless abuse surface." Correct instinct; that column is client-writable.

### The migration is purely additive — this matters because `profiles` holds real users

`profiles` has 74 rows including last week's event signups, so I checked for anything that could touch existing data. There is nothing:

```sql
ALTER TABLE profiles ADD COLUMN IF NOT EXISTS segment    SMALLINT;
ALTER TABLE profiles ADD COLUMN IF NOT EXISTS goal       TEXT;
ALTER TABLE profiles ADD COLUMN IF NOT EXISTS daily_goal SMALLINT;
-- + 3 CHECKs, each guarded by an IF NOT EXISTS on pg_constraint
```

All three nullable with no default, so the `ADD COLUMN`s are metadata-only rewrites. The CHECKs are NULL-safe (NULL → unknown → not false), so all 74 existing rows pass with no backfill and no violation. No `UPDATE`, no `DROP`, no policy DDL.

### The self-write claim, verified

The body is upfront that these columns become self-writable through the existing own-row `UPDATE` policy, "whole-row, because RLS gates rows not columns." That is the right framing and the honest risk to check, because the same policy is what the `role` escalation-lockdown trigger defends against.

Verified the migration issues **no DDL whatsoever** against `role`, the trigger, or the function — not `ALTER COLUMN role`, not `CREATE/DROP TRIGGER`, nothing. And the widened surface is harmless by design: `segment` picks an entry course, `goal` selects framing copy, `daily_goal` is stored and read by nothing yet. Nothing reward-bearing, nothing privilege-bearing, and all three CHECK-bounded to closed sets.

The guard test is string-assertion based (RLS can't be exercised in unit tests), but its assertions are the useful *negative* kind — it fails if any future migration in this file touches `role` or a trigger. That is a real regression guard rather than a tautology.

Applying to prod now and verifying by attempting the escalation, not by reading the schema back. Results below.

### One caveat worth recording, surfaced by #697

#697 ("mirror prod wallet-write lock trigger into schema.sql") means **prod carries at least one trigger `schema.sql` does not**. This PR's guard test treats `schema.sql` as the mirror of prod and asserts the role trigger is present there. That assertion still holds, but the general principle it relies on — schema.sql faithfully mirrors prod — is currently false. Not a defect in this PR; it does mean a schema.sql-based guard is weaker evidence than it looks, and it is why I am verifying against the live database rather than the snapshot.

thomgabriel:
## Applied to prod + exploit-verified — and a correction to my own gate note

Migration `add_profiles_segment_state` applied to `pywhtmidcrptomrabbrw`. Prod `profiles` now carries `segment`, `goal`, `daily_goal`, all NULL on the existing 74 rows (untouched, no backfill, no rewrite), with all three CHECKs live.

### Verified by attempting the writes as an impersonated real user

| # | Attempt | Expected | Result |
|---|---|---|---|
| 0 | `auth.uid()` = impersonated user | true | **true** (two distinct users) |
| 1 | self-write `segment=2, goal='build', daily_goal=3` | allowed | **1 affected** |
| 2 | `segment = 4` (outside 1..3) | rejected | **CHECK violation** |
| 3 | `goal = 'hacker'` | rejected | **CHECK violation** |
| 4 | `daily_goal = 9999` | rejected | **CHECK violation** |
| 5 | `daily_goal = 0` (lower bound) | rejected | **CHECK violation** |
| 6 | write **another** user's intake state | 0 affected | **0 affected** |
| 7 | overwrite own `wallet_address` | blocked | **blocked — "wallet_address may only be changed by service…"** |

Rolled back; nothing persisted. Test 5 confirms the post-review tightening (`> 0` → `BETWEEN 1 AND 20`) is doing real work at both ends.

### Correction: the safety argument in this PR cites a protection that does not exist on prod

I accepted the reasoning "whole-row self-write is safe because the escalation-lockdown trigger is column-specific to `role`". While probing, that failed with:

```
ERROR: 42703: column "role" does not exist
```

Prod `profiles` has **no `role` column**, and `enforce_profile_role_write` **does not exist** (`pg_proc` count = 0). The only trigger on `profiles` is `trg_enforce_profile_wallet_write`.

**The conclusion still holds, but for a different reason than the one given.** Whole-row self-write is safe here because there is nothing privilege-bearing on the table to escalate *at all* — not because a trigger is standing guard. And prod is not the broken side: the only reference to `profiles.role` left in app code calls it retired —

> `SP1 PR-2 Task 7: replaces the retired profiles.role / course.author`

— so `role` was deliberately removed. **`supabase/schema.sql` is the stale artifact**, still defining the retired column and its trigger function (6 references).

### Consequence for this PR's guard test

`segment-state-rls-guard.test.ts` asserts *"schema.sql still defines the role trigger exactly once (mirror intact)"*. That assertion passes — and guards nothing, because the trigger it pins is not deployed. It is a green test standing in for a protection that does not exist in production.

Its **other** assertions remain genuinely valuable: the negative checks that this migration issues no `CREATE/DROP TRIGGER`, no `ALTER COLUMN role`, no `UPDATE profiles SET role` would catch a real regression. Only the mirror-intact assertion is hollow.

Not a reason to revert — the applied state is verified safe on every axis that exists. But the test should stop asserting a trigger prod does not have, and the migration header comment should stop citing it as the reason self-write is safe. Filed as **#698**, together with the reverse drift #697 is fixing.

Also verified in passing, and useful for #697: `trg_enforce_profile_wallet_write` is **live and effective** on prod (test 7), while `schema.sql` has zero references to it. So the snapshot is drifted in both directions at once.

PR #697, opened 2026-07-26
fix(schema): mirror prod wallet-write lock trigger into schema.sql

### LINKED ISSUE #408: profiles.wallet_address is self-writable — spoofable identity for Helius XP resolution (and any wallet-keyed auth)

### LINKED ISSUE #696: schema.sql omits the prod-applied wallet-write trigger (#408's fix) — fresh DBs built from it ship vulnerable
**Re-scoped after reconciling with #408 (closed 2026-07-10, prod-verified fixed):** production carries `trg_enforce_profile_wallet_write` (service_role-only wallet_address writes; live attack sim rejected an authenticated self-write). But **`supabase/schema.sql` never got the mirror** — PR #695's adversarial reviewer read the current schema.sql and independently re-derived the full attack (own-row UPDATE policy + no column guard + #620's deploy-save trusting `profiles.wallet_address` as the linked-wallet anchor), which proves the drift is real and materially misleading. Consequences: (1) any FRESH database built from schema.sql (staging, local dev, a future environment) ships WITHOUT the guard — the #408 attack works there, and #620's trust chain with it; (2) every reviewer/agent reading schema.sql reaches wrong security conclusions (it happened tonight). Fix: mirror the trigger + function + REVOKE/GRANT stanza from the applied migration into schema.sql (find the #408 migration file — if it was MCP-applied without a committed migration file, reconstruct from prod via the migration list and commit BOTH), plus a guard test pinning it (the segment-state guard-test pattern). _Filed from the orchestrator loop; supersedes the original duplicate body._

thomgabriel:
## What

`supabase/schema.sql` never mirrored `trg_enforce_profile_wallet_write` — the wallet-write lock that **closed #408 in production on 2026-07-10** (prod-verified: an authenticated self-write of `wallet_address` is rejected with `insufficient_privilege`; a service-role SIWS link still succeeds). This PR mirrors the trigger + function + `REVOKE` into `schema.sql` and adds a guard test pinning it.

`Closes #696`

## Why it matters

The profiles self-service RLS policies (`auth.uid() = id` for INSERT/UPDATE) don't constrain **which** columns are written, so without the guard any authenticated user could overwrite their own `wallet_address` and spoof the wallet identity that Helius XP resolution and #620's linked-wallet trust chain depend on. Because the fix lived only in prod (never in the committed snapshot), **any DB built from `schema.sql` — staging, local dev, a future environment — shipped without the guard**, re-opening the exact vector. PR #695's adversarial reviewer independently re-derived the whole attack from `schema.sql` alone tonight, which is what surfaced the drift.

## Source of truth (which case)

**Committed migration FOUND** — no reconstruction needed. The exact definition is in `supabase/migrations/20260710120000_drop_teacher_role.sql` section **(b)** (that migration closed #408; its header carries the full threat write-up). This PR mirrors that function/trigger/`REVOKE` into `schema.sql` **verbatim** (executable SQL byte-identical; inline comments condensed to match how `schema.sql` already condensed the sibling `enforce_profile_role_write` block).

**No live DB was touched** — prod already has the trigger; these files just make the committed snapshot tell the truth.

## Placement

Inserted as section **5a-bis**, immediately after the role-write lock (section 5a) it mirrors — `enforce_profile_role_write` — and before the leaderboard RPC (5b). The two triggers are column-specific and coexist on `profiles`.

## Guard test

`apps/web/src/lib/supabase/__tests__/wallet-write-lock-guard.test.ts`, following the #569 / #493 guard-test pattern. Asserts, in **both** the migration and the `schema.sql` mirror:

- function + trigger present **exactly once**; drop-then-create (idempotent)
- `SECURITY INVOKER` + `SET search_path = ''` (must observe the caller, not the definer)
- service_role-only privilege check (direct `current_user` **or** PostgREST JWT claim)
- keyed on `wallet_address`: non-service UPDATE that changes it raises `insufficient_privilege`; non-service INSERT coerces it to NULL
- `REVOKE EXECUTE` from `PUBLIC, anon, authenticated`
- coexistence with the role-write lock; no crosstalk into `role` / #695's `segment`/`daily_goal` columns

## Cross-check vs #695

#695's migration (adds `segment`/`goal`/`daily_goal` columns ridden by the same own-row UPDATE policy) is not yet in the repo. The wallet trigger is **column-specific** — it only fires when `wallet_address` changes and never reads/writes those columns — so there is no interaction; the coexistence test pins that. It also coexists with the role trigger.

## Verification

- SQL parse (libpg-query): `schema.sql` **260 statements OK**, migration **8 statements OK**
- `tsc --noEmit`: clean
- Guard test: **16/16**; full web suite: **1365/1365** (151 files)
- `next lint` on the new file: clean (two pre-existing `import/order` warnings in untouched `client.ts`/`server.ts` are not from this PR)

## Out of scope (flagged)

`schema.sql` still carries the `profiles.role` column, `chk_profiles_role`, and `enforce_profile_role_write` — all of which `20260710120000_drop_teacher_role.sql` **drops** in its section (a). Whether prod dropped the role machinery or only applied the wallet stanza (the migration's own apply-gate note defers the role drop until the teacher-authoring repoint ships) is a separate, larger snapshot-vs-migration drift. This PR deliberately leaves the role machinery in `schema.sql` untouched — mirroring the wallet guard beside a still-present role guard reflects the most likely current prod state (both triggers live) and keeps the change scoped to #696.

### DIFF
--- a/apps/web/src/lib/supabase/__tests__/wallet-write-lock-guard.test.ts
+++ b/apps/web/src/lib/supabase/__tests__/wallet-write-lock-guard.test.ts
@@ -1,0 +1,133 @@
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, it, expect } from "vitest";
+
+// #696 (mirrors the prod fix from #408): profiles.wallet_address was
+// self-writable — the self-service profiles RLS policies (auth.uid() = id for
+// INSERT/UPDATE) do not constrain WHICH columns are written, so any
+// authenticated user could overwrite their own wallet_address and spoof the
+// wallet identity that Helius XP resolution and the linked-wallet trust chain
+// depend on. Production carries trg_enforce_profile_wallet_write (live since
+// 2026-07-10) but supabase/schema.sql never mirrored it, so any DB built from
+// the snapshot shipped WITHOUT the guard. RLS/triggers cannot be exercised in
+// unit tests, so — as with the #493 profile guard and the #569 review spine —
+// pin the security-critical SQL invariants in BOTH the migration that applies
+// the fix and the schema.sql mirror new environments are built from. A drift
+// between them is exactly the class of bug this test exists to prevent.
+
+function findRepoRoot(): string {
+  let dir = dirname(fileURLToPath(import.meta.url));
+  while (!existsSync(resolve(dir, "supabase/schema.sql"))) {
+    const parent = dirname(dir);
+    if (parent === dir)
+      throw new Error("repo root (supabase/schema.sql) not found");
+    dir = parent;
+  }
+  return dir;
+}
+
+const repoRoot = findRepoRoot();
+// The committed source of truth for the trigger definition (#408 was closed by
+// this migration; the wallet-lock stanza lives in section (b)).
+const migration = readFileSync(
+  resolve(repoRoot, "supabase/migrations/20260710120000_drop_teacher_role.sql"),
+  "utf8"
+);
+const schema = readFileSync(resolve(repoRoot, "supabase/schema.sql"), "utf8");
+
+const FN = "public.enforce_profile_wallet_write()";
+const TRIGGER = "trg_enforce_profile_wallet_write";
+
+// The guard must hold in the migration (what gets applied) AND in schema.sql
+// (the snapshot fresh environments are built from).
+for (const [label, sql] of [
+  ["migration", migration] as const,
+  ["schema.sql", schema] as const,
+]) {
+  describe(`#696 wallet-write lock — ${label}`, () => {
+    it("defines the trigger function exactly once", () => {
+      const defs = sql.split(`CREATE OR REPLACE FUNCTION ${FN}`).length - 1;
+      expect(defs).toBe(1);
+    });
+
+    it("installs the BEFORE INSERT OR UPDATE trigger exactly once", () => {
+      const creates = sql.split(`CREATE TRIGGER ${TRIGGER}`).length - 1;
+      expect(creates).toBe(1);
+      const start = sql.indexOf(`CREATE TRIGGER ${TRIGGER}`);
+      const body = sql.slice(start, start + 200);
+      expect(body).toContain("BEFORE INSERT OR UPDATE ON public.profiles");
+      expect(body).toContain("FOR EACH ROW");
+      expect(body).toContain(`EXECUTE FUNCTION ${FN}`);
+      // Drop-then-create so re-running the SQL is idempotent.
+      expect(sql).toContain(
+        `DROP TRIGGER IF EXISTS ${TRIGGER} ON public.profiles;`
+      );
+    });
+
+    it("observes the CALLER's role (SECURITY INVOKER, pinned search_path)", () => {
+      const start = sql.indexOf(`CREATE OR REPLACE FUNCTION ${FN}`);
+      const body = sql.slice(start, sql.indexOf("$$;", start));
+      // SECURITY DEFINER here would read the owner's role and defeat the check.
+      expect(body).toContain("SECURITY INVOKER");
+      expect(body).toContain("SET search_path = ''");
+    });
+
+    it("treats only service_role as privileged (direct role OR JWT claim)", () => {
+      const start = sql.indexOf(`CREATE OR REPLACE FUNCTION ${FN}`);
+      const body = sql.slice(start, sql.indexOf("$$;", start));
+      expect(body).toContain("current_setting('request.jwt.claims', true)");
+      expect(body).toContain(
+        "current_user = 'service_role' OR jwt_role = 'service_role'"
+      );
+    });
+
+    it("is keyed on wallet_address — non-service UPDATE that changes it errors", () => {
+      const start = sql.indexOf(`CREATE OR REPLACE FUNCTION ${FN}`);
+      const body = sql.slice(start, sql.indexOf("$$;", start));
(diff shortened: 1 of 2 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (`supabase/**`) → `needs-gate`.** Independent verification: full suite **1365/1365** (incl. the new 16-assertion wallet-lock guard test) re-run first-hand. Low-risk by construction: this MIRRORS an already-prod-applied control (the #408 fix, live since 2026-07-10) into schema.sql — the definition was found COMMITTED in `20260710120000_drop_teacher_role.sql` section (b), so the mirror is verbatim-executable-SQL, no reconstruction; no new migration; nothing to apply. The guard test pins presence-exactly-once, SECURITY INVOKER + empty search_path, service_role-only semantics, wallet_address keying, REVOKEs, and coexistence with the role trigger + #695's pending columns. **One question surfaced for you (also in the PR body):** the SAME migration's section (a) drops `profiles.role` + its trigger behind an apply-gate — has prod actually applied that drop, or does prod still carry both? A one-query prod check settles whether schema.sql's role machinery is accurate or a second drift. Filing it as a tracked question.

thomgabriel:
## Gate: PASS — merging

I verified this against the live database rather than against the PR's description, because the whole point of a mirror fix is fidelity.

### The mirror is byte-faithful to prod

Dumped prod's real definition (`pg_get_functiondef` / `pg_get_triggerdef` on `pywhtmidcrptomrabbrw`) and compared it to what this PR adds. Every semantically load-bearing line matches:

| prod | schema.sql (this PR) |
|---|---|
| `SET search_path TO ''` | `SET search_path = ''` ✓ |
| `jwt_role := NULLIF(current_setting('request.jwt.claims', true), '')::jsonb ->> 'role'` | identical ✓ |
| `is_privileged := COALESCE(current_user = 'service_role' OR jwt_role = 'service_role', false)` | identical ✓ |
| UPDATE: `NEW.wallet_address IS DISTINCT FROM OLD.wallet_address` → `RAISE … USING ERRCODE = 'insufficient_privilege'` | identical ✓ |
| INSERT: `IF NEW.wallet_address IS NOT NULL THEN NEW.wallet_address := NULL;` | identical ✓ |
| `BEFORE INSERT OR UPDATE ON public.profiles FOR EACH ROW` | identical ✓ |

A near-miss here would have been worse than the original gap, because a drifted trigger in the snapshot *looks* mirrored. It isn't drifted.

### The premise is confirmed independently

I hit this an hour ago from the other direction while applying #695's migration. `schema.sql` had **zero** references to `trg_enforce_profile_wallet_write`, and the trigger is live and effective in prod — verified by attempting the write as an impersonated authenticated user:

```
update profiles set wallet_address = '<forged>' where id = <own id>
→ blocked: permission denied: wallet_address may only be changed by service_role
```

So the risk in the PR body is real and correctly stated: any environment built from the snapshot would have had the self-service `profiles` UPDATE policy with **no** column constraint on `wallet_address`, which is the identity the Helius XP resolution and #620's linked-wallet trust chain rest on.

Note the INSERT branch is the subtler half — it silently nulls a supplied `wallet_address` rather than raising, so a fresh environment missing this trigger would let a user *set* their wallet at profile creation without any error to reveal it.

### Finding — the new guard test pins a trigger that does not exist in prod

`wallet-write-lock-guard.test.ts` asserts:

```js
expect(schema).toContain("CREATE TRIGGER trg_enforce_profile_role_write");
```

That trigger is **not deployed**. Prod has no `role` column and no `enforce_profile_role_write` function (`pg_proc` count = 0), because `role` was retired by SP1 — the only surviving reference in app code says so outright: *"replaces the retired `profiles.role`"*.

So this assertion passes while guarding nothing, and it is the **second** test file to carry it — #695's `segment-state-rls-guard.test.ts` has the same one. I filed that as #699 before this PR reached me; updating it to cover both files.

Not blocking: the assertion is inert rather than wrong-in-a-dangerous-direction, and everything else in this test is genuinely load-bearing — it pins the privilege check, the `IS DISTINCT FROM` comparison, the errcode, the INSERT-nulling branch, and the REVOKE. Those are exactly the lines that must not silently change.

Minor note, not worth a change: `REVOKE EXECUTE … FROM PUBLIC, anon, authenticated` is defensive rather than load-bearing — a trigger function fires as part of the table operation, not via a direct call, so EXECUTE isn't required for it to work. Prod's grants are already just `postgres:EXECUTE`, so the REVOKE reproduces the effective state. Harmless and consistent with the repo's convention.

PR #700, opened 2026-07-26
feat(ai): post-pass idiomatic review action (LX-C9)

### LINKED ISSUE #587: AI post-pass idiomatic review
Opt-in review after a passing run; pre-pass suppression untouched (roadmap #13).

**Spec tasks:** `LX-C9`

thomgabriel:
Closes #587 — LX-C9 (launch-experience master spec §3).

## What

Adds an **opt-in AI review of a *passing* solution** for idiomatic improvement. Strictly post-pass: it gates nothing, grants nothing, and leaves the pre-pass no-answer contract byte-identical.

## Action design (`review`)

Rides the existing billed, latched, fail-closed `/api/ai/partner` route as a fourth action, following the ask/hint/propose pattern exactly:

- **Metered** — spends one paid assist via `spendAssist`, records a billed assist, refunds only on a pre-billing throw. Same fail-closed rate limits (per-user + per-IP). No unmetered path.
- **Cap** — `maxTokensFor("review") = 1536` (between `hint` 512 and `propose` 2048). Bounded structured payload; no code/whole-file field, so there is no truncation lever to inflate.
- **Schema** — `{ type: "review", summary: string, notes: string[] }`. `summary` is a 1–2 sentence affirming read; `notes` is 0–6 short idiomatic suggestions and **may be empty** (already-idiomatic → no invented problems). No `edits`/`text`/`proposedCode` field: it is feedback, not a fix or a rewrite. Re-bounded server-side at the money surface.
- **Prompt** — the idiomatic-review persona lives in the **dynamic suffix** (review action only). The cached static prefix, `SYSTEM_PERSONA`, task/tests/solution, and the pre-pass suffix for hint/propose/ask stay byte-identical.

## Button + gating

The "Review my solution" button lives in the existing AI pane (`AiPartnerPane`) and is surfaced **only when the most recent run passed** (`challengeState.status === "success"`, threaded as `solutionPassed`). Never automatic, never pre-pass. It is intentionally **not** gated by the lesson-complete `disabled` (reviewing the solution you just passed is the whole point) but **is** disabled on `budgetExhausted`.

**Quiz suppression is respected for free**: `ChallengeInterface` does not mount the pane at all while a quiz block is unanswered (LX-C1/F18), so a suppressed lesson can never surface the button.

×3 locales (EN/ES/PT-BR), a11y button + labels.

## Tests

- Route: passthrough, empty-notes, billed-path (spend×1 / billed×1 / no refund), budget-exhausted, cap+schema, malformed→502-not-refunded, note re-bounding, assist-log (no user turn).
- Prompt: review cap ordering, schema shape, suffix injection only for `review`, pre-pass suffixes/prefix byte-clean.
- Hook: `review()` POSTs `action:review`.
- UI: message-list renders summary+notes (and empty-notes); pane button gating (hidden pre-pass, shown+clickable post-pass, enabled when complete, disabled on budget-exhausted).

## Verification

- `pnpm --filter web test` — 1370/1370 green
- `pnpm --filter web lint` — 0 errors (cleared the pane's import-order warnings while touching it)
- `prettier --check` on touched files — clean

## Out of scope / notes

- The route trusts the client's `testSummary` as untrusted context (as it already does for every action); the pass gate is client-side, consistent with the rest of the surface. A review of non-passing code is harmless (grants nothing) and still metered.

### DIFF
--- a/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
+++ b/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
@@ -850,4 +850,176 @@ describe("POST /api/ai/partner", () => {
 
     expect(res.status).toBe(502);
   });
+
+  // --- LX-C9: post-pass idiomatic review (`review` action) ---
+
+  const REVIEW_BODY = {
+    ...VALID_BODY,
+    action: "review",
+    testSummary: "3/3 passing",
+  };
+
+  const REVIEW_GEMINI_TEXT = JSON.stringify({
+    type: "review",
+    summary: "Your solution passes and reads clearly.",
+    notes: [
+      "Prefer iter().sum() over the manual loop.",
+      "Name `n` for clarity.",
+    ],
+  });
+
+  it("passes through a well-formed review response for the review action", async () => {
+    stubGeminiFetch(REVIEW_GEMINI_TEXT);
+
+    const { POST } = await import("../partner/route");
+    const res = await POST(makeRequest(REVIEW_BODY));
+    const json = await res.json();
+
+    expect(res.status).toBe(200);
+    expect(json).toEqual({
+      type: "review",
+      summary: "Your solution passes and reads clearly.",
+      notes: [
+        "Prefer iter().sum() over the manual loop.",
+        "Name `n` for clarity.",
+      ],
+    });
+  });
+
+  it("accepts a review with an empty notes array (already idiomatic)", async () => {
+    stubGeminiFetch(
+      JSON.stringify({
+        type: "review",
+        summary: "Already idiomatic — nothing to change.",
+        notes: [],
+      })
+    );
+
+    const { POST } = await import("../partner/route");
+    const res = await POST(makeRequest(REVIEW_BODY));
+    const json = await res.json();
+
+    expect(res.status).toBe(200);
+    expect(json).toEqual({
+      type: "review",
+      summary: "Already idiomatic — nothing to change.",
+      notes: [],
+    });
+  });
+
+  it("is a metered paid action: spends once, records billed once, no refund (billed-path)", async () => {
+    stubGeminiFetch(REVIEW_GEMINI_TEXT);
+
+    const { POST } = await import("../partner/route");
+    const res = await POST(makeRequest(REVIEW_BODY));
+
+    expect(res.status).toBe(200);
+    // The review rides the same billed, fail-closed, assist-metered path as
+    // every other action — never an unmetered generation.
+    expect(spendAssist).toHaveBeenCalledTimes(1);
+    expect(spendAssist).toHaveBeenCalledWith("user-1", "lesson-1");
+    expect(recordBilledAssist).toHaveBeenCalledTimes(1);
+    expect(refundAssist).not.toHaveBeenCalled();
+  });
+
+  it("returns budgetExhausted for review when the assist budget is spent", async () => {
+    spendAssist.mockResolvedValue({ allowed: false, used: 4 });
+    stubGeminiFetch(REVIEW_GEMINI_TEXT);
+
+    const { POST } = await import("../partner/route");
+    const res = await POST(makeRequest(REVIEW_BODY));
+    const json = await res.json();
+
+    expect(res.status).toBe(200);
+    expect(json).toMatchObject({ budgetExhausted: true, used: 4 });
+  });
+
+  it("caps review output below propose and uses the summary+notes schema", async () => {
+    const fetchSpy = vi.fn(async (_url: string, init: { body: string }) => {
+      void init;
+      return {
+        ok: true,
+        text: async () => "",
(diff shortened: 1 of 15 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (billed AI route) → adversarial review → `needs-gate`.** Independent verification: full suite **1370/1370** re-run first-hand. **Adversarial review: PASS, no blocker.** Proven under hostile read: the action is PURELY additive on the money route (billed latch, both failClosed limiters, per-IP bucket, MAX_CODE_CHARS, tutorNotes bounds all byte-identical; zero changes to assist-budget/rate-limit/check-seal); static-prefix cache cleanliness holds (REVIEW_INSTRUCTIONS provably confined to the review-action dynamic suffix — no cache-key contamination of other actions); truncation → 502 billed-not-refunded consistent with #613; the trustless client pass-gate grants nothing and stays metered; quiz suppression is structural (pane unmounted). **One factual correction applied to the PR body** (review's own catch): the original claimed the reference solution is in the public projection — it is server-only; the accurate rationale is that the solution was ALREADY in the Gemini prompt for every existing action, so review adds no new model exposure. **Accepted nice-to-haves (recorded, not blocking):** per-note char cap (output-side, bounded by the 1536 token cap and JSX-escaped — the input-side asymmetry vs #644's tutorNotes cap is deliberate) and two minor test additions.

thomgabriel:
## Gate: PASS — merging

Review is current (17:46:46Z vs head 17:45:42Z, 64s after). I went at the cost surface and the answer-leak surface, since those are the two ways an AI action goes wrong.

### The answer-leak concern dissolves, for a reason worth stating

I started from "can a learner call `review` pre-pass and get the solution?" The route answers it at the source:

> The lesson's **PUBLIC** block projection (post-D4 there is no secret answer key). The AI Partner surfaces a `code` block; the challenge solution + tests are read straight from that block — same projection every reader gets.

Post-D4 there is no confidentiality boundary to breach: the solution and tests are already public to every reader. So the "no-answer contract" is about the model not *handing over* an answer in its response, not about hiding data — and this action's schema structurally cannot: `{ type, summary, notes[] }` with **no** `edits`, `text`, `proposedCode`, or any code-bearing field.

### Cost bounds — verified, not taken on faith

- **Metered with no unmetered path.** `spendAssist(user.id, lesson._id)` runs before any Gemini call and before the action branches, so `review` cannot bypass it. A denied budget short-circuits without a model call.
- **Refund discipline is correct.** `billed` flips on the model's 2xx, and the refund is gated on `if (!billed)`. A successful-but-useless generation is billed and **not** refunded — which is what closes AIE-10/-11, where reliably triggering truncation bought unlimited billed calls against a quota that never decremented.
- **`review: 1536`** is an explicit entry in `MAX_TOKENS`, between `hint: 512` and `propose: 2048`. Not a default fallback.
- **No truncation lever.** Because the schema has no whole-file field, there is no attacker-controllable way to force MAX_TOKENS — the same reasoning that justified dropping `propose` from 8192 → 2048.
- **Double-bounded output.** `MAX_REVIEW_NOTES = 6` is enforced in the response schema (`maxItems`) *and* re-bounded server-side with `.slice(0, MAX_REVIEW_NOTES)` at the money surface, with a malformed payload 502ing rather than forwarding unbounded text.

Against O-1's envelope ($500/mo @ 10k MAU, per-account $0.50 soft / $1.50 hard), a 1536-token bounded action costing one assist is comfortably inside.

### Prompt-cache placement is the right call

`REVIEW_INSTRUCTIONS` is appended **only** when `action === "review"`, in the dynamic suffix — the cached static prefix and `SYSTEM_PERSONA` are untouched. Two things follow, both good: the pre-pass prompt for `hint`/`propose`/`ask` stays byte-identical (so this PR genuinely cannot change pre-pass behaviour), and only `review` calls pay for those tokens instead of every action carrying them.

Also verified all three locales (`en`/`es`/`pt-BR`) carry the review keys.

### Note on my own process

My first check of `maxTokensFor` grepped the **local** file — which is `main`'s copy without this PR's changes — and appeared to show `review` missing from `MAX_TOKENS`. That was my error, not a defect: reading the file at the PR head shows `review: 1536` present with its rationale. Recording it because "grep the working tree while gating an unmerged PR" is a reliable way to invent a finding.

thomgabriel:
**Erratum to my earlier body correction (via #702's adversarial review):** the accurate exposure picture is subtler than either prior phrasing. The solution IS present in the browser — the lesson page is a server component that passes the full lesson object (solutions included) as props into the client tree, so the bytes ride the RSC payload and have since the SP2 flip. What is server-only is the client FETCH path (client-queries deliberately has no full-Lesson read). The #700 security conclusion is unchanged either way: the review action adds no new exposure on any seam (model-side: the solution was already in the Gemini prompt for every action; client-side: already embedded in the page payload).

PR #701, opened 2026-07-26
feat(content): Parsons block type + server-authoritative grader + renderer (LX-C7)

### LINKED ISSUE #577: Parsons block type + ladder rungs + segment entry
Server-authoritative grader; defer segment plumbing until variants exist.

**Spec tasks:** `LX-C7, LX-C8`

thomgabriel:
Closes #577 — Parsons block type + ladder rungs (**LX-C7**). Segment entry-rung plumbing (**LX-C8**) is deferred per the issue ("defer segment plumbing until variants exist"); the seam is noted below.

## What ships

A new `parsons` (arrange-the-lines) block: schema + registry, a server-authoritative deterministic grader, an accessible renderer, and ×3-locale UI chrome. No content authored (content repo is separate) — this ships the capability + tests.

### 1. Block schema (`packages/content-schema`)
`ParsonsBlock` (zod), following the quiz-block pattern:

| field | meaning |
|---|---|
| `prompt` | the task text |
| `lines[]` | `{ id, content, distractor?, pairedWith? }` — the shuffled presentation pool. Correctness is keyed to stable line **ids**, never array index (same invariant `QuizOption` keeps). `content`'s leading whitespace is its displayed indentation. |
| `correctOrder[]` | ordered non-distractor line ids — a **permutation of exactly the non-distractor lines** (refined: unique, all reference non-distractor lines, every solution line used once). The sole grading key. |
| `explanation?` | shown after a check, like `QuizQuestion.explanation`. |

Distractors (`distractor: true`) are optional and excluded from `correctOrder`; `pairedWith` is a distractor-only presentational hint (F13 paired-distractor markers), validated to reference a real non-distractor line. Wired into the `Block` discriminated union + `BLOCK_REGISTRY` (`graded: true, required: true`). **JSON schema regenerated** — only `lesson.schema.json` changes (parsons rides the lesson block union; no standalone doc kind, unlike `*.quiz.yaml`). **content-lint needs no change**: gate-1 validates parsons through the shared `Lesson` schema (verified — the loader only classifies standalone `.quiz.yaml`).

### 2. Grader semantics (`apps/web/src/lib/grading/graders/parsons.ts`)
Pure, deterministic. Proof = `{ order: string[] }` (arranged line ids). Verdict:
- **exact sequence equality** of `proof.order` vs `block.correctOrder` → `{ ok: true }`
- wrong order / included distractor / missing line / length mismatch → **403**
- malformed proof (`order` not a string array) → **403**
- block missing its `correctOrder` array (unjudgeable) → **503**

Registered in the `GRADERS` map (`satisfies Record<GradedBlockType, Grader>`), so the completion gate dispatches it generically and stays fail-closed by construction — a graded type with no grader is a compile error, and the route already denies unknown types. Adds a `parsons_failed` deny reason + localized copy. Client shows instant feedback (`correctOrder` ships to the client per the D4 open-book ruling, like quiz `correct` flags); the server re-grades at completion and stays authoritative.

### 3. Renderer (`parsons-block.tsx`)
Accessible two-panel arrange UI — an "Available lines" bank and an ordered "Your solution" list. **Keyboard-operable, never drag-only**: real `<button>`s for Add / Remove / Move-up / Move-down (move buttons disabled at the ends), each with a line-identifying `aria-label`. Per-attempt Check gives an instant client verdict (mirror of the server rule) with red-highlight on out-of-place slots (F13); the verdict + `explanation` render in an always-mounted `aria-live="polite"` region. Deterministic seeded shuffle of the pool (keyed off block key + line id) — no hydration mismatch, no re-scramble across renders. Registered in `RENDERERS` (`satisfies Record<BlockType, Renderer>`).

### a11y approach
Buttons for every action (drag is not required for any operation); ends-disabled move controls; per-control `aria-label` naming the target line; `aria-live` verdict announced without unmounting anything from under the keyboard user; monospace `pre` preserves indentation as a visual guide.

## Deferred / out of scope
- **LX-C8 (entry-rung-by-segment)** — deferred per the issue; meaningless before rung variants exist. Seam: rung/variant selection would be read in the lesson page once `segment` (LX-A3) and authored variants land.
- **Indent-sensitive grading** — indentation is presentational this version; order is the only graded dimension. `content` leading-whitespace renders as an indent guide; a follow-up can promote it to a graded dimension (proof would carry per-line indent).

## Verification
- `pnpm typecheck` — 6/6 packages pass.
- Full `pnpm --filter @superteam-lms/web test` (vitest) — **159 files / 1414 tests pass**, incl. new grader (7), renderer a11y+reorder (10), and the project golden fixtures (unchanged — parsons fields surface conditionally).
- `content-schema` — **156 tests pass** (schema refinements + registry + regenerated-schema byte-identity). `content-lint` — **59 tests pass**.
- `pnpm lint` — exit 0. `prettier --check` — clean.


thomgabriel:
**Orchestrator routing record — SENSITIVE (new grader on the XP-gate route) → adversarial review → `needs-gate`.** Independent verification: full suite **1414/1414** + content-schema **156/156** re-run first-hand. **Adversarial review: PASS, no must-fix.** Proven under attack: the route change is a 2-line reason-mapper insert with everything else byte-identical (#631/#647/#690 semantics intact); the grader is positional sequence-equality (duplicate-id, trailing-id, cross-lesson-id, empty, and malformed proofs all 403; the function is provably total — cannot throw — with the route's fail-closed 500 as backstop); the 503 branch is a real bundle-drift guard, unreachable in a schema-validated bundle; the refinements make pathological content unconstructable (cycles impossible, permutation-exact, distractor-only pairing); D4 open-book consistency deliberate with server-authoritative minting; the seeded shuffle is deterministic FNV-1a and the solution area starts EMPTY (no initial-state leak); golden fixtures untouched via conditional projection. **Optional (not blocking, deliberately deferred):** a route-level integration test for the `parsons_failed` deny path — the wiring is compile-enforced (typed exhaustive Record), and #570 is actively editing the same route-test files, so adding it now would collide; it can ride #570's PR or a follow-up. LX-C8 segment variants deferred per the issue; no content authored (capability only).

thomgabriel:
## Gate: PASS — merging, both should-fix items filed

Review is current (17:52:39Z vs head 17:50:59Z). I traced the trust boundary independently before reading it, and we agree — with one detail the review established better than I did.

### "Server-authoritative" is accurate, and worth stating precisely

The grader compares the learner's submitted order against `correctOrder` **read from the block**, and the block comes from the server:

```ts
for (const block of gradedLesson.blocks) { … await grader(block, proofs[block.key]) }
```

`gradedLesson` is server-resolved; the client supplies only `proofs[block.key]`. So a learner cannot forge `correctOrder` and cannot claim a pass — they must submit a matching sequence. That is the property that matters.

**It does not mean the answer is secret.** `correctOrder` ships in the client projection (`project.ts` surfaces it conditionally), so a learner can read it. That is the deliberate D4 open-book precedent already set by quiz `correct` flags and challenge solutions — not a new exposure. Flagging the distinction only because "server-authoritative grader" can be misread as "answer is hidden"; the two are independent, and only the first is claimed.

Fail-closed behaviour is genuinely layered:
- unknown block type → **503 denied**, not silently skipped
- missing grader for a graded type → **503**, an independent second path
- missing `correctOrder` → **503** (unjudgeable, not "wrong")
- wrong order / included distractor / missing line / length mismatch / malformed proof → **403**

The review's better detail: `GRADERS` is `satisfies Record<GradedBlockType, Grader>`, so an unregistered graded type is a **compile error**, not a runtime gap. That is stronger than the runtime fail-closed path I had verified, and it is the right way to build it.

Also confirmed the schema pins `correctOrder` to an exact permutation of non-distractor line ids, keyed to stable ids rather than array index — the same invariant `QuizOption` holds, and the reason a shuffled pool can't desync from the key. All three locales carry identical `parsons` key sets.

### Both should-fix items confirmed against the code

**1. `distractor` / `pairedWith` are modelled and never rendered.** I grepped the renderer: it reads **neither** field. Meanwhile the schema not only models them, its comment states the purpose outright —

> `pairedWith` … so the renderer can pair them (F13 "paired-distractor markers"). Presentational; not graded.

— and the PR description credits F13 paired-distractor markers as delivered. Distractors are functionally live (correctly graded wrong when picked) but visually identical to real lines and unconnected to the line each mimics.

Zero grading impact (`pairedWith` is explicitly not graded), so not blocking. But it is a description crediting a capability the diff provisions and then doesn't use — the same class as #690's `Closes` overclaim and #10's coverage table reading as complete. The schema field is the tell: you don't add and validate a presentational field unless something was meant to render it.

**2. `parsons_failed` has no test** while `deny-reasons.test.ts` exhaustively covers `quiz_failed`, `challenge_failed`, `reflection_required`, `enrollment_missing` and the no-reason 503. A future edit collapsing the new three-way ternary back to two would misreport a failed parsons block as `challenge_failed` — wrong localized copy via `completion-error.ts` — with nothing to catch it. Cheap to add, and the surrounding file's exhaustiveness is exactly what makes the gap conspicuous.

Filed together as **#702**. Merging because the grading path — the part that could have been dangerous — is correct, fail-closed, and compile-time-enforced.

Issue #699, opened 2026-07-26
db: schema.sql drifted from prod in both directions (retired profiles.role present, live wallet trigger absent) — makes a guard test hollow


thomgabriel:
Found while applying #695's migration and probing prod. **`supabase/schema.sql` is drifted from prod in both directions at once**, and one of the gaps makes a guard test hollow.

#697 fixes one half. This is the other half plus the consequence.

## Direction 1 — schema.sql defines things prod does not have

Live check against `pywhtmidcrptomrabbrw`:

```
profiles columns: id, wallet_address, google_id, github_id, username, bio, avatar_url,
                  social_links, is_public, name_rerolls_used, wallet_xp_synced_at,
                  created_at, deleted_at, deletion_requested_at, segment, goal, daily_goal

pg_proc 'enforce_profile_role_write' → 0
```

There is **no `role` column** and **no `enforce_profile_role_write` function**. `schema.sql` still carries both (6 references).

**Prod is correct here.** `role` was deliberately retired — the sole surviving reference in app code says so:

`apps/web/src/app/api/teacher/courses/[id]/stats/route.ts:20`
> `SP1 PR-2 Task 7: replaces the retired profiles.role / course.author`

No app code selects or writes `profiles.role`, so nothing is broken at runtime. The snapshot is simply stale.

## Direction 2 — prod has a trigger schema.sql does not (this is #697)

The only trigger on `profiles` is **`trg_enforce_profile_wallet_write`**, and `schema.sql` has **zero** references to it. It is live and effective — verified by attempting the write as an impersonated authenticated user:

```
update profiles set wallet_address = '<forged>' where id = <own id>
→ blocked: permission denied: wallet_address may only be changed by service…
```

## The consequence — a green test guarding nothing

`apps/web/src/lib/supabase/__tests__/segment-state-rls-guard.test.ts` (added by #695) asserts:

> `it("schema.sql still defines the role trigger exactly once (mirror intact)")`

That passes, and it protects nothing: the trigger it pins is not deployed. #695's migration header also cites that trigger as the reason whole-row self-write is safe:

> "The escalation-lockdown trigger is column-specific to `role` and is provably untouched"

[...]

PR #702, opened 2026-07-26
feat: solution-reveal soft-gate + if-then plan + experiment registry

### LINKED ISSUE #582: Solution-reveal soft-gate + if-then prompt + experiment registry
Three small P2s bundled for triage.

**Spec tasks:** `LX-C6, LX-A6, LX-F4`

thomgabriel:
Closes #582.

Three small bundled P2s from the launch-experience master spec §3 (tasks LX-C6, LX-A6, LX-F4). Each composes with what is already on `main`; no existing behavior changes except the additive surfaces below.

## LX-C6 — "View reference solution" soft-gate
The reference solution already ships silently in the lesson payload (`CodeBlockData.solution`, resolved by the content projector) with no UI. This adds a deliberate, gated reveal:
- A low-emphasis **View reference solution** control in the challenge toolbar (only when a solution exists and the lesson isn't already complete).
- Clicking it opens a **confirm nudge**, not the solution — a soft-gate, never a refusal (one-tap override, AIE-27). The learner can back out ("Keep trying") or confirm ("Show solution").
- Confirming: reveals the read-only solution, fires `solution_revealed` (session-deduped, lean payload — never the solution text), and **best-effort reschedules the lesson into the LX-B3 review queue** via a new `POST /api/lessons/reveal-solution` → `schedule_review_item` RPC.
- **XP path untouched**: xpPerLesson is fixed on-chain, so the only consequence is review-scheduling + framing (unified-spec zero-on-chain edge #1). No Codewars-style permanent forfeit.

## LX-A6 — session-end if-then plan
A "when's your next lesson?" implementation-intention prompt (day + time picker) on the dashboard:
- Stored in a **new `profiles.prefs` JSONB** column (`{ nextLesson: { day, time } }`), written self-service via the existing own-row profiles UPDATE RLS policy (no new policy, no service role). Merges into any other prefs keys. **No profiles column is privilege-bearing** (prefs is non-PII UI state, not in `public_profiles`), so the write is bounded by two CHECK constraints — `chk_profiles_prefs_object` (`jsonb_typeof = 'object'`) and `chk_profiles_prefs_size` (`pg_column_size <= 2048`) — which are the SOLE bound on this self-write, not a second line behind a trigger.
- New additive dashboard slot `NextLessonPlan` (LX-B1 split has landed): shows the picker, then the committed if-then sentence back on return visits with a "Change" affordance.
- **v1 is display-only** — no notification channel exists, so nothing delivers a reminder (no notification code shipped). Committing fires `next_lesson_plan_committed` (closed-set weekday id only, never the exact time/PII). Its effect on return is a **pre-registered NULL** in the experiment registry (S4).

## LX-F4 — experiment registry
Grows the #648 substrate into the full report experiment set (extends the existing module; no duplication):
- Adds the missing rows: retrieval on/off, anonymous trial, test-out visibility, nudge threshold, weekly cadence, cohort vs global, and the **challenge-first pilot scoped to one course**. (continue-hero / LinkedIn nudge / endowed first-tick / celebration tiering were already rows.)
- Adds a `preregisteredNull` flag and marks the **S4 planning-prompt row** as a pre-registered null.
- Carries **E6 (incentive-graded XP ladder) as `infeasible-as-designed`** so it is not re-proposed (XP can't vary by rung/level without a program change; see LX-C7 justified drop). The ≥10-week window is already encoded via `EVALUATION_WINDOW_WEEKS`.

## Scope / composition notes
- `review_items` substrate (LX-B3) exists on `main` but had **zero application callers** — this wires the first one (`lib/review/schedule-review.ts`). If a `/review` surface later reads the queue, revealed lessons are already seeded.
- `/start` intake (#695/#566) is **not on main**; per spec, LX-A6 v1 lives entirely on the dashboard, independent of `/start`. The reserved `onboarding_goal_committed` daily-goal event is the future composition point.
- All ×3 locales (en / pt-BR / es), a11y (labeled controls, focus-visible rings, `role="region"` disclosure, `aria-expanded`/`aria-controls`).

## Verification
- `pnpm --filter web test` — **1365 passed** (152 files), incl. new event/registry/soft-gate/plan tests
- `pnpm lint` — green (only pre-existing warnings in untouched files)
- Prettier clean on all changed files
- Migration `20260726160000_add_profiles_prefs.sql` is idempotent (pg_constraint-guarded CHECK adds) with a tested ROLLBACK block; mirrored into `supabase/schema.sql`, and pinned by a guard test (`profiles-prefs-bound-guard`) that asserts the shape+size bound holds in both the migration and the snapshot


thomgabriel:
**Merge-order note:** this PR and #706 BOTH create `apps/web/src/lib/review/schedule-review.ts` (same purpose). Second-to-merge rebases and unifies to one implementation.

thomgabriel:
**Orchestrator routing record — SENSITIVE (`profiles` migration + new authed route calling a service_role RPC) → adversarial review → `needs-gate`.** Independent verification: full suite run first-hand at three heads — **1450/1450** post-rebase, **1460/1460** at the fix head `9aae6ca`. **Adversarial review: PASS with one must-fix, now landed.** Cleared under attack: the reveal route is authenticated, session-scoped (user.id never from the body — now test-asserted), bundle-validated-before-RPC (404 → no call), and idempotent; **the solution-exposure question is RESOLVED — no new exposure on any seam** (the RSC payload has embedded solutions since the SP2 flip; the fetch wrappers are what's server-only; the soft-gate is honest UX, not a security boundary); migration 160000 unique, idempotent, transactional, mirrored, rollback-tested; the double rebase dropped nothing (#695 segment + #700 review machinery verified intact); registry rows match the spec's rulings (S4 preregistered null, E6 infeasible-as-designed). **Must-fix landed in `9aae6ca`:** the reveal copy no longer overclaims — 'Marked for review' renders ONLY on a confirmed 2xx+scheduled; anonymous/failed reschedules get the neutral copy (AIE-27 honesty, ×3 locales, 401-path test). Plus the previously-missing route tests (6) and wrapper never-throws tests (3). **Advisory items accepted on record:** no rate limit on the POST (RPC idempotency bounds harm), prefs JSONB unbounded (social_links precedent), no enrollment check (self-scoped, bundle-bounded). **Merge-order:** the `lib/review/schedule-review.ts` add/add with #706 — second-to-merge UNIONS both exports (confirmed semantically compatible). Migration is dry-run-first yours.

thomgabriel:
## Gate: CHANGES REQUESTED

Cannot merge — `mergeable: CONFLICTING`. Two substantive findings alongside the rebase, so this is one round-trip rather than two.

### Blocking 1 — rebase (6 files, all from PRs merged while this was open)

Merge-base `6a92a28d`, main now `f2ad979`. Overlapping files:

```
apps/web/src/lib/review/schedule-review.ts          ← #706 (failure-capture) landed here
apps/web/src/lib/review/__tests__/schedule-review.test.ts
apps/web/src/messages/{en,es,pt-BR}.json            ← #700, #701, #707 all added keys
supabase/schema.sql                                 ← #695, #697, #707
```

`schedule-review.ts` is the one to be careful with: #706 added `captureReviewFailure` to that file, and this PR adds `scheduleReviewItem`. Both belong — keep both exports. The locale files and `schema.sql` are additive on every side.

### Blocking 2 — `prefs` is an unbounded self-writable JSONB

The migration is:

```sql
ALTER TABLE profiles ADD COLUMN IF NOT EXISTS prefs JSONB NOT NULL DEFAULT '{}';
```

No CHECK, no size bound, no shape constraint — and it is written self-service through the existing own-row `profiles` UPDATE policy. Any authenticated user can store arbitrary JSON of arbitrary size in their own row, repeatedly.

The standard here is your own, from #695 about **an integer**:

> Bounded on BOTH ends: a daily lesson goal above ~20 is not a real target, and **an unbounded self-writable integer is a needless abuse surface**.

An unbounded self-writable JSONB is strictly a larger version of that surface. The intended payload is `{ nextLesson: { day, time } }` — tens of bytes.

Good news on blast radius: I checked, and `prefs` is **not** in `public_profiles` (which exposes only `wallet_address, username, avatar_url, bio, social_links`), so this is storage/cost, not user-controlled content served to others. That is the difference between "bound it" and "block it", and it is why this is a bound-it.

Please cap it. A size CHECK is the cheap version; if you want shape too, `jsonb_typeof(prefs) = 'object'` plus a size cap covers the realistic abuse without trying to whitelist keys in SQL. A `BEFORE UPDATE` trigger is the certain path if a CHECK expression gives you immutability grief on the jsonb→text cast.

### Blocking 3 — the PR body repeats a premise that is false in production

> "written self-service via the existing profiles UPDATE RLS policy (**the role trigger guards only `role`**; no new policy, no service r…)"

**There is no role trigger in production, and no `role` column.** Verified live: `profiles` has no `role`, and `pg_proc` has zero `enforce_profile_role_write`. It was dropped by migration `20260711152518_drop_teacher_role_lock_wallet_writes`, and the only surviving reference in app code calls it "the retired `profiles.role`". `schema.sql` still defining it is the drift I filed as #699.

This is the **third** place the premise has propagated — #695's migration header, #695's and #697's guard tests, and now this body. Each time it is cited as the reason a widened self-write surface is safe.

The safety conclusion is still fine, but the actual reason is different and it matters here: nothing on `profiles` is privilege-bearing, so a whole-row self-write is bounded **only by the column constraints you write**. Which is precisely why finding 2 is blocking rather than cosmetic — with no role trigger, the CHECK constraints are not a second line of defence, they are the only one.

Please restate it as "no column on `profiles` is privilege-bearing; the write is bounded by CHECK constraints", and add the bound.

### Not blocking — the rest is well built

- **Reveal-solution route**: user-scoped client for auth only, then `scheduleReviewItem` → `createAdminClient().rpc(...)`, so the service_role-only RPC is reached correctly. `user.id` comes from the session and the lesson is resolved server-side from slugs, so a client cannot enqueue an arbitrary `item_key` — the same forgery class #701 handles, handled the same way. Slug lengths bounded.
- **Soft-gate framing is right.** The solution already ships in the payload post-D4, so a reveal control is friction and honesty, not a security boundary — and the PR treats it that way (one-tap override, no permanent forfeit, XP untouched because `xpPerLesson` is fixed on-chain). Rescheduling the lesson into the review queue instead of penalising is the better lever.

thomgabriel:
## Gate: PASS — all three blocking findings resolved. Merging.

CI green, MERGEABLE/CLEAN on `d66d5641`.

**1. Rebase — done.** The six-file overlap with #706/#695/#697/#707 is resolved and `schedule-review.ts` retains both `captureReviewFailure` (#706) and this PR's `scheduleReviewItem`.

**2. `prefs` is bounded — and I verified the bound actually bites, not just that it exists.** Two CHECKs now: `chk_profiles_prefs_object` (`jsonb_typeof(prefs) = 'object'`) and `chk_profiles_prefs_size` (`pg_column_size(prefs) <= 2048`).

`pg_column_size` is STABLE rather than IMMUTABLE, so a CHECK using it can be rejected at creation depending on the server — a migration that will not apply is worse than a loose one. Tested on prod in a rolled-back transaction:

| attempt | result |
|---|---|
| create both CHECKs | **both created** |
| `{"nextLesson":{"day":"mon","time":"19:00"}}` (the intended payload) | **accepted** |
| 8,540-byte JSONB | **rejected** |
| `[1,2,3]` (array, not object) | **rejected** |
| `"just a string"` (scalar) | **rejected** |

So the abuse surface is genuinely closed, and the intended payload still fits with room to spare.

**3. The false premise is corrected, precisely.** The body and the migration header no longer cite a role trigger. They now say:

> **No profiles column is privilege-bearing** … so the write is bounded by two CHECK constraints … which are the **SOLE bound** on this self-write, **not a second line behind a trigger**.

That is the accurate statement, and it is the reason finding 2 was blocking rather than cosmetic: with no trigger behind them, those CHECKs are the only thing standing between a self-writable column and arbitrary content. Getting the reasoning right is what made the bound obviously necessary rather than optional.

Worth noting this is the **third** place that premise had propagated (#695's migration header, #695+#697's guard tests, this body) — #699 tracks removing it from the two tests, and the retired `role` machinery from `schema.sql`.

**Also fixed, unprompted:** `d9c31d03` — "honest reveal copy gated on confirmed reschedule". Good instinct: telling the learner their lesson was rescheduled for review before the reschedule is confirmed would be a claim the system cannot back.

The parts I passed on round 1 are unchanged and still right: the reveal route's client split (user-scoped for auth, `createAdminClient()` inside the helper for the service_role RPC), server-side lesson resolution so a client cannot enqueue an arbitrary `item_key`, and the soft-gate framing — post-D4 the solution already ships in the payload, so this is friction and honesty rather than a security boundary, with rescheduling instead of a forfeit.

**Migration not yet applied to prod.** It is additive and now bounded; it can go in the next apply batch alongside #707's `review_quest_kind`, which is also merged-but-unapplied.

thomgabriel:
**Orchestrator: all 3 gate blockers addressed at `d66d564` — independently verified, re-queuing for gate re-review.**

1. **Rebase**: branch now based on main tip `cd094ab`; `lib/review/schedule-review.ts` unions BOTH exports — #706's `captureReviewFailure` (completion-deny capture) and this PR's `scheduleReviewItem` (reveal path) over the same idempotent `schedule_review_item` RPC; unified test file, 8 tests.
2. **Prefs bound**: `chk_profiles_prefs_object` (`jsonb_typeof='object'`) + `chk_profiles_prefs_size` (`pg_column_size ≤ 2048`), idempotent pg_constraint-guarded, byte-mirrored into schema.sql, pinned by an 11-test guard (`profiles-prefs-bound-guard`). CHECK chosen over trigger deliberately — `pg_column_size` needs no immutability in a CHECK; trigger fallback documented in-header.
3. **False premise corrected** in migration header, schema.sql comment, and the PR body: no profiles column is privilege-bearing in prod (role column dropped by 20260711152518), so these CHECKs are the SOLE bound on the self-write — stated as such, no phantom second line of defense. schema.sql's role-trigger snapshot drift intentionally untouched (that is #699's scope; #566 guard still asserts it).

**My independent verification at `d66d564`** (not the implementer's claim — re-run in a clean clone): `pnpm typecheck` exit 0; full web suite **177 files / 1561 tests, exit 0**; prefs CHECKs confirmed present in BOTH `20260726160000_add_profiles_prefs.sql` and `schema.sql` by direct inspection. Caveat carried honestly from the implementer: no live SQL apply anywhere (SQL-only per standing policy; owner dry-runs).

Note: main has since advanced to `04b3d260` (#714, content-bump — content.lock/generated/goldens only, zero file overlap with this diff).

@claude re-review the current head please.

Issue #703, opened 2026-07-26
content: parsons distractor/pairedWith modelled but never rendered (F13 unshipped) + parsons_failed untested


thomgabriel:
Two follow-ups from gating #701 (merged). Neither affects grading — the server-authoritative path is correct, fail-closed and compile-time-enforced.

## 1. `distractor` / `pairedWith` are modelled, validated, and never rendered

`packages/content-schema/src/blocks/parsons.ts` defines both, and states the purpose in its own comment:

```ts
/** A line that does NOT belong in the solution (F13 paired-distractor). */
distractor: z.boolean().default(false),
/** … so the renderer can pair them (F13 "paired-distractor markers").
    Presentational; not graded. */
pairedWith: z.string().min(1).optional(),
```

`parsons-block.tsx` reads **neither field** — verified by grep. `bankOrder` shuffles `lines` unfiltered and every bank item renders through identical markup, so a distractor is:

- **functionally live** — picking one produces a non-matching sequence and is correctly graded wrong, and
- **visually indistinguishable** from a real line, with no connection to the line it mimics.

#701's description credits "F13 paired-distractor markers" as delivered. It isn't. Zero grading impact (`pairedWith` is explicitly not graded), so this is a UX gap plus a description overclaim rather than a defect.

**Why it's worth doing rather than dropping the fields:** the whole point of a *paired* distractor is that the learner sees two plausible neighbours and has to discriminate — that discrimination is the learning, and it's what distinguishes a Parsons problem from "drag the lines into the only order that compiles". Without the marker the pairing is invisible, so the pedagogy the field was added for doesn't happen. The alternative — deleting both fields — is worse: it removes the capability and silently narrows what authors can express, right as C1/C4 are about to use `parsons-completion` as a ladder rung.

**Fix:** give a `distractor` line a visual treatment in the bank, and connect it to its `pairedWith` target. Add a renderer test asserting the marker appears. Keep it accessible — the marker must not be colour-only, since the existing renderer was built for keyboard/AT use.

## 2. `parsons_failed` deny-reason has no test

`route.ts` now maps:

```ts
type === "quiz" ? "quiz_failed" : type === "parsons" ? "parsons_failed" : "challenge_failed"
```

`apps/web/src/app/api/lessons/complete/__tests__/deny-reasons.test.ts` exhaustively covers `quiz_failed`, `challenge_failed`, `reflection_required`, `enrollment_missing` and the no-reason 503 case — but has no `parsons_failed` case.

[...]

Issue #693, opened 2026-07-26
lessons/complete: concurrent duplicates still each run the grading suite (residual of #651)


thomgabriel:
Residual from #690 (merged), which closed the reverted-tx-fee half of #651 but not the grading half.

## What #651 asked for vs what shipped

#651 named two harms from the non-atomic read-then-submit in `/api/lessons/complete`:

> one confirms, the rest revert on the program double-complete guard — but the backend keypair pays the base fee **and the code grader runs for all N**

and prescribed a guard **"before grading/submission"**.

#690 placed the in-flight guard **after** grading, immediately before `onChainCompleteLesson`. **That placement is correct and should not be changed** — guarding before grading would throttle a learner's legitimate quiz/challenge *re-attempts* to one per 30s window and make the "already being processed" copy false. That is a worse bug than the one being fixed.

The consequence is that N concurrent same-lesson requests still each run the full server grading suite before one of them takes the token. Only the on-chain submission is deduplicated.

## Why this is much smaller than the original

- Cost is **server CPU**, not SOL — no reverted-tx fees, which was the expensive half.
- Already bounded by the existing volume gates: 40/hr per user and 1200/hr per IP.
- No correctness exposure — the program's double-complete guard is unaffected, and #690's exploit surface was verified closed.

So this is a cost/efficiency follow-up, not a security or correctness item.

## Possible shapes (deliberately not prescribing one)

The tension to respect: **any dedup that sits before grading must not block a legitimate re-attempt of a failed submission.** That is what rules out the naive version.

1. **Dedup on submission identity, not on (user, lesson).** Key the guard on a hash of the graded payload (the answers/code actually submitted). A genuine re-attempt carries *different* proofs and therefore a different key, so it is never blocked; a true duplicate carries identical proofs and short-circuits. This distinguishes "the same request twice" from "a second, different attempt", which is the actual requirement.
2. **Cache the grading result** per (user, lesson, payload hash) for a short TTL and serve concurrent duplicates from it, so the suite runs once and the rest are cheap. Also improves the honest-retry path.
3. **Accept it.** Given the bounds above, documenting the residual and closing may be the right economic answer.

Option 1 or 2 also removes the need for the guard to be a *throttle* at all — the current 1-token/30s window is a coarse lock, and the code comment is explicit that two requests straddling a window boundary can both pass.

## Also worth folding in: the in-flight token is never released

From gating #690, non-blocking:

[...]

PR #706, opened 2026-07-26
feat(review): failure-capture feed on the completion deny path (LX-B4)

### LINKED ISSUE #570: Review spine 2/3: failure-capture feed

thomgabriel:
Closes #570 — **Review spine 2/3: failure-capture feed** (LX-B4). Builds on #689's `review_items` substrate.

## What this does
Stops discarding per-test grading results and turns a **failed completion** into a spaced-repetition review item, keyed by lesson `_id`.

### 1. Plumb `ServerTestResult[]` through `fromRun` (`graders/code.ts`)
`fromRun` collapsed every run to `{ok,status}`, throwing away `run.results`. Now a `403` carries the **failing** `ServerTestResult[]` in `GradeResult.failedTests` (omitted — not `[]` — when the run produced no per-test rows, e.g. a degenerate rust/build verdict, so the fail shape stays `{ok,status}`). `failedTests` never gates completion; `status` alone does.

### 2. Capture on the deny path (`/api/lessons/complete`)
On a **graded miss only** — `quiz_failed` / `challenge_failed` (403) — the route best-effort enqueues the lesson into the review substrate. Explicitly **not** captured:
- **503** (executor outage / could-not-judge) — not a learning signal.
- **enrollment_missing**, **reflection_required** — grading already passed / not a graded miss; these say nothing about what the learner knows.

Capture is isolated two ways: `captureReviewFailure` never throws by contract (swallows + logs internally), **and** the call site adds `.catch(() => {})`. A review-write hiccup can never change the deny response the learner sees.

### 3. Cap-before-feed primitive (`lib/review/due-items.ts`)
Per #570's pinned hard constraint (from #689's adversarial review): the #569 seed floods box-1 items due-tomorrow for the whole cohort, so the first learner-facing read must be capped. `getDueReviewItems` is the **one** read path for every review surface; it clamps any requested limit to `REVIEW_SESSION_CAP = 6` (upper bound of the spec's "3–6 item" review strip). #571's /review page and the dashboard strip both go through it, so **neither can ship an uncapped read**. No caller in this PR yet — it's the shared, tested primitive #571 consumes.

Box math stays entirely in #689's `schedule_review_item` / `record_review_result` RPCs — not re-implemented here.

## Verification
- `pnpm typecheck` (turbo, 6/6 packages) — clean
- `pnpm test` (apps/web) — **1433 passed**; the real-QuickJS integration tests (`gradeCode (executor present)`) ran (not skipped), proving `failedTests` flows end-to-end from the actual executor through `fromRun`
- `pnpm lint` — clean (only pre-existing import/order warnings in untouched files)

New tests: `fromRun` plumbing (unit + real-executor); deny-path capture fires on qualifying denials only + best-effort isolation; `captureReviewFailure` never-throws contract; `getDueReviewItems` cap clamping.

## Out of scope / notes
- No new caller for `getDueReviewItems` — #571 (/review) and the dashboard strip consume it.
- `/api/lessons/validate-challenge` remains a verified orphan; wiring it is optional per the spec, not done.
- v1 feeds on **submit-failure only** — practice "Run" stays client-side; no new rate-limit exposure.

### DIFF
--- a/apps/web/src/app/api/lessons/complete/__tests__/gate.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/gate.test.ts
@@ -57,6 +57,10 @@ vi.mock("@/lib/grading/graders", () => ({
   GRADERS: { code: () => codeGrader() },
 }));
 
+vi.mock("@/lib/review/schedule-review", () => ({
+  captureReviewFailure: vi.fn().mockResolvedValue(undefined),
+}));
+
 vi.mock("@/lib/ai/check-seal", () => ({
   openAttestation: () => openAttestation(),
 }));
--- a/apps/web/src/app/api/lessons/complete/__tests__/inflight-guard.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/inflight-guard.test.ts
@@ -65,6 +65,9 @@ vi.mock("@/lib/content/queries", () => ({
   getCourseById,
 }));
 vi.mock("@/lib/grading/graders", () => ({ GRADERS: {} }));
+vi.mock("@/lib/review/schedule-review", () => ({
+  captureReviewFailure: vi.fn().mockResolvedValue(undefined),
+}));
 vi.mock("@/lib/ai/check-seal", () => ({ openAttestation: () => true }));
 vi.mock("@/lib/solana/academy-program", () => ({
   isOnChainProgramLive,
--- a/apps/web/src/app/api/lessons/complete/__tests__/review-capture.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/review-capture.test.ts
@@ -1,9 +1,9 @@
 /* eslint-disable import/order -- vi.mock calls must precede importing the route. */
-// The 403 `reason` discriminator (#564): the route's 403 bodies must carry a
-// machine string the client can map to the RIGHT localized copy — without it a
-// failed quiz rendered the "verify enrollment" message. Companion to
-// gate.test.ts, with BOTH graders registered (gate.test.ts deliberately omits
-// `quiz` to prove the no-grader 503 path).
+// LX-B4: a GRADED miss on the completion route's deny path is captured into the
+// review substrate; a "could-not-judge" 503 and the non-learning denials
+// (enrollment / reflection) are NOT. Capture is best-effort — a capture failure
+// must never change the deny response. Shares the mock harness with
+// deny-reasons.test.ts.
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { NextRequest } from "next/server";
 import type { GradeResult } from "@/lib/grading/types";
@@ -22,6 +22,7 @@ const {
   isRateLimited,
   isCourseInMaintenance,
   isPlatformFrozen,
+  captureReviewFailure,
 } = vi.hoisted(() => ({
   getUser: vi.fn<() => Promise<unknown>>(),
   singleProfile: vi.fn<() => Promise<unknown>>(),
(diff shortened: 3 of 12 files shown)


thomgabriel:
**Merge-order note:** this PR and #702 BOTH create `apps/web/src/lib/review/schedule-review.ts` (same purpose: admin-client wrapper for the `schedule_review_item` RPC). Whichever merges second must rebase and unify to ONE implementation (they should be semantically identical — the adversarial review is confirming). Flagged on both PRs.

thomgabriel:
**Orchestrator routing record — SENSITIVE (completion-route deny path) → adversarial review → `needs-gate`.** Independent verification: full suite **1433/1433** re-run first-hand. **Adversarial review: PASS, no must-fix.** Proven under hostile trace: capture fires ONLY on graded-403s (every other deny site traced clean — 503s/reflection/enrollment/429/409 untouched), the deny response shape is byte-identical, and the double best-effort guard is throw-tested; farming is structurally impossible (PK + ON CONFLICT DO NOTHING; box math untouchable from capture; raw-`_id` keying means a fail in lesson A can only ever touch lesson A); `failedTests` never reaches the DB or the client (ids-only logging — no hidden-test leak, no payload blowup); the ≤6 cap primitive is the table's ONLY reader and every surface (#571) must consume it; the QuickJS integration tests genuinely execute (not mocked). **Two notes for the gate:** (1) the graded-miss deny now AWAITS one admin RPC before returning — durability-over-latency, the right serverless call, but a hung RPC would ride the platform timeout (optional hardening: a timeout wrapper); (2) the `lib/review/schedule-review.ts` add/add conflict with #702 is CONFIRMED semantically compatible but has distinct export names — whoever rebases second must **UNION both exports, not pick one**. Pre-existing inherited note: the executor suite's skipIf pattern would silently skip (not fail) on a WASM-load regression — a latent CI blind spot worth a future issue.

thomgabriel:
## Gate: PASS — merging

Review spine 2/3, on the substrate I applied and exploit-verified for #689, so the contract was concrete to check against.

### The thing that would have made this silently dead

`schedule_review_item` is `REVOKE`d from `PUBLIC, anon, authenticated` and `GRANT`ed only to `service_role` — I verified that live on prod (an impersonated learner gets `permission denied for function`). Combined with capture being best-effort and **swallowing every error**, a user-scoped client here would have produced a feature that appears to work, logs quietly, and writes nothing. Ever.

It uses `createAdminClient()`. Correct, and this is the one place where "best-effort + swallowed" and "wrong client" would have compounded into an invisible no-op.

### Scoping of what counts as a learning signal is right

Capture fires on `result.status === 403` inside the graded-block loop only. Deliberately excluded, and each exclusion is correct:

- **503** — could-not-judge (executor outage). Not a miss. Capturing it would poison the queue with items the learner may know perfectly well.
- **enrollment_missing** / **reflection_required** — grading either already passed or never ran; neither says anything about what the learner knows.

Idempotency is inherited rather than reimplemented: `schedule_review_item` is `ON CONFLICT DO NOTHING`, so re-failing a lesson never resets hard-won box progress, and the box math stays entirely inside the RPCs. That is the right seam — and it means #693's residual (concurrent duplicates each running the grading suite) cannot produce duplicate review rows either.

Double isolation is real: the helper never throws by contract *and* the call site adds `.catch(() => {})`.

### Finding — the capture reason has silently diverged from the deny reason

`route.ts:240` decides the captured reason with a **two-way** ternary:

```ts
reason: type === "quiz" ? "quiz_failed" : "challenge_failed",
```

while the deny response two lines later — from #701, merged ~20 minutes ago — is **three-way**, including `parsons_failed`. `ReviewFailureCapture.reason` is typed `"quiz_failed" | "challenge_failed"`, a closed two-value union, so `parsons` is not merely unhandled; it is unrepresentable.

**Impact is small and I want to be precise about why.** `reason` never reaches the database — `schedule_review_item` takes only `p_user_id` and `p_item_key`, so a failed parsons block **is** captured, with an identical review row. `reason` appears only in the `logError` context, and only when capture *fails*. So the observable consequence is a mislabeled telemetry field on an error path. Not a data defect.

What makes it worth filing is the *shape*: the `else` branch absorbs every future graded type into `challenge_failed`, so the type system actively hides the divergence instead of surfacing it. Contrast #701's own `GRADERS satisfies Record<GradedBlockType, Grader>`, where a missing entry is a **compile error**. Same PR family, opposite failure posture.

This is also a near-miss confirmation of #703 item 2, which I filed 20 minutes ago warning that "a future edit collapsing this back to a two-way ternary would silently misreport a failed parsons block." The next PR did write a two-way ternary — on the capture path rather than the deny path, and harmlessly — but the mechanism predicted is exactly the mechanism that appeared.

And #706 **modified `deny-reasons.test.ts`** without adding a `parsons_failed` case (still 0 mentions), so the guard that would have flagged this still does not exist. Updating #703 rather than filing a duplicate.

Merging: review rows are correct for every graded type including parsons, the deny response is correct, and nothing here can change a completion outcome.

PR #2, opened 2026-02-12
feat: Superteam Academy LMS frontend (healer-sol-learning)

yeshealer:
- Next.js 14 App Router, TypeScript strict, Tailwind, Radix UI
- i18n: PT-BR, ES, EN; light/dark theme; Solana Wallet Adapter (Phantom, Solflare)
- 10 core pages: landing, courses, course detail, lessons, dashboard, profile, leaderboard, settings, certificates
- LearningProgressService interface + stub (localStorage) for progress/XP/streak/leaderboard
- Course enroll flow: Enroll & start / Continue with progress; lesson Mark complete persists XP
- Docs: README, ARCHITECTURE, CMS_GUIDE, CUSTOMIZATION

[superteam-academy-o4z12tere-yeshealers-projects.vercel.app](https://superteam-academy-o4z12tere-yeshealers-projects.vercel.app/)

PR #715, opened 2026-07-26
ci: enforce + exercise SHA-pin guard, cover composite actions (#662, #681)

### LINKED ISSUE #662: CI: reject non-SHA-pinned GitHub Actions (the guard half of #635)
#659 pinned all six remaining mutable actions — verified: **0 tag-referenced `uses:` left in any of the six workflow files**. It deliberately scoped out #635's *second* acceptance criterion, and `Closes #635` would bury it, so it lives here.

## Why the guard matters more than the pins

#635's own history is the argument: #479 SHA-pinned the actions, then **six usages silently drifted back to mutable tags** and nobody noticed until a dependabot bump exposed one of them while being gated. Pins without enforcement are a state, not a property — they decay on the next PR that adds a `uses:` line.

Two of the six sat in the `Detect changes` job that runs on **every pull request**, so the decayed state meant a repointed upstream tag was arbitrary code execution in CI. That is the failure this guard prevents recurring.

## Options, cheapest first
- **grep step** in the existing CI: fail if any `uses:` in `.github/workflows/` lacks a 40-hex SHA. ~3 lines, no new dependency, catches 100% of the regression class.
- **[zizmor](https://github.com/zizmorcore/zizmor)** or **actionlint** — real workflow linters; more signal (also catches injection patterns, over-broad permissions), one more tool to keep pinned.
- **CODEOWNERS on `.github/workflows`** — social, not mechanical. Weakest, since the last decay happened under review.

Recommend the grep step now and consider zizmor later; the point is that it must be mechanical.

## Also in scope — the last inconsistent pin
`ci.yml:242` (the `build-server-image` job from #634) still pins `actions/checkout@9c091bb2… # v7.0.0` while every other checkout in the file is now `3d3c42e5… # v7.0.1`. Not a security gap — it *is* SHA-pinned — but I flagged this exact straggler when gating #634 (the branch predated #538). One-line bump; do it in the same PR so the guard lands on a fully consistent file.

## Done when
- [ ] A CI step fails any `uses:` in `.github/workflows/` that isn't SHA-pinned
- [ ] The step is proven by a deliberately-unpinned line in a scratch commit (don't ship an unexercised guard — see #614)
- [ ] `ci.yml:242` bumped to the `v7.0.1` SHA

Source: #659's review raised both; #635 verified complete on the pins themselves.

### LINKED ISSUE #681: ci: SHA-pin guard does not cover composite actions (.github/actions/**)
Found while gating #680. Non-blocking, latent — nothing is broken today.

## The gap

`scripts/check-action-pins.sh` scans exactly one glob:

```bash
files=("$workflow_dir"/*.yml "$workflow_dir"/*.yaml)   # .github/workflows only
```

and `action-pin-guard.yml` triggers on exactly two paths:

```yaml
paths:
  - ".github/workflows/**"
  - "scripts/check-action-pins.sh"
```

**Composite actions are outside both.** A composite action at `.github/actions/<name>/action.yml` can itself contain `uses:` steps, and today those would be:

1. **not scanned** — the glob never reaches `.github/actions/**`, and
2. **not even triggered** — adding one touches no path in the filter, so the guard doesn't run at all.

That is a double miss: an unpinned third-party action could enter the repo through a composite action with the guard green.

## Why it is not blocking

`.github/` currently contains only `workflows/` — there are zero composite actions, so the scan surface is complete as of #680 and the guard is fully effective today. I verified #680's detection is otherwise exact: mutable tags, branch refs, missing refs, short SHAs, 40-char non-hex, quoted refs and mutable reusable-workflow refs are all caught; comments, `run:`-block prose, `./` local refs and `docker://` refs correctly pass; and the SHA length boundary is exact (39 rejected, 40 accepted, 41 rejected).

This is only about the day someone adds the first composite action — at which point the guard silently stops covering the new surface rather than failing loudly.

## Fix

Two lines, both needed (either alone leaves the hole):

1. Extend the scan to composite actions:
   ```bash
   files+=("$repo_root"/.github/actions/*/action.yml "$repo_root"/.github/actions/*/action.yaml)
   ```
2. Extend the trigger so adding one actually runs the guard:
   ```yaml
   paths:
     - ".github/workflows/**"
     - ".github/actions/**"
     - "scripts/check-action-pins.sh"
   ```

Note the script's existing "no workflow files found → error" guard is a good pattern to preserve: the scan should fail loudly if its globs ever match nothing, rather than pass vacuously.

Related: #661 (the guard), #635 / #659 (the pinning this protects), #662.

thomgabriel:
Closes #662. Closes #681.

## What

Completes the SHA-pin guard so it is **enforcing, exercised, and covers composite actions**:

1. **#681 — composite-action coverage.** `scripts/check-action-pins.sh` now also scans `.github/actions/**/action.yml`, not just `.github/workflows/`. A composite action can carry its own `uses:` steps; an unpinned one there was arbitrary code execution in CI with the guard green. Collected via `find` (not a `**` glob) so it runs under the bash 3.2 that ships on macOS / `.husky` pre-commit.
2. **#662 — exercise the guard.** New `scripts/__tests__/check-action-pins.test.sh`, run as a step in `action-pin-guard.yml` *before* the real scan, so every PR proves the guard rejects an unpinned ref in a workflow **and** in a composite action, and accepts a fully-pinned tree. No unexercised guard (#662 / #614). A `CHECK_ACTION_PINS_ROOT` env seam lets the guard point at fixture trees.
3. **#662 — last inconsistent pin.** `build-server-image`'s checkout was still `9c091bb2…` (v7.0.0) while every other checkout in `ci.yml` is `3d3c42e5…` (v7.0.1). Already SHA-pinned (not a security gap), bumped for consistency.

## Already resolved, not re-done

- **The guard already enforces** (`set -euo pipefail`, `exit 1`) — landed in #680, made a registerable required check in #682. This PR does not re-add enforcement.
- **#681 fix #2 (extend `on.paths` trigger) is moot** — #682 removed the `on.paths` filter entirely so the guard runs on every PR regardless of touched paths. Nothing to change there.

## Coverage matrix

| Surface | Scanned before | Scanned now | Exercised by self-test |
|---|---|---|---|
| `.github/workflows/*.yml` | yes | yes | yes (pinned pass + unpinned fail) |
| `.github/actions/**/action.yml` (composite) | **no** | **yes** | yes (pinned pass + unpinned fail) |
| `./`-local refs | excluded | excluded | yes (pass) |
| `docker://…@sha256` | excluded | excluded | yes (pass) |
| branch ref / short SHA | rejected | rejected | yes (fail) |
| empty `.github/` (broken checkout) | fails loudly | fails loudly | yes (fail) |

### Pinned fixes in this PR
- `ci.yml` `build-server-image` checkout: `9c091bb2… # v7.0.0` → `3d3c42e5… # v7.0.1`.
- No unpinned (mutable-tag) `uses:` existed anywhere in the repo; the real guard passes clean.

## Verification (local, bash 3.2.57 on macOS)

```
$ ./scripts/__tests__/check-action-pins.test.sh
ok   — fully pinned workflow + composite action + ./ and docker:// exclusions (exit 0)
ok   — unpinned tag in workflow is rejected (exit 1)
ok   — unpinned tag in composite action is rejected (#681) (exit 1)
ok   — empty .github fails loudly (no vacuous pass) (exit 1)
ok   — branch ref / short SHA rejected (exit 1)
check-action-pins.test.sh: all cases passed ✓

$ ./scripts/check-action-pins.sh
All GitHub Actions `uses:` refs are SHA-pinned across 8 file(s) in .github/workflows/ and .github/actions/. ✓
```

No TS touched, so no typecheck run. Changes are limited to `.github/**` and `scripts/**`.

> SENSITIVE (`.github/workflows/**`) — do not merge without adversarial review + gate.

### DIFF
--- a/.github/workflows/action-pin-guard.yml
+++ b/.github/workflows/action-pin-guard.yml
@@ -31,5 +31,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      # Prove the guard actually rejects unpinned refs (in workflows AND composite
+      # actions) and accepts pinned trees before trusting its verdict — #662 /
+      # #614 "don't ship an unexercised guard".
+      - name: Self-test the guard against fixtures
+        run: ./scripts/__tests__/check-action-pins.test.sh
       - name: Assert every `uses:` is SHA-pinned
         run: ./scripts/check-action-pins.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,7 +246,7 @@ jobs:
       github.event.action != 'labeled' &&
       (needs.changes.outputs.buildserver == 'true' || github.event_name == 'workflow_dispatch')
     steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       # Needed for the GHA layer cache below (the default docker driver cannot
       # export type=gha).
--- a/scripts/__tests__/check-action-pins.test.sh
+++ b/scripts/__tests__/check-action-pins.test.sh
@@ -1,0 +1,111 @@
+#!/bin/bash
+# Exercises scripts/check-action-pins.sh against fixture trees so the guard is
+# never shipped unexercised (#662 "prove it fails on a bad line"; #614). Runs in
+# the Action pin guard workflow *before* the real scan, so every PR proves the
+# guard both accepts a fully-pinned tree and rejects an unpinned ref — in a
+# workflow file AND in a composite action (#681).
+#
+# Pure bash, no test framework: the guard itself is pure bash and CI runs the
+# monorepo vitest via `pnpm -r test`, not the root `scripts/` suite, so a vitest
+# spec here would not gate. Self-contained + exit-code assertions do.
+set -uo pipefail
+
+guard="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)/check-action-pins.sh"
+sha="1111111111111111111111111111111111111111"   # 40 hex, shape-valid
+fails=0
+tmproot=""
+
+cleanup() { [ -n "$tmproot" ] && rm -rf "$tmproot"; }
+trap cleanup EXIT
+
+# run <expected_exit> <description> — invokes the guard against $fixture (built by
+# the caller) and asserts its exit code. On mismatch, dumps output and flags.
+run() {
+  local expected="$1" desc="$2" out rc
+  out="$(CHECK_ACTION_PINS_ROOT="$fixture" bash "$guard" 2>&1)"
+  rc=$?
+  if [ "$rc" -eq "$expected" ]; then
+    printf 'ok   — %s (exit %s)\n' "$desc" "$rc"
+  else
+    printf 'FAIL — %s: expected exit %s, got %s\n%s\n' "$desc" "$expected" "$rc" "$out"
+    fails=$((fails + 1))
+  fi
+}
+
+new_fixture() {
+  tmproot="$(mktemp -d)"
+  fixture="$tmproot/repo"
+  mkdir -p "$fixture/.github/workflows" "$fixture/.github/actions"
+}
+
+# 1. Fully pinned workflow + pinned composite action + legit exclusions → pass.
+new_fixture
+cat > "$fixture/.github/workflows/ci.yml" <<EOF
+jobs:
+  a:
+    steps:
+      - uses: actions/checkout@$sha # v7.0.1
+      - uses: ./.github/actions/local-thing
+      - uses: docker://alpine@sha256:$sha
+EOF
+mkdir -p "$fixture/.github/actions/local-thing"
+cat > "$fixture/.github/actions/local-thing/action.yml" <<EOF
+runs:
+  using: composite
+  steps:
+    - uses: actions/setup-node@$sha # v4
+EOF
+run 0 "fully pinned workflow + composite action + ./ and docker:// exclusions"
+cleanup
+
+# 2. Unpinned tag in a WORKFLOW → fail.
+new_fixture
+cat > "$fixture/.github/workflows/ci.yml" <<EOF
+jobs:
+  a:
+    steps:
+      - uses: actions/checkout@v4
+EOF
+run 1 "unpinned tag in workflow is rejected"
+cleanup
+
+# 3. Unpinned tag in a COMPOSITE ACTION → fail (the #681 regression this closes).
+new_fixture
+cat > "$fixture/.github/workflows/ci.yml" <<EOF
+jobs:
+  a:
+    steps:
+      - uses: actions/checkout@$sha # v7.0.1
+EOF
+mkdir -p "$fixture/.github/actions/build"
+cat > "$fixture/.github/actions/build/action.yml" <<EOF
+runs:
+  using: composite
+  steps:
+    - uses: actions/setup-node@v4
+EOF
+run 1 "unpinned tag in composite action is rejected (#681)"
+cleanup
+
+# 4. Empty .github (no workflow or action files) → fail loudly, not vacuous pass.
+new_fixture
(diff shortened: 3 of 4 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (`.github/workflows/**`) → adversarial review complete → queuing for gate.**

**My independent verification** (clean clone at `214e0c40`): self-test 5/5 cases pass locally; real guard green across all 8 files; claude-review read in full (no blocking findings, two nits noted).

**Adversarial review: PASS** — fresh hostile reviewer, full report highlights:
- Composite-action scanning demonstrated working via fixture trees (unpinned `action.yaml` at depth → rc=1 with correct file:line). Forward-looking (repo has no `.github/actions/` yet) — exercised by the self-test only, acceptable.
- Self-test genuinely gates: runs the REAL guard before the real scan, no `continue-on-error`, exits 1 on case failure.
- **The v7.0.1 SHA verified against upstream**: `git ls-remote actions/checkout refs/tags/v7.0.1` = `3d3c42e5…` exactly; all 16 checkout call sites now identical. Not a supply-chain swap.
- No scope creep: ci.yml diff is the single SHA line; permissions/triggers/secrets untouched.
- **One MAJOR finding, pre-existing and out of scope**: flow-style YAML (`steps: [{uses: actions/checkout@v4}]`) bypasses the line-anchored matcher — demonstrated with exit-0 repro. The grep line is UNCHANGED by this PR (verified not in any diff hunk), so it's filed as follow-up **#723**, not a blocker. Two MINORs (symlinked action.yml skipped; 40-hex-branch indistinguishable offline) recorded there too.

Recommend merge as-is; #723 tracks making the matcher YAML-aware.

thomgabriel:
## Gate: PASS — merging. The composite hole is genuinely closed; I attacked it rather than reading it.

Review current (20:03:23Z vs head 20:02:08Z, 75s after). CI green including `SHA-pin guard` (5s), which now runs the self-test step *before* the real scan.

### #681 verified by attack, using the seam this PR added

`CHECK_ACTION_PINS_ROOT` made this testable, so I built fixture trees and ran the new script against each. Two of the six are traps I designed specifically to break the implementation:

| fixture | expected | result |
|---|---|---|
| clean tree, workflow pinned | pass | **pass** |
| composite with a **pinned** `uses:` | pass | **pass** |
| composite with an **unpinned** `uses:` — the #681 hole | fail | **fail — caught** |
| unpinned **nested two levels deep** (`actions/deep/nested/action.yml`) | fail | **fail — caught** |
| composite referencing a local `./` action | pass | **pass** (correctly excluded) |
| unpinned in **`action.yaml`**, not `.yml` | fail | **fail — caught** |

The nested case is why `find` was the right call over a glob — a `*` pattern would have missed it, and the PR says so explicitly rather than leaving it implicit. The `.yaml` extension case is the one most implementations forget.

**And I ran it under `bash 3.2.57`** — the exact version the comment claims it must support (macOS ships 3.2, and `.husky` pre-commit runs there). So "globstar is bash 4+ only, use `find`" is a verified constraint, not a guess. A `**` glob would have silently matched nothing under 3.2, which is the worst possible failure for a security guard: green, and scanning zero files.

### #662 — the guard is now self-exercising, which is the durable version of what I did by hand

When I gated #680 I mutation-tested the script manually because a pin-guard passing on an already-pinned repo proves nothing. `scripts/__tests__/check-action-pins.test.sh` makes that permanent and runs it on **every PR**, before the real scan, proving the guard rejects an unpinned ref in both a workflow and a composite action. That is the #614/#662 lesson properly discharged: the guard no longer depends on a human remembering to attack it.

### The "moot" claim is correct, and it is moot because of #682

The body says #681's second fix — extending `on.paths` to `.github/actions/**` — is moot. Verified: the workflow's trigger is now `push: branches ["**"]` + `pull_request: branches [main]` with **no paths filter at all**, because #682 removed it so the check could be registered as required. So there is no filter left to extend, and adding one back would re-break enforceability. Right conclusion, and good that it was checked rather than mechanically implementing both halves of my issue.

### Remaining, unchanged, and still yours

`SHA-pin guard` is enforcing, exercised, and complete in coverage — but **still not in branch protection**. Required checks on `main` are `Frontend (lint · typecheck · test)` and `On-chain (fmt · clippy · cargo test)` only. Until `SHA-pin guard` is added there, an unpinned action still merges past a red ✗. That is a repo-settings action I deliberately do not take.

Everything else about the guard is now closed. Closes #662, closes #681.

PR #718, opened 2026-07-26
fix(lessons/complete): dedup grading suite for concurrent duplicate submissions (#693)

### LINKED ISSUE #693: lessons/complete: concurrent duplicates still each run the grading suite (residual of #651)

thomgabriel:
Closes #693.

## What

Adds a **pre-grading submission-dedup guard** to `/api/lessons/complete` so N concurrent duplicates of the *same* submission run the server grading suite (code executors, quiz checks) **once**, not N times — implemented as a true **in-progress lock**, not a fixed window.

## Why

Residual of #651 / #690. #690's per-`(user, lesson)` in-flight guard sits **after** grading — correctly, since moving it earlier would throttle a learner's legitimate re-attempts. So #690 only dedups the on-chain submit; the graders still ran for every concurrent duplicate. This follows **Option 1** from the issue: dedup on **submission identity**, not on `(user, lesson)`.

## Adversarial-review fix (v2 on this branch)

The first cut acquired a 1-token / 30s key **before** grading and never released it. That leaked a false `429 completion_in_progress` onto every honest failure-then-resubmit path (nothing was actually "in progress"). Fixed by making the guard a real **acquire / release lock**:

- **Acquire** the fingerprint key atomically (unchanged — the mechanism the review verified sound).
- The guarded section is wrapped in **`try / finally`**, and the key is **released on every exit** — every deny/`503`/`409`, the already-completed `200`, the success `200`, and thrown errors — via a new `releaseRateLimit(namespace, key)` (service-role `DELETE` of the `rate_limits` row, fail-soft).
- The `429` for a genuine twin returns **outside** the `try/finally`: a request that did **not** acquire must never release the twin's lock.
- **Release is awaited** (release-before-response) so an honest re-click can't race the delete. Released on success too — the on-chain bit + #651's guard already cover post-success duplicates, so holding it would only risk a false 429 on a genuine retry.

## Before / after concurrency semantics

| Scenario | Before this PR | After this PR |
|---|---|---|
| N concurrent duplicates of the **same** submission | grading runs N times; #690 dedups only the on-chain tx | grading runs **once**; the rest `429 completion_in_progress` **while a twin is genuinely in flight** |
| Legitimate re-attempt with **different** answers/code | grading runs (correct) | grading runs — different fingerprint → different key → **never blocked** |
| Failed grading → **identical** resubmit (re-click Submit) | `403` + reason | lock released after the first failure → **re-grades**, real `403` + reason (**not** a false 429) |
| `wallet-not-connected` → connect → resubmit same answers | `400` wallet reason | `400` wallet reason again (**not** a false 429) |
| Thrown on-chain submit (`500`) → resubmit | `500`, retryable | `finally` released the lock → resubmit **re-grades** and can succeed |
| #690 in-flight guard (on-chain submit) | dedups `(user, lesson)` | **unchanged** — still dedups valid payloads the fingerprint guard lets through |

## How the fix differs from — and preserves — #690's guard

- **Key.** #690: `${user}:${lessonId}` (coarse, per-lesson). New guard: `${user}:${lessonId}:${sha256(canonical(proofs))}` — keyed on the **proofs fingerprint**, so a corrected re-attempt (different proofs) is never blocked.
- **Placement.** #690 stays **after** grading (dedups the tx). New guard sits **before** grading (dedups the CPU). Both coexist; #690 is untouched.
- **Lifecycle.** #690 is a coarse 30s window. The new guard is an acquire/release lock that is handed back the moment the request stops processing — so it means "a twin is processing right now", which is exactly what its `completion_in_progress` copy claims.
- **Response.** Identical to #690 — `429` + `reason: "completion_in_progress"` + `Retry-After: 30`. No new client contract.
- **Fail mode.** Acquire fails **open** (store blip → grading runs). Release fails **soft** (a leaked lock self-heals when its window expires). Never a blocked completion.

The fingerprint canonicalises `proofs` (recursively sorted object keys, array order preserved) so key-spelling order can't split a true duplicate, while any real content difference always yields a different key.

## Verification

- `pnpm install --frozen-lockfile` — clean
- `pnpm --filter web test` — **1573 passed** (178 files)
- `/api/lessons/complete` suite + `rate-limit` unit suite — **66 passed** (6 files). New/updated coverage: concurrent twin → 429, grading skipped, twin's lock **not** released; acquiring request releases the same fingerprint key; **failed grading → identical resubmit re-grades + real 403** (inverts the prior test); wallet-not-connected resubmit → wallet reason (not 429); thrown submit → `finally` releases → resubmit re-grades; fingerprint keying (64-hex, identical vs different, order-independent); fail-open; `releaseRateLimit` unit tests (exact key, fail-soft on error and on throw).
- eslint clean on changed files; zero `any`, strict TS.

### DIFF
--- a/apps/web/src/app/api/lessons/complete/__tests__/deny-reasons.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/deny-reasons.test.ts
@@ -38,6 +38,7 @@ const {
 
 vi.mock("@/lib/rate-limit", () => ({
   isRateLimited: (ns: string) => isRateLimited(ns),
+  releaseRateLimit: () => Promise.resolve(),
   getClientIp: () => "203.0.113.7",
 }));
 
--- a/apps/web/src/app/api/lessons/complete/__tests__/gate.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/gate.test.ts
@@ -31,6 +31,7 @@ const {
 
 vi.mock("@/lib/rate-limit", () => ({
   isRateLimited: (ns: string) => isRateLimited(ns),
+  releaseRateLimit: () => Promise.resolve(),
   getClientIp: () => "203.0.113.7",
 }));
 
--- a/apps/web/src/app/api/lessons/complete/__tests__/grading-dedup.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/grading-dedup.test.ts
@@ -1,0 +1,219 @@
+/* eslint-disable import/order -- vi.mock calls must precede importing the route. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest } from "next/server";
+import type { GradeResult } from "@/lib/grading/types";
+
+vi.mock("server-only", () => ({}));
+
+// Pre-grading submission-dedup guard (#693). Residual of #651/#690: the
+// per-(user,lesson) in-flight guard sits AFTER grading, so N concurrent
+// duplicates of the SAME submission each ran the full grading suite before one
+// took the on-chain token. This guard dedups them BEFORE grading, keyed on the
+// SUBMISSION FINGERPRINT — so a true duplicate short-circuits while a genuine
+// re-attempt (different answers → different fingerprint) is never blocked.
+
+const {
+  getUser,
+  singleProfile,
+  getLessonByIdForGrading,
+  getCourseById,
+  codeGrader,
+  isRateLimited,
+  isOnChainProgramLive,
+  isCourseInMaintenance,
+  isPlatformFrozen,
+} = vi.hoisted(() => ({
+  getUser: vi.fn<() => Promise<unknown>>(),
+  singleProfile: vi.fn<() => Promise<unknown>>(),
+  getLessonByIdForGrading: vi.fn(),
+  getCourseById: vi.fn(),
+  codeGrader: vi.fn<() => Promise<GradeResult>>(),
+  // Records BOTH (namespace, key) so tests can assert the guard is keyed on the
+  // proofs fingerprint, not on (user, lesson) alone.
+  isRateLimited: vi.fn<(ns: string, key: string) => Promise<boolean>>(),
+  isOnChainProgramLive: vi.fn<() => Promise<boolean>>(),
+  isCourseInMaintenance: vi.fn<() => Promise<boolean>>(),
+  isPlatformFrozen: vi.fn<() => Promise<boolean>>(),
+}));
+
+vi.mock("@/lib/rate-limit", () => ({
+  isRateLimited: (ns: string, key: string) => isRateLimited(ns, key),
+  getClientIp: () => "203.0.113.7",
+}));
+
+vi.mock("@/lib/supabase/server", () => ({
+  createClient: async () => ({ auth: { getUser } }),
+}));
+
+vi.mock("@/lib/supabase/admin", () => ({
+  createAdminClient: () => ({
+    from: () => ({
+      select: () => ({ eq: () => ({ single: singleProfile }) }),
+    }),
+  }),
+}));
+
+vi.mock("@/lib/content/queries", () => ({
+  getLessonByIdForGrading,
+  getCourseById,
+}));
+
+vi.mock("@/lib/grading/graders", () => ({
+  GRADERS: { code: () => codeGrader() },
+}));
+
+vi.mock("@/lib/review/schedule-review", () => ({
+  captureReviewFailure: vi.fn().mockResolvedValue(undefined),
+}));
+vi.mock("@/lib/ai/check-seal", () => ({ openAttestation: () => true }));
+vi.mock("@/lib/solana/academy-program", () => ({
+  isOnChainProgramLive,
+  completeLesson: vi.fn(),
+  getConnection: vi.fn(),
+  getProgramId: vi.fn(),
+}));
+vi.mock("@/lib/solana/academy-reads", () => ({
+  fetchEnrollment: vi.fn(),
+  fetchCourse: vi.fn(),
+}));
+vi.mock("@/lib/solana/bitmap", () => ({ isLessonComplete: vi.fn() }));
+vi.mock("@/lib/courses/lesson-index", () => ({ findLessonIndex: () => 0 }));
+vi.mock("@/lib/logging", () => ({ logError: vi.fn() }));
+vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
+vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
+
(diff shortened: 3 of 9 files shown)


thomgabriel:
**Orchestrator routing record — adversarial review: FAIL (1 MAJOR). Fix dispatched; do not merge yet.**

Independent hostile review (full suite re-run: 178 files / 1568 green; canonicalization probed empirically; atomicity + #690 interaction + row growth all verified sound). Two of my own briefing premises were corrected: the limiter is Supabase-backed and cross-instance (not in-memory), and there is no acquire/release — it's a fixed-window counter.

**MAJOR (confirmed, user-reachable):** the guard spends its 30s token at route.ts:253-258, BEFORE grading and before every downstream failure exit. On main, none of those exits had consumed a token, so an identical retry re-ran cleanly. On this head, a byte-identical resubmit within 30s of: graded failure (403 quiz/challenge/parsons), reflection_required, **wallet-not-connected → connect → resubmit**, program-not-live 503, maintenance 503, pending-re-sync 409, or a thrown 500 — all return `429 completion_in_progress`, which is FALSE in every one of those states (nothing is in progress). The PR's own test at grading-dedup.test.ts:168-196 codifies the failed-resubmit case, contradicting the body's 'failed-grading 403s unchanged' row (true only for first submission / different-answer retries). Not a security hole, no state corruption, self-heals in ≤30s — but the honest retry paths (re-click Submit on failing code; wallet-connect flow) get a misleading answer.

**Required fix (dispatched):** make the guard mean what its reason says — an in-progress lock, not a 30s window. Release the fingerprint key on every exit after the guarded section (graded-failure exits, all pre-submit failure exits, thrown errors — a finally), so concurrent duplicates 429 only while a twin request is genuinely processing, and an honest sequential identical resubmit re-grades and gets its true reason. Tests to add: failed grading → identical resubmit → real 403 reason + grader re-runs; concurrent identical → one grades, rest 429; release also happens on thrown errors.

MINORs (recorded, no action): higher one-shot key churn in rate_limits (bounded by the 1% sweep); full-blob sync hashing (pre-existing exposure).

thomgabriel:
@claude Re-review please — this pushes the adversarial-review fix for the MAJOR (false `429 completion_in_progress` on failure-then-resubmit paths).

Head `92e20c7`. The pre-grading dedup guard is now a true **acquire/release in-progress lock**, not a 30s window:
- Acquire the 1-token fingerprint key atomically (unchanged), then **release it in a `try/finally` on every exit** — all deny/503/409, already-completed 200, success 200, and thrown errors — via new `releaseRateLimit()` (service-role DELETE, fail-soft).
- The twin-429 return stays **outside** the try/finally (a non-acquirer must never release the twin's lock).
- Release is awaited (release-before-response) so an honest re-click can't race the delete.

Please focus on: (1) no exit path leaks the lock; (2) the twin-429 path does not release; (3) the concurrent-twin race is still correctly a 429 while genuinely in flight. Full suite green: 1573 passed / 178 files; typecheck 6/6.

thomgabriel:
**Orchestrator: adversarial re-verification PASS at `92e20c76` — merging.**

Re-verified per-exit by the ORIGINAL reviewer (warm context): all 7 leak rows from the FAIL table now release via the finally (span 287–617, acquire at 266, twin-429 at 274–282 structurally before the try); twin cannot release the holder's lock (asserted in tests); release awaited pre-response; thrown-submit path proven by test (500, then identical resubmit re-grades and succeeds). Two informational tradeoffs recorded, both bounded: cross-window release can at worst allow one extra grading run in a >30s straddle (#690's guard + the on-chain double-complete check still prevent double-submit), and sequential identical floods re-grade once per processing-time instead of per-30s (bounded by the #459 volume gates) — the deliberate price of truthful 429s. claude[bot] re-review at this head: approved, no blocking findings, independently traced the same three properties. My own re-run at head: 178 files / 1573 tests, exit 0.

PR #707, opened 2026-07-26
feat(quests): review quest kind (enum + SQL branch + card deep-links)

thomgabriel:
Closes #572 — **LX-B7 · Review quest kind** (unified launch spec §3, `docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md`).

Adds a `review` daily-quest kind end to end. This is the **monorepo half**; the quest content YAML lives in `solanabr/courses-academy` and ships as a follow-up content PR (shape documented below).

## What changed

| Layer | Change |
|---|---|
| `packages/content-schema` | `review` appended to `QUEST_TYPES`. The `Quest` zod schema (`z.enum(QUEST_TYPES)`) and the generated `quest.schema.json` follow automatically; the SQL-enum-match test updated. |
| `packages/types` | New `QuestType` literal union; `DailyQuest` gains a `type: QuestType` field so the card can pick a per-type deep-link. |
| `apps/web` API | `/api/quests/daily` now passes `type: q.type` into each `DailyQuest`. |
| SQL | Additive `ELSIF v_type = 'review'` branch in `get_daily_quest_state`, shipped as a migration mirrored into `schema.sql`. |
| `apps/web` UI | Quest cards in `dashboard-identity-panel` become locale-prefixed links via a new `questHref()` helper. |

## SQL branch design

```sql
ELSIF v_type = 'review' THEN
  SELECT COUNT(*)::INTEGER INTO v_current
  FROM public.review_items
  WHERE user_id = p_user_id
    AND last_result = true
    AND last_reviewed_at::date = CURRENT_DATE;
```

- **"Cleared today" = passed today.** `record_review_result` (#689 / LX-B3) sets `last_result` and `last_reviewed_at`; a pass advances the box, a **miss resets to box 1 with `last_result = false`** and must not count toward the quest.
- **Purely additive.** The branch only computes `v_current` and falls through to the **shared generic upsert** — the single owner of the `xp_granted` flip + the atomic `pending_onchain_actions` enqueue. So the durable-delivery / atomicity invariant (CS-7, #353) is inherited verbatim and **every existing quest type is byte-for-byte unchanged**. Quest XP rides the existing `quest_xp` rail (`retryPendingOnchainActions → award_xp`) untouched.
[...]

PR #720, opened 2026-07-26
feat(ai): fail-closed daily spend ledger + degrade-first caps for the AI tutor (#591)

### LINKED ISSUE #591: AI tutor: no spend ledger — add fail-closed per-account/per-IP/global daily caps

thomgabriel:
Closes #591.

## What & why
The AI Partner spends a platform-funded Gemini key with **no record of what it costs**. `challenge_assists` counts *turns*, not spend, so there was no ceiling and no way to spot an abusive account before the bill. This adds an `ai_spend_ledger` that records actual micro-USD usage per billed generation and enforces daily ceilings **fail-closed**, degrading before it denies. Thresholds are derived per the owner decision recorded in #591.

## Cap semantics & defaults
Windows are **America/Sao_Paulo** days. Three dimensions, each with a soft (degrade) and hard (deny) threshold — **any** dimension over hard denies; over soft degrades:

| dimension | soft (degrade) | hard (deny) | env var |
|---|---|---|---|
| global | $14/day | $25/day | `AI_SPEND_GLOBAL_SOFT_USD` / `..._HARD_USD` |
| per-account | $0.50/day | $1.50/day | `AI_SPEND_ACCOUNT_SOFT_USD` / `..._HARD_USD` |
| per-IP | $2/day | $5/day | `AI_SPEND_IP_SOFT_USD` / `..._HARD_USD` |

Gemini pricing (→ micro-USD from `usageMetadata`) is also config: `AI_SPEND_INPUT_USD_PER_MTOK` (0.3), `AI_SPEND_OUTPUT_USD_PER_MTOK` (2.5). **Thresholds are config, not constants** — they move without a migration. All documented in `apps/web/CLAUDE.md`.

### Enforcement (route: `/api/ai/partner`)
- **`checkAiSpend` gates BEFORE the model call**, and before `spendAssist`, so a hard-cap / outage denial burns no paid assist.
  - under every soft cap → **full** (normal output budget)
  - over a soft cap → **degrade**: halve the output budget (the dominant cost lever), floored at 256. Degrade is a shorter output, not a model swap — the cheaper tier is gated for new keys (route already documents the 404).
  - over a hard cap → **deny** (503, `spendCapped: true`, `reason: "spend_cap"`)
  - **ledger unreachable ⇒ 503** (`reason: "ledger_unavailable"`) — the `assist-budget`/#590 fail-closed contract, *not* `check_rate_limit`'s fail-open.
- **`recordAiSpend` books actual usage** after Gemini bills us, on every billed path (including empty/non-JSON/truncated): prompt + candidate + **thinking** tokens, thinking billed at the output rate.
- Hard-cap 503 surfaces dedicated **localized copy** in the pane (en / es / pt-BR); degrade-first, then stop.

### Data layer
`ai_spend_ledger` (integer micro-USD), RPCs `record_ai_spend` / `check_ai_spend` / `get_ai_spend_today` (admin burn observability). All follow the `challenge_assists` hardening: `SECURITY DEFINER`, `SET search_path = ''`, RLS-on table with **no policies**, REVOKEd from PUBLIC/anon/authenticated, GRANTed to `service_role` only. Byte-mirrored into `supabase/schema.sql`.

## Migration
`supabase/migrations/20260726180000_ai_spend_ledger.sql` — timestamp verified unique (latest existing was `20260726170000`; open PR #702 uses `160000`). Idempotent (`IF NOT EXISTS` / `CREATE OR REPLACE`), explicit `BEGIN;`/`COMMIT;`, manual-rollback block. **SQL-only — NOT applied to any database**; the owner dry-runs before it lands.

## Verification
- `pnpm install --frozen-lockfile` — OK
- `pnpm typecheck` — 6 packages successful, zero errors, zero `any`.
- `pnpm --filter web test` — **175 files, 1567 tests passed** (+113 new/changed across 6 touched test files; re-verified green post-format).
- `pnpm --filter web lint` — exit 0 (only pre-existing import/order warnings, none in changed files).

New tests: SQL guard (migration + schema mirror invariants), spend-ledger unit (cost math incl. thinking-at-output-rate, degrade helper, tier decisions, BIGINT-as-string coercion, fail-closed on error/missing-row/non-numeric/throw), route (checks with user+IP, under-cap full + records spend, soft-cap degrades to 1024, hard-cap 503 no-assist-no-model, ledger-error 503 fail-closed, no solution leak), hook + pane (`spendCapped` copy vs generic error).

## Deviations from the issue's proposal (with reasons)
1. **Decision logic lives in TS, not the RPC.** `check_ai_spend` returns the three daily totals; the tier (full/degrade/deny) is picked in `lib/ai/spend-ledger.ts`. Reason: the issue also requires "thresholds are config, not constants" — keeping the numbers in env means no migration when they move. The wrapper is still fail-closed (any error → deny).
2. **Degrade = shorter output budget, not a cheaper model.** The cheaper tier is gated for new keys (documented 404), so a model swap isn't reliably available; halving the output budget is the real, always-available lever.
3. **Admin observability is the `get_ai_spend_today` RPC + wrapper, not a UI page.** The data primitive is shipped and tested; wiring an `/admin` surface is a small follow-up (no admin metrics page exists to hook into today).

Do not merge — for gate review + owner dry-run of the migration.

### DIFF
--- a/apps/web/CLAUDE.md
+++ b/apps/web/CLAUDE.md
@@ -78,6 +78,21 @@ AI_PARTNER_SEAL_SECRET=            # Optional dedicated key for sealing the comp
                                     # SUPABASE_SERVICE_ROLE_KEY.
 AI_PARTNER_DEBUG=                  # Set to "1" to log per-call prompt-cache token counts
                                     # from /api/ai/partner. Default off (quiet in production).
+
+# Optional — AI tutor daily spend caps (#591). Whole/decimal USD, per
+# America/Sao_Paulo day, enforced by the ai_spend_ledger (lib/ai/spend-ledger).
+# Derived DOWNWARD from the $500/mo sponsor commitment (O-1); config, not
+# constants — bump when the commitment moves. Over a SOFT cap the tutor degrades
+# (shorter output budget); only a HARD cap denies (503, fail-closed). Defaults
+# shown; unset/"" uses the default.
+AI_SPEND_GLOBAL_SOFT_USD=          # Global degrade threshold      (default 14)
+AI_SPEND_GLOBAL_HARD_USD=          # Global hard deny              (default 25)
+AI_SPEND_ACCOUNT_SOFT_USD=         # Per-account degrade           (default 0.5)
+AI_SPEND_ACCOUNT_HARD_USD=         # Per-account hard deny         (default 1.5)
+AI_SPEND_IP_SOFT_USD=              # Per-IP degrade                (default 2)
+AI_SPEND_IP_HARD_USD=              # Per-IP hard deny              (default 5)
+AI_SPEND_INPUT_USD_PER_MTOK=       # Gemini input price  $/1M tok  (default 0.3)
+AI_SPEND_OUTPUT_USD_PER_MTOK=      # Gemini output price $/1M tok  (default 2.5; thinking bills here)
 OPENENDED_AI_REPLY=                # Set to "1" to enable the best-effort AI reply on
                                     # /api/lessons/reflect (openEnded reflections). Default OFF:
                                     # the reflection SEAL is always returned regardless; the reply
--- a/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
+++ b/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
@@ -27,6 +27,16 @@ vi.mock("@/lib/rate-limit", () => ({
   getClientIp,
 }));
 
+// #591 spend ledger. checkAiSpend gates the call fail-closed; recordAiSpend
+// books the cost. degradedMaxTokens is the real pure helper (halve, floor 256).
+const checkAiSpend = vi.fn();
+const recordAiSpend = vi.fn();
+vi.mock("@/lib/ai/spend-ledger", () => ({
+  checkAiSpend,
+  recordAiSpend,
+  degradedMaxTokens: (base: number) => Math.max(256, Math.floor(base / 2)),
+}));
+
 const getLessonBySlug = vi.fn();
 vi.mock("@/lib/content/queries", () => ({
   getLessonBySlug,
@@ -150,6 +160,8 @@ beforeEach(() => {
   appendAssistLog.mockReset();
   isRateLimited.mockReset();
   getLessonBySlug.mockReset();
+  checkAiSpend.mockReset();
+  recordAiSpend.mockReset();
 
   // Happy-path defaults; individual tests override as needed.
   getUser.mockResolvedValue({ data: { user: { id: "user-1" } }, error: null });
@@ -159,6 +171,9 @@ beforeEach(() => {
   refundAssist.mockResolvedValue(undefined);
   recordBilledAssist.mockResolvedValue(undefined);
   appendAssistLog.mockResolvedValue(undefined);
+  // Under every spend cap by default → serve at full budget.
+  checkAiSpend.mockResolvedValue({ decision: "full" });
+  recordAiSpend.mockResolvedValue(undefined);
 });
 
 afterEach(() => {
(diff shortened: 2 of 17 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (migration + service_role RPCs + billing path) → adversarial review PASS → queuing for gate.**

**My independent verification** (clean clone at `dee3c526`): `pnpm typecheck` exit 0; full web suite **175 files / 1567 tests, exit 0** — matches the implementer's claim. Migration `20260726180000` present, 22 `ai_spend` references mirrored in schema.sql.

**Adversarial review: PASS** — highlights of what was attacked and held:
- **schema.sql mirror byte-identical** to the migration (mechanical diff, 78/78 code lines).
- **SQL guard test proven effective by mutation**: removing `authenticated` from one REVOKE in a scratch copy failed the guard as required.
- All 3 RPCs: SECURITY DEFINER + `search_path=''` + RLS-on/no-policies + complete REVOKE/GRANT.
- São Paulo day-window expression identical in record+check; no UTC off-by-one; guard forbids `CURRENT_DATE`.
- Scope integrity: account = server-verified `user.id`; IP from Vercel-trusted header chain; `''` scope keys can't collide (scope in PK).
- **Fail-closed proven in tests**: gate before `spendAssist` (denial burns no assist, no model call); ledger error → 503; env parsing rejects garbage/negatives at boot (NaN-cap attack closed).

**Findings, all routed:**
- MAJOR (out of this PR's scope, pre-existing): `reflection-reply.ts` spends the same sponsored key outside the ledger, fail-open — default-OFF behind `OPENENDED_AI_REPLY`. Filed **#724** (with the MINORs folded in: record-after-json-parse skip, undocumented bounded TOCTOU overshoot, cap=0 footgun).
- The envelope guarantee holds exactly while that flag stays off — stated here so the gate rules with eyes open.

**Owner note (unchanged from the PR):** migration is SQL-only, applied nowhere; owner dry-run required before any apply. Caps default from O-1 ($500/mo commitment → $14/$25 global daily soft/hard).

thomgabriel:
## Gate: PASS — merging. Implements O-1 exactly as decided.

Review current (20:13:16Z vs head 20:09:29Z). CI green.

### The caps match the owner decision line for line

Checked against the O-1 record from #591, not against the PR's own summary:

| dimension | decided | implemented |
|---|---|---|
| global | $14 soft / $25 hard | ✓ |
| per-account | $0.50 soft / $1.50 hard | ✓ |
| per-IP | $2 soft / $5 hard | ✓ |
| window | America/Sao_Paulo days | ✓ |
| tunability | **config, not constants** | ✓ env-driven, no migration to move a threshold |

The per-account and per-IP dimensions are the ones that make $500/mo actually safe: the pathological single-account peak in that analysis was ~$73/day, ~4.4× the entire daily envelope. A global cap alone would have let one account eat everyone's budget before tripping.

### Strictest security surface in the codebase, and correctly so

`ai_spend_ledger` has **RLS on with no policies at all** — not even an own-row SELECT. Every path goes through three `SECURITY DEFINER` RPCs (`record_ai_spend`, `check_ai_spend`, `get_ai_spend_today`), each `SET search_path = ''`, each `REVOKE ALL … FROM PUBLIC, anon, authenticated` + `GRANT EXECUTE … TO service_role`. That is stricter than `review_items` (which has an own-row read) and right for a table whose contents are billing data no client needs.

### Fail-closed is reasoned, not copied — and that is the notable part

The module says it outright:

> It fails CLOSED: any check error denies (503) … **the opposite of `check_rate_limit`'s fail-open.**

That asymmetry is correct, and it is the same distinction I drew gating #690:

- **#690's in-flight guard** is a cost *optimisation* protecting reverted-tx fees → fail **open**, because a store blip must never stop a learner completing a lesson.
- **This** is a spend *ceiling* on a platform-funded key → fail **closed**, because a store blip must never wave unmetered traffic through.

Copying either pattern blindly would have been wrong in one of the two places. `SpendDenyReason` also separates `spend_cap` from `ledger_unavailable`, so an outage is distinguishable from a real cap in logs — which matters, because those demand opposite responses.

**And the blast radius of fail-closed is bounded**: I checked, and there is no spend gate in `/api/lessons/complete`. A denial disables the AI tutor, it cannot block finishing a lesson. If the ledger were reachable from the completion path this would be a much harder call.

### Two more good calls

- **`checkAiSpend` runs before `spendAssist`**, so a hard-cap or outage denial burns **no paid assist**. Without that ordering, a capped learner would lose an assist for a request that never reached the model.
- **Degrade halves the output budget (floored at 256) rather than swapping models.** Output tokens are the dominant cost lever here (`$2.50` vs `$0.30` per Mtok), and a shorter answer degrades honestly — a quietly worse model would not.

Migration not yet applied. Batching it with #707's `review_quest_kind` and #702's `prefs`, both also merged-and-pending.

PR #721, opened 2026-07-26
feat(credentials): capstone-gated credential (LX-E2, #561)

### LINKED ISSUE #561: Capstone-gated credential (split webhook handler)

thomgabriel:
Closes #561 · spec task **LX-E2** (`docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md` §3).

## What

Withhold the **capstone course** credential until the learner's graded deploy is **verified** — a `deployed_programs` row for the capstone deploy lesson, which is written only by the on-chain-verified `/api/deploy/save` (#620 / LX-E1). Every other course's credential, and all XP, are unchanged.

## Why

Owner decision **O-2**: the capstone is follow-along; the credential must attest a real deployed program ("deployed a Solana program"), not attendance. `deployed_programs` is on-chain-verified (executable + upgrade-authority-is-the-learner), so a row is a genuine artifact.

## How — three mint paths, one shared gate

New `apps/web/src/lib/credentials/capstone-gate.ts`: an app-side capstone constant (`course-building-first-program` / `lesson-bfsp-m4-capstone`) + `checkCapstoneCredentialGate`, which is **fail-closed** (`indeterminate` never returns `allowed`). The constant is asserted against the committed content bundle so a content rename fails the test instead of silently denying every capstone.

1. **Webhook** (`event-handlers.ts` `tryIssueCredential`): `finalize_course` + completion-bonus XP proceed on lesson completion **as today**; only `issue_credential` gains the gate. `deploy_required`/`indeterminate` → throw → deferred onto the existing `certificate` queue.
2. **Manual mint** (`POST /api/certificates/mint`): returns a stable `capstone_deploy_required` code (403) — localized client-side in `course-completion-mint.tsx` across en / pt-BR / es — or fail-closed 503 on `indeterminate`.
3. **Login drainer** (`onchain-queue.ts` `certificate` case): the same gate runs here too, so a queued capstone certificate can **never** mint ungated on the next drain. Defers **without burning the retry budget** (mirrors the maintenance defer), so it lands the moment the verified deploy row appears — and never before.

**Grandfathering:** all three paths short-circuit on an already-issued credential (`credential_asset` / existing `certificates` row) **before** the gate, so pre-gate devnet certs stay valid and are never re-evaluated.

No program change; server-route + queue logic only.

## LX-E2 acceptance criteria

- [x] Credential mints only with a verified deploy (capstone course)
- [x] Completion-XP timing unchanged (`finalize_course` + bonus XP proceed; gate guards only `issue_credential`) — regression test on `handleCourseFinalized`
- [x] Manual mint route returns a specific, localized "deploy required" state (`capstone_deploy_required` → `certificates.deployRequired`, x3 locales)
- [x] Fail-closed on indeterminate deploy state (defer/deny, never issue)
- [x] Manual mint gated **and** the login-drainer mint path gated (else a deferred capstone mints ungated on next login)
- [x] Capstone identity is an app-side constant, asserted against the content bundle
- [x] Grandfather pre-gate devnet certs
- [x] No program / on-chain change

## Verification

- `pnpm install --frozen-lockfile` ✓
- `pnpm typecheck` ✓ (6/6 packages)
- `pnpm --filter web test` ✓ — **176 files, 1545 tests passing**
- `pnpm lint` ✓ (exit 0, no new errors)
- New tests (21): gate unit + bundle-constant assertions (9), webhook gate + XP-untouched regression (5), manual-mint gate (4), drainer defer/allow (3)

Zero `any`, strict TS. Do **not** merge — awaiting gate review + owner sign-off.

### DIFF
--- a/apps/web/src/app/api/certificates/mint/__tests__/capstone-gate.test.ts
+++ b/apps/web/src/app/api/certificates/mint/__tests__/capstone-gate.test.ts
@@ -1,0 +1,151 @@
+/* eslint-disable import/order -- vi.mock('server-only') must be hoisted above
+   the route import so the `server-only` graph loads under vitest. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest } from "next/server";
+
+vi.mock("server-only", () => ({}));
+
+// LX-E2 — the manual mint route must apply the same capstone deploy gate as the
+// webhook: no verified deploy → a specific "deploy required" state (localized on
+// the client via a stable code); an unreadable deploy state → fail-closed 503.
+
+const {
+  getUser,
+  isCourseInMaintenance,
+  isPlatformFrozen,
+  isRateLimited,
+  fetchEnrollment,
+  fetchCourse,
+  getCourseById,
+  issueCredential,
+  checkCapstoneCredentialGate,
+} = vi.hoisted(() => ({
+  getUser: vi.fn<() => Promise<unknown>>(),
+  isCourseInMaintenance: vi.fn<() => Promise<boolean>>(),
+  isPlatformFrozen: vi.fn<() => Promise<boolean>>(),
+  isRateLimited: vi.fn<() => Promise<boolean>>(),
+  fetchEnrollment: vi.fn(),
+  fetchCourse: vi.fn(),
+  getCourseById: vi.fn(),
+  issueCredential: vi.fn(),
+  checkCapstoneCredentialGate: vi.fn(),
+}));
+
+vi.mock("@/lib/supabase/server", () => ({
+  createClient: async () => ({ auth: { getUser } }),
+}));
+vi.mock("@/lib/supabase/admin", () => {
+  // A valid base58 pubkey so the route's `new PublicKey(wallet_address)` works.
+  const chain: {
+    eq: () => typeof chain;
+    single: () => Promise<{ data: { wallet_address: string } }>;
+    maybeSingle: () => Promise<{ data: null }>;
+  } = {
+    eq: () => chain,
+    single: async () => ({
+      data: { wallet_address: "11111111111111111111111111111111" },
+    }),
+    maybeSingle: async () => ({ data: null }),
+  };
+  return {
+    createAdminClient: () => ({ from: () => ({ select: () => chain }) }),
+  };
+});
+vi.mock("@/lib/content/queries", () => ({ getCourseById }));
+vi.mock("@/lib/solana/academy-program", () => ({
+  issueCredential,
+  getConnection: () => ({}),
+  describeProgramError: () => "unknown",
+}));
+vi.mock("@/lib/solana/academy-reads", () => ({ fetchEnrollment, fetchCourse }));
+vi.mock("@/lib/solana/pda", () => ({ getProgramId: () => "program-id" }));
+vi.mock("@/lib/solana/arweave", () => ({ uploadCertificateMetadata: vi.fn() }));
+vi.mock("@/lib/solana/credential-metadata", () => ({
+  capCredentialName: (n: string) => n,
+}));
+vi.mock("@/lib/rate-limit", () => ({
+  isRateLimited,
+  getClientIp: () => "203.0.113.7",
+}));
+vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
+vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
+vi.mock("@/lib/platform/freeze-http", () => ({
+  platformFrozenResponse: () =>
+    new Response(JSON.stringify({ maintenance: true }), { status: 503 }),
+}));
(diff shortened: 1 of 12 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (credential issuance) → adversarial review PASS → queuing for gate.**

**My independent verification** (clean clone at `11572241`): `pnpm typecheck` exit 0; full web suite **176 files / 1545 tests, exit 0**.

**Adversarial review: PASS** — attack surfaces closed with evidence:
- **All 3 mint paths gated, no fourth exists** (grep-proven): webhook `tryIssueCredential`, manual mint route, queue drainer `certificate` case. The only other `certificates` writer (`admin/resync`) mirrors already-minted on-chain credentials, never mints.
- **Gate source un-forgeable**: `deployed_programs` post-#620 = RLS-on with the two client policies DROPPED; only writer is `/api/deploy/save` (admin client, after on-chain verify). Gate query scoped by all three keys (user_id + course_id + lesson_id) — another user's or another lesson's row does not satisfy it.
- **No deadlock**: deferrals don't burn the retry budget; queued capstone credentials re-derive the gate every drain (no stale-allow replay) and land on the next post-deploy login-drain; manual mint gives an immediate post-deploy path.
- **XP/finalize byte-untouched** — regression test proves completion-bonus XP lands even when the credential is withheld.
- **Rename-safe**: the bundle assertion (course exists, lesson hosts the deploy panel, exactly one such lesson repo-wide) fails CI LOUDLY on the upcoming C3 bfsp renames — no silent denial. (C3 workflow: expect this test to go red when your renames land; update the constants in `lib/credentials/capstone-gate.ts` in the same PR.)
- i18n parity for the 403 code across en/es/pt-BR (enforced by the parity test).

**Findings (recorded, non-blocking):**
- INFO — conscious acceptance, inherited from LX-E1/#620, matches owner decision O-2: the gate attests 'a real deployed program under the learner's authority', not vault correctness; `/api/deploy/save` accepts arbitrary courseId/lessonId against a verified deploy. If the owner ever wants correctness attestation, that's a #620-side change, not this PR.
- MINOR — the manual-mint 503 `indeterminate` body is English-only (the 403 is localized). Transient copy; fold into any follow-up i18n sweep.

thomgabriel:
## Gate: PASS — merging. O-2's "sole remaining P0-class item".

Review current (20:15:02Z vs head 20:13:24Z). CI green, MERGEABLE.

### The gate is unforgeable, and I can say that from evidence rather than reading

It gates on a `deployed_programs` row, which is written **only** by `/api/deploy/save` via a service-role client — because that table has no client INSERT/UPDATE policies, per the lockdown migration `20260726120000_lockdown_deployed_programs_rls`. I applied that migration to prod earlier today and **verified it by attempting the forgery** as an impersonated authenticated user: the INSERT was blocked, 0 rows. So "a row is a genuine artifact" is a checked claim, not an assumed one. Nice compounding — #620's hardening is what makes this credential gate meaningful.

Fail-closed is right here and correctly implemented (`indeterminate` never returns `allowed`): a credential is a permanent on-chain attestation, so the failure that matters is issuing one that is not earned, not delaying one that is.

### The constant survives C3 — I checked, because this was the live collision risk

The gate hardcodes `course-building-first-program` / `lesson-bfsp-m4-capstone`. My C3 transform is in flight **on that exact course** and renames most of its lessons, so a stale constant would have meant "nobody ever gets the capstone credential".

It survives: `lesson-bfsp-m4-capstone` is the id of `what-you-built` — a CARRY lesson — and it sits at slot 15 in C3's regenerated `slots.lock.json`. It is also present in the shipped bundle (99 lessons). So the constant holds before and after C3.

Asserting the constant against the committed bundle is the right call for exactly this reason: a content rename now **fails a test** instead of silently denying every capstone. That is the difference between a loud break and a permanently withheld credential.

### Zero retroactive effect — confirmed against prod

```
certificates total:                4
certificates for the capstone course: 0
completed capstone enrolments:        0
deployed_programs rows:               0
```

Nobody loses a credential they already hold, and nobody is mid-flight waiting on one. That is the answer to the only question that could have made this blocking.

### ⚠️ The real risk: this gates on an artifact that has never been produced in production

`deployed_programs` has **0 rows, ever**. So `/api/deploy/save` — the on-chain-verified path that writes them — has no successful production run behind it. The gate is correct, but its dependency is unexercised where it counts.

The failure mode is nasty because it is silent and looks like success: if `/api/deploy/save` has a bug, the symptom is "no learner can ever earn the capstone credential", and from the outside that is indistinguishable from the gate working as designed. The webhook path defers onto the `certificate` queue rather than erroring, so it would not even be loud.

Not blocking — the gate is what was asked for, and nothing is broken today because nobody has reached the capstone. But **before a learner reaches it, `/api/deploy/save` needs one real end-to-end run** producing a row. Filing that rather than leaving it implicit.

Issue #712, opened 2026-07-26
Author C3 — building-your-first-solana-program (16 → 15, KEEP + anchor-framework merge) — UNTRACKED, blocks C4


thomgabriel:
**Nothing tracks C3, and C4 is blocked on it.** #674 says "blocked on C3 shipping"; there is no issue for C3 shipping. Searched all issues for `building-your-first-solana-program` — only tangential hits (#683, #606, #686). This is a missing node on the launch critical path.

C3 is the **only** target course that already exists as a real course, so it is the cheapest of the four remaining — but it is not zero, and it gates C4.

## Current vs target

`building-your-first-solana-program` — **16 lessons today, target 15 (3/4/4/4)**, same slug, same exercise ladder. CATALOG §3 marks it **KEEP**, not rewrite: 12 of 15 lesson ids carry over.

## What CATALOG §3 says carries — do not re-author these

- `your-first-build` — **carry verbatim**, including its 3 hints
- `wire-up-initialize` — the `// BUG:` DEBUG rung, called "the crown jewel" in the ledger
- `define-counter-account` — renamed to the vault
- `airdrop-fund-wallet`, `deploy-program-devnet`, `interact-with-program`, `what-you-built`
- prose from `anatomy-anchor-program`, `on-chain-state`, `account-constraints-deep-dive`

## What must be destroyed or fixed

- **Slot-11 `deploy-to-devnet`** — five files byte-identical (md5 `9b77d6d5…`) to slot 10, no `deployable: true`, teaches CLI deploy. Delete.
- **The opener's false claim** — "You've written Anchor programs in the previous courses". Untrue for C3's actual entrants.
- **Merge in `anchor-framework`'s three prose assets** — `anchor-accounts` (478w), `pda-deep-dive` (599w), `cpi-overview` (623w) — **re-authored against Anchor 1.x**. Its five exercises are `DELETE, do not port, do not cite`: none imports `anchor_lang`, and two derive PDAs with `DefaultHasher`, accepting a bump when `hash % 2 == 0`. That is an actively false model of the most important Solana primitive, taught twice.

## Two things that make C3 harder than "edit 16 lessons down to 15"

[...]

Issue #683, opened 2026-07-26
content: tutorNotes missing on two CARRY launch-path challenges (your-first-build, deploy-program-devnet)


thomgabriel:
Found while gating courses-academy#10 (merged), which authored `tutorNotes` for 11 challenges. Two launch-path challenges that the salvage ledger marks **CARRY** were left without notes.

## Why it matters

`tutorNotes` feeds the `[TUTOR_NOTES]` section of the AI tutor prompt (#644, merged). With no authored notes the section never fires, so the learner gets the unmitigated tutor — and teacher-authored common-mistakes were half the evidenced mitigation for the −17% unassisted-exam harm (AIE-26). The schema makes `tutorNotes` optional and absent notes leave the prompt unchanged, so this degrades gracefully rather than breaking; it is a coverage gap, not a defect.

### 1. `building-your-first-solana-program/lessons/your-first-build` — the priority

Salvage ledger: **CARRY**, carry-verbatim (`your-first-build/lesson.yaml:18` already ships 3 `hints:`). `buildType: buildable`, `language: rust`, has `exercise/starter.rs`.

This is the learner's **first successful Rust build**. It is the highest-mistake, highest-anxiety moment in the course and the point where an unguided tutor is most likely to hand over a corrected file instead of teaching the learner to read a compiler error. It is the single worst place in the catalog to have no notes.

Candidate mistakes to author against (verify before writing — the graded block is compile-only):
- editing the wrong file / expecting `cargo build` rather than the sbf toolchain
- missing or misnamed `#[program]` module, so nothing is exported
- treating a warning as the failure and chasing it instead of the error
- pasting a fix without reading which line the compiler names

### 2. `building-your-first-solana-program/lessons/deploy-program-devnet`

Salvage ledger: **CARRY** (`lesson.yaml:15` `deployable: true`, `:22` Resume hint). Weaker case — deploy failures are environmental rather than code — but the same reasoning covers insufficient SOL, wrong cluster, and a stale program id after a rebuild.

## Correctly excluded — do not "fix" these

- `deploy-to-devnet` — ledger says `DELETE (retired)`, slot-11 duplicate of the module-4 deploy.
- `add-instruction` — ledger says `PROSE-ONLY`; the scrambled-fragment exercise is reshaped for C3, so notes written against today's block would be discarded. Author them when C3's `add-an-instruction` is written.
[...]

Issue #606, opened 2026-07-25
Devnet on-chain window: 7 operations, after the track-ladder decision


thomgabriel:
One batched devnet window using **shipped instructions only** (`close_course` + `create_course`). Blocked on the track-ladder decision (O-3) because `trackId`/`trackLevel` are immutable after creation. **Read item 49 first: recreate is no longer progress-safe** — the Pinocchio program's generation check invalidates stale enrollments (error 6034) and re-enrolling zeroes lesson flags.

**Unified spec item 50** — full detail, evidence and acceptance criteria in [`docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md`](docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md) §3.

_Filed from the unified launch spec; this work existed in the plan but had no issue._

thomgabriel:
## OWNER DECISION (O-3) — track ladder: trackId 1, trackLevel 1-5 sequential

Owner delegated to the gate's recommendation. Recording it as decided, since `trackId`/`trackLevel` are immutable after on-chain creation:

| course | trackId | trackLevel |
|---|---|---|
| `solana-for-web-devs` (C1) | 1 | 1 |
| `rust-for-program-devs` (C2) | 1 | 2 |
| `building-your-first-solana-program` (C3) | 1 | **3** (currently 2 — needs the close+recreate) |
| `dapp-and-sdk-with-kit` (C4) | 1 | 4 |
| `stablecoin-agentic-payments` (C5) | 1 | 5 |

This is the catalog spec's own ordering: **one track, five sequential levels**, matching the single "Zero to Deployed" path rather than the current five parallel one-course tracks.

### ⚠️ Before executing the window — read #607 first
The recreate is **no longer progress-safe**. The Pinocchio program bumps a course *generation*, and `complete_lesson`/`finalize_course`/`issue_credential`/`upgrade_credential` each reject a stale enrollment with **error 6034**; re-enrolling re-initialises in place and **zeroes lesson flags**.

Mitigating facts, both verified: the 38 pre-move enrollments are **already stranded** on `7NeJaSRy…` and unreachable from `Dsro2Cd9…`, and the new instance holds **0 enrollments**. So the C3 recreate costs nothing that isn't already lost — but do it **before** learners accumulate progress on the new instance, not after.

[...]

PR #722, opened 2026-07-26
content: bump content.lock to 1b74e4a6 — C2 staged + CARRY tutorNotes + C5/template explanations

thomgabriel:
## What

Pins `courses-academy@1b74e4a6`, picking up four merged content PRs:

| Content PR | What lands | Learner-visible? |
|---|---|---|
| solanabr/courses-academy#12 | **C2 `rust-for-program-devs`** (14 lessons) | No — staged; not on-chain, hidden until owner course-creation (PB-7 instructor wallets) |
| solanabr/courses-academy#14 | tutorNotes on the two CARRY deploy-path challenges (`your-first-build`, `deploy-program-devnet`) | **Yes** — live Zero-to-Deployed path, feeds the AI tutor's [TUTOR_NOTES] |
| solanabr/courses-academy#15 | 15 C5 quiz explanations + `_template` quiz-bar fix | C5 hidden until on-chain window; template is authoring-time |
| solanabr/courses-academy#13 | CATALOG.md PB-7 currency amendment | Docs-only, not compiled |

Bundle counts: courses 7→**8**, lessons 85→**99**, paths/quests/achievements unchanged.

## Golden fixtures

C2's arrival triggers the coverage tests, so `courses.json`, `lessons.json`, and `quests-raw.json`'s two derived fields (`challengeLessonIds`, `moduleLessonMap`) were regenerated from the bundle through the real projectors — existing doc order preserved, raw quest docs untouched, delta documented in the test header (waves 1–3 now enumerated). All other fixtures untouched.

- compile-content: `courses:8, learningPaths:8, lessons:99, quests:5, achievements:10`
- Spot-checks: both CARRY lessons carry tutorNotes; 140 explanation fields across the bundle (was 97); C2 present with 14 lessons
- `pnpm typecheck` exit 0; `pnpm --filter web test` **179 files / 1571 tests, exit 0**
- Visibility safety: C2 (like C5) has no on-chain course → `isSynced` false → hidden from catalog; nothing learner-facing changes except the two tutorNotes lessons

## Routing

`area:content`, SAFE lane.

### DIFF
--- a/apps/web/content.lock
+++ b/apps/web/content.lock
@@ -1,4 +1,4 @@
 {
   "repo": "solanabr/courses-academy",
-  "sha": "e1190680421e8190ab4deb9b41376667abc7a398"
+  "sha": "1b74e4a60f8813374219451b60436af832df2d91"
 }
--- a/apps/web/src/content/generated/courses.json
+++ b/apps/web/src/content/generated/courses.json
@@ -418,6 +418,174 @@
     "xpReward": 1000
   },
   {
+    "_id": "course-rust-for-program-devs",
+    "_type": "course",
+    "creator": "B7o8NfV81HzjuZFWQTTx3Xdvh77Dqoajwib3kWEnvzJF",
+    "creatorRewardXp": 30,
+    "description": "You can read a Solana account's bytes. You cannot yet write the struct those bytes came from. You finish this course holding `vault_core.rs` — a `VaultState` with `owner`, `balance` and `bump`, `deposit` and `withdraw` built on `checked_add`/`checked_sub` so a `u64` can never silently wrap, a `VaultError` enum with four named variants, and not one `unwrap()` — compiled green against the real Anchor 1.x toolchain on our build server. Nothing is installed locally — no rustup, no cargo, no `.so` upload. Ownership, borrowing and lifetimes are taught here rather than deferred, because `Account<'info, Vault>` is unreadable without them. Your file deliberately carries no `#[program]` and no `#[derive(Accounts)]` — that omission is the handoff, and Course 3 is what wraps around it.",
+    "difficulty": "intermediate",
+    "duration": 12,
+    "minCompletionsForReward": 10,
+    "modules": [
+      {
+        "_key": "rpd-rust-that-compiles",
+        "_type": "courseModule",
+        "description": "Get a green build before learning any syntax, then meet exactly the types the vault needs — `u64` lamports, `u8` bump, `i64` timestamps — and the compiler's opinions about mutability, tail expressions and the missing `return`. You end holding a file of your own — two `const fn` helpers that add and subtract lamports and return `None` on overflow or underflow instead of wrapping or panicking, verified by compile-time assertions. That arithmetic becomes `VaultState::withdraw` nine lessons later.",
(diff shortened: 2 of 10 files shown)


thomgabriel:
## Gate: PASS — merging. The launch catalog picks up C2 (staged) and the tutorNotes go live.

### The CI "failure" is a cancelled duplicate, not a failure

`gh pr checks` shows `Dependency audit fail 0` and `Frontend fail 0`. Both belong to run `30218667068`, whose conclusion is **`cancelled`** — every job cancelled at `20:17:57Z` by the concurrency group when the `push` and `pull_request` events raced on the same commit. The `push` run (`30218665035`) on the **same SHA `3250b600`** succeeded, `Frontend (lint · typecheck · test)` in 3m1s.

That matters because `Frontend` carries the **Content bundle freshness** step — it recompiles the bundle from `content.lock` and fails on any drift from what is committed. For a bundle-regeneration PR that is *the* check, and it passed. Recording the diagnosis because a cancelled job rendering as "fail" is the same class of misleading signal as the stale search index.

### Pin verified against the source of truth

```
content main : 1b74e4a60f8813374219451b60436af832df2d91
#722 pins    : 1b74e4a60f8813374219451b60436af832df2d91
```

Exact match. Bundle: 7→**8** courses, 85→**99** lessons.

### C2 lands staged — verified, not assumed

C2 is in the bundle with 14 lessons across 4 modules, and it will **not** render. The catalog gate is `status === "synced" && (is_active ?? true)` over `onchain_deployments`; I queried prod earlier and that table holds exactly **6 course rows** — the old courses. C2 has none, same as C5. So "staged, hidden until owner course-creation" is accurate for a checkable reason, not a hopeful one.

### ⚠️ This bump encodes the duplicate rung into shipped content

Confirmed from the generated bundle at this head:

```
building-your-first-solana-program   trackId=1  trackLevel=2
rust-for-program-devs (C2)           trackId=1  trackLevel=2   ← same rung
stablecoin-agentic-payments (C5)     trackId=1  trackLevel=5
```

**Inert today** — C2 cannot render, so nothing collides for a learner. But the collision is no longer theoretical or local; it is in the compiled bundle on `main`. The moment C2 is created on-chain, two courses occupy trackId 1 / level 2, and `trackLevel` has **no on-chain setter** (`update_course`'s mutable set is `content_tx_id`, `is_active`, `xp_per_lesson`, `creator_reward_xp`, `collection`, `active_lessons`) — so undoing it costs a close+recreate, not an edit.

So the on-chain window (#606) must do C3's 2→3 recreate **before or in the same batch as** C2's creation. C3's transform (#712, in flight) sets `trackLevel: 3` in content, which is the other half. Neither alone is sufficient.

### What actually becomes learner-visible

Only courses-academy#14's tutorNotes, on `your-first-build` and `deploy-program-devnet` — both live on the Zero-to-Deployed path. Those feed the AI tutor's `[TUTOR_NOTES]` section, which was the whole point of #683: without authored notes that section never fires, and the learner gets the unmitigated tutor. C5's 15 new explanations and the `_template` fix are authoring-time or gated behind the on-chain window.

Golden fixtures regenerated through the real projectors with doc order preserved, and the bundle-freshness gate is what proves the regeneration was not stale.

PR #726, opened 2026-07-26
feat(lessons): render parsons paired-distractor markers + one exhaustive deny-reason map (#703)

### LINKED ISSUE #703: content: parsons distractor/pairedWith modelled but never rendered (F13 unshipped) + parsons_failed untested

thomgabriel:
Closes #703.

Two follow-ups from gating #701, exactly as the issue prescribed (render the modelled fields — do not rip them out; add the missing test; and, per the item-2 update, collapse the divergent reason ternaries into one exhaustive map). Neither touches grading.

## 1. Render parsons paired-distractor markers (F13) — issue item 1

`distractor` / `pairedWith` were modelled, validated, and documented in `packages/content-schema` but `parsons-block.tsx` read neither — a paired distractor was visually indistinguishable from a real line, so the *discrimination* that is the whole point of a paired distractor never happened.

**Before:** every bank line rendered through identical markup; the pairing was invisible.

**After:** each distractor and its `pairedWith` target share one **`Look-alike N`** marker.
- **Symmetric** — the marker sits on *both* members and never reveals which of the pair is the trap (revealing it would hand the learner the answer).
- **Not colour-only** — an icon + the group text carry the signal, so it survives for AT and colour-blind learners (the renderer was built for keyboard/AT use).
- **Follows the line** — the marker shows in both the bank and the solution.
- **Grading untouched** — sequence equality already scores an included distractor wrong, client and server.
- A short hint appears only when a block authors paired distractors. i18n added for `en` / `pt-BR` / `es`.

## 2. One exhaustive graded-deny-reason map — issue item 2 (a + b)

Two independent ternaries mapped a graded block type → deny reason and had **already diverged** (#706's capture path folded a failed `parsons` into `challenge_failed`, and `ReviewFailureCapture.reason` was a closed two-value union in which `parsons` is unrepresentable).

- **(b)** New `DENY_REASONS` keyed by `GradedBlockType` with `satisfies Record<GradedBlockType, CompletionDenyReason>` — the same compile-time inversion `GRADERS` already uses. A new graded type now fails the build until its reason is declared. The route's capture + deny both read it; `ReviewFailureCapture.reason` widens to the derived `GradedDenyReason`.
- **(a)** Adds the missing `parsons_failed` case to `deny-reasons.test.ts` — the exhaustiveness the file already had for every other reason.

## Verification

- `pnpm typecheck` — 6/6 packages green.
- `pnpm --filter web test` — **179 files / 1577 tests pass** (was 1568; +9 new: 5 renderer pairing cases incl. a distractor-inclusion-grades-wrong assertion, 1 no-pairing case, 1 `parsons_failed` deny case, and the widened suite still green).
- `eslint` clean on all changed files.

Does **not** merge.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
@@ -270,4 +270,38 @@ describe("ParsonsBlock — paired distractors (F13)", () => {
       screen.getByText("Not quite — some lines are out of order.")
     ).toBeInTheDocument();
   });
+
+  it("two look-alikes for ONE real line all share a single group (no orphan)", () => {
+    // Two traps for `ret` is legitimate authoring; grouping by target keeps the
+    // target + both distractors in one group of three, never orphaning the first
+    // distractor into its own marker.
+    const twoTraps: ParsonsBlockData = {
+      _type: "parsons",
+      key: "p-two-traps",
+      prompt: "Arrange the function.",
+      lines: [
+        { id: "sig", content: "function add(a, b) {" },
+        { id: "ret", content: "  return a + b;" },
+        { id: "close", content: "}" },
+        {
+          id: "bad1",
+          content: "  return a - b;",
+          distractor: true,
+          pairedWith: "ret",
+        },
+        {
+          id: "bad2",
+          content: "  return b - a;",
+          distractor: true,
+          pairedWith: "ret",
+        },
+      ],
+      correctOrder: ["sig", "ret", "close"],
+    };
+    renderWithIntl(<ParsonsBlock block={twoTraps} ctx={makeCtx()} />);
+    // ret + bad1 + bad2 = three members, one shared marker, and only one group
+    // exists (no stray "Look-alike 2").
+    expect(screen.getAllByText("Look-alike 1")).toHaveLength(3);
+    expect(screen.queryByText("Look-alike 2")).not.toBeInTheDocument();
+  });
 });
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
@@ -187,6 +187,15 @@ describe("ParsonsBlock — arrange interaction", () => {
     ).toBeDisabled();
   });
 
+  it("no paired distractor → no look-alike marker and no pairing hint", () => {
+    // The default block's distractor carries no `pairedWith`, so nothing pairs.
+    renderWithIntl(<ParsonsBlock block={parsonsBlock} ctx={makeCtx()} />);
+    expect(screen.queryByText(/Look-alike/)).not.toBeInTheDocument();
+    expect(
+      screen.queryByText(/Some lines are look-alikes/)
+    ).not.toBeInTheDocument();
+  });
+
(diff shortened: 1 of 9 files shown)


thomgabriel:
@claude re-review please. Addressed the paired-distractor grouping defect from the last review (now at head a709fe4, and the branch has also been rebased onto main over #718's grading-dedup lock).

Fix: `pairGroups` in `parsons-block.tsx` now groups by TARGET line id instead of per distractor. Previously two distractors authoring the same `pairedWith` target overwrote the target's group — orphaning the first look-alike into its own marker. Now the target's group is minted once and every distractor pointing at it joins it, so N look-alikes + their one target share a single marker with no orphan, regardless of authoring order.

Schema: left as-is on purpose — two traps for one real line is legitimate authoring (it sharpens the discrimination paired-distractors exist for) and the schema already permits repeated `pairedWith`, so no 1:1 `.refine()` was added. Added a renderer test: two distractors → one shared group of three, no stray group.

Verification: `pnpm typecheck` 6/6 green; `pnpm --filter web test` 185 files / 1654 tests pass.

PR #630, opened 2026-07-26
feat(db): xp_transactions.source typed column, derived from reason prefix (LX-B9a)

thomgabriel:
Closes #557 (LX-B9a, master spec §3).

Adds a CHECK-constrained `source` column to `xp_transactions` so future league scoring (LX-B9b) filters by a typed column — able to exclude creator/community XP — instead of parsing free-text reasons. **No league code in this PR.** SQL is **not applied anywhere**: it ships here for the gate/owner to apply dry-run-first.

## Call-site inventory (the spec's "6–7")

Every XP award write goes through exactly **two** SQL functions that INSERT into `xp_transactions` (`award_xp`, `award_community_xp`). Their callers:

**`award_xp` RPC — 6 call sites**
| # | Call site | Reason written | → source |
|---|-----------|----------------|----------|
| 1 | `lib/helius/event-handlers.ts:196` (handleLessonCompleted) | `Completed lesson: <lessonId \| index:N>` | `lesson` |
| 2 | `lib/helius/event-handlers.ts:263` (handleCourseFinalized bonus) | `Course completion bonus: <courseId>` | `course_completion` |
| 3 | `lib/helius/event-handlers.ts:289` (handleCourseFinalized creator) | `Creator reward: <courseId>` | `creator_reward` |
| 4 | `lib/helius/event-handlers.ts:368` (handleAchievementAwarded) | `Achievement reward: <achievementId>` | `achievement` |
| 5 | `lib/helius/event-handlers.ts:404` (handleXpRewarded) | arbitrary on-chain memo, or `XP reward`; historically incl. `daily_quest:<id>` | `platform` (or `quest` for daily_quest memos) |
| 6 | `lib/solana/onchain-queue.ts:554` (creditXpAndSettle — shared by the `xp`, `course_finalize`, and `quest_xp` queue rows) | payload reason, or fallbacks `Completed lesson: …` / `Completed course: <courseId>` / `daily_quest:<questId>` | `lesson` / `course_completion` / `quest` |
| +| `app/api/admin/resync/route.ts:257` (admin resync replay of #4) | `Achievement reward: <achievementId>` | `achievement` |

**`award_community_xp` RPC — 5 call sites, all `community:%` reasons → `community`**
[...]

PR #727, opened 2026-07-26
feat(gamification): streak forgiveness across all streak paths (#573)

### LINKED ISSUE #573: Streak forgiveness: one PR, all streak paths
Freeze storage + 3 DB writers (award_xp, award_community_xp, quest branch) + display/calendar merge + weekly cadence.

**Spec tasks:** `LX-B8`

thomgabriel:
Closes #573 (LX-B8). One coordinated PR across every streak code path: a missed day consumes a **streak freeze** instead of hard-resetting the streak, applied in **all three** persisted-state writers so a frozen streak survives whichever path fires first.

## Streak paths found (and touched)

| Path | Kind | What it computes | Change |
|------|------|------------------|--------|
| `award_xp` (schema.sql / migration) | DB writer | headline streak on `user_xp.current_streak` (learning XP) | gap now defers to shared `next_streak_value` (freeze-or-reset) |
| `award_community_xp` | DB writer | the **same** headline streak, from the community-XP path | replaced its inline `ELSE 1` CASE reset with the same shared decision (the writer the original audit missed) |
| `get_daily_quest_state` `login_streak` branch | DB writer | the login-streak **quest's own** per-quest streak (`user_daily_quests` period_start machine) | Case 3 (gap) consults the shared freeze log before breaking; first completion **grants** a freeze |
| `lib/gamification/streaks.ts` (`generateWeekCalendar` / `generateStreakCalendar` / `updateStreak`) | Display only | weekly circles / month calendar | renders freeze days as snowflakes; **no state write** (confirmed zero state-writing callers) |
| `dashboard-identity-panel.tsx` `buildHeatmapData` | Display only | 270-day activity heatmap | merges freeze days as snowflake cells |
| `hybrid-progress-service.getStreak` | Read | assembles `StreakData` | reads `streak_freezes` + `streak_freezes_used` (own-row RLS) |

The three DB writers agree **by construction**: they route through shared SECURITY DEFINER helpers (`next_streak_value`, `cover_missed_days_with_freezes`), not three copies.

## Forgiveness semantics

| Situation | Result |
|-----------|--------|
| Active today / yesterday | unchanged (keep / +1) |
| Gap, every missed day covered by a freeze | streak **survives** (+1); each missed day logged as a snowflake |
| Gap, not enough freezes | streak **resets** to 1 |
| Missed day already frozen by another path | **not re-charged** (idempotent per `(user_id, frozen_date)`) |
| Freeze inventory | `user_xp.streak_freezes`, **capped at 2** (CHECK 0..2 + `LEAST(..+1, 2)`) |
| Earning | server-side only — login_streak quest completion banks 1 (never client-minted) |
| Consumption | single guarded `UPDATE ... WHERE streak_freezes >= needed` — atomic, never over-consumes, never negative |

## Migration (SQL-only, NOT applied to prod)

`supabase/migrations/20260726190000_streak_forgiveness.sql` — idempotent (ADD COLUMN / CREATE TABLE IF NOT EXISTS, DROP CONSTRAINT/POLICY before create, CREATE OR REPLACE), explicit `BEGIN/COMMIT`, byte-mirrored into `supabase/schema.sql`, every helper `SECURITY DEFINER` + `SET search_path=''` + REVOKE/GRANT, own-row RLS SELECT on `streak_freezes_used` with **no** client write policy. Ships a **tested ROLLBACK** block that drops the new objects and restores both rewritten XP writers' prior bodies (get_daily_quest_state restored by re-applying its prior migration). Timestamp `20260726190000` — unique vs `supabase/migrations/` and the only open PR (#718, no migration).

## Display

`StreakData` gains `frozenDays: string[]` + `freezesRemaining: number`; snowflakes merge into both calendars and the heatmap (**activity always wins** over a freeze flag); new i18n strings `streakFrozenDay` / `streakFreezesRemaining` in **en / pt-BR / es**; dedicated `--freeze` cool-blue token (light + dark); snowflake cells carry `title`/`aria-label`.

## Verification

- `pnpm install --frozen-lockfile && pnpm typecheck` — green (6/6 packages)
- `pnpm --filter web test` — **183 files / 1641 tests pass** (includes 27 new: SQL guard + calendar-merge display tests)
- `pnpm --filter web lint` — clean (only pre-existing import-order warnings, none in touched files)

## Deviation

- **Weekly-cadence mode (spec F32)** intentionally **not built** — it is gated on the owner's A/B decision (pedagogy open question #2) and is not part of the LX-B8 acceptance criteria. Deferring it lets forgiveness ship now; documented in the migration header.

---

## Post-review update (commit b69bf0e)

**Concurrency fix (MAJOR, from adversarial review):** `award_xp` and `award_community_xp` hold *different* per-writer advisory locks, so on a gap-boundary day with exactly one freeze both could read `v_needed=1` from `cover_missed_days_with_freezes`' unlocked `SELECT COUNT` before either consumed; the loser's guarded decrement then found 0 freezes, returned FALSE, and **spuriously reset the streak while the freeze was burnt**. Fixed by taking a **shared `pg_advisory_xact_lock('streak:<uid>')` before the count** inside `cover_missed_days_with_freezes` — one choke point covering all three writers' gap paths (the two XP writers via `next_streak_value`, the quest path directly). The loser now runs after the winner commits, sees the day already in `streak_freezes_used` (`v_needed=0`), returns TRUE, and both streaks survive on one freeze. No deadlock: `streak:<uid>` is the only cross-writer lock and is always acquired after any per-writer key. Guard test extended to pin the exact key + lock-before-count ordering (migration + schema.sql). Also reverted `next_streak_value`'s `>= p_today` to `= p_today` for byte-parity with the base writers' reset-on-future-date behavior.

**Owner decision to flag — shared freeze inventory:** freezes live in a single `user_xp.streak_freezes` pool shared by the headline streak and the login_streak quest streak. Because the quest streak also consumes from this pool on its own gaps, a login_streak-quest gap can burn a freeze the headline streak later needs (and vice-versa). This is intentional for launch (one inventory, one mental model), but if you'd rather the two streaks not compete for freezes, that's a **separate-pools** follow-up — calling it out now so it's a conscious choice, not a surprise.

### DIFF
--- a/apps/web/src/components/gamification/__tests__/dashboard-identity-panel-quests.test.tsx
+++ b/apps/web/src/components/gamification/__tests__/dashboard-identity-panel-quests.test.tsx
@@ -15,6 +15,8 @@ const STREAK: StreakData = {
   longestStreak: 0,
   lastActivityDate: "",
   streakHistory: {},
+  frozenDays: [],
+  freezesRemaining: 0,
 };
 
 function quest(overrides: Partial<DailyQuest>): DailyQuest {
--- a/apps/web/src/components/gamification/dashboard-identity-panel.tsx
+++ b/apps/web/src/components/gamification/dashboard-identity-panel.tsx
@@ -34,6 +34,8 @@ interface HeatmapCell {
   level: 0 | 1 | 2 | 3 | 4;
   isToday: boolean;
   count: number;
+  /** An inactive day a streak freeze saved (LX-B8) — rendered as a snowflake. */
+  frozen: boolean;
 }
 
 interface HeatmapData {
@@ -86,11 +88,16 @@ function ordinal(n: number): string {
   }
 }
 
-function formatCellTooltip(dateStr: string, count: number): string {
+function formatCellTooltip(
+  dateStr: string,
+  count: number,
+  frozen: boolean
+): string {
   if (!dateStr) return "";
   const d = new Date(dateStr + "T00:00:00");
   const month = MONTH_FULL[d.getMonth()]!;
   const day = ordinal(d.getDate());
+  if (frozen) return `Streak frozen on ${month} ${day}`;
   if (count === 0) return `No activity on ${month} ${day}`;
   if (count === 1) return `1 lesson completed on ${month} ${day}`;
   return `${count} lessons completed on ${month} ${day}`;
@@ -104,7 +111,11 @@ function countToLevel(count: number): 0 | 1 | 2 | 3 | 4 {
   return 4;
 }
 
-function buildHeatmapData(streakHistory: Record<string, number>): HeatmapData {
+function buildHeatmapData(
+  streakHistory: Record<string, number>,
+  frozenDays: readonly string[] = []
+): HeatmapData {
+  const frozenSet = new Set(frozenDays);
   const today = new Date();
   const todayStr = today.toISOString().split("T")[0]!;
   const todayDow = today.getDay();
@@ -121,7 +132,13 @@ function buildHeatmapData(streakHistory: Record<string, number>): HeatmapData {
       const daysAgo = startOffset - cellIdx;
 
       if (daysAgo < 0) {
-        colCells.push({ date: "", level: 0, isToday: false, count: 0 });
+        colCells.push({
+          date: "",
+          level: 0,
+          isToday: false,
+          count: 0,
+          frozen: false,
+        });
         continue;
       }
 
@@ -131,8 +148,10 @@ function buildHeatmapData(streakHistory: Record<string, number>): HeatmapData {
       const isToday = dateStr === todayStr;
       const count = streakHistory[dateStr] ?? 0;
       const level = countToLevel(count);
+      // Activity always wins: a day is only "frozen" if it has no real activity.
+      const frozen = count === 0 && frozenSet.has(dateStr);
 
-      colCells.push({ date: dateStr, level, isToday, count });
+      colCells.push({ date: dateStr, level, isToday, count, frozen });
 
       if (row === 0 && !monthLabelMap.has(col)) {
         const month = d.getMonth();
@@ -376,8 +395,8 @@ export function DashboardIdentityPanel({
   }, []);
 
   const heatmap = useMemo(
-    () => buildHeatmapData(streak.streakHistory),
-    [streak.streakHistory]
+    () => buildHeatmapData(streak.streakHistory, streak.frozenDays),
+    [streak.streakHistory, streak.frozenDays]
   );
 
   // Tap-to-show tooltip state (mobile touch support)
@@ -543,6 +562,7 @@ export function DashboardIdentityPanel({
                                   cell.level === 2 && "l2",
                                   cell.level === 3 && "l3",
                                   cell.level === 4 && "l4",
+                                  cell.frozen && "frozen",
                                   cell.isToday && "today"
                                 )}
                                 onClick={(e) => {
@@ -559,7 +579,11 @@ export function DashboardIdentityPanel({
                                 sideOffset={6}
                                 side="top"
                               >
-                                {formatCellTooltip(cell.date, cell.count)}
+                                {formatCellTooltip(
+                                  cell.date,
+                                  cell.count,
+                                  cell.frozen
+                                )}
                                 <Tooltip.Arrow className="fill-[var(--card)]" />
                               </Tooltip.Content>
                             </Tooltip.Portal>
--- a/apps/web/src/components/gamification/streak-display.tsx
+++ b/apps/web/src/components/gamification/streak-display.tsx
@@ -2,7 +2,7 @@
 
 import { useTranslations } from "next-intl";
 import { useLocale } from "next-intl";
-import { Fire } from "@phosphor-icons/react";
+import { Fire, Snowflake } from "@phosphor-icons/react";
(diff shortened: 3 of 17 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (migration rewrites the two core XP-writer RPCs) → adversarial review + fix round + re-verify PASS → queuing for gate.**

**My independent verification** at `b69bf0e5`: typecheck exit 0; full web suite **183 files / 1643 tests, exit 0**.

**Adversarial round 1 (at `2aa93a9`): PASS with 1 MAJOR caveat.** Held under attack: XP-writer rewrites proven **behavior-preserving by normalized diff against base** (cap math, #630 source, idempotency-key path, level calc, RETURNS, errors all byte-equivalent — only the streak branch changed); rollback bodies **token-identical to base** (rollback can't brick XP); freeze farming blocked (grant RPC service_role-only, cap enforced twice, user_xp SELECT-only, log table no write policy); consume all-or-nothing, never negative, never double per (user, frozen_date); mirror logic-identical; guard mutation-tested. The MAJOR: the two XP writers held different advisory locks, so a freeze-consume race could spuriously reset the loser's streak and waste the freeze.

**Fix round (`b69bf0e5`) re-verified by the same reviewer: PASS.** Shared `streak:uid` xact-lock inside `cover_missed_days_with_freezes` BEFORE the count — one choke point for all three writers; race replayed: loser now blocks, sees the logged day, returns +1 (both streaks survive on one freeze). Deadlock-free (streak:uid is the only cross-writer lock, always after per-writer keys; quest path holds none). Guard extended and mutation-tested both ways (split-back and lock-after-count both FAIL the guard). Byte-parity of non-gap branches restored to base.

**For the gate + owner:**
- Migration `20260726190000` is SQL-only, applied nowhere — owner dry-run before any apply (it rewrites `award_xp`/`award_community_xp`; rollback block restores base bodies verbatim).
- Owner note (in PR body): freezes are ONE shared pool across the headline streak and the login_streak quest — a quest gap can burn a freeze the headline streak needs. Intentional for launch; separate pools possible later.
- Known limitation: no live-DB behavioral test harness in repo — SQL behavior pinned by static guard + the reviewer's semantic analysis, not executed SQL.

thomgabriel:
## Gate: PASS — merging.

### I checked the coverage claim independently, because it is a coverage claim

The title says "across **all** streak paths", and I have twice today under-asked "what else is in this category?" (the ungated `reflection-reply` in #720, the flow-style bypass in #715). So I enumerated the writers myself rather than reading the table.

Parsing `schema.sql` for functions whose body writes `current_streak`/`longest_streak`:

```
award_xp                writes_streak=True
award_community_xp      writes_streak=True
```

Exactly two SQL writers, both in the PR's table. The third streak-write site I found (`schema.sql:1683`) turned out to be **inside** `award_community_xp` — the inline `ELSE 1` CASE this PR replaces. Plus the quest-state `login_streak` branch and the display layer. Coverage confirmed.

Credit where due: the body notes `award_community_xp` was "the writer the original audit missed", which is the honest version of exactly the failure I keep making.

### The shared decision fixes a latent inconsistency, not just adds forgiveness

Before this, `award_xp` and `award_community_xp` each decided independently whether a streak broke. Two writers, one `user_xp.current_streak` — so which path fired first could change the answer. Routing both through `next_streak_value` makes that impossible, and that is worth more than the forgiveness feature itself.

### Freezes are not client-grantable

`streak_freezes_used`: RLS on, **own-row SELECT only** (a learner can see which days were frozen — correct UX), and **no write policy**. All five helpers are `SECURITY DEFINER`, `SET search_path = ''`, REVOKEd from PUBLIC/anon/authenticated, GRANTed to `service_role` only — including `grant_streak_freeze`, so a client cannot mint itself forgiveness.

### The unreviewed commit is the important one, again

Review is **15m3s stale** (20:54:18Z vs head 21:09:21Z). The uncovered commit is `fix: serialize freeze decision with a shared lock` — +23/−6 on the migration — and it fixes a real concurrency bug:

`award_xp` holds `'award_xp:<uid>'` and `award_community_xp` holds `'award_community_xp:<uid>'` — **different** advisory locks — and the quest path holds none. Concurrently, both could read the freeze count before either consumed; the loser then sees 0 freezes, returns FALSE, and **resets a streak while the freeze it needed was already burnt by the winner**. The learner loses the streak *and* the freeze.

The fix takes a single shared `pg_advisory_xact_lock(hashtext('streak:' || uid))` **before** the count, so the loser runs after the winner commits, sees the day already in `streak_freezes_used` (`v_needed = 0`), and returns TRUE — both streaks survive on one freeze.

I checked lock ordering for deadlock risk: each writer takes its own per-writer lock first, then the shared `streak:` lock. Consistent acquisition order across both paths, no inversion.

That is the **third** concurrency defect surfaced in this codebase today (#689's box advance, #692's weekly cap, this) — and unlike the first two, this one was caught by the orchestrator's own adversarial pass before it reached me.

### Research grounding

The body makes **no causal-evidence claim**, which is the right call: the unified spec marks PED-12 **WEAKENED** ("the Peru RCT tested message salience, not freezes … do not cite Peru as proof streaks teach"). No misattribution to correct. Weekly-cadence mode (F32) is explicitly deferred pending your A/B decision rather than quietly built.

Migration not applied — batching with #707, #702 and #720's, all merged and pending.

PR #728, opened 2026-07-26
ci: make SHA-pin guard YAML-aware — close flow-style bypass (#723)

### LINKED ISSUE #723: ci: SHA-pin guard is blind to flow-style YAML uses: (demonstrated bypass, pre-dates #715)
Found by the adversarial review of #715 (which PASSed — the gap pre-exists that PR and is not widened by it; the grep line is unchanged in its diff).

## The bypass (demonstrated, reproducible)

`scripts/check-action-pins.sh:72` matches `uses:` only line-anchored (`^[[:space:]]*(-[[:space:]]+)?uses:`). Flow-style YAML — which GitHub Actions' parser accepts — is invisible to it. Both of these pass the guard green (exit 0, 'All … SHA-pinned ✓') while unpinned:

```yaml
    steps: [{uses: actions/checkout@v4}]
```
```yaml
    steps:
      - { uses: actions/checkout@v4 }
```

Repro: place either in `$FX/.github/workflows/x.yml`, run `CHECK_ACTION_PINS_ROOT=$FX bash scripts/check-action-pins.sh`.

## Also noted (minor)
- Symlinked composite `action.yml` skipped (`find -type f` excludes symlinks; the workflow glob DOES follow them — asymmetric).
- A 40-hex branch name is indistinguishable from a SHA offline (inherent limit; note only).

## Fix shape
Make the matcher YAML-aware (parse with a real YAML parser and walk every `uses` key at any nesting — the self-test's fixture seam makes this safely testable), add flow-style reject cases to the self-test, and include symlinked action files. At minimum, stop the self-test header claiming it proves rejection of *all* unpinned refs.

Severity: an author would have to deliberately write flow-style to sneak past — this guards against malice, not accident, which is exactly what the guard exists for (#635).

thomgabriel:
Closes #723.

## The bypass (from #723, now fixed)

`scripts/check-action-pins.sh` matched `uses:` with a **line-anchored grep**, so flow-style YAML — which the GitHub Actions parser accepts — was invisible. Both of these passed the guard green while unpinned:

```yaml
steps: [{uses: actions/checkout@v4}]
```
```yaml
- { uses: actions/checkout@v4 }
```

An author had to write flow-style deliberately to sneak past — precisely the malice this guard exists to stop (#635).

## Fix

- **YAML-aware matcher.** New `scripts/check-action-pins.py` parses each file with a real YAML parser (PyYAML) and walks **every `uses` key at any nesting, in block and flow style**. `check-action-pins.sh` keeps file discovery + the `CHECK_ACTION_PINS_ROOT` test seam and delegates matching to it.
- **Symlinked composite actions (#723 minor).** `find -L` so a symlinked `action.yml` is scanned — symmetric with the workflow glob, which already follows symlinks. The old `find -type f` skipped them.
- **Self-test header (#723 minor).** No longer claims to prove rejection of *all* unpinned refs; it now states it exercises representative surfaces, and new evasions should land as a new failing case first.

## Runtime choice: python3 + PyYAML — and why

The runner (`ubuntu-latest`) is guaranteed to have **python3**, **node 22**, and **`yq`**. I chose python3 + PyYAML because:
- It's a **real YAML parser** (handles flow + block), unlike node (no built-in YAML) which would need an `npm install`.
- **No install at guard time and no third-party Action to pin** — so the guard can't reintroduce the unpinned-dependency risk it exists to prevent (node/`js-yaml` or `pip install` would).
- PyYAML ships with the **pre-installed `yamllint`** on the runner, and it's present on typical dev machines (verified locally, PyYAML 6.0.3). `yq` is guaranteed on the runner but absent locally, so I couldn't verify against it.
- **Fail-closed:** a missing PyYAML or an unparseable file exits non-zero — never a vacuous pass.

## Verification (local, bash 3.2.57 + python3.13/PyYAML 6.0.3 on macOS)

```
$ ./scripts/__tests__/check-action-pins.test.sh
ok   — fully pinned workflow + composite action + ./ and docker:// exclusions (exit 0)
ok   — unpinned tag in workflow is rejected (exit 1)
ok   — unpinned tag in composite action is rejected (#681) (exit 1)
ok   — empty .github fails loudly (no vacuous pass) (exit 1)
ok   — branch ref / short SHA rejected (exit 1)
ok   — flow-style mapping unpinned is rejected (#723) (exit 1)
ok   — flow-style sequence unpinned is rejected (#723) (exit 1)
ok   — flow-style correctly pinned is accepted (exit 0)
ok   — symlinked composite action.yml is scanned (#723) (exit 1)
check-action-pins.test.sh: all cases passed ✓

$ ./scripts/check-action-pins.sh
All GitHub Actions `uses:` refs are SHA-pinned across 8 file(s) in .github/workflows/ and .github/actions/. ✓
```

Also ran: the exact #723 flow-style repro → now exits 1 with a correct line number; quoted refs / `docker://…@sha256` / local reusable-workflow `uses:` → correctly pass; `shellcheck` on both shell files → clean (rc 0); `python3 -m py_compile` on the helper → OK.

Scope is `scripts/**` only (the guard workflow needed no change — python3 is pre-installed). No TS/Rust touched, so no web suite / typecheck.

> SENSITIVE (guards `.github/workflows/**`) — do not merge without adversarial review + gate. The `CHECK_ACTION_PINS_ROOT` seam lets a reviewer re-attack it directly.

### DIFF
--- a/.github/workflows/action-pin-guard.yml
+++ b/.github/workflows/action-pin-guard.yml
@@ -31,6 +31,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      # The guard's parser needs PyYAML. It already ships with the runner's
+      # pre-installed yamllint, so this is a fallback that pins the dependency
+      # explicitly rather than relying on the image incidentally carrying it (a
+      # future image change would otherwise break the guard). apt's distro
+      # package, not `pip install --user`: ubuntu-24.04's system python is
+      # externally-managed (PEP 668) and refuses --user installs.
+      - name: Ensure PyYAML (guard parser dependency)
+        run: |
+          python3 -c 'import yaml' 2>/dev/null || {
+            sudo apt-get update -qq
+            sudo apt-get install -y python3-yaml
+          }
       # Prove the guard actually rejects unpinned refs (in workflows AND composite
       # actions) and accepts pinned trees before trusting its verdict — #662 /
       # #614 "don't ship an unexercised guard".
--- a/scripts/__tests__/check-action-pins.test.sh
+++ b/scripts/__tests__/check-action-pins.test.sh
@@ -1,13 +1,17 @@
 #!/bin/bash
 # Exercises scripts/check-action-pins.sh against fixture trees so the guard is
 # never shipped unexercised (#662 "prove it fails on a bad line"; #614). Runs in
-# the Action pin guard workflow *before* the real scan, so every PR proves the
-# guard both accepts a fully-pinned tree and rejects an unpinned ref — in a
-# workflow file AND in a composite action (#681).
+# the Action pin guard workflow *before* the real scan.
 #
-# Pure bash, no test framework: the guard itself is pure bash and CI runs the
-# monorepo vitest via `pnpm -r test`, not the root `scripts/` suite, so a vitest
-# spec here would not gate. Self-contained + exit-code assertions do.
+# These are representative cases, not a proof of completeness: they show the
+# guard accepts a fully-pinned tree and rejects unpinned refs across the surfaces
+# that have bitten us — block style in a workflow AND a composite action (#681),
+# and flow style `[{uses: ...}]` / `- { uses: ... }` and symlinked composite
+# files (#723). New evasions should arrive here as a new failing case first.
+#
+# Pure bash, no test framework: CI runs the monorepo vitest via `pnpm -r test`,
+# not the root `scripts/` suite, so a vitest spec here would not gate.
+# Self-contained + exit-code assertions do.
 set -uo pipefail
 
 guard="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)/check-action-pins.sh"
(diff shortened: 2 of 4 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (`.github/workflows/**` + the guard itself) → adversarial review PASS + fix round verified → queuing for gate.**

**Adversarial review (same reviewer that demonstrated the #723 bypasses): PASS.** Both original bypasses now rejected; full evasion matrix held — anchors/aliases (resolved via compose), merge keys, duplicate keys, multi-document files, tag-wrapped scalars, reusable-workflow `jobs.<id>.uses` (correctly treated as pinnable), `run: |` prose (correctly NOT flagged — old blind spot gone), docker://+./ exclusions in flow style. Fail-closed proven on every dependency path: python3 missing → rc=127; PyYAML import failure → rc=2 with explicit error; corrupt YAML → rc=1 'failing closed'. No vacuous-pass route exists.

**Fix round at `cf543c3b` (both review MINORs), independently re-verified by me locally:**
- Over-match fixed: matching scoped to step contexts + workflow-level reusable calls; `env:`/`with:` keys named `uses` no longer false-reject (new self-test case) and unpinned reusable-workflow calls still reject (new case). Self-test 11/11 green; real guard green across all 8 files.
- PyYAML dependency made explicit: guard workflow ensures it via `python3 -c 'import yaml' || apt-get install python3-yaml` (PEP 668-aware — system python refuses --user pip). The added step introduces no new action refs; checkout remains SHA-pinned v7.0.1.

**PR diff vs main is exactly the 4 guard files** (the branch's older commits are already in main — GitHub's computed diff is clean).

COSMETIC recorded, no action: alias violations report the anchor-definition line; merge-key case double-annotates.

thomgabriel:
## Gate: PASS — merging. This closes a bypass in a guard **I twice declared verified by attack.**

### The bypass is real — I reproduced it against `main` before reading the fix

```
- uses: actions/checkout@v4                     FAIL  (caught, as expected)
- {uses: actions/checkout@v4}                   PASS  ← bypass
steps: [{uses: actions/checkout@v4}]            PASS  ← bypass
- {name: x, uses: actions/checkout@v4}          PASS  ← bypass
```

GitHub Actions accepts flow-style YAML, so an unpinned action written that way sailed straight past a green guard.

**I mutation-tested this guard twice** — 8 violation classes on #680, 6 composite-action cases on #715 — and never tried flow style. My attack surface was shaped by what I imagined an author would *write*, not by what the YAML spec *permits*. That is exactly the gap a grep-based checker has and a parser does not, and it is the argument for this rewrite independent of any single bypass.

### Fix verified — 11/11, no regressions

Bypass closed (all three flow variants now caught, and a **pinned** flow-style ref correctly passes — no false positive), and the whole prior suite still holds: block-style pinned/unpinned, short SHA, `./` local, `docker://`, comments, and `uses:` inside `run:` prose.

### I attacked the unreviewed commit specifically

The review is **11m27s stale** (20:59:13Z vs head 21:10:40Z), and the uncovered commit is not cosmetic: **+67/−15 on the parser**, "scope pin guard to step/job contexts". Scoping can narrow detection as easily as it removes false positives, so I tested both directions:

| still caught | |
|---|---|
| job-level reusable workflow (`jobs.<id>.uses`) | ✓ |
| step inside a `matrix` job | ✓ |
| second job unpinned, first pinned | ✓ |

| correctly ignored | |
|---|---|
| `uses:` as a `with:` **input value** | ✓ |
| `uses:` inside `env:` | ✓ |

So the scoping distinguishes a real step/job `uses:` from a string that merely happens to be keyed `uses` — which is the thing a naive YAML walk gets wrong, and which the old grep could never have distinguished at all.

### The dependency trade is handled correctly

This replaces "pure bash + grep — no third-party tool, so the guard can't itself reintroduce an unpinned dependency" with python3 + PyYAML. That was a **stated design property**, so it deserves scrutiny — and the important part is what happens when the parser is missing. I simulated an `ImportError`:

```
exit=2
::error::check-action-pins: PyYAML is required but `import yaml` failed … The guard fails closed rather than skip its check.
```

**Fails closed.** A guard that silently skips when its parser is unavailable would be strictly worse than the grep version it replaces; this one refuses to run instead. The workflow also ensures the dep via `sudo apt-get install -y python3-yaml`, with the reasoning recorded (ubuntu-24.04's system python is externally-managed, so `pip install --user` would fail) — and no third-party Action and no registry install at guard time, so the spirit of the original property survives.

Self-test still runs before the real scan.

PR #729, opened 2026-07-26
fix(ai): bring reflection-reply under the #591 spend ledger + 3 review minors (#724)

### LINKED ISSUE #724: ai: reflection-reply spends the sponsored Gemini key outside the #591 ledger (fail-open, uncapped, under-reported)
Found by the adversarial review of #720 (which PASSed — this path is outside that PR's declared /api/ai/partner scope, and its own comment defers spend accounting to #590's territory).

## The gap

`apps/web/src/lib/ai/reflection-reply.ts:71` (`maybeGenerateReflectionReply`, called from `/api/lessons/reflect`) spends the same `GEMINI_API_KEY` (gemini-3.5-flash, ≤512 output tokens/call) with **no `checkAiSpend` and no `recordAiSpend`**. Its only throttle is a per-user 10/min limiter that is **fail-open** (`reflection-reply.ts:57`, no `failClosed:true`) — the exact posture #591 eliminated on cost paths.

Consequences while enabled: the sponsor envelope from #720 is incomplete (this spend is uncapped by the new caps), and `get_ai_spend_today()` under-reports true key burn — defeating 'visible before the invoice'.

**Mitigant:** default-OFF behind `OPENENDED_AI_REPLY !== "1"` (`reflection-reply.ts:47`). The #720 envelope guarantee is true exactly as long as that flag stays off.

## Fix
1. Wrap the reflection reply in the same `checkAiSpend` (account + IP + global) / `recordAiSpend` pair, fail-closed (a skipped nicety reply is free — this path degrades gracefully by design).
2. Flip its limiter to `failClosed: true`.

## Fold in the review's MINORs (same files, same PR)
- `route.ts`: `recordBilledAssist` runs before `response.json()` — a 200 with a non-JSON body records the assist but not the spend. Record spend in a finally / before parse, and record a conservative estimate when `usageMetadata` is absent (currently books $0 for a real call).
- Document the bounded TOCTOU residual in-code: N in-flight requests can overshoot the hard cap by ~in-flight × $0.006 (bounded by the 20/min user + 600/min IP limiters).
- `usdNum` accepts 0 (self-DoS: denies every request). Fail-closed so safe, but boot-warn or reject 0 for the hard caps.

Related: #590 (billed latch), #591 (the ledger, PR #720).

thomgabriel:
Closes #724. Found by the adversarial review of #720.

## What & why
The reflection-reply path (`lib/ai/reflection-reply.ts`, called from `/api/lessons/reflect`) spends the **same** sponsored Gemini key as the AI Partner but sat **outside** the #591 ledger: no `checkAiSpend`, no `recordAiSpend`, and a **fail-open** 10/min limiter. Latent today (default-OFF behind `OPENENDED_AI_REPLY !== "1"`), but the moment it's enabled the sponsor envelope would be incomplete and `get_ai_spend_today()` would under-report true key burn.

## Main fix — reflection reply now runs under the ledger
- Gate the reply through `checkAiSpend` (account + IP + global) and book actual usage via `recordAiSpend` — the same ledger as the Partner.
- **Fail-closed by SKIPPING**, not 503: a hard-cap denial or a ledger outage (`checkAiSpend` fails closed to `denied` on its own) returns `null`, and the seal still ships. The reflect route's **receipt-first separation is preserved** — reflection SAVING never depends on the reply (verified: the seal is computed and returned before the reply, in its own try/catch).
- Over a **soft cap**, degrade to a shorter reply (512 → 256 output tokens) instead of skipping.
- Flip the reply limiter to `failClosed: true`.
- Thread the client IP from the route into the helper (per-IP dimension — free wallet signup means per-user keys can't bound a Sybil).

## Folded-in review minors (same files)
1. **Record spend on every billed call.** `route.ts` parsed `response.json()` after flipping `billed = true`, so a 200 with a non-JSON body recorded the assist but not the spend. Now the envelope is parsed **defensively** (can't throw past the booking), and `recordAiSpend` books a **conservative fallback** (full output budget + a generous input estimate) when `usageMetadata` is absent — instead of $0. Same fallback wired into the reflection path.
2. **TOCTOU residual documented in-code** at the gate: N in-flight requests can overshoot a hard cap by ~N × per-call cost (a full-budget propose ≈ $0.006), bounded by the fail-closed 20/min user + 600/min IP limiters — an accepted, self-limiting residual, not a hole.
3. **Reject a hard-cap env of 0 at boot** (`usdNumHard`, `gt(0)`): a $0 hard cap denies every request (self-DoS). Soft caps of 0 still boot (always-degrade is legitimate).

## No migration
Reuses the #591 RPCs (`check_ai_spend` / `record_ai_spend`). No DB change.

## Verification
- `pnpm typecheck` — 6 packages successful, zero errors, zero `any`.
- `pnpm --filter web test` — **186 files, 1667 tests passed**.
- `pnpm --filter web lint` — exit 0 (only pre-existing import/order warnings).

New/updated tests: reply denied at cap → skipped + reflection still saves; ledger down → no model call + reflection still saves; spend recorded on reply (usageMetadata → ledger); conservative-fallback booking when usageMetadata absent; fail-closed reply limiter (flag + store-throw); soft-cap degrade to 256; route passes the client IP; the `recordAiSpend`-ordering fix (a non-JSON 200 still books spend + billed-assist, 502, no refund); hard-cap `=0` boot rejection (all three hard caps) while a soft cap of 0 still boots.

Do not merge — for gate review.

### DIFF
--- a/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
+++ b/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
@@ -1079,13 +1079,54 @@ describe("POST /api/ai/partner", () => {
       generationConfig: { maxOutputTokens: number };
     };
     expect(sent.generationConfig.maxOutputTokens).toBe(2048);
-    // The actual usage is booked against the ledger with the request's IP.
+    // The actual usage is booked against the ledger with the request's IP, plus
+    // a conservative fallback (full output budget + generous input) used only
+    // when usageMetadata is absent (#724).
     expect(recordAiSpend).toHaveBeenCalledTimes(1);
-    expect(recordAiSpend).toHaveBeenCalledWith("user-1", "203.0.113.9", {
-      promptTokenCount: 1000,
-      candidatesTokenCount: 500,
-      thoughtsTokenCount: 0,
-    });
+    expect(recordAiSpend).toHaveBeenCalledWith(
+      "user-1",
+      "203.0.113.9",
+      {
+        promptTokenCount: 1000,
+        candidatesTokenCount: 500,
+        thoughtsTokenCount: 0,
+      },
+      { promptTokens: 6000, outputTokens: 2048 }
+    );
+  });
+
+  it("records spend on a 200 with a NON-JSON body (billed but unparseable, #724)", async () => {
+    // The 200 billed us, but response.json() throws — the ordering fix must still
+    // book the spend (with the fallback) and the billed-assist, then 502.
+    checkAiSpend.mockResolvedValue({ decision: "full" });
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(async () => ({
+        ok: true,
+        text: async () => "",
+        json: async () => {
+          throw new SyntaxError("Unexpected token < in JSON");
+        },
+      }))
+    );
+
+    const { POST } = await import("../partner/route");
+    const res = await POST(makeRequest(VALID_BODY));
+
+    expect(res.status).toBe(502);
+    // Billed audit + spend both recorded despite the unparseable body.
+    expect(recordBilledAssist).toHaveBeenCalledTimes(1);
+    expect(recordAiSpend).toHaveBeenCalledTimes(1);
+    // usageMetadata is undefined (no parse), so the conservative fallback carries
+    // the estimate — never $0 for a real billed call.
+    expect(recordAiSpend).toHaveBeenCalledWith(
+      "user-1",
+      "203.0.113.9",
+      undefined,
+      { promptTokens: 6000, outputTokens: 2048 }
+    );
+    // Not refunded — Gemini billed us.
+    expect(refundAssist).not.toHaveBeenCalled();
   });
 
   it("soft cap: DEGRADES to a shorter output budget but still serves (200)", async () => {
--- a/apps/web/src/app/api/ai/partner/route.ts
+++ b/apps/web/src/app/api/ai/partner/route.ts
@@ -14,6 +14,7 @@ import {
   recordAiSpend,
   degradedMaxTokens,
 } from "@/lib/ai/spend-ledger";
+import type { GeminiUsageMetadata } from "@/lib/ai/spend-ledger";
 import { sealCheck } from "@/lib/ai/check-seal";
 import {
   buildStaticPrefix,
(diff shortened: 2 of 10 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (billing path, area:ai+db) → adversarial review PASS → queuing for gate.**

**My independent verification** (clean clone at `a385ea0a`): `pnpm typecheck` exit 0; full web suite **186 files / 1667 tests, exit 0**.

**Adversarial review: PASS, claim-by-claim** (same hostile reviewer that found #724 on #720 — warm context):
- Reflection reply gated BEFORE the model call; denied/ledger-down/limiter-throw → reply skipped (null), **seal still saves** (computed first, reply in its own try/catch) — the receipt-first separation held under attack.
- IP server-derived (same Vercel-trusted chain as partner), userId from `getUser()` — neither client-suppliable.
- Partner-route ordering fix verified non-vacuously: non-JSON 200 → books spend (conservative fallback) + billed assist + 502, no refund; success path byte-equivalent; over-booking quantified at ≈ real cost ($0.0069 vs ~$0.006) — no false cap trips.
- Boot guard: hard-cap 0/negative/garbage all reject; soft-cap 0 boots (always-degrade); partial config safe.
- Full #720 regression sweep clean (gate-before-spendAssist, awaited records, fail-closed 503, refund only when unbilled; no migration in this PR).

**Residuals recorded (all advisory, none blocking):** abort-mid-call books nothing (bounded, pre-existing); partial usageMetadata under-books (Gemini-controlled); **rate-var envs (`AI_SPEND_*_USD_PER_MTOK`) still accept 0 → every booking $0 → caps never trip** — config-only footgun, suggest extending the gt(0) guard to rates in a future config-hardening pass (gate's call whether to demand it now); env test lacks explicit negative/empty hard-cap cases (correct by construction).

Closes the #720 envelope hole: both Gemini-spending paths now run under the same fail-closed ledger.

thomgabriel:
## Gate: PASS — merging. This closes a gap **I missed** gating #720.

### Owning the miss first

I gated #720 and verified: the caps matched the O-1 record line for line, the RLS surface (no policies, three service_role-only `SECURITY DEFINER` RPCs), that fail-closed was the *right direction* for a spend ceiling versus `check_rate_limit`'s fail-open, that `checkAiSpend` ran before `spendAssist` so a denial burns no assist, and that no spend gate sat in the completion path.

What I never asked was the coverage question: **is this every path that spends the key?** It was not. On `main` right now:

```
app/api/ai/partner/route.ts     gated
lib/ai/reflection-reply.ts      NOT GATED   ← same sponsored Gemini key
```

That is the same lens I applied to a *content* PR two hours earlier — I audited #10's tutorNotes coverage across every launch-catalog challenge and found four gaps. I applied "did it cover everything?" to content and not to cost. Worth recording, because a spend ceiling that covers one of two paths is not a ceiling.

### Severity is genuinely low, and I want to be accurate rather than dramatic

The path is **default-OFF**: `reflection-reply.ts:74` returns `null` unless `OPENENDED_AI_REPLY === "1"`, and that variable is not set in the environment. So nothing was leaking. This was a latent hole that would have opened the moment the flag flipped — at which point `get_ai_spend_today()` would have under-reported true key burn while looking authoritative, which is the worse failure: a ledger that is confidently wrong.

### The fix is right, including the part that is easy to get wrong

- **Fail-closed by skipping, not 503.** A hard-cap denial or ledger outage returns `null` and **the seal still ships**. Verified in the source: the header states it, and the reflect route computes and returns the seal before the reply, in its own try/catch.

  That distinction is the whole game. `checkAiSpend` failing closed is correct for spend; making *reflection saving* fail with it would have been wrong. A learner's own submission must never be blocked because an AI extra was denied — same principle as #720 not sitting in the completion path.
- **Soft cap degrades** (512 → 256 output tokens) rather than skipping outright, matching the Partner's degrade-first behaviour.
- **Limiter flipped to `failClosed: true`**, consistent with the ledger it now sits behind.
- **Client IP threaded from the route**, so the per-IP dimension actually applies here — without it the per-IP cap would have been silently unenforced on this path, which is precisely the class of bug being fixed.

Post-#729, both Gemini paths are gated. Verified against the PR head, not the description.

Credited to "the adversarial review of #720" — which is the system working: the automated review caught the coverage question I did not ask.

Issue #711, opened 2026-07-26
content: retired DeFi course still renders in the catalog — path draft:true does not hide its courses (D-1, confirmed live)


thomgabriel:
**Confirmed at runtime, not inferred.** With the activated bundle (`content.lock` = `012cd03`, #679) running locally, `GET /en/courses` returns 200 and the page contains:

| course / path | in catalog |
|---|---|
| Zero to Deployed (path) | **yes** ✓ expected |
| Solana Fundamentals | **yes** ✓ expected |
| **DeFi** | **yes** ← the problem |
| Getting Paid / Stablecoin (C5) | no ✓ correct — awaits on-chain sync |
| Rust for Program Developers (C2) | no ✓ correct — #12 unmerged |

## Why this is a problem

`paths/defi.yaml` was set to `draft: true` with `courses: []` as the mechanism for retiring the DeFi track (courses-academy #7). The salvage ledger (`docs/superpowers/specs/2026-07-26-course-salvage-ledger.md` §5) retires `defi-on-solana` outright — **no successor** — salvaging only `token-standards` prose into C5 L2, and marks all five of its challenges `DELETE`.

But the course still renders in the catalog, because **the catalog lists courses independently of paths**. Drafting the path hides the *path*, not its courses. So a learner can still enrol in and work through a course the curriculum has retired, whose challenges are slated for deletion and whose prose has known currency problems (that is why courses-academy #9 is trying to fix its Token-2022 table).

This was flagged as open decision **D-1** and carried as a hypothetical — "*may* still render". It renders.

## Why it matters more now than an hour ago

Before #679 the bundle was pinned pre-activation, so this was latent. #679 activated the catalog, which means **this is the live state a learner sees today**.

It also changes the priority of courses-academy#9: I bounced that PR partly arguing the Token-2022 fix mattered because learners might still reach the page. They can, so that reasoning is now confirmed rather than precautionary.

## Options

1. **Set `isActive: false` (or equivalent) on `defi-on-solana`'s course doc** so the catalog gate excludes it, the same mechanism that keeps C5 hidden until it syncs. Cleanest — content-only, no code, reversible, and it does not touch the on-chain account.
2. **Close the course on-chain** as part of the devnet window (#606/#607). Heavier and irreversible-ish; also the enrolment data question applies.
3. **Accept it for launch** and let it sit as a legacy course. Defensible only if someone decides that deliberately — right now it is live *by omission*, which is the part worth fixing regardless of which option wins.

[...]

Issue #578, opened 2026-07-25
Test-out for segment 2 (server-internal batch completion)


thomgabriel:
10-question course challenge → retroactive per-lesson XP via existing rails.

**Spec tasks:** `LX-A5`

Full context — what to build (file-level), acceptance criteria, effort, dependencies and wave — is in [`docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md`](docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md), section 3 (Workstreams) under the task ids above; section 4 has the delivery sequencing.

Derived from the 2026-07-25 research program (pedagogy / UI-UX / personalization+launch reports in the same directory), code-audited before filing.

thomgabriel:
Done via PR #730 (merged by the gate after its own adversarial pass — the auto-close reference was lost in a PR-body edit). Shipped: 10-question deterministic per-(user,course) challenge, server-authoritative grading, batch completion through the existing completeLesson primitive (XP + credential ride the normal webhook rail — capstone credential still requires a real deploy, verified by the gate), scoped server-side to live path-member courses (staged C2 and retired defi NOT eligible), min-pool floor of 8, fail-closed XP cap. Residual for the owner (flagged in the PR body, does not block): explicit ack of the open-book posture — a pass grants the same XP as completing every lesson, which the gate assessed as no-arbitrage (retroactive, equal XP, no leaderboard advantage).

PR #730, opened 2026-07-26
feat(onboarding): segment-2 test-out via server-internal batch completion (LX-A5, #578)

thomgabriel:
## What

Implements **LX-A5 / #578** — a segment-2 **test-out**: a 10-question course challenge that, on pass, batch-completes the course's lessons server-side through the existing per-lesson rails and grants the same retroactive per-lesson XP. An **offer, not a gate** — skipping never costs leaderboard position (I5).

## Why

Per the launch-experience master spec §3 (LX-A5): a learner who already knows a course's material should be able to skip it without losing the XP they'd have earned lesson-by-lesson. The XP is retroactive and equal to normal completion, so there is nothing to game.

## How it works

- **`lib/courses/test-out.ts`** (pure) — assembles the challenge from the course's authored quiz-block questions (LX-D2 substrate). Deterministic per-`(user, course)` selection via a seeded shuffle, so the GET (fetch) and POST (grade) halves agree on the same set with **no stored state** and the client can't cherry-pick questions. The answer key is **stripped** for transport; grading is server-authoritative set-equality, identical to the in-lesson quiz grader — the client verdict is never trusted.
- **`lib/lessons/batch-complete.ts`** — drives the existing `completeLesson` primitive once per incomplete lesson (any-order via the backend signer; **zero on-chain change**). Idempotent (re-reads the on-chain bitmap and skips completed lessons), fail-closed, and a single lesson's on-chain failure is isolated into `pending` rather than aborting the batch. Enforces the spec's XP cap.
- **`app/api/courses/test-out/route.ts`** — `GET` returns the challenge; `POST` re-grades and, on pass, batch-completes **service-role-side** and routes the learner onward to the next course in the path.
- Test-out **page** + challenge **UI** (a11y: `fieldset`/`legend`, radio/checkbox, `aria-live`), **i18n ×3** (en / pt-BR / es).

## XP & on-chain semantics

| Concern | How it's handled |
|---|---|
| Per-lesson XP | Minted by the **existing `LessonCompleted` Helius webhook** off each on-chain event (`source = lesson`). The batch never calls `award_xp` → **no double-credit surface**. |
| Course-completion bonus + credential | Ride the same webhook rail once the final lesson lands; credential stays capstone-gated (LX-E2) — test-out grants lesson XP + completion, not the capstone credential. |
| Idempotency | On-chain bitmap re-read skips completed lessons; the program's own double-complete guard is the ultimate backstop. Re-submitting a passed test-out retries only the still-unset lessons. |
| #459 volume gate | The batch is invoked internally with the service-role client and **never traverses** `/api/lessons/complete`, so it is exempt **by construction**, not by loosening the gate. Normal traffic is untouched. |
| XP cap | Refuses when `xpPerLesson × liveLessonCount > 10000` (spec acceptance). |

## Anti-abuse

- Server-graded (forged client verdict cannot force a pass — test-covered).
- Attempt gates: per-user `courses:test-out` (10/hr) + per-IP `courses:test-out:ip` (200/hr) + a per-`(user,course)` in-flight lock, released in `finally`.
- Same platform-freeze / program-live / course-maintenance guards as the complete route.

## LX-A5 acceptance checklist

- [x] Segment-2 user passes test-out → all live lessons complete + XP minted via existing rails (webhook, `source = lesson`).
- [x] `#459` gate unchanged for normal traffic (batch never uses that route).
- [x] Capped by `xpPerLesson × lessonCount ≤ 10000`.
- [x] Offer, not gate; retroactive XP; onward routing to the next course.
- [x] i18n en/pt-BR/es; a11y challenge UI.

## Verification

`pnpm install --frozen-lockfile && pnpm typecheck && pnpm --filter web test` — **typecheck exit 0**, **1678/1678 tests pass** (188 files), including ~32 new tests: pass grants XP once + idempotent on re-submit, fail/partial grants nothing, rate limit, server re-grade rejects forged results, extraneous-id rejection, clamp to on-chain lesson count, enrollment_missing/xp_cap mapping, segment routing after pass.

## Migration

**None.** Completions use `user_progress`, XP uses the existing webhook rail, attempts use the existing `rate_limits` table — no new tables or columns.

## Deviations / notes

- **Enrollment required:** `completeLesson` needs the on-chain enrollment PDA, so test-out returns `403 enrollment_missing` if the learner isn't enrolled — same as `/api/lessons/complete`. The offer surface should ensure enrollment first.
- **Surfacing:** ships the mechanism + a dedicated page at `/courses/[slug]/test-out` and onward routing. The "Test out" **entry button** on the path/course-detail surfaces belongs to LX-A7 / LX-B14 (which own those contested files); left unwired here to avoid collisions.
- **Course-general:** built for any course with a quiz pool (not hardcoded to fundamentals) — the offer surface decides who sees it, and the per-module follow-on reuses the same rails.
- **`maxDuration = 300`** on the route: a pass drives N sequential confirmed on-chain txs; noted as an operational consideration.

---

## Update — adversarial-review follow-ups (commit 3b510a1)

- **Server-side scoping** (`isCourseTestOutEligible`): test-out is now refused unless the course is **live (synced+active)** AND a **learning-path member** — the spec's flagship-path scope. Enforced in the route (GET + POST) and the page; UI-only scoping would be bypassable. Closes the held-defi concern (#711). Not hardcoded to a course id, so the per-module follow-on and a future JS/TS entry-rung course stay in scope automatically. Stable reason `test_out_unavailable`.
- **Minimum pool floor** (`MIN_TEST_OUT_POOL = 8`): refuse test-out when the authoritative question pool `< 8`, so a sparse-quiz course can't gate all its lessons' XP behind 1–2 open-book answers. Stable, localized reason `insufficient_questions`.
- **Cap fail-closed:** `batchCompleteCourse` now refuses with `xp_unavailable` (→ 409) when `fetchCourse` yields a missing/zero `xp_per_lesson`, instead of `?? 0` silently zeroing the product and skipping the XP cap.

New tests: scoping (ineligible → 404 `test_out_unavailable`, GET+POST), pool-floor refusal, cap fail-closed (undefined + zero xp), `isCourseTestOutEligible` unit. Full suite: **189 files / 1691 tests green**, typecheck exit 0.

## Owner ack requested — open-book quiz posture

Test-out is assembled from the course's authored quiz questions, whose `correct` flags are **client-visible by design (D4 open-book)**. A determined learner can therefore read the answers and pass without knowing the material. This is **consistent with the accepted XP posture** and deliberately left as-is:

- A pass grants **exactly the same XP** as completing every lesson normally — there is no prize for skipping, so nothing to farm.
- The **capstone credential stays separately gated** by the on-chain deploy check (LX-E2); test-out grants lesson XP + completion, never the credential.
- The **min-pool floor (≥8 @ 80%)** now bounds the worst case so a course can't be skipped on 1–2 answers.

Requesting explicit owner/gate acknowledgement of this tradeoff rather than adding answer-hiding/anti-cheat machinery (a non-goal per spec §5.4).


thomgabriel:
**Orchestrator routing record — XP-issuance amplifier (one quiz → course-scale XP + N backend-signed devnet txs) → adversarial review + hardening round → queuing for gate.**

**My independent verification** at `3b510a17`: typecheck exit 0; full web suite **189 files / 1691 tests, exit 0** (and 188/1678 at the pre-fix head).

**Adversarial review (at `05dcded`): PASS — the security-critical guards are on-chain, where they belong.**
- **No double-mint**: the program's `LessonAlreadyCompleted` guard (complete_lesson.rs:129) is the backstop — replay, test-out×2, and test-out-vs-normal races all revert the second tx; batch counts reverts as pending, no event, no XP.
- **No answer forgery**: key stripped from GET; POST re-derives the authoritative question set and grades set-equality server-side; **deterministic sha256(userId:courseId) selection kills the retake-resample attack** (same 10 questions every attempt).
- **Capstone credential unreachable**: batch rides the LessonCompleted webhook, which runs the #721 deploy gate — no deployed_programs row, no credential (deferred, correctly).
- **Fee-drain bounded**: enrollment is learner-signed+paid, so each Sybil funds itself before touching the backend signer; 10/hr user + 200/hr IP on attempts.

**Hardening round (`3b510a17`), verified by me:** server-side scoping (LIVE synced+active AND learning-path member — closes the reviewer's any-course/unsynced/defi-#711 concern; spec-faithful reading of LX-A5, one-line predicate to tighten if owner prefers), MIN_TEST_OUT_POOL=8 floor (sparse-quiz course can't gate all XP behind 1-2 open-book answers), batch XP cap now fail-closed on missing/zero xp_per_lesson (409, not silent cap-skip).

**⚠️ OWNER ACK REQUESTED (in PR body, deliberate, not a defect):** a test-out pass grants the same XP as completing every lesson INCLUDING sandboxed challenges — via an open-book quiz (quiz `correct` flags are client-visible by D4 design). Reviewer's assessment: consistent with the accepted XP posture (completion proofs are already open-book-replayable) and the credential is separately protected; the floor bounds the worst case. Gate: merge is fine per posture-consistency, but the ack should be recorded — this is O-2-adjacent.

MINORs recorded: in-flight lock is cost/UX only (on-chain guard is the boundary — documented); serverless 300s hard-kill can strand the 120s lock briefly (idempotent retry, UX nit).

thomgabriel:
## Gate: PASS — merging.

Batch lesson completion is the highest-leverage forgery surface in the codebase, so I went at it from the attacker's side.

### Every lever I tried is closed

- **Client verdict never trusted.** POST accepts only `selections`; the server re-derives the question set and grades with `gradeTestOut` (set-equality, same as the in-lesson quiz grader).
- **No rerolling for easy questions.** The set comes from `mulberry32(seedFor(userId, course._id))` — derived server-side from the session user and course id. A learner cannot request a different draw.
- **Answer key stripped for transport** — `correct`, `feedback` and `explanation` never leave the server.
- **Eligibility is enforced on POST, not just GET.** `resolveChallenge` is the shared gate, and its own comment names the reason: *"UI scoping is bypassable"*. Both halves route through it.
- **Fail-closed on a thin pool** — below `MIN_TEST_OUT_POOL` the offer is refused rather than served as a trivially-passable challenge.

### The interaction I most wanted to check: test-out × capstone credential

If batch completion issued credentials, a learner could test out of C3 and receive "deployed a Solana program" without deploying anything — defeating #721 an hour after it merged.

It doesn't. `batch-complete.ts` **never calls `award_xp`** and never mints. It drives the existing `completeLesson` primitive per lesson; XP and the credential ride the Helius `LessonCompleted` webhook exactly as for a normal completion — and that path is **capstone-gated**. So a test-out still cannot produce the capstone credential without a real `deployed_programs` row. The comment says so, and the absence of any credential call in the module confirms it.

That also removes the double-credit surface: one XP source, unchanged.

### On the "nothing to game" claim

Post-D4 the quiz `correct` flags are public in the content bundle, so stripping them from transport does not make the challenge unguessable — anyone can read the answers. The real defence is economic, and it holds: the XP granted is **retroactive and equal to normal completion**, so passing by reading the bundle buys *time*, not XP advantage, and I5 means skipping costs no leaderboard position either. There is no arbitrage. Worth stating explicitly because "answer key stripped" could otherwise read as the security property, and it isn't.

### Unreviewed commit — and it is the one that matters tonight

Review is **17m53s stale** (21:20:53Z vs head 21:38:46Z, eighth today). The uncovered commit is "scope test-out to live path courses" — +67/−27 on the route plus a new `path-order.ts`.

That scoping is load-bearing given today's state: eligibility requires the course to be a member of a learning path. **C2 is staged and in no path** → not eligible. **`defi-on-solana`'s path is `draft: true` with `courses: []`** → not eligible, even though the course itself still renders (#711). So a learner cannot test out of a staged course or a retired one. Without this commit they could have.

Per-lesson on-chain failures isolate into `pending` rather than aborting the batch, and the whole thing re-reads the on-chain bitmap so re-running skips completed lessons.

Issue #731, opened 2026-07-26
ops: merged migrations auto-deploy before being applied — caused an AI-tutor outage and zeroed every streak tonight


thomgabriel:
**Vercel auto-deploys from `main` the moment a PR merges. Migrations are applied by hand, later.** So between merge and apply, production runs code that expects schema which does not exist. Tonight that window produced two live regressions before anyone looked.

## What actually happened

Four migrations sat merged-but-unapplied. I found them while reconciling, and two were already causing user-visible damage:

**1. The AI tutor was denying every request.** #720's `checkAiSpend` calls `check_ai_spend`; the function did not exist; a missing RPC is an error; the module **fails closed** — correctly, by design — so every AI Partner request 503'd. The fail-closed choice is right for a spend ceiling. It just means an unapplied migration converts directly into a total outage of that feature.

**2. Every learner's streak displayed as 0.** #727's `hybrid-progress-service.getStreak()` runs:

```ts
const [{ data }, { data: frozen }] = await Promise.all([
  supabase.from("user_xp").select("current_streak, longest_streak, last_activity_date, streak_freezes"),
  supabase.from("streak_freezes_used").select("frozen_date"),
]);
return { currentStreak: data?.current_streak ?? 0, ... };
```

`streak_freezes` (column) and `streak_freezes_used` (table) were both absent, so PostgREST 400'd both queries. **The destructuring discards the errors**, and every field falls through to its `?? 0` / `?? []` default. No throw, no log, no alert — just every learner's streak silently rendering as zero.

That second one is the worse shape: the first fails loudly in a way a user reports, the second looks like the learner lost their streak.

Both are now fixed (applied and verified: `check_ai_spend` returns zeros instead of erroring; both streak reads resolve; forging a `streak_freezes_used` row is denied by RLS).

## Why this will recur

Nothing links the two halves. A PR containing `supabase/migrations/*.sql` merges, deploys, and serves traffic without anything checking that the migration ran. The gap is unbounded in time and invisible while it lasts.

It also compounds #708: applying by hand through MCP stamps its own ledger version, so "is this migration applied?" cannot be answered by comparing filenames to the ledger. Tonight I had to check for the **objects** instead.

## Options, cheapest first

1. **A CI check that fails when a PR adds a migration whose objects are absent from prod** — or more simply, a post-merge job that lists unapplied migrations and opens/updates a tracking issue. Detection beats nothing.
[...]

Issue #713, opened 2026-07-26
Retire the 5 superseded courses (60 lessons still live after #679 activation) — UNTRACKED


thomgabriel:
**Nothing tracks retiring the old courses**, and #679's activation made them live. The 5-course catalog is not reached by authoring C1–C4 — it is reached by authoring them **and removing five courses**. Only the second half is untracked.

#711 covers the immediate symptom for one of them (DeFi still renders). This is the full job.

## What is actually live right now

```
anchor-framework                     12 lessons  trackId 3 / level 1   → RETIRE (prose → C3)
defi-on-solana                       12 lessons  trackId 5 / level 1   → RETIRE (token-standards → C5)
rust-for-solana                      12 lessons  trackId 2 / level 1   → RETIRE (superseded by C2)
solana-frontend                      12 lessons  trackId 4 / level 1   → RETIRE (superseded by C4)
solana-fundamentals                  12 lessons  trackId 1 / level 1   → RETIRE (superseded by C1)
building-your-first-solana-program   16 lessons  trackId 1 / level 2   → KEEP as C3 (#712)
stablecoin-agentic-payments           9 lessons  trackId 1 / level 5   → C5 ✓
```

Five courses, **60 lessons**, all currently reachable by learners.

## Why "draft the path" was not enough

`paths/defi.yaml` was set `draft: true` with `courses: []` to retire the DeFi track. The course still renders, because **the catalog lists courses independently of paths** — confirmed at runtime (#711): `/en/courses` returns it. Drafting a path hides the path, not its courses. The same is true of the other four.

## Two things that make this more than deleting directories

**1. They are deployed on-chain, and `trackId`/`trackLevel`/`creator` are immutable.** `update_course` can only touch `content_tx_id`, `is_active`, `xp_per_lesson`, `creator_reward_xp`, `collection`, `active_lessons`. So there are two levers:

- **`is_active: false`** — mutable, reversible, no recreate. Hides the course without touching the account. This is almost certainly the right launch move.
- **close + recreate** — needed only if the slot itself must be reused. Note `solana-fundamentals` occupies **trackId 1 / level 1**, which is C1's target slot, and C3 must move 2 → 3. So at least those two interact with the ladder and belong in the #606 window.

[...]

PR #732, opened 2026-07-26
feat(courses): course changelog on course pages (#654)

### LINKED ISSUE #654: Course changelog on courses pages
## Summary

Add a **course changelog** to the courses pages so learners (and admins) can see how a course has evolved after deployment — lessons added, removed, or reordered, XP changes, and content re-syncs.

## Motivation

Courses are mutable after deployment. Lessons are tracked as a 256-bit live-lesson bitmap (`active_lessons: [u64; 4]`), and `update_course` can add / retire / reorder lessons and bump the content version without closing the course. Today these changes happen silently — a learner mid-course has no signal that lessons were added or retired, and admins have no in-app history of what changed when.

An on-chain trail already exists to source this from:
- `CourseUpdated` event — emitted on every `update_course` with `version`, `collection`, `timestamp` (`onchain-academy/programs/onchain-academy-pinocchio/src/events.rs`, `emit_course_updated`).
- `course.version` bumps whenever `content_tx_id` changes.
- `content.lock` / `apps/web/src/content/generated/meta.json` bumps on each content pin.

## Proposed scope

1. **Data source** — derive changelog entries from `CourseUpdated` events (via Helius) joined with content-pin metadata. Decide: index into Supabase vs. read on demand.
2. **UI** — a "Changelog" / "What's changed" section on the course page (`apps/web/src/app/[locale]/(platform)/courses/...`) showing dated entries: version bump, lessons added/removed (diff the `active_lessons` mask between versions), XP-per-lesson changes.
3. **Learner-facing signal** — badge/notice when a course changed since a learner enrolled (compare enrollment generation/version to current).
4. **i18n** — strings in `en.json` / `es.json` / `pt-BR.json`.

## Open questions

- Persist derived changelog (Supabase table) or compute from on-chain events at read time?
- Show retired-lesson entries to learners, or admin-only?
- How to render a mask diff into human-readable "Lesson N added/removed" (mask bit index → lesson slot → content title).

## References

- `onchain-academy/programs/onchain-academy-pinocchio/src/instructions/update_course.rs` (`new_active_lessons`)
- `onchain-academy/programs/onchain-academy-pinocchio/src/state/course.rs` (`active_lessons`, `live_lesson_count`, `version`)
- `apps/web/src/lib/helius/event-handlers.ts` (event ingestion)
- `apps/web/src/app/api/admin/courses/sync/route.ts` (where content-hash commits + mask writes happen)

thomgabriel:
Closes #654.

## What

A **course changelog** on course pages: a learner-visible log of how a course evolved after deployment — lessons added / retired (diffed from the on-chain `active_lessons` bitmap), XP-per-lesson changes, and content re-syncs. Post-deployment mutation is no longer silent.

Rendered below the curriculum + discussions. Entries newer than the learner's enrollment are flagged "new since you enrolled" (issue point 3).

## Data-source design

**Server-side capture at mutation time** — not Helius `CourseUpdated` replay.

The issue floats two options (index events vs. read on demand); I took the higher-fidelity third path the mutation model implies. The admin sync route (`api/admin/courses/sync`) is the *sole* mutator of a course's on-chain state, and at mutation time it already holds **both** the prior on-chain state (`fetchCourse`) **and** the intended new state (the committed `slots.lock` mask + xp). So it computes the human-readable diff there and writes `course_changelog` rows via `service_role` — mirroring how `onchain_deployments` is written from the same route. The `CourseUpdated` event carries only `version`/`collection`/`timestamp` and **cannot** say which lesson slots moved, so replay would be lossy.

Lesson titles are **snapshotted** into `detail` at capture time: a changelog is an immutable historical record, a retired lesson may later vanish from the content bundle, and a renamed lesson should read in the log under the name it had when it changed. The read seam therefore never touches the content bundle.

One row per `(mutation, change-kind)`: a single sync that adds a lesson *and* changes XP produces two rows sharing `tx_signature` + `version`, rendered under one dated update. Kinds: `deployed`, `lessons_added`, `lessons_removed`, `xp_changed`, `content_updated`.

**Design tradeoff — `content_updated` is suppressed when the live-lesson set changed in the same sync.** A `content_updated` row is emitted only for a *pure* re-sync (content `tx_id` bumped, no add/retire). When lessons were added or retired in the same update, those `lessons_added`/`lessons_removed` rows already carry the version bump, so a separate `content_updated` would be redundant noise on the same timestamp. This is a deliberate choice (chosen, not missed): the tradeoff is that a "lessons changed **and** other content edited" sync surfaces only the lesson rows, not a distinct "content refreshed" line — acceptable because the version bump is already visible on the lesson rows. Reorder is intentionally not a kind: the bitmap tracks liveness, not display order, so reordering isn't derivable from it.

## Migration

`20260726210000_course_changelog.sql` (+ `schema.sql` mirror):
- **RLS on, fail-closed SELECT policy**: a row is visible only when its course currently has a `synced` AND active deployment — the exact catalog gate (`isSynced` = `status = 'synced' AND COALESCE(is_active, true)`). The `EXISTS` targets the anon-readable `public_onchain_deployments` **view** (owner-privilege), not the RLS-blocked base table. A course can be on-chain-deployed yet hidden (deactivated / synced-then-deactivated); a blanket `USING (true)` would let any anon PostgREST caller enumerate its snapshotted lesson titles. No deployment / unsynced / inactive ⇒ hidden.
- **No write policy** → `service_role`-only writes (the sync route is the sole writer).
- **`tx_signature` NOT NULL**: it is the dedup key, and Postgres treats NULLs as distinct in a UNIQUE, so a nullable column would silently defeat idempotency. Dedup `UNIQUE (course_id, kind, tx_signature)` + `ON CONFLICT DO NOTHING` → re-running a sync never double-logs.
- Idempotent (`IF NOT EXISTS`, `DROP POLICY IF EXISTS`), explicit `BEGIN/COMMIT`, tested rollback block. `course_id` is the content `_id` verbatim (PDA-seed convention).

No live-DB changes — SQL-only, per repo conventions.

## Learner signal (issue pt.3)

Entries newer than the learner's enrollment are flagged "new since you enrolled" via a date compare against `enrollments.enrolled_at` — no on-chain generation read needed.

## i18n / a11y

Strings in `en` / `es` / `pt-BR` with ICU plurals. Section is `aria-labelledby` a heading, entries are an `<ol>`, dates are `<time datetime>`.

## Verification

- typecheck: 6/6 packages pass
- **web suite: 191 files, 1720 tests pass**

Tests: mask diff, capture per mutation kind + idempotency + null-title fallback + non-fatal error, RLS/schema-mirror guard (pins the scoped predicate; mutation: `USING(true)` fails the guard; NOT NULL pinned), RLS-safe read + defensive decode (drops malformed rows), component rendering + empty state + since-enrolled signal.

## Review fixes (adversarial review of #732)

Second commit (`183fc2c`) addresses:
- **MINOR-1** — scoped the SELECT policy fail-closed (see Migration above).
- **MINOR-2** — `tx_signature` NOT NULL; route drops the `!` assertions for truthy guards.
- **MINOR-3** — documented the `content_updated` suppression tradeoff (see Data-source design above).

PR diff is clean against origin/main (merge-base `5892eb9`); the branch carries only this change.

Do not merge — for review.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/course-detail-client.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/course-detail-client.tsx
@@ -26,7 +26,9 @@ import { useOnChainEnroll } from "@/hooks/use-on-chain-enroll";
 import { findEnrollmentPDA } from "@/lib/solana/pda";
 import { ThreadList } from "@/components/community/thread-list";
 import { CreateThreadModal } from "@/components/community/create-thread-modal";
+import { CourseChangelog } from "@/components/course/course-changelog";
 import type { PublicProfile } from "@/lib/profiles/public-profile";
+import type { CourseChangelogEntry } from "@/lib/courses/changelog-types";
 
 interface CourseDetailClientProps {
   course: Course;
@@ -36,11 +38,17 @@ interface CourseDetailClientProps {
    * so the instructor section never has to fetch client-side.
    */
   instructorProfile: PublicProfile | null;
+  /**
+   * The course's post-deployment evolution log (issue #654), read server-side
+   * and passed in already decoded. Rendered below the curriculum.
+   */
+  changelog: CourseChangelogEntry[];
 }
 
 export function CourseDetailClient({
   course,
   instructorProfile,
+  changelog,
 }: CourseDetailClientProps) {
   const t = useTranslations("courses");
   const tCommon = useTranslations("common");
@@ -73,6 +81,9 @@ export function CourseDetailClient({
   const walletAddress = authProfile?.wallet_address ?? null;
 
   const [isEnrolled, setIsEnrolled] = useState(false);
+  // ISO enrollment timestamp — lets the changelog flag entries newer than the
+  // learner's enrollment (issue #654 point 3). Null when signed-out/unenrolled.
+  const [enrolledAt, setEnrolledAt] = useState<string | null>(null);
   const [completedLessons, setCompletedLessons] = useState<string[]>([]);
   const [isLoading, setIsLoading] = useState(true);
   const [discussionModalOpen, setDiscussionModalOpen] = useState(false);
@@ -92,10 +103,12 @@ export function CourseDetailClient({
       // (enrollment PDA exists on-chain but Supabase has no record).
       const { data: enrollment } = await supabase
         .from("enrollments")
-        .select("id")
+        .select("id, enrolled_at")
         .eq("user_id", userId)
         .eq("course_id", course._id)
         .maybeSingle();
+
+      setEnrolledAt(enrollment?.enrolled_at ?? null);
 
       let enrolled = !!enrollment;
 
@@ -374,6 +387,9 @@ export function CourseDetailClient({
           />
         )}
       </div>
+
+      {/* Changelog (issue #654) — how the course has evolved since deployment. */}
+      <CourseChangelog entries={changelog} enrolledAt={enrolledAt} />
     </div>
   );
 }
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/page.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/page.tsx
@@ -1,8 +1,9 @@
 import { notFound } from "next/navigation";
-import { CourseDetailClient } from "./course-detail-client";
 import { getCourseBySlug } from "@/lib/content/queries";
 import { resolvePublicProfileByWallet } from "@/lib/profiles/public-profile";
+import { getCourseChangelog } from "@/lib/courses/changelog";
 import { createClient } from "@/lib/supabase/server";
+import { CourseDetailClient } from "./course-detail-client";
 
 interface CourseDetailPageProps {
   params: Promise<{ slug: string }>;
@@ -23,7 +24,15 @@ export default async function CourseDetailPage({
     ? await resolvePublicProfileByWallet(await createClient(), course.creator)
     : null;
 
+  // Course changelog (#654) — the post-deployment evolution log, read server-
+  // side through the public RLS policy and passed down already decoded.
+  const changelog = await getCourseChangelog(course._id);
+
   return (
-    <CourseDetailClient course={course} instructorProfile={instructorProfile} />
+    <CourseDetailClient
+      course={course}
+      instructorProfile={instructorProfile}
+      changelog={changelog}
+    />
   );
 }
--- a/apps/web/src/app/api/admin/courses/sync/route.ts
+++ b/apps/web/src/app/api/admin/courses/sync/route.ts
@@ -327,17 +327,20 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
 
     // Course changelog (#654): record the first on-chain deploy. Non-fatal —
     // the course is already live; a missing log entry must not fail the sync.
-    try {
-      await recordCourseDeployed({
-        courseId,
-        txSignature: result.signature!,
-        lessonCount: course.lessonCount,
-      });
-    } catch (changelogErr) {
-      console.error(
-        "[admin/courses/sync] changelog (deploy) failed:",
-        changelogErr
-      );
+    // tx_signature is NOT NULL (the dedup key), so only log with a real sig.
+    if (result.signature) {
+      try {
+        await recordCourseDeployed({
+          courseId,
+          txSignature: result.signature,
+          lessonCount: course.lessonCount,
+        });
+      } catch (changelogErr) {
+        console.error(
+          "[admin/courses/sync] changelog (deploy) failed:",
+          changelogErr
+        );
+      }
     }
 
     // The course is now "synced" in Sanity — purge the catalog cache so it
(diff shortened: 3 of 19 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (new migration + admin-sync write path) → adversarial review PASS + fix round verified → queuing for gate.**

**My independent verification** at `183fc2c3`: typecheck exit 0; full web suite **191 files / 1720 tests, exit 0** (and 191/1718 at the pre-fix head).

**Adversarial review (at `196b6b36`): PASS.** Held under attack: migration hygiene (mirror byte-identical, idempotent, correct rollback; REVOKE posture correct — IDENTITY PK, no grantable sequence); capture genuinely non-fatal (a changelog failure can never fail a landed on-chain mutation — verified both call sites); mask diff correct across the 4 u64 words incl. retired slots; the camelCase/snake_case mixed access chased to ground as a FALSE alarm (DecodedCourse normalizer intentionally exposes both); read seam defensive (drops malformed rows, LIMIT 50, single query); since-enrolled boundary strict (entry at enrollment instant not flagged); i18n parity 15/15 keys ×3 locales; guard mutation-tested both ways; suite reproduced.

**Fix round (`183fc2c3`), independently verified:**
- **SELECT policy scoped by construction** (was `USING(true)` = anon enumeration of deployed-but-hidden courses' snapshotted lesson titles — the C2/C5 window scenario): now EXISTS against the anon-readable `public_onchain_deployments` VIEW (synced + active), matching the catalog's isSynced predicate exactly. The view-not-base-table choice is load-bearing (base table is RLS-no-policy; an EXISTS against it would hide everything) — implementer caught and documented it. Guard pins the predicate; USING(true) reversion fails the guard.
- **tx_signature NOT NULL** (UNIQUE+NULL dedup hole closed), types tightened end-to-end, `!` assertions replaced with guards.
- content_updated suppression documented as a chosen tradeoff in the body.

**Branch shape note for the gate:** fixes landed as a second commit, not a rebase — the classifier (correctly) blocks history rewrites on teammate request; GitHub's three-dot diff is clean (merge-base ancestors only) and squash-merge collapses it anyway. Verified `git diff origin/main...HEAD --stat` = only this PR's files.

**Apply-before-merge callout (#731 lesson):** this PR carries migration `20260726210000_course_changelog.sql`. Merging deploys capture code + a UI that reads a table that won't exist until applied. The read seam returns [] on error (course page unaffected — verified defensive), and capture is non-fatal (sync unaffected), so the gap window here is BENIGN-degraded, not an outage — but per the new discipline: **apply at merge time or note the window.**

thomgabriel:
## Gate: PASS — merging, with one coverage finding filed.

### The RLS scoping is the best-reasoned policy I have gated today

Not `USING (true)`. It scopes reads to the catalog's own gate:

```sql
EXISTS (SELECT 1 FROM public.public_onchain_deployments d
        WHERE d.content_id = course_changelog.course_id
          AND d.kind = 'course' AND d.status = 'synced'
          AND COALESCE(d.is_active, true))
```

Two details worth calling out:

1. It targets the **anon-readable view**, not the base table — and says why: *"RLS-on/no-policy would make this subquery see nothing → hide all"*. A policy that queries an RLS-protected base table silently evaluates to false for everyone. Easy to get wrong, and the failure would have looked like "the changelog never renders" rather than an error.
2. It reuses the exact expression the catalog gate uses (`deployments.ts:92`), so a course hidden from the catalog is hidden from the changelog **by construction** rather than by a parallel rule that could drift.

That property lands well against #713: when the five retired courses get `is_active = false`, their changelogs stop being readable automatically. No write policy, with the reason stated.

The unreviewed commit (review is 18m18s stale, 21:35:16Z vs head 21:53:34Z) **is** this RLS fix — +43/−19 on the migration. Verified directly.

### Finding — "the sole mutator" is not the sole mutator

The design rests on this:

> The admin sync route (`api/admin/courses/sync`) is the *sole* mutator of a course's on-chain state

I checked, because a coverage claim is exactly what I have twice failed to check today. It is not sole. Two other routes mutate on-chain course state and write **no** changelog entry:

- **`api/admin/courses/reactivate/route.ts:44`** — `updateCoursePda({ courseId, newIsActive: true })` followed by `writeCourseActive(courseId, true)`.
- **`lib/admin/recreate-course.ts`** — imports `closeCoursePda` + `updateCoursePda` and performs the WS-2 close-and-recreate.

(`lessons/complete` matched my first grep but is a false positive — line 454 is a comment about `update_course.new_lesson_count`, not a call.)

**Why this matters beyond tidiness:** the feature's stated purpose is that "post-deployment mutation is no longer silent", and the two loudest mutations available are exactly the silent ones. A recreate is the most dramatic thing that can happen to a course — it closes the account and makes a new one — and it would leave no trace in the log a learner reads. And #713's retirement plan works by flipping `is_active`, which is the `reactivate` path's mirror image, so retiring five courses would produce no changelog entries either.

Not blocking: the capture that exists is correct, the feature is additive, and nothing here is wrong — it is incomplete. Filed as **#738**.

The design choice itself is right, incidentally — capturing at mutation time beats replaying `CourseUpdated` events, because the sync route holds both the prior on-chain state and the intended new state in the same call, which an event replay cannot reconstruct. The gap is that two other callers hold that same information and do not use it.

PR #733, opened 2026-07-26
fix(streak+ops): honest reads + schema-expectation health check (#731)

### LINKED ISSUE #731: ops: merged migrations auto-deploy before being applied — caused an AI-tutor outage and zeroed every streak tonight

thomgabriel:
Closes #731 (options 3 + 2; option 1 CI/post-merge job stays with owner/gate).

Post-incident fix. My #727 merged and Vercel auto-deployed **before** the SQL was applied; `getStreak`'s `Promise.all` destructuring discarded the PostgREST 400s into `?? 0` / `?? []`, so every learner's streak silently rendered as **0** for the whole merge→apply window. This makes the reads honest and adds deploy-time detection.

## Option 3 — honest reads (primary)

`StreakData` gains **`available: boolean`**. `getStreak` now captures *both* queries' errors:
- A real read failure → `available: false` (logged), which the UI renders as **em-dash/hidden**, never "0 day streak".
- A genuinely absent row (`PGRST116`) → stays a **true `available` zero** (new users legitimately have no `user_xp` row).

**Sweep** — every error-discarding read in `hybrid-progress-service.ts`, each now captures `error` and routes it to `logError` (no more silent defaults):

| Method | Read | Before | After |
|--------|------|--------|-------|
| `getStreak` | `user_xp` + `streak_freezes_used` | `?? 0` / `?? []`, error discarded | **`available:false` degraded state**, both errors logged |
| `getXP` | `user_xp.total_xp` | `?? 0` | logs (skips `PGRST116` no-row); on-chain value still bounds the answer |
| `getCredentials` | `certificates` | `if (!certs) return []` | logs error |
| `getProgress` | `user_progress` | `?? []` | logs error |
| `getWalletForUser` | `profiles` | `?? null` | logs (skips `PGRST116`) |
| `getProfileByWallet` | `profiles` | `?? null` | logs (skips `PGRST116`) |
| `getLeaderboardFromSupabase` | `get_leaderboard` rpc | `if (!data) return []` | logs error |

UI/consumers: profile hero + `StreakDisplay` render em-dash when `!available`; `use-dashboard-data` default + `updateStreak` passthrough + `profile-data` thread the flag; new i18n `streakUnavailable` in **en / pt-BR / es** (gamification + profile namespaces). The dashboard activity heatmap is unaffected (its `streakHistory` comes from an independent `xp_transactions` query).

## Option 2 — schema-expectation health check (detection)

`lib/supabase/schema-expectations.ts`: a **colocated, documented** list of the DB objects this release depends on, probed cheaply and **side-effect-free**:
- **table** → `select * limit 1` (missing → PGRST205/42P01; RLS only filters rows, never errors on existence)
- **column** → `select <col> limit 1` (missing → 42703)
- **rpc** → call with inert args via the **anon** server client, which lacks `EXECUTE` on the service_role-only functions — so a **present** function returns permission-denied (42501) **without running its body**, and a **missing** one returns PGRST202. Bodies are never executed.

Seeded with the three objects this incident hit: `user_xp.streak_freezes` (#727), `streak_freezes_used` (#727), `check_ai_spend` (#720) — each carrying the migration that creates it. Migration authors extend the list (documented in the file header; existence-only, so it catches "never applied", not a stale body).

`GET /api/health/schema`: **200** when complete, **503** (naming the missing objects + their migrations) when the gap is open. Uses the existing anon server client — **no new secrets**.

## Verification

- `pnpm install --frozen-lockfile && pnpm typecheck` — green (6/6)
- `pnpm --filter web test` — **191 files / 1713 tests pass** (10 new)
- `pnpm --filter web lint` — clean

New tests: `getStreak` degrades (not zeroes) on each query's error **independently** and stays a true zero on `PGRST116`; the health check reports a missing column/table/rpc with its migration and treats permission-denied as **present**; happy paths clean; route maps ok→200 / missing→503.

## Notes / scope
- Option 1 (CI or post-merge job listing unapplied migrations) needs prod DB access from Actions — a secrets decision left to the owner/gate. The health check gives the detection half without it.
- Not touched: `check_ai_spend`'s fail-closed behavior is correct by design; this only makes the missing-object case *detectable*.


thomgabriel:
**Orchestrator: both non-blocking review findings fixed at the new head** (by me — trivial polish): the degraded em-dash in `streak-display.tsx` now carries `aria-label`/`title` from a new descriptive `gamification.streakUnavailableLabel` key (en/es/pt-BR, parity test green), and `/api/health/schema` is documented in the API CLAUDE.md inventory under a new Health section. Full suite re-run at the fixed head: 191 files / 1713 tests, exit 0.

PR #740, opened 2026-07-26
content: bump content.lock to 7a49747 — C3 transform live (16→15 lessons, ids stable)

thomgabriel:
## What

Pins `courses-academy@7a49747` — the C3 transform of the LIVE flagship course `building-your-first-solana-program` (courses-academy C3 authoring, merged by the gate): **16 → 15 lessons** (4 retired incl. DELETE-slated `deploy-to-devnet`, 3 added incl. `pre-flight-check`), plus absorbed anchor-framework prose per the salvage ledger.

## The property that matters (verified in the compiled bundle)

- **Zero surviving lesson ids changed** — the C3 'renames' are path/title-level; slot-keyed ids are stable, so the on-chain progress bitmaps of the 38 live enrollments are untouched by construction.
- **Capstone intact**: `lesson-bfsp-m4-capstone` present; the #721 capstone-gate bundle assertions (course id, deploy-panel lesson, exactly-one-panel) pass unchanged — no gate-constant updates needed.
- DELETE-slated `lesson-bfsp-deploy-to-devnet` gone; `lesson-bfsp-pre-flight-check` added; all 13 tutorNotes lessons carried through (11 from #10 + 2 CARRY from #14).
- Counts: courses 8, lessons 99→**98**, paths/quests/achievements unchanged.

## Golden fixtures

C3 reshaped a live course, so five fixtures regenerated through the real projectors (courses, lessons, quests-raw derived fields, paths, course-summaries — existing doc order preserved, raw quest docs and authored path fields untouched); wave-4 delta documented in the test header.

`compile-content` clean; `pnpm typecheck` exit 0; `pnpm --filter web test` **200 files / 1803 tests, exit 0** (includes the #721 capstone-gate assertions and #730 test-out path-eligibility against the new bundle).

## Routing

`area:content`, SAFE lane.

### DIFF
--- a/apps/web/content.lock
+++ b/apps/web/content.lock
@@ -1,4 +1,4 @@
 {
   "repo": "solanabr/courses-academy",
-  "sha": "1b74e4a60f8813374219451b60436af832df2d91"
+  "sha": "7a497475e14f3f53e5353cd752476eb0b2120096"
 }
--- a/apps/web/src/content/generated/courses.json
+++ b/apps/web/src/content/generated/courses.json
@@ -133,7 +133,7 @@
     "_type": "course",
     "creator": "B7o8NfV81HzjuZFWQTTx3Xdvh77Dqoajwib3kWEnvzJF",
     "creatorRewardXp": 30,
-    "description": "Take your Anchor knowledge from theory to practice. Write, compile, and deploy real Solana programs using the Superteam Academy Build Server — no local toolchain setup required. By the end, you'll have a working counter program deployed to Devnet.",
+    "description": "You arrive holding `vault_core.rs` — a `VaultState` and checked `deposit`/`withdraw` that compile, with no `#[program]` and no `#[derive(Accounts)]`. This course adds exactly that. You wrap your own Rust core in Anchor 1.x, create a per-user vault PDA with the space arithmetic derived rather than copied, and move real lamports: deposit through a System Program CPI, withdraw by debiting the program-owned vault directly above an enforced rent floor — and you learn why signer seeds are not the answer there. Then you write the account constraints that stop three named attacks, read the LiteSVM test that would catch what compiling cannot, deploy from a browser tab, publish the IDL on-chain, and call your own live program. Nothing is installed locally. You finish with a devnet program id, a fetchable IDL, and a build write-up.",
     "difficulty": "intermediate",
     "duration": 12,
     "minCompletionsForReward": 10,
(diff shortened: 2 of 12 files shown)


thomgabriel:
## GATE: blocked — the safety premise doesn't hold

Authored by my subagents (C3 itself), so per #712 this is not "gated by whoever did not author it" — flagging that, though the finding below is independent of provenance.

**What I verified as true.** Every id-level claim in the description checks out. I diffed `slots.json` independently: all 12 surviving lessons keep their exact slots (`your-first-build` 1, `capstone` 15, …), retired is exactly `[0,2,11,14]`, new are 16/17/18, `next` 19. `lesson-bfsp-m4-capstone` is present. Counts 99→98. Zero surviving lesson ids changed.

**What doesn't follow.** The description reasons: *"slot-keyed ids are stable, so the on-chain progress bitmaps of the 38 live enrollments are untouched by construction."* The premise is right and the conclusion is wrong, because **the bitmap is not slot-keyed**. `/api/lessons/complete` derives the on-chain index with `findLessonIndex`, which returns the lesson's position in the flattened module array ([lesson-index.ts:14](apps/web/src/lib/courses/lesson-index.ts:14), via [route.ts:52](apps/web/src/app/api/lessons/complete/route.ts:52) at :449). `slots.json` is read by exactly two files, both admin-only (`admin/courses/sync`, `admin-signer`) — nothing on the learner path reads it.

So the slot lockfile that this PR keeps perfectly stable is not what the write path consults. Applying the transform, **9 of 12 survivors change their bitmap index**:

`your-first-build` 1→0 · `add-instruction` 3→1 · `adding-instructions` 8→2 · `on-chain-state` 4→3 · `wire-up-initialize` 7→6 · `build-increment` 9→8 · `complete-counter` 10→9 · `account-constraints` 6→10 · **`m4-capstone` 15→14**

**This lands on live data.** In prod, `lesson-bfsp-m4-interact` — retired by C3 — is completed with a real on-chain tx at **bit 14**. The capstone moves to index 14. The idempotency check (`isLessonComplete(flags, 14)`) reads that set bit, so the capstone gets marked complete without the learner touching it, with no XP and no tx. The capstone is the credential gate (#721). Separately, new content tops out at index 14 while the on-chain mask still has bit 15, so `flags & active == active` can never hold — `finalize_course` becomes unreachable for everyone.

Also worth knowing before the on-chain window: the `lessonIndex >= liveLessonCount` guard at route.ts compares an array index against `popcount(active_lessons)`. Once the C3 mask goes sparse those are different spaces (popcount 15, live slots to 18), and slots 15–18 would 409 as "recently extended". The code comment there already predicts this.

Root cause and fix filed as **#741** — the completion path needs to resolve `lessonId → slot` from the committed `slots.json`, plus the guard becomes a bit test and `user_progress.lesson_index` needs a backfill decision.

**Disposition:** hold #740 until #741 lands. Merging auto-deploys from main (#731), so there is no dry-run window between merge and live. The content itself is good — this is purely an ordering dependency.

Blast radius is small right now (1 enrollment, 5 completed rows, 2 learners, 0 certificates on this course), which is the argument for fixing it before launch traffic rather than after. Nothing needs re-authoring; #740 should be mergeable unchanged once #741 is in.

thomgabriel:
⚠️ This merged at 22:52:55Z, ~3 minutes before my blocking review posted — so the block never had a chance to hold it. Reverting in #744; root cause in #741.

The content is fine and needs no re-authoring. The problem is that the on-chain bitmap index is the flattened-array position (`findLessonIndex`), not the slot, so the transform remaps 9 of 12 survivors — `m4-capstone` 15→14, colliding with the retired `m4-interact`'s already-set on-chain bit 14 in prod. Re-land this unchanged once #741 lands.

PR #742, opened 2026-07-26
fix(content-lint): gate-3 base-ref fallback + non-destructive slots remedy (#737)

### LINKED ISSUE #737: ci: gate-3's error tells you to run pnpm content:slots — which would renumber slots and break 38 learners' progress bitmaps
**`gate-3`'s error message instructs you to run a command that would destroy learner progress on a deployed course.** Found while gating courses-academy#17 (C3).

## The message

```
[error] gate-3 …/slots.lock.json: slots.lock.json does not match regeneration —
a slot was changed, reused, or a lesson is missing. Run `pnpm content:slots`.
Expected {"slots":{"lesson-bfsp-your-first-build":0, … ,"lesson-bfsp-m4-capstone":14},"retired":[],"next":15}
```

### 1. The failure is spurious outside CI

`gate3-slots.ts` computes the expectation with `assignSlots(baseLock(...))`. Without a base git ref `baseLock` is `null`, so it regenerates the course from scratch as a fresh 0–N and reports any real lock as a mismatch. CI is fine — `validate-content.yml` sets `GITHUB_BASE_REF` with `fetch-depth: 0`. Verified both ways against C3:

```
bare invocation           → exit 1, gate-3 error
GITHUB_BASE_REF=main      → exit 0, 0 errors
```

So anyone linting locally, or any agent doing so, sees a red gate on a correct lock. That alone is just noise.

### 2. The suggested remedy is destructive, and that is the real problem

`pnpm content:slots` would write the regenerated 0–14 map. For `building-your-first-solana-program` that renumbers **every** lesson:

| | committed (correct) | what the message tells you to write |
|---|---|---|
| surviving lessons | keep original slots (0 moved) | all renumbered 0–14 |
| retired | `[0, 2, 11, 14]` | `[]` |
| next | 19 | 15 |

Learner progress is a **slot-keyed on-chain bitmap** and this course has **38 live enrolments**. Renumbering silently re-points every learner's completed-lesson bits at different lessons. `slots.ts:5-8` states the invariant the message violates: *"Assigned once, never renumbered, never reused."*

The dangerous shape is the combination: a **spurious** error that fires exactly where a **destructive** fix is suggested. Someone trusting the tool clears a red gate and corrupts progress, and nothing downstream would catch it — the lock would be internally consistent and lint would go green.

## Fix

1. **Make the bare invocation honest.** When `baseLock` is `null`, gate-3 cannot decide whether the lock is correct — it should say so (notice/warning: "slot immutability not checked — no base ref") rather than assert a mismatch. A check that cannot run should not report a failure, same principle as #614/#694.
2. **Stop recommending `pnpm content:slots` unconditionally.** It is only safe for a course with no assigned slots. For a deployed course the correct action is to *keep* the lock and investigate, never to regenerate. If the message keeps the suggestion, it needs the condition attached.
3. **Consider refusing to regenerate a lock that has non-empty `retired[]` or `next > count`**, which are the signatures of a course that has already shipped slots.

## Note

This is close to #708's family: tooling that is confidently wrong about production state. There the ledger disagreed with the repo; here the linter disagrees with a correct lock and offers to "fix" it destructively.

Related: courses-academy#17 (where it surfaced), #712, #708.

thomgabriel:
Fixes #737.

`gate-3` (slots-lock immutability) had two coupled defects: a **spurious failure** outside CI, sitting exactly where it suggested a **destructive** remedy. A remediation agent nearly ran that remedy on a deployed course.

## Problem 1 — spurious failure outside CI
`gate3-slots.ts` computed its expectation with `assignSlots(baseLock(...))`. Without `GITHUB_BASE_REF` (any bare local run), `baseLock` was `null`, so `assignSlots(null, …)` regenerated a fresh `0–N` map and reported **any correct lock** as a mismatch. CI was fine (`validate-content.yml` sets `LINT_BASE_REF` with `fetch-depth: 0`); everyone else saw a red gate on a good lock.

**Fix:** when no CI ref is set, resolve a remote-tracking base (`origin/HEAD` → `origin/main`/`origin/master`) via the new `resolveLocalBase()` and diff against it, exactly like CI. If **no** base ref is resolvable (a repo with no fetched remote), gate-3 emits a warning — `slot immutability not checked — no base ref` — and skips the immutability comparison instead of fabricating a mismatch. It never invents `HEAD` as its own base. Same principle as #614/#694: a check that cannot run must not report a failure.

## Problem 2 — destructive remedy text
The mismatch error said *"Run `pnpm content:slots`."* On a deployed course that renumbers **every** slot; slots are permanent on-chain progress-bit positions and `building-your-first-solana-program` has **38 live enrolments**, so regeneration silently re-points every learner's completed-lesson bits.

**Fix:** the mismatch message no longer suggests any regenerate command. It states the invariant (surviving lessons keep their slot; retired slots are never reused; `next` only grows), tells the author to **repair the lock by hand** (with the exact `git show <base>:<file>` to diff), and carries a **DANGER** note that regeneration is only ever for a brand-new, never-deployed course. The missing-lock branch now distinguishes a genuinely-absent lock (new course, safe to generate) from a present-but-unparseable one (flag for a human, don't regenerate).

## CI is unchanged
When `ctx.baseRef` is set (CI), `resolveLocalBase` is never called and the merge-base / diff path is byte-identical to before, including the shallow-base warning.

## `pnpm content:slots` — the guard has no home
The command referenced in the old message **does not exist as a runnable script** anywhere: not in this monorepo's `package.json`s, and courses-academy has no `package.json` at all. It appears only in docs/specs and in these two error strings. The only real slot generator is `assignSlots` in `content-schema` (used correctly by the gate, which passes the base lock so it *preserves*) and the one-shot migration `scripts/cs8-extraction/extract.ts`. There is therefore no `content:slots` script to add a `--force-new` refusal guard to. A guard must **not** be added to `assignSlots` itself — gate-3 legitimately calls it with a base lock that has non-empty `retired[]`/`next`, so refusing that would break the gate. Routing the "build a guarded `content:slots`" half separately per the task brief.

## courses-academy validation matrix
Linter (this branch) run against `courses-academy@main` (read-only clone):

| invocation | before this PR | after this PR |
|---|---|---|
| bare (no base ref), `origin/main` present | exit 1, gate-3 error on every lock | **exit 0**, no gate-3 diagnostic (resolves `origin/main`, locks match) |
| CI-simulated (`GITHUB_BASE_REF=main`) | exit 0 | **exit 0** (byte-identical) |
| bare, no `origin` remote at all | exit 1, gate-3 error | **exit 0**, gate-3 **warning** "no base ref" |

## Tests
`pnpm --filter @superteam-lms/content-lint test` → 85 passed (gate3: 8). `content-schema` → 170 passed. `typecheck` clean on both touched packages (web is not in content-lint's dependency graph). New/updated gate-3 coverage: no-base-ref → warning not error; `origin/main` local resolution; base-ref mismatch still errors; remedy text asserts **no** `content:slots` / regenerate suggestion.

**Do not merge** — SENSITIVE (ci/lint semantics). Needs adversarial + gate review.

### DIFF
--- a/packages/content-lint/src/__tests__/gate3.test.ts
+++ b/packages/content-lint/src/__tests__/gate3.test.ts
@@ -22,6 +22,13 @@ title: ${id}
 blocks: [{ type: prose, key: intro, src: intro.md }]
 `;
 
+const correctLock = JSON.stringify({
+  version: 1,
+  slots: { "lesson-a": 0, "lesson-b": 1 },
+  retired: [],
+  next: 2,
+});
+
 describe("gate 3 — slots", () => {
   const tree = (lock: string) => ({
     "courses/x/course.yaml": course,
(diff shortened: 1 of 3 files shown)


thomgabriel:
## GATE: approved

Both halves verified independently.

**The CI path is genuinely untouched.** `resolvedRef = ctx.baseRef ?? resolveLocalBase(ctx.root)` — short-circuits before `resolveLocalBase` whenever a ref is present, and `validate-content.yml:59` sets `LINT_BASE_REF: ${{ github.base_ref }}`, which GitHub populates on `pull_request` events. So the gate that actually blocks a content PR keeps byte-identical behavior; the softening only reaches runs with no base ref at all. `github.base_ref` isn't author-controllable, so a PR can't suppress the ref to downgrade the gate to a warning.

**Fail-open check.** The `!base` branch skips only the immutability comparison and still errors on a missing lock — which needs no base. That's the right split: a check that cannot run must say so rather than fabricate a verdict (#614/#694 precedent). Worst case on a `push`-triggered call is a self-comparison that passes trivially, which is no weaker than the pre-PR behavior of failing spuriously.

**The destructive remedy is gone**, and the missing-lock branch now distinguishes absent (new course, safe to generate) from present-but-unparseable (human repair). That distinction is the actual fix — the old text invited regeneration in exactly the case where it destroys progress.

Good independent catch that `pnpm content:slots` has no runnable implementation anywhere, and correct call not to bolt a refusal guard onto `assignSlots` — gate-3 itself calls it with a non-empty base lock, so that guard would have broken the gate.

CI green at this head; content-lint 85 / content-schema 170.

### One nit, not blocking

The new DANGER string says regeneration *"re-points 38+ learners' completed-lesson bitmaps."* That figure doesn't hold for prod: `enrollments` totals **7 rows across all six courses**, with **1** on `course-building-first-program`. The 38 is the devnet on-chain enrollment count from the reset wave, across courses. The warning's argument doesn't need a number — one learner's corrupted bitmap is already disqualifying — and a citable-but-wrong figure is the kind of thing that gets the whole message discounted. Suggest dropping the count.

Merging; the nit can ride along with any later touch to this file.

Issue #725, opened 2026-07-26
ops: /api/deploy/save has never written a row in prod — the capstone credential now depends on it


thomgabriel:
Found gating #721 (capstone-gated credential, merged). **`deployed_programs` has 0 rows in production — ever.** So `/api/deploy/save`, the on-chain-verified path that writes them, has no successful production run behind it. #721 now makes the capstone credential depend on exactly that row.

## Why this matters more than it looks

The failure mode is silent and **indistinguishable from correct behaviour**:

- If `/api/deploy/save` has a bug, no row is ever written.
- No row → `checkCapstoneCredentialGate` returns `deploy_required`.
- The webhook path (`tryIssueCredential`) throws and **defers onto the existing `certificate` queue** rather than erroring loudly.
- Observed result: learners complete the capstone and never receive a credential, and the system reports a working gate withholding an unearned credential.

That is the correct design — a credential is a permanent on-chain attestation, so fail-closed is right — but it means the *only* signal that the write path is broken is the absence of something, which nobody alerts on.

## Current prod state

```
deployed_programs rows:               0
certificates total:                   4
certificates for the capstone course: 0
completed capstone enrolments:        0
```

Nothing is broken today, because nobody has reached the capstone. This is a pre-launch item, not an incident.

## What needs to happen

**One real end-to-end run** producing a `deployed_programs` row, before any learner reaches the capstone deploy lesson. Specifically:

1. Deploy a program through the actual challenge UI on the capstone deploy lesson (`lesson-bfsp-m4-deploy` / whatever C3's transform lands on) — not a synthetic API call, since the point is to exercise the whole path.
2. Confirm a row appears with the right `user_id`, `course_id`, `lesson_id`, `program_id`, `network`.
3. Confirm `checkCapstoneCredentialGate` then returns `allowed` for that learner.
4. Confirm the credential actually mints — including the queued path, since that is the one the webhook uses.

Worth doing on devnet during the #606 on-chain window, when the authority keypair is already in hand.

## Related risk in the same path

`/api/deploy/save` verifies the program is **executable** and that **upgrade authority is the learner**. Both are on-chain reads. If either read is against a stale or wrong RPC the verification silently fails closed too. Worth confirming during the same run that a *legitimate* deploy is accepted — a gate that rejects everything is as broken as one that accepts everything, and only the first looks safe.

[...]

PR #743, opened 2026-07-26
feat(ops): make /api/deploy/save write path observable (#725)

thomgabriel:
Closes part of #725.

## The indistinguishability problem
`deployed_programs` has **0 rows in production, ever**, so `/api/deploy/save` — the on-chain-verified path that writes them — has no successful prod run behind it. #721 now makes the capstone credential depend on exactly that row. The failure mode is silent and identical to correct behaviour:

- Broken save path → no row → `checkCapstoneCredentialGate` returns `deploy_required`.
- The webhook (`tryIssueCredential`) defers the certificate onto the retry queue rather than erroring.
- Result: learner completes the capstone, never gets a credential, and the system *looks like* a working gate withholding an unearned one.

Fail-closed is the right design (a credential is a permanent on-chain attestation), but it means the only signal of a broken write path is the **absence** of something — which nobody alerts on. This PR makes that absence loud, and adds the tests the issue asks for.

**Scope note:** #725 also calls for one real end-to-end devnet deploy producing a row. That requires on-chain transactions + live DB and is out of scope here (a separate ops step during the #606 window); this PR is the observability + test half so that when that run happens — or fails — it is unambiguous.

## What is now observable
- **`GET /api/admin/capstone-funnel`** (new, admin-authed, service-role aggregate reads only, no PII): counts the four funnel stages for the capstone course — deploy-lesson **completions** vs verified **deploys** (`deployed_programs`) vs **deferrals** (unresolved `certificate` queue rows parked by the capstone gate) vs issued **certificates** — plus a `verdict`. `completions > 0 AND deploys = 0 AND deferrals > 0` ⇒ `write_path_suspect` (write path broken, not merely idle). A failed read ⇒ `unavailable`, never a misleading `healthy`. On `write_path_suspect` it also emits an error-level structured log, so the signal reaches log-based alerting, not only whoever opens the route.
- **Structured logging in `/api/deploy/save` on every deny/exception/success path.** Previously the outer `catch {}` swallowed *all* unexpected throws into a silent generic 500, and the RPC-unreachable 503 was silent too — precisely the invisible failures #725 is about. Now: `deploy-save.rejected` (with the on-chain reason, still generic to the client per #560), `deploy-save.rpc-unavailable`, `deploy-save.write-failed`, `deploy-save.unhandled`, and `deploy-save.saved` (info) so the first real prod success is visible.

## Bug found in the save path
No functional bug in verification/write logic — the #620 verification is sound. The concrete defect fixed is the **silent `catch {}`**: any unexpected throw returned a generic 500 with zero logging, making a broken save indistinguishable from a working one in the logs. That is the failure mode #725 describes, so it is fixed with a test that forces a throw and asserts the `deploy-save.unhandled` log.

## Test evidence
- New `capstone-funnel.test.ts` (unit): per-table count scoping (asserts the exact eq/is/like filters) + every `verdict` transition + `unavailable` on read error.
- New admin route test: 401 without admin auth, returns counters, alert log only on `write_path_suspect`.
- Extended `deploy/save` route test: asserts the structured log on success + each deny/exception path, and a test pinning that a save stamps **exactly the (user_id, course_id, lesson_id) keys `checkCapstoneCredentialGate` reads** — using the real `CAPSTONE_CREDENTIAL` constants, so a content rename that moves the gate breaks this test too.

```
pnpm install --frozen-lockfile   # ok (frozen)
pnpm typecheck                   # 6 packages, all successful
pnpm --filter web test           # 202 files, 1821 tests passed
```

SENSITIVE (credentials-adjacent): do not merge without adversarial + gate review. Head: e6b6d7b8909703f2636dac9291b20b25c7bbb5e3

### DIFF
--- a/apps/web/src/app/api/admin/capstone-funnel/__tests__/route.test.ts
+++ b/apps/web/src/app/api/admin/capstone-funnel/__tests__/route.test.ts
@@ -1,0 +1,99 @@
+/* eslint-disable import/order -- vi.mock('server-only') must be hoisted above
+   the route import so the `server-only` graph loads under vitest. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import type { NextRequest } from "next/server";
+
+vi.mock("server-only", () => ({}));
+
+const { requireAdminAuth, getCapstoneFunnel, logEvent } = vi.hoisted(() => ({
+  requireAdminAuth: vi.fn(),
+  getCapstoneFunnel: vi.fn(),
+  logEvent: vi.fn(),
+}));
+
+vi.mock("@/lib/admin/auth", () => ({
+  AdminAuthError: class extends Error {},
+  adminUnauthorizedResponse: () => new Response("{}", { status: 401 }),
+  requireAdminAuth,
+}));
+vi.mock("@/lib/supabase/admin", () => ({
+  createAdminClient: () => ({}) as never,
+}));
+vi.mock("@/lib/credentials/capstone-funnel", () => ({ getCapstoneFunnel }));
+vi.mock("@/lib/logging", () => ({ logEvent }));
+
+import { GET } from "../route";
+import { AdminAuthError } from "@/lib/admin/auth";
+
+function reqFor(): NextRequest {
+  return new Request(
+    "https://x/api/admin/capstone-funnel"
+  ) as unknown as NextRequest;
+}
+
+const healthyFunnel = {
+  courseId: "course-building-first-program",
+  deployLessonId: "lesson-bfsp-m4-capstone",
+  completions: 5,
+  deploys: 3,
+  deferrals: 0,
+  certificates: 3,
+  verdict: "healthy" as const,
+};
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  requireAdminAuth.mockReturnValue(undefined);
+  getCapstoneFunnel.mockResolvedValue(healthyFunnel);
+});
+
+describe("GET /api/admin/capstone-funnel", () => {
+  it("401s when admin auth fails, without reading the funnel", async () => {
+    requireAdminAuth.mockImplementation(() => {
+      throw new AdminAuthError();
+    });
+
+    const res = await GET(reqFor());
+
+    expect(res.status).toBe(401);
+    expect(getCapstoneFunnel).not.toHaveBeenCalled();
+  });
+
+  it("returns the funnel counters for an authed admin", async () => {
+    const res = await GET(reqFor());
+
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual(healthyFunnel);
+  });
+
+  it("does not raise an alert log when the funnel is healthy", async () => {
+    await GET(reqFor());
+    expect(logEvent).not.toHaveBeenCalled();
+  });
+
+  it("emits a loud error log when the funnel signals a broken write path", async () => {
+    getCapstoneFunnel.mockResolvedValue({
+      ...healthyFunnel,
+      completions: 4,
+      deploys: 0,
+      deferrals: 2,
+      certificates: 0,
+      verdict: "write_path_suspect",
+    });
+
+    const res = await GET(reqFor());
+
+    expect(res.status).toBe(200);
+    expect(logEvent).toHaveBeenCalledWith(
+      expect.objectContaining({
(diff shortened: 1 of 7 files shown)


thomgabriel:
## GATE: approved

**Premise confirmed against prod**, not taken on faith: `deployed_programs` = **0 rows**, `certificates` = 4, unresolved certificate queue rows = 0, queue total 7. The route that writes `deployed_programs` has indeed never succeeded in production, and #721 now hangs the capstone credential off exactly that row.

**The fail-direction is right where it matters.** `countOrNull` returns `null` on any read error and the verdict collapses to `unavailable` — a monitor that cannot read must never report `healthy`. `decideVerdict` only reaches `healthy` on `deploys > 0`; everything else lands on `idle` / `watch` / `write_path_suspect`. That's the correct asymmetry for an alarm (same shape as #720's spend ceiling failing closed).

**Checked the one silent coupling.** The deferral count greps `last_error LIKE 'capstone-gate-%:<courseId>'`, which is a string contract with a different file. It matches the writer exactly — `onchain-queue.ts:683` writes `capstone-gate-${reason}:${courseId}`. Worth knowing that if that format ever drifts, deferrals read 0 and the verdict softens `write_path_suspect` → `watch`; it degrades toward "look at this" rather than toward "healthy", so the drift is tolerable, but the two are now coupled by an unenforced string.

**Auth/PII**: `requireAdminAuth` before any read, service-role client, head+count only (no rows transferred), response is four integers plus ids. `logEvent` emits ids and a reason — the `console.warn` it replaces already logged `user.id`, so this is no wider, and the client still gets the generic `VERIFICATION_FAILED` per #560.

The real find here is the silent `catch {}` in `/api/deploy/save`: every unexpected throw became a generic 500 with no log, which is *precisely* the indistinguishability #725 describes. Fixing that with a test that forces the throw and asserts `deploy-save.unhandled` is the right shape.

Accurate scope note, and I'd underline it: **this is the observability half only.** #725 stays open until a real devnet deploy produces the first `deployed_programs` row during the #606 window. What this PR buys is that when that run happens, success or failure is unambiguous instead of silent.

CI green — 202 files / 1821 tests, typecheck clean across 6 packages.

thomgabriel:
**Orchestrator routing record — SENSITIVE (credentials-adjacent deploy/save + new admin route) → adversarial review PASS → queuing for gate.**

**My independent verification** at `e6b6d7b`: typecheck exit 0; full web suite **202 files / 1821 tests, exit 0**.

**Adversarial review (the same reviewer that verified #721's gate): PASS.**
- **Logging is the ONLY behavioral delta** — proven by grep-filtering the diff to non-logging lines: verification calls, retry, upsert keys + onConflict, every early return/status byte-identical to main; #620's verify untouched; the `saved` log sits after the upsert.
- Client bodies stay generic per #560 (reason only in logs, test-pinned); funnel response is aggregate-only, zero PII; admin auth = the real convention (signed `admin_session` HMAC cookie, timingSafeEqual, fails closed 401 before any read).
- **Verdict logic enumerated and correct**: deploys>0 ⇒ healthy (partial success and late-clearing queues both healthy — the path demonstrably works); all-zeros pre-launch ⇒ idle, not suspect; single-count read error ⇒ `unavailable`, never misleadingly healthy. Deferral LIKE prefix verified EXACT against #721's `deferForCapstoneGate` write (`capstone-gate-${reason}:${courseId}`, no drift, no false positives).
- Key-pinning test imports the REAL `CAPSTONE_CREDENTIAL` constants — a content rename that moves the gate breaks this test too.

**INFOs recorded (neither blocks):**
1. Latent-not-live: `.rpc-unavailable`/`.unhandled` log raw `err.message`; on Vercel's runtime (undici) the Helius key does NOT leak, but the node-fetch fallback's FetchError would embed the full `?api-key=` URL. Trivial hardening (strip query / log err.name) — gate's call whether to demand it inline or as polish.
2. Pre-existing, unchanged, O-2-consistent (also flagged in the #721 review): the gate attests 'a real deploy under the learner's authority', not 'the capstone deploy specifically' — deploy/save takes courseId/lessonId from the body and verifies the program only.

**Scope note:** #725's step 1 (one real end-to-end devnet deploy producing a row) requires on-chain txs — deliberately left for the owner's #606 window; this PR makes that run's outcome unambiguous either way. No migration.

PR #745, opened 2026-07-26
fix(content-lint): fail closed on unresolvable base ref; restore-not-regenerate deleted deployed lock (#737 follow-up)

thomgabriel:
Follow-up to #742 (already squash-merged). Adversarial review of that PR found two latent holes in the base-ref handling shared by gate-2 (id immutability) and gate-3 (slots immutability). Both are fixed here.

## 1. MAJOR — fail-open on an explicit-but-unresolvable base ref
#742 collapsed two distinct states into one warning-and-skip:
- **bare local, no base ref** — warning + skip is correct (nothing to diff against).
- **base ref EXPLICITLY set (CI's `LINT_BASE_REF`) but unresolvable** — e.g. the ref was never fetched. `mergeBase` returns null → warning + skip + **exit 0**.

The second case regressed the hard gate to **fail OPEN**: a future workflow that stopped doing `fetch-depth: 0` would silently disable slot/id immutability, and gate-2 skipped on a null base too, so nothing backstops. Reproduced: baseRef set, `origin/main` absent, slot renumbered → lint green.

**Fix:** distinguish the two. No baseRef at all (bare local) stays a warning-skip. A baseRef that was *provided* but does not resolve is now a hard **ERROR** — `CI misconfiguration: told to diff … against "<ref>" but it is not resolvable — fetch it (fetch-depth: 0) or fix LINT_BASE_REF`. Applied identically to `gate2-ids.ts` and `gate3-slots.ts`.

## 2. MINOR — destructive advice for a deleted deployed lock
`missingLockDiag` branched only on working-tree absence, so deleting a **deployed** course's `slots.lock.json` errored with *"a new course needs one generated from its lesson order"* — reintroducing the renumber-and-corrupt class for the accidental-deletion case.

**Fix:** gate-3 now consults the base ref. If the base HAD a lock at that path, the message is *"slots.lock.json was deleted for a deployed course — restore it from the base ref (`git show <base>:<path>`); do NOT regenerate it"*. New-course guidance is kept only when the base has no lock either. A present-but-unparseable lock keeps its existing fix-by-hand message.

## Tests
`pnpm --filter @superteam-lms/content-lint test` → **88 passed** (gate2: 5, gate3: 10). `content-schema` → 170. `typecheck` clean on both touched packages. New coverage: explicit-but-unresolvable base → ERROR (both gate-2 and gate-3), no warning; deleted-deployed-lock → restore message with no "generate" text; the bare-no-base warning path is unchanged.

## courses-academy blast matrix (linter @ this branch vs courses-academy@main, read-only)
| invocation | result |
|---|---|
| bare, `origin/main` present | exit 0, no gate-2/gate-3 diagnostics (resolves origin/main) |
| CI-sim `GITHUB_BASE_REF=main` | exit 0 (byte-identical to prod) |
| bare, no `origin` remote | exit 0, gate-3 **warning** "no base ref" |
| **`LINT_BASE_REF=origin/does-not-exist`** (explicit but unresolvable) | **exit 1**, gate-2 AND gate-3 **error** "CI misconfiguration" (fail closed) |

### DIFF
--- a/packages/content-lint/src/__tests__/gate2.test.ts
+++ b/packages/content-lint/src/__tests__/gate2.test.ts
@@ -106,6 +106,31 @@ describe("gate 2 — ids", () => {
     ).toBe(false);
   });
 
+  it("fails CLOSED when an explicit base ref is set but unresolvable — #737 (no fail-open)", async () => {
+    // LINT_BASE_REF points at a ref that was never fetched. Id immutability cannot
+    // be checked; a silent skip would fail OPEN. It must ERROR.
+    const root = makeTempRepo({
+      "courses/a/course.yaml": course("course-x"),
+    });
+    const git = (...args: string[]) => execFileSync("git", args, { cwd: root });
+    git("init", "-q");
+    git("config", "user.email", "t@t.t");
+    git("config", "user.name", "t");
+    git("add", "-A");
+    git("commit", "-qm", "c1");
+    git("branch", "-M", "main");
+
+    const r = await runLint(root, { baseRef: "origin/does-not-exist" });
+    expect(
+      r.diagnostics.some(
+        (d) =>
+          d.gate === "gate-2" &&
+          d.severity === "error" &&
+          /misconfiguration/i.test(d.message)
+      )
+    ).toBe(true);
+  });
+
   it("emits a warning (never silent) when no exact merge-base exists", async () => {
     // An orphan base branch shares NO history with HEAD, so `git merge-base`
     // fails and the check degrades to the base tip — which must be surfaced.
--- a/packages/content-lint/src/__tests__/gate3.test.ts
+++ b/packages/content-lint/src/__tests__/gate3.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from "vitest";
 import { execFileSync } from "node:child_process";
-import { writeFileSync } from "node:fs";
+import { writeFileSync, unlinkSync } from "node:fs";
 import { join } from "node:path";
 import { runLint } from "../lint";
 import "../checks/gate1-schema";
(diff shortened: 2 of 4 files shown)


thomgabriel:
@claude please re-review. Addressed the Blocking finding on the gate-3 DANGER note: the hardcoded "38+ learners" literal (which fired even for a zero-enrollment new course) is replaced with the generic "re-points every already-enrolled learner's completed-lesson bitmap" — no fabricated count — and pinned in the message test (`gate3.test.ts`). Also in this branch: the fail-open ERROR on an explicit-but-unresolvable base ref (gate-2 + gate-3) and the restore-not-regenerate message for a deleted deployed lock. Head is 9fa6b9a; content-lint 88 tests + content-schema 170 green, typecheck clean.

thomgabriel:
**Orchestrator routing record — SENSITIVE (lint-gate semantics) → adversarial re-verify PASS → queuing for gate.**

Context: #742 merged before its fix round, so all three review findings landed here, cherry-picked onto main (diff vs main = exactly the 4 content-lint files; `git.ts` untouched — verified no double-application).

**Adversarial re-verify at `9fa6b9af` (same reviewer, same fixture): PASS, per-attack:**
- ATTACK D (the MAJOR): explicit-but-unresolvable baseRef + tampered slot → **exit 1, both gate-2 AND gate-3** emit the 'CI misconfiguration … was NOT verified' ERROR (was exit-0 fail-open). Bare-local no-baseRef keeps the warning (row 3 unchanged).
- ATTACK B: deleted deployed lock → restore-from-base message, zero 'generate' advice; genuinely-new course still gets new-course guidance (`baseHadLock` draws the line correctly).
- claude[bot]'s Blocking: '38+ learners' literal gone, generic phrasing pinned by a no-count-adjacent-to-learner regex.
- 4-row blast matrix green (bare+origin / CI-sim / no-remote / explicit-unresolvable).

**My independent verification** at `9fa6b9af`: content-lint **88/88**, typecheck exit 0, grep '38+' = 0. Bot re-review was in progress at queue time — read it before merging.

thomgabriel:
Rebased over #746 and reopened as **#747** — same commits, authorship preserved.

I resolved the conflict myself because I caused it: #746 merged the warn-and-skip pattern into `gate2-ids.ts` while this was in flight. Force-pushing to this branch was blocked by a permission classifier, so #747 is a new branch rather than an update in place.

**Your MAJOR finding is confirmed independently.** `summarize()` sets `ok` from errors only, so a warning exits 0 — the gate really did pass green without running. #746 had just widened that hole to gate-2. The conflict resolution keeps three distinct states rather than picking a side: explicit-but-unresolvable → ERROR (yours), no ref at all → warning (mine), resolvable → run the check.

90 tests pass with both semantics asserted by name. Closing this in favor of #747.

Issue #627, opened 2026-07-26
content-lint: verify achievement award refs; rule on the dead path draft flag


thomgabriel:
Two lint gaps surfaced by #559 (courses-academy PR #7):

1. **Achievement award refs are not lint-verified** (award.path / award.course). Retiring path-solana-core would have shipped achievements/full-stack-solana.yaml permanently unearnable — buildUserState requires path.courseIds.length > 0 (apps/web/src/lib/gamification/achievements.ts:164-171) — and only a manual catch prevented it. A gate-4 extension resolving award refs against live, non-empty paths/courses closes this class.

2. **paths draft: true is dead code end-to-end** — the compiler carries it into the bundle (compile/projector.ts:166-178) but zero runtime reads exist; hiding is empty-courses-only, and course visibility filters only on isSynced (queries.ts:171-189), so a course in a drafted-but-populated path leaks into All Courses/search/detail. Either wire a runtime filter or lint-forbid draft:true without courses:[] and document empty-courses as the sole mechanism.

Related: #596 (skills-tag gate) is the same family of registry-resolution lint work.

_Filed from the orchestrator loop; discovered implementing #559._

thomgabriel:
**First half done** (#749 merged: gate-4a verifies achievement award refs resolve to earnable content — course existence, path exists+non-empty per buildUserState's semantics, exhaustive over award kinds so future ref-bearing kinds fail the build; proven to catch the #559 empty-path case on real content). **Remaining half = the path `draft:true` flag ruling** — design decision tangled with #711/D-1 (the flag drafts the path but not its courses), owner's call. Relabeled needs-human.

thomgabriel:
**Owner ruling 2026-07-28: draft:true + empty courses becomes a lint ERROR.** The implicit retirement pattern is retired. Implementation spec (loop-actionable now):
1. Add an explicit `retired: true` field to the path schema (courses-academy) — semantically distinct from draft ('coming soon') vs retired ('permanently emptied, id preserved').
2. Migrate the two current draft+empty paths (`frontend`, `defi`) to `retired: true` (+ the 4 always-empty scaffold paths as appropriate — audit which are 'future' vs 'retired').
3. content-lint: ERROR on draft+empty-courses unless `retired: true`; `retired` paths must have empty courses (error otherwise).
4. App runtime already hides empty-course shelves; verify no runtime change needed (the runtime never read `draft` — confirm it also ignores `retired`).
[...]

Issue #596, opened 2026-07-25
content-lint: gate skill tags against skills.yaml


thomgabriel:
`content-lint` does not validate skill-tag vocabulary; `checkSkillVocabulary` only runs at sync time, so a course can pass CI and then fail to activate (exactly what nearly happened with C5). Cheapest item on the blocker list.

**Unified spec item 7** — full detail, evidence and acceptance criteria in [`docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md`](docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md) §3.

_Filed from the unified launch spec; this work existed in the plan but had no issue._

Issue #559, opened 2026-07-25
Launch content PR: Zero-to-Deployed path + capstone Earn terminus + defi hold


thomgabriel:
One courses-academy PR, one content.lock bump, four atomic changes.

**Spec tasks:** `LX-D1`

Full context — what to build (file-level), acceptance criteria, effort, dependencies and wave — is in [`docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md`](docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md), section 3 (Workstreams) under the task ids above; section 4 has the delivery sequencing.

Derived from the 2026-07-25 research program (pedagogy / UI-UX / personalization+launch reports in the same directory), code-audited before filing.

thomgabriel:
LX-D1 content PR is up: **https://github.com/solanabr/courses-academy/pull/7** (owner-gated — do not merge without sign-off; no content.lock bump in this repo yet by design).

**Exact two-step activation sequence:**

1. **Merge courses-academy PR #7** to its `main` (owner-gated). Nothing changes in production at this point — content only stages.
2. **Monorepo activation PR**: bump `apps/web/content.lock` to the merged courses-academy `main` SHA and commit the regenerated bundle in the same PR (`pnpm --filter web compile-content`; CI byte-verifies). One lock bump activates all four changes atomically.

**Flagged risk (verified, file:line in the PR body):** the runtime never reads the path `draft` flag — shelves hide only via empty `courses` (done in the PR) — and `getAllCourses()` gates only on synced+active deployment (`apps/web/src/lib/content/queries.ts:171-181`), so after activation `course-defi-on-solana` **still shows in the All Courses tab/search/course page** (it only leaves the paths tab, pills, and landing explorer). Fully hiding it requires deactivating the course on-chain via the existing admin gate — intentionally not done (owner-gated); optional step 3 if the leak is unacceptable for launch.

One consequence fix rode along, flagged for veto in the PR: `full-stack-solana`'s award retargeted `path-solana-core → path-zero-to-deployed`, because the retired (emptied) path can never complete and the badge would ship permanently unearnable.

No slot renumbering in the diff; no file overlap with courses-academy PRs #5/#6; content-lint OK (0 errors) against monorepo-main tooling.

thomgabriel:
[...]

PR #749, opened 2026-07-26
feat(content-lint): gate-4a — verify achievement award refs resolve to earnable content (#627)

thomgabriel:
Implements the **first** half of #627: content-lint verification that achievement `award` refs resolve to earnable content. The second half (the dead `path draft:true` flag ruling) is deliberately left open — see below.

## Problem
An achievement's `award` rule is its **only** unlock path. If the rule names a course/path id that doesn't resolve — or a `path-completed` rule points at an **empty** path — the achievement is minted but can **never** be unlocked, and nothing downstream catches it: the doc is schema-valid, the ref is just dead. This is the class that nearly shipped `full-stack-solana` permanently unearnable in #559 (courses-academy#7) when `path-solana-core` was emptied; only a manual catch saved it.

## Gate 4a (new, `checks/gate4a-award-refs.ts`, error tier)
Mirrors the runtime consumer `buildUserState` (`apps/web/src/lib/gamification/achievements.ts`) exactly:

| award kind | check |
|---|---|
| `course-completed`, `lessons-completed-in-course` | course must exist (courses always have ≥1 lesson per schema `modules.min(1)`/`lessons.min(1)`, so existence is sufficient — `completedCourseIds` is keyed by real course ids) |
| `path-completed` | path must exist **and be non-empty** — `buildUserState` only adds a path to `completedPathIds` when `courseIds.length > 0` (achievements.ts:164-171), so an award on an empty path can never fire |
| `lessons-completed`, `streak`, `user-number`, `community-stat`, `manual` | reference no content id — ignored |

Error tier because a dangling ref is authoring breakage, not style. Registered via the existing `registerCheck` convention and exported from `index.ts`.

## Tests (`__tests__/gate4a.test.ts`, both directions)
- pass: `course-completed` / `lessons-completed-in-course` / `path-completed` naming existing (and non-empty) content
- error: missing course; missing path
- error: **existing but empty path** — the #559 unearnable case (drafted/emptied path)
- pass: the five non-ref award kinds are ignored

`pnpm --filter @superteam-lms/content-lint test` → 94 passed (gate4a: 7). `content-schema` → 170. `typecheck` clean.

## courses-academy blast check — GREEN
Ran the linter (this branch) against `courses-academy@main`, bare and CI-simulated — **0 gate-4a diagnostics** both ways (gate-4a is git-independent, so the two are identical). The gate is **not** vacuously green: it actively resolves 4 real refs and all pass —

| achievement | award | resolves to | ok |
|---|---|---|---|
| anchor-expert | course-completed | course-anchor-framework | ✓ |
| course-completer | course-completed | course-solana-fundamentals | ✓ |
| rust-rookie | lessons-completed-in-course | course-rust-for-solana | ✓ |
| full-stack-solana | path-completed | path-zero-to-deployed (4 courses) | ✓ |

Notably `path-solana-core` in that repo is `draft: true, courses: []` (the emptied path from #559). `full-stack-solana` was retargeted off it to the non-empty `path-zero-to-deployed`, so the gate is green — had that retarget not happened, gate-4a would have flagged it. So the gate reproduces the #627 protection on the exact content that motivated it.

## Out of scope (stays open on #627)
The second half — **ruling on the dead `path draft:true` flag** (compiler carries it into the bundle but no runtime reads it; a drafted-but-populated path leaks its courses into All Courses/search) — is a design decision entangled with #711 / D-1 (owner) and is intentionally **not** addressed here. #627 should remain open for it.

Related: #596 (skills-tag gate) is the same registry-resolution family.

### DIFF
--- a/packages/content-lint/src/__tests__/gate4a.test.ts
+++ b/packages/content-lint/src/__tests__/gate4a.test.ts
@@ -1,0 +1,146 @@
+import { describe, it, expect } from "vitest";
+import { runLint } from "../lint";
+import "../checks/gate1-schema";
+import "../checks/gate4a-award-refs";
+import { makeTempRepo } from "./helpers";
+
+const course = (id: string) => `id: ${id}
+slug: ${id.replace("course-", "")}
+title: ${id}
+difficulty: beginner
+duration: 1
+xpPerLesson: 10
+xpReward: 100
+modules: [{ key: m, title: M, lessons: [lesson-a] }]
+`;
+
+const path = (id: string, courses: string[]) => `id: ${id}
+slug: ${id.replace("path-", "")}
+title: ${id}
+difficulty: beginner
+courses: [${courses.join(", ")}]
+`;
+
+/** A schema-valid empty path — allowed only as an explicit draft. */
+const draftEmptyPath = (id: string) => `id: ${id}
+slug: ${id.replace("path-", "")}
+title: ${id}
+difficulty: beginner
+draft: true
+`;
+
+const achievement = (id: string, award: string) => `id: ${id}
+name: ${id}
+category: progress
+xpReward: 50
+award: ${award}
+`;
+
+const g4a = (
+  r: Awaited<ReturnType<typeof runLint>>
+): { severity: string; message: string }[] =>
+  r.diagnostics.filter((d) => d.gate === "gate-4a");
+
+describe("gate 4a — achievement award refs", () => {
+  it("passes when course-completed names an existing course", async () => {
+    const root = makeTempRepo({
+      "courses/x/course.yaml": course("course-x"),
+      "achievements/a.yaml": achievement(
+        "achievement-a",
+        "{ kind: course-completed, course: course-x }"
+      ),
+    });
+    expect(g4a(await runLint(root))).toEqual([]);
+  });
+
+  it("errors when course-completed names a missing course", async () => {
+    const root = makeTempRepo({
+      "achievements/a.yaml": achievement(
+        "achievement-a",
+        "{ kind: course-completed, course: course-ghost }"
+      ),
+    });
+    const d = g4a(await runLint(root));
+    expect(
+      d.some((x) => x.severity === "error" && /course-ghost/.test(x.message))
+    ).toBe(true);
+  });
+
+  it("errors when lessons-completed-in-course names a missing course", async () => {
+    const root = makeTempRepo({
+      "achievements/a.yaml": achievement(
+        "achievement-a",
+        "{ kind: lessons-completed-in-course, course: course-ghost, gte: 3 }"
+      ),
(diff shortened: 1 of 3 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (lint semantics) → adversarial review PASS + MINOR folded in → queuing for gate.**

**Adversarial review at `ccf4f395`: PASS.** Fidelity to `buildUserState` verified case-by-case (course existence sound because the schema forbids empty courses AND the enrollment-flag completion path is independent of the live slot set — the fully-retired-course edge chased and closed); path exists+non-empty matches the runtime's `courseIds.length > 0` credit exactly; per-course path membership covered by the registered gate-4 (two-gate contract noted). Mutation-tested (breaking the empty-path check fails exactly the #559 test). Wired e2e through the REAL CLI (dangling ref → exit 1). **Blast on courses-academy: green and non-vacuous** — resolves 4 real refs, and re-pointing full-stack-solana at the #559-emptied `path-solana-core` errors exactly as designed; this gate reproduces the protection that motivated #627.

**MINOR folded in by me at the new head:** the award-kind switch is now exhaustive — non-ref kinds enumerated explicitly + a `never` default, so a future ref-bearing kind fails the BUILD here instead of being silently skipped (the #742/#745 fail-open class). Re-verified: content-lint 94/94, typecheck exit 0.

#627 stays open for its second half (the path draft-flag ruling — owner/D-1 territory).

thomgabriel:
## GATE: approved

Verified the coverage question myself, since a lint gate that silently ignores a kind is worse than no gate: `AWARD_KINDS` has exactly 8 members, and this accounts for all 8 — 2 course-ref, 1 path-ref, 5 correctly ignored as carrying no content id.

Confirmed the `path-completed` non-empty rule against the runtime rather than the description: `buildUserState` gates on `path.courseIds.length > 0 && every(...)` ([achievements.ts:164](apps/web/src/lib/gamification/achievements.ts:164)), so an award on an empty path genuinely can never fire. The gate mirrors it exactly.

The blast check is the part I'd highlight — it resolves 4 real refs against `courses-academy@main` and passes, and you showed *why* it isn't vacuous: `path-solana-core` is still `draft: true, courses: []`, and had `full-stack-solana` not been retargeted off it, gate-4a would have caught the #559 case. A green gate that demonstrates what it would have caught is worth more than the green.

Scope is honest — "first half of #627", no `Closes`, draft-flag ruling left with #711/D-1.

Taking the reviewer's exhaustiveness point as a follow-up (#753): the `switch (award.kind)` has no `never` guard, so a future ref-carrying kind falls through silently — the exact failure this gate exists to prevent. `achievements.ts` enforces it structurally with `satisfies Record<AwardKind, Predicate>`; this should too. Not blocking, all 8 covered today.

CI green.

PR #751, opened 2026-07-26
fix(lessons): derive on-chain bitmap index from slot lock, not array position (#741)

### LINKED ISSUE #741: Lesson bitmap index is array position, not slot — slots.lock.json has no learner-path reader
## Summary

The on-chain lesson bitmap index is derived from a lesson's **position in the flattened module array**, not from its slot in `slots.lock.json`. The entire slot-immutability discipline — `slots.lock.json`, gate-3, "assigned once, never renumbered, never reused" — exists to keep that bitmap stable, but **no learner-facing code path reads `slots.json`**. The two have never been connected.

This is latent today (every deployed course is dense, so `arrayIndex == slot` by accident) and becomes live the first time a deployed course retires or inserts a lesson. #740 is that first time.

## The code

`/api/lessons/complete` derives the index it sends on-chain:

```ts
// apps/web/src/app/api/lessons/complete/route.ts:52-61
async function deriveLessonIndex(courseId: string, lessonId: string): Promise<number> {
  const course = await getCourseById(courseId);
  ...
  const index = findLessonIndex(course, lessonId);
```

```ts
// apps/web/src/lib/courses/lesson-index.ts:14-17
export function findLessonIndex(course: CourseLike, lessonId: string): number {
  const allLessons = (course.modules ?? []).flatMap((m) => m.lessons ?? []);
  return allLessons.findIndex((l) => l._id === lessonId);   // <-- array position
}
```

`getCourseById` → `projectCourse(..., {fullLessons: true})` preserves module and lesson order, so this is literally the ordinal of the lesson in the authored course structure.

Readers of `slots.json`, exhaustively: `app/api/admin/courses/sync/route.ts` and `lib/solana/admin-signer.ts` — both admin-only, both for deriving the `active_lessons` mask to send on-chain. Nothing on the learner path.

## Why it's a correctness bug, not a style issue

Retiring a slot is supposed to be safe: the slot stays burned, survivors keep their numbers, `finalize_course` does a subset test (`flags & active == active`) so a retired bit never blocks completion. That contract holds on-chain. It is broken on the way in, because the index the client sends is recomputed from array order every time the bundle changes.

Concretely, applying the C3 transform (#740) to `course-building-first-program`: **9 of 12 surviving lessons change their bitmap index** while their slot is stable —

| lesson | slot (stable) | old idx | new idx |
|---|---|---|---|
| `your-first-build` | 1 | 1 | 0 |
| `add-instruction` | 3 | 3 | 1 |
| `adding-instructions` | 8 | 8 | 2 |
| `on-chain-state` | 4 | 4 | 3 |
| `wire-up-initialize` | 7 | 7 | 6 |
| `build-increment` | 9 | 9 | 8 |
| `complete-counter` | 10 | 10 | 9 |
| `account-constraints` | 6 | 6 | 10 |
| `m4-capstone` | 15 | 15 | 14 |

The failure modes follow directly from the idempotency check at `route.ts` (`isLessonComplete(flags, lessonIndex)` against the *new* index over the *old* bitmap):

1. **Silent false completion.** New index lands on a bit set by a different, now-retired lesson → the route treats the lesson as already done, skips the on-chain tx, writes `user_progress.completed = true`, awards no XP.
2. **Duplicate XP + wrong slot marked.** New index lands on an unset bit → `complete_lesson` fires for a lesson the learner already finished, re-awarding XP and setting the wrong on-chain bit.
3. **Credential permanently unreachable.** New content tops out at index 14; while the on-chain mask is the dense 16-slot one, bit 15 can never be set, so `flags & active == active` never holds and `finalize_course` can never pass.

## This is already true of live data

Prod (`pywhtmidcrptomrabbrw`), `course-building-first-program`:

```
lesson-bfsp-your-first-build   idx  1  completed  tx: no
lesson-bfsp-m4-interact        idx 14  completed  tx: YES
lesson-bfsp-from-code-to-chain idx  -  completed  tx: no
lesson-bfsp-anatomy-anchor-program idx - completed tx: no
lesson-bfsp-on-chain-state     idx  -  completed  tx: no
```

`m4-interact` is **retired** by C3 and has a real on-chain completion at **bit 14**. C3 moves `m4-capstone` to index 14. That learner's capstone would be marked complete on first visit without doing it — and the capstone is exactly what gates the credential (#721).

## Fix

Make the completion path slot-aware: resolve `lessonId → slot` from the committed `slots.json` (the same lockfile `admin/courses/sync` already trusts) and send that as the bitmap index. `findLessonIndex`'s array-position result is fine for *display* ordering but must stop feeding the on-chain call. A test should assert `slot(lessonId)` is what reaches `completeLesson`, with a fixture whose array order deliberately disagrees with its slots.

Two things to settle in the same change:

- **The count guard is in the wrong space.** `route.ts` compares `lessonIndex >= onChainCourse.liveLessonCount`, and `liveLessonCount` is `popcount(active_lessons)`. Once a mask is sparse those are different spaces: with the C3 mask, `popcount = 15` while live slots run to 18, so slots 15–18 (capstone included) would 409 as "recently extended". The existing comment predicts this ("Once a sparse-mask course can exist, this must become a per-slot bit test"). It needs to become a bit test against `activeLessons`.
- **Backfill.** `user_progress.lesson_index` holds old-space values (and nulls). Decide whether it is re-derived or left as historical record.

## Blast radius

Small today — 1 enrollment, 5 completed rows, 2 learners, 0 certificates on this course in prod — which is the argument for fixing it now rather than after launch traffic.

Blocks #740.

thomgabriel:
Closes #741. Prerequisite for re-landing #740 (reverted in #744).

## What was wrong

The on-chain bitmap index (sent to `complete_lesson`, tested against `lesson_flags`, written to `lesson_index`) was derived from a lesson's **position in the flattened module array**, not its **slot** in `slots.lock.json`. The slot lock exists precisely to keep that bit stable across reorders/retirements, but **no learner-facing code read it**. Dense courses made `arrayIndex == slot` by accident; a restructure (#740 C3) remaps 9 of 12 survivors → a `lessonId` lands on a different (possibly retired) bit: silent false-completion, double XP, or an unreachable credential.

## Fix — every on-chain index consumer is now slot-aware

New `lib/courses/lesson-slot.ts`:
- `getLessonSlot(courseId, lessonId)` — reads the committed slot lock (`slotsByCourseId`). **FAIL-CLOSED**: throws on a missing lock or unslotted lesson; no array-index fallback.
- `slotToLiveLessonId(courseId)` — reverse map slot→lessonId over live slots (retired absent), for reconstructing progress from a bitmap.

Consumers converted:
- **`lessons/complete/route.ts`** — slot → `complete_lesson`, slot-bit idempotency, slot → `lesson_index`. Count guard rewritten from `slot >= popcount(active_lessons)` to a **per-slot bit test** against `activeLessons` (a sparse mask makes popcount != highest live slot, so the count compare would 409 a valid capstone at slot 15).
- **`admin/resync/route.ts`** — reconstructs `user_progress` by iterating live slots + `isLessonComplete(flags, slot)` instead of `bit i → allLessons[i]`.
- **`lib/helius/resolvers.ts` (`resolveLessonId`)** — *found by the adversarial reviewer.* The Helius webhook mapped the event index via `allLessons[index]`; the event index IS the slot, so on a sparse course it wrote a phantom `user_progress` row for the WRONG lesson + an XP reason naming it, on **every** webhook-mirrored completion. Now `slotToLiveLessonId(courseId).get(index)`; a retired/unsynced slot → null, which `handleLessonCompleted` already handles gracefully (skips upsert, XP reason falls back to the index). Never throws.
- **`lib/lessons/batch-complete.ts` (test-out, #730)** — *found by the adversarial reviewer.* Loop index over `flatMap` was sent to `completeLesson`, tested against the bitmap, and written to `lesson_index`. All three now use `getLessonSlot` (fail-closed → the lesson is isolated into `pending` with a logged reason, batch continues). The count-based `total = min(len, popcount)` bound is replaced by a **per-slot `is_active_slot` bit test** so a live high slot is attempted and an un-activated slot is skipped without a reverted-tx fee.
- **`lib/courses/lesson-index.ts`** — `findLessonIndex` docstring corrected to DISPLAY-ONLY; no non-test callers remain.

## Call-site disposition (full sweep)

| Call site | Path | Disposition |
|---|---|---|
| `api/lessons/complete/route.ts` | learner on-chain completion | **FIXED → slot**; count guard → per-slot bit test |
| `api/admin/resync/route.ts` | admin progress reconstruction from bitmap | **FIXED → slot** (reverse map) |
| `lib/helius/resolvers.ts` `resolveLessonId` | **Helius webhook** → progress + XP attribution | **FIXED → slot** (reviewer-found); retired slot → null, graceful skip |
| `lib/lessons/batch-complete.ts` | **test-out batch** (#730) | **FIXED → slot** ×3 (reviewer-found); count bound → per-slot `is_active_slot` gate |
| `lib/courses/lesson-index.ts` `findLessonIndex` | pure helper | **docstring → display-only**; no non-test callers |
| `api/admin/courses/sync/route.ts`, `lib/solana/admin-signer.ts` | admin `active_lessons` mask derivation | **no change** — already read slots.lock |
| `lesson-client.tsx` prev/next, `continue-learning.ts` `findNextIncompleteLesson` | display navigation | **no change** — lessonId-keyed, order-safe display |
| `lib/solana/academy-program.ts` `completeLesson(lessonIndex)` | instruction sink | **no change** — forwards the number it's given (now the slot) |

Credit to the adversarial reviewer for catching the webhook + batch consumers that the first pass missed.

## Re-grep evidence (no other consumers)

`grep -rn "allLessons\[" src` → only `lesson-client.tsx` prev/next nav (display) + a comment in `resolvers.ts`.
`grep -rn "completeLesson(" src` → sink builder (`academy-program.ts`), `batch-complete.ts` (now slot), `continue-learning`/`hybrid-progress-service` (unrelated: nav / throwing stub).
`grep -rn "resolveLessonId" src` → single caller `event-handlers.ts:161` passing `event.lessonIndex` (already the slot; `lesson_index: event.lessonIndex` write is correct).
`grep -rn "lesson_index:|lessonIndex" src` → remaining hits are the bitmap helper (sink), the event decoder (reads the on-chain field), generated DB types, and the fixed files.

## Backfill

New `user_progress.lesson_index` writes are slot-space; historical rows left as a record (no live-DB backfill — self-correct on next completion/resync).

## Why safe to ship before #740 re-lands

Every current course is dense (slot == arrayIndex), proven by `lesson-slot-equivalence.test.ts` over the committed bundle. One deliberately-landmined sub-test asserts the bundle is all-dense today; #740's re-land relaxes that one test while the dense⟹equivalence and sparse-fixture tests remain the guarantees.

## Tests

- `lesson-slot.test.ts`, `lesson-slot-equivalence.test.ts` — slot resolution + bundle-wide equivalence.
- `slot-aware.test.ts` (route, real bitmap) — slot (not array pos) reaches `completeLesson`; retired bit-14 never false-completes the capstone; per-slot count guard 409s only when the slot isn't live and does NOT 409 the capstone at popcount==15; missing slot → 500, no tx.
- `resolvers.test.ts` — event slot → correct lesson when array order disagrees; retired slot → null; never throws.
- `batch-complete.test.ts` — capstone sent as slot 15 (not array pos); not-live slots skipped as `pending`; unslotted lesson isolated as `pending`, batch continues.
- `bitmap.test.ts` — C3 `isCourseComplete` cases (capstone slot 15 required; retired slot-14 burned bit ignored).

## Verification

- `pnpm --filter web test` — **1890 passed** (208 files)
- eslint clean on all changed files; zero `any`, strict TS.

Does NOT merge — SENSITIVE (on-chain trust path) → adversarial + gate.

### DIFF
--- a/apps/web/src/app/api/admin/resync/route.ts
+++ b/apps/web/src/app/api/admin/resync/route.ts
@@ -18,7 +18,8 @@ import {
   fetchEnrollment,
   fetchAchievementReceiptData,
 } from "@/lib/solana/academy-reads";
-import { decodeLessonBitmap } from "@/lib/solana/bitmap";
+import { isLessonComplete } from "@/lib/solana/bitmap";
+import { slotToLiveLessonId } from "@/lib/courses/lesson-slot";
 import { getAllCourses, getAllAchievements } from "@/lib/content/queries";
 import { calculateLevel } from "@/lib/gamification/xp";
 
@@ -149,32 +150,28 @@ export async function POST(req: NextRequest) {
       );
       results.enrollments++;
 
-      // Decode lesson bitmap → sync individual lesson progress. Display order is
-      // the array position now (spec §10.1), so the flattened order IS the
-      // on-chain bitmap order — matching findLessonIndex.
-      const allLessons = (course.modules ?? []).flatMap((m) => m.lessons ?? []);
-
-      const lessonCount = enrollment.lesson_count
-        ? Number(enrollment.lesson_count)
-        : allLessons.length;
-
-      if (enrollment.lesson_flags && lessonCount > 0) {
-        const bitmap = decodeLessonBitmap(enrollment.lesson_flags, lessonCount);
-
-        for (let i = 0; i < bitmap.length && i < allLessons.length; i++) {
-          const lesson = allLessons[i];
-          if (bitmap[i] && lesson) {
+      // Reconstruct lesson progress from the on-chain bitmap by SLOT, not array
+      // position (#741). A bit at index N means "slot N is complete"; slot N maps
+      // to a lessonId via the committed slot lock, which is stable across
+      // restructures. Iterating array position instead would mis-map every live
+      // lesson once a course is sparse (retire/insert), and would resurrect a
+      // retired slot's stale bit onto whichever lesson now sits at that array
+      // index. A retired slot has no entry in the live map, so its bit is ignored.
+      if (enrollment.lesson_flags) {
+        for (const [slot, lessonId] of slotToLiveLessonId(course._id)) {
+          if (isLessonComplete(enrollment.lesson_flags, slot)) {
             await supabase.from("user_progress").upsert(
               {
                 user_id: profile.id,
                 course_id: course._id,
-                lesson_id: lesson._id,
+                lesson_id: lessonId,
                 completed: true,
                 completed_at: enrollment.enrolled_at
                   ? new Date(
                       Number(enrollment.enrolled_at) * 1000
                     ).toISOString()
                   : new Date().toISOString(),
+                lesson_index: slot,
               },
               { onConflict: "user_id,lesson_id", ignoreDuplicates: true }
             );
--- a/apps/web/src/app/api/lessons/complete/__tests__/deny-reasons.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/deny-reasons.test.ts
@@ -88,7 +88,7 @@ vi.mock("@/lib/solana/academy-reads", () => ({
   fetchCourse: vi.fn(),
 }));
 vi.mock("@/lib/solana/bitmap", () => ({ isLessonComplete: vi.fn() }));
-vi.mock("@/lib/courses/lesson-index", () => ({ findLessonIndex: () => 0 }));
+vi.mock("@/lib/courses/lesson-slot", () => ({ getLessonSlot: () => 0 }));
 vi.mock("@/lib/logging", () => ({ logError: vi.fn() }));
 vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
 vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
--- a/apps/web/src/app/api/lessons/complete/__tests__/gate.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/gate.test.ts
@@ -77,7 +77,7 @@ vi.mock("@/lib/solana/academy-reads", () => ({
   fetchCourse: vi.fn(),
 }));
 vi.mock("@/lib/solana/bitmap", () => ({ isLessonComplete: vi.fn() }));
-vi.mock("@/lib/courses/lesson-index", () => ({ findLessonIndex: () => 0 }));
+vi.mock("@/lib/courses/lesson-slot", () => ({ getLessonSlot: () => 0 }));
 vi.mock("@/lib/logging", () => ({ logError: vi.fn() }));
 vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
 vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
--- a/apps/web/src/app/api/lessons/complete/__tests__/grading-dedup.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/grading-dedup.test.ts
@@ -87,7 +87,7 @@ vi.mock("@/lib/solana/academy-program", () => ({
 }));
 vi.mock("@/lib/solana/academy-reads", () => ({ fetchEnrollment, fetchCourse }));
 vi.mock("@/lib/solana/bitmap", () => ({ isLessonComplete }));
-vi.mock("@/lib/courses/lesson-index", () => ({ findLessonIndex: () => 0 }));
+vi.mock("@/lib/courses/lesson-slot", () => ({ getLessonSlot: () => 0 }));
 vi.mock("@/lib/logging", () => ({ logError: vi.fn() }));
 vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
 vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
--- a/apps/web/src/app/api/lessons/complete/__tests__/inflight-guard.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/inflight-guard.test.ts
@@ -78,7 +78,7 @@ vi.mock("@/lib/solana/academy-program", () => ({
 }));
 vi.mock("@/lib/solana/academy-reads", () => ({ fetchEnrollment, fetchCourse }));
 vi.mock("@/lib/solana/bitmap", () => ({ isLessonComplete }));
-vi.mock("@/lib/courses/lesson-index", () => ({ findLessonIndex: () => 0 }));
+vi.mock("@/lib/courses/lesson-slot", () => ({ getLessonSlot: () => 0 }));
 vi.mock("@/lib/logging", () => ({ logError: vi.fn() }));
 vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
 vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
--- a/apps/web/src/app/api/lessons/complete/__tests__/review-capture.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/review-capture.test.ts
@@ -75,7 +75,7 @@ vi.mock("@/lib/solana/academy-reads", () => ({
   fetchCourse: vi.fn(),
 }));
 vi.mock("@/lib/solana/bitmap", () => ({ isLessonComplete: vi.fn() }));
-vi.mock("@/lib/courses/lesson-index", () => ({ findLessonIndex: () => 0 }));
+vi.mock("@/lib/courses/lesson-slot", () => ({ getLessonSlot: () => 0 }));
 vi.mock("@/lib/logging", () => ({ logError: vi.fn() }));
 vi.mock("@/lib/content/deployments", () => ({ isCourseInMaintenance }));
 vi.mock("@/lib/platform/freeze", () => ({ isPlatformFrozen }));
--- a/apps/web/src/app/api/lessons/complete/__tests__/slot-aware.test.ts
+++ b/apps/web/src/app/api/lessons/complete/__tests__/slot-aware.test.ts
@@ -1,0 +1,211 @@
+/* eslint-disable import/order -- vi.mock calls must precede importing the route. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest } from "next/server";
+
+vi.mock("server-only", () => ({}));
+
+// Slot-aware completion path (#741). These tests drive the route through to the
+// on-chain submit with a C3-shaped sparse course: the target lesson's SLOT (15,
+// the capstone) differs from what its array position would be, and slot 14 is a
+// retired lesson whose bit may still be set on an old enrollment. They assert
+// the SLOT — not array position — is what reaches the bitmap read, the count
+// guard, and completeLesson. The bitmap helpers are REAL (not mocked) so the
+// slot arithmetic is exercised end-to-end.
+
+const {
+  getUser,
+  singleProfile,
+  upsertProgress,
+  getLessonByIdForGrading,
+  isOnChainProgramLive,
+  isRateLimited,
+  isCourseInMaintenance,
+  isPlatformFrozen,
+  completeLesson,
+  fetchEnrollment,
+  fetchCourse,
+  getLessonSlot,
+} = vi.hoisted(() => ({
+  getUser: vi.fn<() => Promise<unknown>>(),
+  singleProfile: vi.fn<() => Promise<unknown>>(),
+  upsertProgress: vi.fn<(row?: unknown) => Promise<unknown>>(),
+  getLessonByIdForGrading: vi.fn(),
+  isOnChainProgramLive: vi.fn<() => Promise<boolean>>(),
+  isRateLimited: vi.fn<() => Promise<boolean>>(),
+  isCourseInMaintenance: vi.fn<() => Promise<boolean>>(),
+  isPlatformFrozen: vi.fn<() => Promise<boolean>>(),
+  completeLesson: vi.fn<() => Promise<string>>(),
+  fetchEnrollment: vi.fn<() => Promise<unknown>>(),
+  fetchCourse: vi.fn<() => Promise<unknown>>(),
+  getLessonSlot: vi.fn<() => number>(),
+}));
+
+vi.mock("@/lib/rate-limit", () => ({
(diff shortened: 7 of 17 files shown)


thomgabriel:
## GATE: blocked — one call site left in array space, on the webhook write path

The fix itself is right and well-built. `getLessonSlot` is genuinely fail-closed (throws, no array fallback — I checked for a silent `?? findLessonIndex` and there isn't one), the count guard correctly becomes a per-slot bit test against `activeLessons`, and catching `admin/resync` was a good find — I missed that one when I filed #741.

But the sweep isn't complete, and the miss is on a write path.

## `lib/helius/resolvers.ts:97`

```ts
const allLessons = (course.modules ?? []).flatMap((m) => m.lessons ?? []);
const lessonId = allLessons[lessonIndex]?._id ?? null;   // array position
```

`resolveLessonId(courseId, lessonIndex)` is called from `handleLessonCompleted` ([event-handlers.ts:161](apps/web/src/lib/helius/event-handlers.ts:161)) with `event.lessonIndex` — the index the program emitted, i.e. **the slot**. The resolved `lessonId` goes straight into a `user_progress` upsert at :167.

The disposition table calls itself a "full sweep of `findLessonIndex` / `deriveLessonIndex` / array-position bitmap." This site doesn't call `findLessonIndex` — it re-implements the same `flatMap` + index inline, so a grep on the helper name passes right over it. Same defect, different spelling.

## Why this is blocking rather than a follow-up

This PR makes the two paths **disagree** for any sparse course. After it merges, `/api/lessons/complete` sends the slot, so the program emits the slot — and the webhook then interprets that slot as an array position. On C3 post-restructure, slot 15 (capstone) resolves to `allLessons[15]`, which is out of range on a 15-lesson course → `null` → the `if (lessonId)` guard at :166 skips the upsert entirely. A learner completes the capstone on-chain and `user_progress` never records it. Slot 16/17/18 likewise. Lower slots resolve to the *wrong* lesson and record a completion the learner never made.

The webhook is the primary on-chain→DB sync path, so this reintroduces exactly the corruption #741 exists to prevent — which means re-landing #740 on top of this PR is still unsafe. That's the one thing this PR is a prerequisite for.

Latent today for the same reason as everything else: every deployed course is dense, so `slot == arrayIndex`. Your own `lesson-slot-equivalence.test.ts` is the proof of that — and the reason this hasn't already broken.

## Fix

`slotToLiveLessonId` is already in this PR and is exactly the right tool — it's what `admin/resync` now uses for the identical "bitmap index → lessonId" direction:

```ts
const lessonId = slotToLiveLessonId(courseId).get(lessonIndex) ?? null;
```

A retired slot's set bit then has no live lesson and is correctly ignored, matching the resync semantics. Worth a test asserting the webhook resolves slot 15 to the capstone on a C3-shaped fixture whose array order disagrees with its slots — the same fixture `lesson-slot.test.ts` already builds.

Also worth deciding deliberately: `resolveLessonId` currently returns `null` on a miss and the caller silently skips the upsert. For a genuine slot that should have resolved, silence is the wrong outcome — that's a dropped completion with no signal. A structured log there would fit what #743 just established.

## Everything else checks out

- No non-test `findLessonIndex` callers remain (verified independently — only its own definition and a docstring reference).
- The other `allLessons[...]` hit is `lesson-client.tsx` prev/next navigation, which is display-only and correct.
- Backfill call is right: slot-space for new writes, historical rows left alone, self-correcting on next completion/resync.
- The landmined all-dense assertion is a good idea — it forces a conscious decision at the moment a restructure lands rather than silently weakening the equivalence proof.

Re-request when the resolver is slot-aware; nothing else needs changing.

thomgabriel:
**Orchestrator: adversarial review FAIL (2 blockers) — fix round dispatched, do not merge.**

The three fixes in this PR verified correct (route slot-derivation + guard mirroring the program's is_active_slot exactly, resync reverse-map, honest all-lessons equivalence test, lesson_index proven write-only). But the central claim — every learner-facing on-chain path slot-aware — is false:

**BLOCKER-1:** the Helius LessonCompleted webhook maps the on-chain event index back to a lessonId by ARRAY POSITION (`resolvers.ts:94-97` `allLessons[lessonIndex]`, called from `event-handlers.ts:161`). The event index IS the slot; on a sparse course this upserts a PHANTOM completed row for the WRONG lesson (new (user_id, lesson_id) key — not a correction) + mis-attributed XP reason, corrupting progress/review-spine/attribution — the webhook mirrors EVERY completion. The identical reverse-map this PR fixed in resync. Fix: `slotToLiveLessonId(courseId).get(lessonIndex)`.

**BLOCKER-2:** `batch-complete.ts:140-155` (test-out) drives `completeLesson` by flatMap loop index — array position on-chain. One correction to the reviewer's scoping note: **#730 IS merged — batch-complete is live on main** (verified `git cat-file origin/main`), so this is a live main bug in this PR's scope, not a stacked-PR sibling.

Both fixes + disposition-table completion + sparse webhook/batch tests dispatched to the implementer on this branch. Ship-safe-today argument still holds (dense bundle ⇒ both missed paths coincidentally correct) — but the C3 re-land gate is BOTH blockers closed.

thomgabriel:
@claude Re-review please — round 2 addresses the adversarial FAIL (two missed on-chain index consumers). Head `0e93d1b`.

Both now slot-aware on the same branch:
1. **Helius webhook** `resolveLessonId` (`lib/helius/resolvers.ts`) — was `allLessons[eventIndex]` (array pos); the event index IS the slot. Now `slotToLiveLessonId(courseId).get(index)`; retired/unsynced slot → null → handler skips gracefully (no throw → no webhook retry loop).
2. **Test-out batch** (`lib/lessons/batch-complete.ts`, #730) — loop index → `completeLesson`/bitmap/`lesson_index` all → `getLessonSlot` (fail-closed → `pending`, batch continues); count bound `min(len,popcount)` → per-slot `is_active_slot` bit test.

Re-grep evidence for any other consumer is in the PR body — remaining `allLessons[]`/`lessonIndex` uses are display nav (lessonId-keyed) and the instruction sink. Please focus on: no exit path in either consumer falls back to array index; webhook null-path stays graceful; batch accounting (newly+already+pending==total) holds. Full suite: 1890 passed / 208 files; typecheck 6/6.

thomgabriel:
**Orchestrator routing record — SENSITIVE (on-chain trust path) → adversarial FAIL → fix round → re-verify PASS → queuing for gate.**

**My independent verification** at `0e93d1b2`: typecheck exit 0; full web suite **208 files / 1890 tests, exit 0** (and 207/1885 at the round-1 head).

**Round 1 (`df9c5572`): FAIL** — route/resync/guard fixes verified correct, but two on-chain index consumers were missed: the Helius webhook's `resolveLessonId` (array-position reverse map ⇒ phantom completed rows for the WRONG lesson on any sparse course, mirrored on every completion) and `batch-complete` (test-out driving `completeLesson` by flatMap index ⇒ wrong on-chain bit).

**Round 2 (`0e93d1b2`): re-verify PASS by the same reviewer, per-attack:**
- Original phantom-row repro replayed — dead. Null path traced independently: skip-upsert + fallback XP reason + console.warn, never throws (no webhook retry-loop). The bundle-older-than-chain drift case chased to ground: the route's own fail-closed slot lookup means an unmappable slot can't be minted normally; out-of-band completions keep their XP (tx-idempotent, source classification unaffected) with only the DB mirror lagging until the slot-aware resync — logged and recoverable.
- Batch: all three sites slot-space, fail-closed isolation to `pending`; the count bound replaced with per-slot `is_active_slot` tests; accounting proven exhaustive (every lesson lands in exactly one bucket; finalize's subset semantics unaffected by not-yet-activated slots).
- Independent consumer re-grep at head: ZERO array-position-against-the-bitmap paths remain; `findLessonIndex` has no callers.
- claude[bot] re-review at this head: 'both blockers closed, no new issues.'

**For the gate:** this is THE prerequisite for re-landing the reverted C3 bump (#740/#744). Ship-safe today (dense bundle ⇒ byte-identical behavior, pinned by the all-lessons equivalence test whose one landmine sub-test flips on the re-land). After this merges I'll re-open the C3 bump with the sparse bundle.

thomgabriel:
## GATE: approved — both blockers closed, and you found a fourth site I hadn't

The Helius resolver is fixed the right way: `slotToLiveLessonId(courseId).get(lessonIndex)`, wrapped so it **never throws** — correct instinct, since a throwing webhook handler retries into a loop — and it now warns on a miss instead of dropping the completion in silence.

**`lib/lessons/batch-complete.ts` was a fourth call site that neither my sweep nor the original disposition table caught.** The test-out path had both halves of the same bug: `isLessonComplete(flags, index)` on array position, and a `Math.min(lessons.length, liveLessonCount)` count truncation that would have attempted array positions `0..popcount-1` and skipped a live capstone at slot 15. Now per-lesson slot resolution, per-slot bit test, `completeLesson(…, slot)`, `lesson_index: slot`. That one mattered — test-out mints retroactive XP in a batch, so a wrong slot there is wrong XP plus a wrong bit, at volume.

The fail-closed shape in that loop is right too: an unslotted lesson is isolated into `pending` with a logged reason and the batch continues, with no index fallback in the catch path. And `activeLessons[word] ?? 0n` defaults a missing word to not-live, so it fails closed rather than open.

### Checks I ran rather than took on faith

- **`total = lessons.length`** replacing the count bound: `lessons` is `course.modules.flatMap(m => m.lessons)`, i.e. the bundle's live lessons, so the XP-cap basis can only move *up* relative to the old `min(...)`. A cap that refuses more is the safe direction.
- **The backfill question** from round 1 — whether anything reads stale array-space `lesson_index`. It doesn't: `lesson_index` appears in four write sites, the generated types, the IDL, and docstrings. **No reader anywhere** — no display, no filter, no join. Historical rows are inert, so leaving them is correct rather than merely tolerable.
- **Remaining `allLessons[...]`**: only `lesson-client.tsx` prev/next nav (lessonId-keyed display) and a docstring. No runtime array-position bitmap consumer left.
- **`resolvers.test.ts`** asserts the real repro — slot 15 → capstone, retired slot 14 → null, thrown lookup → null.

### What this unblocks

#740 can now be re-landed unchanged. The chain from here is: re-land #740 → C3 reaches trackLevel 3 → the C3/C2 level-2 collision clears → C2 creation becomes safe on-chain (see my note on #606 for why that ordering is load-bearing and not cosmetic).

Merging. Nice work running the sweep wider than the blocker I named — the fourth site is the one that would have hurt most.

PR #752, opened 2026-07-26
feat(courses): changelog for reactivate / deactivate / recreate mutators (#738)

### LINKED ISSUE #738: content: course changelog misses two on-chain mutators (reactivate, recreate) — including the is_active flips #713 retires with
#732 (merged) captures changelog entries in `api/admin/courses/sync`, on the stated basis that it is *"the sole mutator of a course's on-chain state"*. It is not. Two other paths mutate on-chain course state and write no entry.

## The uncovered mutators

| path | what it does | changelog |
|---|---|---|
| `api/admin/courses/reactivate/route.ts:44` | `updateCoursePda({ courseId, newIsActive: true })` then `writeCourseActive(courseId, true)` | **none** |
| `lib/admin/recreate-course.ts` (via `api/admin/courses/recreate`) | `closeCoursePda` + re-create — the WS-2 close-and-recreate | **none** |

(`api/lessons/complete` looks like a match on a naive grep but is not one — line 454 is a comment about `update_course.new_lesson_count`.)

## Why it matters more than "one more call site"

The feature's premise is that **post-deployment mutation is no longer silent**. The two loudest mutations available are the silent ones:

- A **recreate** closes the on-chain account and creates a new one. It is the most dramatic thing that can happen to a course, and a learner reading the changelog would see nothing.
- **`is_active` flips are how #713 retires the five superseded courses.** `reactivate` is that path's mirror image, so the retirement of `solana-fundamentals`, `rust-for-solana`, `anchor-framework`, `solana-frontend` and `defi-on-solana` would produce no changelog entries at all — on exactly the courses whose learners most need to know something changed.

There is a mild irony worth noting: #732's own RLS policy is scoped to `status='synced' AND COALESCE(is_active, true)`, so a deactivated course's changelog stops being readable. Combined with the gap, deactivating a course both hides its history **and** never records the deactivation.

## The design is right — extend it, don't replace it

Capturing at mutation time is the better choice and should stay. Replaying Helius `CourseUpdated` events cannot reconstruct intent: the sync route holds the prior on-chain state (`fetchCourse`) and the intended new state in the same call, which is what makes a meaningful diff possible.

The fix is that **both other callers hold the same information and simply do not record it**:

1. `reactivate` knows the course id and that it is flipping `is_active` false→true. That is a one-line entry.
2. `recreate-course.ts` computes the full before/after (it already reasons carefully about what `create_course` resets vs preserves — see its header) and is the richest diff available.

Suggest a shared `recordCourseChange` helper called from all three, rather than three inline writers — otherwise the fourth mutator added later repeats this.

## Also worth deciding

Should a **retire** entry be written *before* `is_active` flips false? Under the current RLS the entry becomes unreadable the moment the flip lands, so a retirement would be recorded and immediately invisible. If the intent is that a learner enrolled in a retired course can still see why it went away, the policy needs a carve-out for that case — or retirement entries need a different surface.

Related: #732 (the feature), #713 (retirement via `is_active`), the WS-2 recreate work.

thomgabriel:
Closes #738.

## What

#654 captured changelog entries only in `api/admin/courses/sync`, on the premise that it was "the sole mutator of a course's on-chain state". It was not. This PR extends the changelog to **every** on-chain course mutator, so post-deployment mutation is genuinely no longer silent.

## Mutator enumeration (every caller of the course-mutating signer fns)

| path | signer call | changelog |
|---|---|---|
| `api/admin/courses/sync` (deploy) | `deployCoursePda` | `deployed` (already, #654) |
| `api/admin/courses/sync` (update) | `updateCoursePda` | `lessons_added` / `lessons_removed` / `xp_changed` / `content_updated` (already, #654) |
| `api/admin/courses/deactivate` | `deactivateCoursePda` | **`deactivated` (NEW)** |
| `api/admin/courses/reactivate` | `updateCoursePda({newIsActive:true})` | **`reactivated` (NEW)** |
| `lib/admin/recreate-course` (via `api/admin/courses/recreate`) | `closeCoursePda` + `deployCoursePda` | **`recreated` (NEW)** |
| `sync` / `recreate` collection bind | `setCourseCollectionPda` | none — **by design** (Metaplex collection pointer plumbing, not a learner-facing course change) |
| `recreate` internal `restoreAfterCreate` | `updateCoursePda({newIsActive:false})` | none — **by design** (restores pre-close state as a sub-step of the recreate, already captured by the `recreated` entry) |

`api/lessons/complete` is not a mutator (the grep hit is a comment about `update_course.new_lesson_count`).

## Design — extends #654, does not replace it

- **Shared `recordCourseChange` seam.** All recorders delegate to one low-level writer, so the next mutator added has one obvious call site rather than a fresh inline supabase write (the exact repeat the issue warns about).
- **Three new kinds** via migration `20260726220000_course_changelog_status_kinds.sql`: an idempotent CHECK-constraint swap (DROP IF EXISTS + ADD, BEGIN/COMMIT, tested rollback), the `schema.sql` inline CHECK now lists all eight, plus supabase types + a guard test.
- **Non-fatal service_role write-back** in each mutator, matching #654. Status flips don't bump `version`, so the routes read the current on-chain version and skip the entry if it can't be read (never log a wrong version). Recreate records `version = 1` (create resets it) with the H3-preserved `lessonCount`, keyed on the create signature, recorded **last** — after the course is up, restored, and the maintenance gate is cleared.
- **i18n** en/es/pt-BR + component rendering (icons/title/body) for all three new kinds.

## RLS / surfacing decision (the issue's "Also worth deciding")

The #654 fail-closed policy (`status='synced' AND COALESCE(is_active,true)`) is **unchanged**. Consequence: a `deactivated` entry is written but not publicly readable while the course is inactive — it becomes visible again if the course is reactivated (the deactivate→reactivate history then shows together). `reactivated` and `recreated` land on synced+active courses, so they are immediately visible (the high-value cases).

Surfacing a **retired** course's log to enrolled learners is **intentionally deferred**: it requires un-gating the deactivated-course *page* (which 404s today via the catalog `isSynced` gate), not just the RLS. An RLS carve-out alone would be dead code — there is no surface to render it on. That un-gating is #713 / visibility-policy scope. The write side is complete regardless (full audit trail; visible on reactivation).

## Verification

- typecheck: 6/6 packages
- **web suite: 205 files, 1882 tests pass**

New/updated tests: shared-seam + three recorders (incl. non-fatal), read-decode of the new kinds, component rendering, the new migration guard (CHECK swap + rollback + 8-kind schema mirror), and recreate wiring (records **last**, full call-ordering assertion).

Do not merge — for review.

### DIFF
--- a/apps/web/src/app/api/admin/courses/deactivate/route.ts
+++ b/apps/web/src/app/api/admin/courses/deactivate/route.ts
@@ -2,6 +2,7 @@ import "server-only";
 
 import { revalidateTag } from "next/cache";
 import { NextRequest, NextResponse } from "next/server";
+import { Connection } from "@solana/web3.js";
 import {
   requireAdminAuth,
   adminUnauthorizedResponse,
@@ -9,8 +10,12 @@ import {
 } from "@/lib/admin/auth";
 import { isPlatformFrozen } from "@/lib/platform/freeze";
 import { platformFrozenResponse } from "@/lib/platform/freeze-http";
+import { serverEnv } from "@/lib/env.server";
 import { deactivateCoursePda } from "@/lib/solana/admin-signer";
 import { writeCourseActive } from "@/lib/content/deployment-writes";
+import { recordCourseDeactivated } from "@/lib/content/changelog-writes";
+import { fetchCourse } from "@/lib/solana/academy-reads";
+import { getProgramId } from "@/lib/solana/pda";
 import { COURSES_CACHE_TAG } from "@/lib/content/queries";
 
 export async function POST(req: NextRequest): Promise<NextResponse> {
@@ -64,6 +69,31 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
     );
     warning =
       "Deactivated on-chain, but the catalog flag didn't update — the course may still show until re-synced.";
+  }
+
+  // Course changelog (#738): record the deactivation — the path #713 retires the
+  // superseded courses with. Non-fatal. deactivate does not close the account or
+  // bump version, so the course is still readable on-chain; record its current
+  // version. NOTE: under the #654 RLS (synced+active only) this entry is not
+  // publicly readable while the course is inactive — it becomes visible if the
+  // course is later reactivated. See #738 on deferring the retired-course surface.
+  if (result.signature) {
+    try {
+      const connection = new Connection(serverEnv.SOLANA_RPC_URL, "confirmed");
+      const onChain = await fetchCourse(courseId, connection, getProgramId());
+      if (onChain) {
+        await recordCourseDeactivated({
+          courseId,
+          txSignature: result.signature,
+          version: onChain.version,
+        });
+      }
+    } catch (changelogErr) {
+      console.error(
+        "[admin/courses/deactivate] changelog write failed:",
+        changelogErr
+      );
+    }
   }
 
   return NextResponse.json({
--- a/apps/web/src/app/api/admin/courses/reactivate/route.ts
+++ b/apps/web/src/app/api/admin/courses/reactivate/route.ts
@@ -2,6 +2,7 @@ import "server-only";
 
 import { revalidateTag } from "next/cache";
 import { NextRequest, NextResponse } from "next/server";
+import { Connection } from "@solana/web3.js";
 import {
   requireAdminAuth,
   adminUnauthorizedResponse,
@@ -9,8 +10,12 @@ import {
 } from "@/lib/admin/auth";
 import { isPlatformFrozen } from "@/lib/platform/freeze";
 import { platformFrozenResponse } from "@/lib/platform/freeze-http";
+import { serverEnv } from "@/lib/env.server";
 import { updateCoursePda } from "@/lib/solana/admin-signer";
 import { writeCourseActive } from "@/lib/content/deployment-writes";
+import { recordCourseReactivated } from "@/lib/content/changelog-writes";
+import { fetchCourse } from "@/lib/solana/academy-reads";
+import { getProgramId } from "@/lib/solana/pda";
 import { COURSES_CACHE_TAG } from "@/lib/content/queries";
 
 export async function POST(req: NextRequest): Promise<NextResponse> {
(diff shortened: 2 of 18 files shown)


thomgabriel:
## GATE: approved — and the migration is already applied to prod

Independently enumerated every caller of `deployCoursePda` / `updateCoursePda` / `deactivateCoursePda` / `closeCoursePda` / `setCourseCollectionPda`. My sweep surfaced two files your table routes differently, and both check out: `api/admin/courses/recreate/route.ts` delegates to `recreateCourse` (the grep hits are an import and a comment), and `recreate-preflight.ts` names `closeCoursePda` only in a docstring promising it never calls it. Enumeration is complete — which matters, because "sole mutator" is exactly the claim #654 got wrong and #738 exists to correct.

## 🔴 The migration could not have applied on its own — fixed before merging

`course_changelog` **did not exist in production.** #732's `20260726210000_course_changelog.sql` was never applied, so this PR's `ALTER TABLE public.course_changelog` would have failed outright with `relation does not exist` — not a widened constraint, an error.

Worse, it means **#732's changelog has been dead in prod since it merged**: the writes are non-fatal by design, so every entry has been silently discarded into a missing table with no signal. Another instance of #699/#708/#731 — the feature is merged, the schema snapshot describes it, production never had it.

Applied both in order, before merging this:

1. `20260726210000_course_changelog` — table + index + RLS + the synced-active SELECT policy (correctly scoped through the `public_onchain_deployments` **view**; a base-table subquery in that policy would evaluate false for everyone).
2. `20260726220000_course_changelog_status_kinds` — the CHECK widen.

Verified live: table present, RLS on, 1 policy, and the constraint now lists all 8 kinds.

Ordering matters here beyond tidiness — merge auto-deploys (#731), and code writing `deactivated` against the 5-kind CHECK would have been a constraint violation on a real admin action.

## Rest of the review

- RLS unchanged: no write policy, service_role only, fail-closed `synced+active` SELECT.
- The deferred decision is the right one — surfacing a retired course's log needs the *page* un-gated, not just RLS, so an RLS carve-out alone would be dead code. Correctly left to #713.
- `version: 1` on recreate is a verified on-chain invariant (`write_fresh_course` always sets version 1), not an assumption.
- Changelog write is step 8, after the maintenance gate clears, and is skipped when no `createSignature` lands — tested both ways.

CI green (the three `fail` rows are all run `30225932534`, conclusion `cancelled`; the live push run passed).

Nit from the reviewer worth picking up whenever this file is next touched: `recordCourseRecreated`'s `detail as unknown as Json` double-cast is probably unnecessary — `{ lessonCount: number }` should be directly assignable.

thomgabriel:
**Orchestrator: retroactive adversarial record (the merge beat the round-trip — posting for the audit trail).**

Adversarial review (same reviewer as #732): **PASS.** Verified independently: mutator enumeration exhaustive (all 5 signer fns + every caller grepped; the two NONE-by-design calls proven unable to change learner-facing state — restoreAfterCreate is a sub-step of 'recreated', collection re-bind only ever re-binds the SAME collection); migration CHECK-swap correct (the constraint name matches Postgres's auto-name so the DROP actually hits, ACCESS EXCLUSIVE makes the swap atomic — no unchecked-write window), rollback restores the 5-kind CHECK, guard mutation-tested; recreate ordering correct and keyed on the stable content id (not the changing PDA); deactivated-entry invisibility live-evaluated (no RLS snapshot staleness), no alternate read path leaks; suite reproduced 205/1882. Also confirmed this PR incorporated both #732 MINORs (fail-closed policy + NOT NULL tx_signature).

**One MINOR, fixed post-merge in #754:** the deactivate/reactivate version-unreadable skip was silent — now warn-logged per the #731 honesty standard.

PR #754, opened 2026-07-26
fix(courses): log the changelog version-unreadable skip (#752 review MINOR)

thomgabriel:
Post-merge follow-up to #752 (the gate merged before the adversarial round-trip completed — findings land here per the branch-lifecycle rule).

**Adversarial review of #752 was PASS** with one MINOR: in the deactivate/reactivate routes, when `fetchCourse` returns null after a landed mutation, the changelog entry is skipped to avoid recording a wrong version (correct call) — but the skip itself was SILENT: no changelog row AND no trace. Per the #731 honesty standard (a silent miss is an invisible audit gap), the `else` now emits a `console.warn` naming the course, the skipped kind, and the tx signature. `recreate-course` already logged its failures loudly — these two routes now match.

Two-line-per-route change, no behavior delta on the happy path. Verified: typecheck exit 0, admin route tests 51/51.

Full #752 adversarial record (enumeration exhaustiveness verified independently, migration CHECK-swap proven atomic under ACCESS EXCLUSIVE, RLS invisibility semantics confirmed live-evaluated, guard mutation-tested) is on the #752 thread.

### DIFF
--- a/apps/web/src/app/api/admin/courses/deactivate/route.ts
+++ b/apps/web/src/app/api/admin/courses/deactivate/route.ts
@@ -87,6 +87,12 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
           txSignature: result.signature,
           version: onChain.version,
         });
+      } else {
+        // #731 honesty standard: skipping (to avoid logging a wrong version)
+        // must itself be visible — a silent miss is an invisible audit gap.
+        console.warn(
+          `[course-changelog] ${courseId}: 'deactivated' entry SKIPPED — on-chain course unreadable after mutation (tx ${result.signature})`
+        );
       }
     } catch (changelogErr) {
       console.error(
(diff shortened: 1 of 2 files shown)


thomgabriel:
## GATE: approved

Right fix, right severity. `warn` not `error` is correct — the mutation succeeded, only the audit entry is missing, and it matches the existing convention in this file family where `error` is reserved for actual write failures.

Verified the one thing that could bite: `result.signature` is `signature?: string` on the signer return type, but both new `console.warn`s sit inside the pre-existing `if (result.signature)` guard, so the template literal can't render `undefined`. No happy-path delta — the `else` was previously empty.

Worth flagging that this closes a *third* variant of the same silent-skip pattern from tonight: #743 made `/api/deploy/save`'s swallowed `catch {}` visible, #752's routes skipped the changelog entry silently, and #741's webhook resolver drops a completion silently on a `null` lesson id. Same shape each time — a deliberate, correct decision not to act, with no trace that the decision was made. The correctness was never the problem; the silence was.

One nit for whenever these routes are next touched: #743 introduced `logEvent` for exactly this (structured, greppable `[event] {context}`). These two use raw `console.warn`, which matches local convention but not the newer helper. Not worth a change on its own.

CI green.

PR #755, opened 2026-07-27
content: re-land the C3 transform bump — now safe on slot-aware completion (#741)

thomgabriel:
Re-lands #740, which I reverted in #744. The blocker it was reverted for — **#741** — is fixed and merged (#751), so the transform is now safe.

## Why it was reverted, and what changed

#740's premise was *"slot-keyed ids are stable ⇒ bitmaps untouched."* The ids were stable; the bitmap simply wasn't slot-keyed. The on-chain index came from the lesson's position in the flattened module array, and the transform moved 9 of 12 survivors — `m4-capstone` 15→14, straight onto retired `m4-interact`'s already-set bit in prod.

#751 closed that across **four** call sites: the completion route, `admin/resync`, the Helius webhook resolver, and `lib/lessons/batch-complete` (test-out). Every on-chain path now resolves `lessonId → slot` from the committed lock and fails closed on a miss. That is exactly the prerequisite this bump was waiting on.

## The one test that had to change

#751 planted a deliberate landmine: an assertion that the whole bundle is dense, with a comment saying to relax it when a restructure lands. This is that moment — `course-building-first-program` is now legitimately sparse (retired `[0,2,11,14]`, added 16-18).

I narrowed it rather than deleting it. It now pins the sparse set to an `EXPECTED_SPARSE` allowlist, so it still fires on an *unexpected* sparse course — an unreviewed restructure, or a hand-edited `slots.lock`. The `dense ⟹ equivalence` test above it is untouched and still passes.

## ⚠️ Known interim state until the on-chain mask is updated

Content now declares 15 live lessons at slots `{1,3,4,5,6,7,8,9,10,12,13,15,16,17,18}`. The Course PDA still carries the **dense `0..15`** mask. Until an `update_course` re-syncs it:

- **Existing lessons work normally** — their slots are live in both spaces.
- **The three new lessons (slots 16-18) return a clean 409** *"pending an admin re-sync"* from the per-slot bit test. Fail-closed and correct — no wrong-bit write, which is precisely what #751 bought.
- **Course completion is unreachable** — `finalize_course` requires `flags & active == active`, and the on-chain mask still demands bits `0,2,11,14`, whose lessons no longer exist in content.

This degradation is safe and self-clearing; it ends when the mask update lands (`update_course.new_active_lessons`, derived from the committed `slots.lock` — non-destructive, no recreate). I'm running that immediately after this merges, and current exposure is 1 enrollment / 2 learners / 0 certificates on this course.

**No recreate is needed for the trackLevel fix.** `trackLevel` is read from the content bundle for display and path ordering (`paths-explorer`, `courses-client`), not from chain — the on-chain value only feeds admin sync-diff comparisons. So this bump alone clears the C3/C2 both-at-level-2 collision documented on #606.

## Verification

`pnpm --filter web test` → **209 files / 1902 tests, all passing** against the re-landed bundle, including #721's capstone-gate assertions, the golden fixtures, and #751's slot-aware suites.

### DIFF
--- a/apps/web/content.lock
+++ b/apps/web/content.lock
@@ -1,4 +1,4 @@
 {
   "repo": "solanabr/courses-academy",
-  "sha": "1b74e4a60f8813374219451b60436af832df2d91"
+  "sha": "7a497475e14f3f53e5353cd752476eb0b2120096"
 }
--- a/apps/web/src/content/generated/courses.json
+++ b/apps/web/src/content/generated/courses.json
@@ -133,7 +133,7 @@
     "_type": "course",
     "creator": "B7o8NfV81HzjuZFWQTTx3Xdvh77Dqoajwib3kWEnvzJF",
     "creatorRewardXp": 30,
-    "description": "Take your Anchor knowledge from theory to practice. Write, compile, and deploy real Solana programs using the Superteam Academy Build Server — no local toolchain setup required. By the end, you'll have a working counter program deployed to Devnet.",
+    "description": "You arrive holding `vault_core.rs` — a `VaultState` and checked `deposit`/`withdraw` that compile, with no `#[program]` and no `#[derive(Accounts)]`. This course adds exactly that. You wrap your own Rust core in Anchor 1.x, create a per-user vault PDA with the space arithmetic derived rather than copied, and move real lamports: deposit through a System Program CPI, withdraw by debiting the program-owned vault directly above an enforced rent floor — and you learn why signer seeds are not the answer there. Then you write the account constraints that stop three named attacks, read the LiteSVM test that would catch what compiling cannot, deploy from a browser tab, publish the IDL on-chain, and call your own live program. Nothing is installed locally. You finish with a devnet program id, a fetchable IDL, and a build write-up.",
     "difficulty": "intermediate",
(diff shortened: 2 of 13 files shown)


thomgabriel:
Closing in favor of **#756** — we both opened a re-land 21 seconds apart when #751 merged.

Theirs is the better PR, for a specific reason: my version carried #740's original golden-test comment forward verbatim, including *"slot-keyed progress bitmaps for the 38 live enrollments are untouched by construction."* That sentence is the **discredited premise that caused the incident** — the bitmaps were not slot-keyed, which is the whole reason #744 had to revert. And the 38 is a devnet on-chain figure; prod has 1 enrollment on that course. #756 replaces it with an accurate note that names the sparse map and records the revert→#751→re-land chain.

The reviewer caught that on my PR, which I'd have had to fix anyway — and it is exactly the fabricated-specificity problem I blocked #742 for a few hours ago, this time in my own diff.

Our equivalence-test relaxations are equivalent in intent (I named an `EXPECTED_SPARSE` const, they inline the set); no reason to prefer mine.

Carrying the one thing mine documented that theirs doesn't — the interim on-chain state — over to #756 as a gate note.

PR #756, opened 2026-07-27
content: re-land content.lock bump to 7a49747 — C3 live on the slot-aware path

thomgabriel:
## What

Re-lands the reverted #740: pins `courses-academy@7a49747` (C3 transform of the LIVE flagship course, 16 → 15 lessons, SPARSE slot map: retired 0/2/11/14 burned, new 16-18, next=19).

## Why it's safe NOW (and wasn't at #740)

#740's premise — 'ids stable ⇒ bitmaps untouched' — was wrong: the runtime derived bitmap indices from ARRAY POSITION (`findLessonIndex`), never from slots.lock. Your block (#741) + revert (#744) were correct. Since then, **#751 (merged, `d6e58331`) made every on-chain index path slot-aware**: completion route (slot to the tx, slot-bit idempotency, per-slot `is_active_slot` guard), the Helius webhook reverse map, admin resync reconstruction, and test-out batch — adversarially FAILed once for the two missed consumers, fixed, re-verified PASS with an independent consumer grep showing ZERO array-position-against-the-bitmap paths remaining.

So on this bundle: `your-first-build` completes as slot 1 (not array pos 0), the capstone as slot 15 (not 14), the #741 learner's stale retired-slot-14 bit can never false-complete the capstone, and the webhook resolves event indices through the slot lock.

## Changes vs the reverted #740

Identical content pin + golden regeneration (courses/lessons/quests-raw-derived/paths/summaries through the real projectors), PLUS the documented relaxation of #751's landmine test — it now pins the EXACT sparse set (`course-building-first-program` only), so any future course going sparse is still a deliberate reviewed event.

`compile-content`: courses 8 / lessons 98; capstone id + deploy-panel assertions (#721) pass unchanged; **full suite 209 files / 1902 tests, exit 0** — including #751's sparse-fixture tests, the capstone-gate bundle assertions, and test-out path-eligibility against the sparse bundle.

## Routing

Deliberately `needs-gate` despite being area:content — you blocked the last attempt; the re-land decision is yours.

### DIFF
--- a/apps/web/content.lock
+++ b/apps/web/content.lock
@@ -1,4 +1,4 @@
 {
   "repo": "solanabr/courses-academy",
-  "sha": "1b74e4a60f8813374219451b60436af832df2d91"
+  "sha": "7a497475e14f3f53e5353cd752476eb0b2120096"
 }
--- a/apps/web/src/content/generated/courses.json
+++ b/apps/web/src/content/generated/courses.json
@@ -133,7 +133,7 @@
     "_type": "course",
     "creator": "B7o8NfV81HzjuZFWQTTx3Xdvh77Dqoajwib3kWEnvzJF",
     "creatorRewardXp": 30,
-    "description": "Take your Anchor knowledge from theory to practice. Write, compile, and deploy real Solana programs using the Superteam Academy Build Server — no local toolchain setup required. By the end, you'll have a working counter program deployed to Devnet.",
+    "description": "You arrive holding `vault_core.rs` — a `VaultState` and checked `deposit`/`withdraw` that compile, with no `#[program]` and no `#[derive(Accounts)]`. This course adds exactly that. You wrap your own Rust core in Anchor 1.x, create a per-user vault PDA with the space arithmetic derived rather than copied, and move real lamports: deposit through a System Program CPI, withdraw by debiting the program-owned vault directly above an enforced rent floor — and you learn why signer seeds are not the answer there. Then you write the account constraints that stop three named attacks, read the LiteSVM test that would catch what compiling cannot, deploy from a browser tab, publish the IDL on-chain, and call your own live program. Nothing is installed locally. You finish with a devnet program id, a fetchable IDL, and a build write-up.",
     "difficulty": "intermediate",
     "duration": 12,
     "minCompletionsForReward": 10,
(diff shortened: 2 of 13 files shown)


thomgabriel:
## GATE: approved

I opened a duplicate re-land (#755) 21 seconds before this one and closed it in favour of this. Yours is better on the point that matters: my version carried #740's original golden-test comment forward, including *"slot-keyed progress bitmaps for the 38 live enrollments are untouched by construction"* — the discredited premise that caused the incident, plus a devnet figure (prod has **1** enrollment on this course). Yours names the sparse map and records the revert→#751→re-land chain instead. That comment is what the next restructure author will read.

The landmine relaxation is right: narrowed to the exact expected sparse set rather than deleted, so a future unreviewed restructure still trips it. `dense ⟹ equivalence` untouched.

### Verified independently

- **The premise of the re-land holds.** All four #741 call sites are slot-aware on `main`: `lessons/complete/route.ts`, `lib/lessons/batch-complete.ts`, `lib/helius/resolvers.ts`, `admin/resync/route.ts`.
- **Slots are correct in the bundle**: `m4-capstone` stays at **15** (the original bug moved it to 14), retired `[0,2,11,14]`, new `16/17/18`, `next: 19`, and `course-building-first-program` is the only course whose lock changed.
- I ran the full suite against this exact bundle on my branch before closing it: **209 files / 1902 tests green**, including #721's capstone-gate assertions and the golden fixtures. CI green here too.

## ⚠️ Interim on-chain state this creates — carrying this over from #755

Content now declares 15 live lessons at slots `{1,3,4,5,6,7,8,9,10,12,13,15,16,17,18}`. The Course PDA still carries the **dense `0..15`** mask. Until `update_course.new_active_lessons` re-syncs it:

- **Existing lessons work normally** — live in both spaces.
- **The three new lessons (16-18) return a clean 409** *"pending an admin re-sync"* from the per-slot bit test. Fail-closed — no wrong-bit write. This is exactly what #751 bought, and it is the correct behaviour, not a bug.
- **Course completion is unreachable** — `finalize_course` requires `flags & active == active`, and the mask still demands bits `0,2,11,14`, whose lessons no longer exist in content.

Safe and self-clearing. The mask update is non-destructive (`update_course`, no recreate) and derived from the committed `slots.lock`. Exposure while it stands: 1 enrollment, 2 learners, 0 certificates on this course. I'm doing the mask update immediately after this merges.

**No recreate is needed for the trackLevel fix** — `trackLevel` is read from the content bundle for display and path ordering (`paths-explorer`, `courses-client`); the on-chain value only feeds admin sync-diff. So this bump alone clears the C3/C2 both-at-level-2 collision I documented on #606.

PR #758, opened 2026-07-27
docs(migrations): self-contained streak-forgiveness rollback (#750)

### LINKED ISSUE #750: streak_forgiveness rollback block is not an inverse for prod — it assumes 20260726170000 was applied
Fifth instance of the #699/#708/#731/#737 theme: an authoritative record that is wrong about production. This one is a **rollback procedure**, which is the worst place for it.

## The claim

`supabase/migrations/20260726190000_streak_forgiveness.sql` ends with a block labelled "a tested ROLLBACK block". Steps 2 and 3 inline full restore bodies for `award_xp` and `award_community_xp`. Step 4 does not:

```
-- 4. Restore get_daily_quest_state: re-apply the migration that last defined it
--    before forgiveness — 20260726170000_review_quest_kind.sql (which carries the
--    full function body with the original Case-3 reset and no freeze grant).
```

## Why it does not hold here

`20260726170000_review_quest_kind` **was never applied to this database**. Verified against prod (`pywhtmidcrptomrabbrw`):

```sql
select position('review' in prosrc) > 0 as has_review_branch,
       position('streak_freezes' in prosrc) > 0 as has_freeze_logic
from pg_proc where proname = 'get_daily_quest_state';
-- has_review_branch = false, has_freeze_logic = false
```

Prod's live `get_daily_quest_state` predates both. So running the documented rollback would not restore the prior state — it would install the review-quest-kind version, which this database **has never run**. The rollback moves the function forward, not back.

It is "tested" against a database where 170000 had been applied. Ours isn't one.

## Second-order consequence

Because 190000 carries 170000's review branch, applying it is a **two-migration behavioral jump** on prod, not the single "add streak freezes" change the filename suggests. It simultaneously activates the review-quest branch in the daily-quest state machine. That may well be fine — but it should be a deliberate decision, not a side effect nobody priced in.

## Current prod state (for whoever applies this)

- Applied: the schema half — `user_xp.streak_freezes` column, `streak_freezes_used` table + RLS.
- Not applied: all six function bodies — `cover_missed_days_with_freezes`, `next_streak_value`, `grant_streak_freeze` do not exist; `award_xp`, `award_community_xp`, `get_daily_quest_state` are still the pre-forgiveness versions.
- This split is safe: prod runs the old streak behavior correctly. Forgiveness is a feature, not a fix — the streak-renders-0 regression was fixed separately in the frontend.

## What to fix

1. Replace step 4 with an inlined restore body for `get_daily_quest_state`, the way steps 2 and 3 do, so the rollback does not depend on which prior migrations a given database happens to have.
2. Before applying 190000 anywhere, snapshot `pg_get_functiondef` for all three functions — `CREATE OR REPLACE` is only reversible if the previous body was captured.
3. Decide explicitly whether activating the review-quest branch at the same time is intended.

thomgabriel:
Closes #750. Comment/doc-only fix to the **already-merged** `20260726190000_streak_forgiveness.sql` — makes its ROLLBACK block a real inverse regardless of a database's migration history, and stops it from silently blocking the owner applying the streak-forgiveness function bodies.

## The bug
Rollback step 4 said: restore `get_daily_quest_state` by re-applying `20260726170000_review_quest_kind.sql`. But **prod never ran 170000** (verified in #750: `position('review' in prosrc)=false`, `position('streak_freezes' in prosrc)=false`). Re-applying 170000 would move the function **forward** into a version prod never had — the opposite of a rollback. It was "tested" only against a DB where 170000 had been applied.

## The fix (SQL is comment-only; executable half untouched)

1. **Step 4 is now self-contained** — no cross-file re-apply. A probe the operator runs **before** applying:
   ```sql
   SELECT position('review' in prosrc) > 0 AS had_review_branch
   FROM pg_proc WHERE proname = 'get_daily_quest_state';
   ```
   forks between two **inlined, verbatim** prior bodies pulled from git history:
   - **4a** `had_review_branch = TRUE` → 170000's body (review branch, original login_streak Case-3 reset, no freeze grant), from `20260726170000_review_quest_kind.sql`.
   - **4b** `had_review_branch = FALSE` (current prod) → the body that preceded it (neither review nor freeze), from `20260709120000_quest_xp_durable_delivery.sql`.
   Preferred path stated first: restore the `pg_get_functiondef` snapshot captured before applying (exact by construction).

2. **Header warning added**: applying 190000 on a pre-170000 DB is a deliberate **two-migration jump** — it also activates the review-quest daily-quest branch — and `CREATE OR REPLACE` is only reversible if the prior body was snapshotted first. So the applier prices both in.

3. **Guard test updated** (`streak-forgiveness-guard.test.ts`): pins the new shape — the probe string, exactly **two** inlined `get_daily_quest_state` restore bodies, the header warning — and **forbids** the old `re-apply the migration` cross-reference.

## Executable-half-untouched proof (required invariant)
Editing an applied-elsewhere migration's rollback **comment** block is safe only if the executable upper half is byte-identical. Verified against `origin/main`:

```
BEGIN..COMMIT executable region: BYTE-IDENTICAL vs origin/main (615 lines)
non-comment changed lines vs origin/main: 0
```

Diff lands in exactly three comment regions (hunk headers):
```
@@ -37,6 +37,18 @@                                              # header warning
@@ -660,9 +672,12 @@ GRANT ... get_daily_quest_state ...      # rollback-header rewrite
@@ -781,10 +796,511 @@ COMMIT;                                  # step-4 inlined fork bodies
```
No `schema.sql` change (the rollback lives only in the migration). No live-DB action.

## Verify
- `pnpm --filter web test` — 209 files / 1903 tests pass (guard test 24, +1 new self-contained-step-4 assertion)

### DIFF
--- a/apps/web/src/lib/supabase/__tests__/streak-forgiveness-guard.test.ts
+++ b/apps/web/src/lib/supabase/__tests__/streak-forgiveness-guard.test.ts
@@ -192,8 +192,7 @@ describe("#573 streak forgiveness — migration-only guarantees", () => {
 
   it("ships a rollback that drops the new objects AND restores the writers", () => {
     // Function-modifying migration: the rollback must both drop what it created
-    // and restore the two rewritten writers' prior bodies (get_daily_quest_state
-    // is restored by re-applying its prior migration).
+    // and restore the rewritten writers' prior bodies.
     expect(migration).toContain(
       "DROP FUNCTION IF EXISTS public.grant_streak_freeze(UUID)"
     );
@@ -209,11 +208,28 @@ describe("#573 streak forgiveness — migration-only guarantees", () => {
     expect(migration).toContain(
       "ALTER TABLE public.user_xp DROP COLUMN IF EXISTS streak_freezes"
     );
-    // Restores the two rewritten XP writers, and points at the quest migration.
+    // Steps 2-3 restore the two rewritten XP writers verbatim.
     expect(migration).toMatch(/Restore award_xp to its pre-forgiveness body/);
     expect(migration).toMatch(
       /Restore award_community_xp to its pre-forgiveness body/
     );
-    expect(migration).toContain("20260726170000_review_quest_kind.sql");
+  });
+
+  it("step 4 is SELF-CONTAINED: inlines BOTH prior bodies with a probe fork (#750)", () => {
+    // get_daily_quest_state's correct prior body is DB-state-dependent (whether
+    // 20260726170000 had been applied). The rollback must NOT tell the operator
+    // to "re-apply" another migration — prod never ran 170000, so that would
+    // move it FORWARD, not back. Instead: a probe + two inlined fork bodies.
+    expect(migration).not.toMatch(/re-apply the migration/i);
+    // The probe the operator runs (before applying) to choose the fork.
+    expect(migration).toContain("position('review' in prosrc)");
+    // Both prior bodies are inlined (commented), one per fork branch (4a/4b).
+    const restoreDefs =
+      migration.match(
+        /^-- CREATE OR REPLACE FUNCTION get_daily_quest_state\(/gm
+      ) ?? [];
+    expect(restoreDefs).toHaveLength(2);
+    // The header prices in the two-migration jump on a pre-170000 DB.
+    expect(migration).toMatch(/TWO-MIGRATION JUMP ON A PRE-170000 DATABASE/);
   });
 });
--- a/supabase/migrations/20260726190000_streak_forgiveness.sql
+++ b/supabase/migrations/20260726190000_streak_forgiveness.sql
@@ -37,6 +37,18 @@
 -- apply. A tested ROLLBACK block is at the very bottom.
 -- Mirrored into supabase/schema.sql (the full-schema snapshot).
 --
+-- WARNING — TWO-MIGRATION JUMP ON A PRE-170000 DATABASE: this migration's
+-- get_daily_quest_state body is built on 20260726170000 (review-quest-kind),
+-- so applying it to a database that never ran 170000 (e.g. prod as of
+-- 2026-07-27, whose live body predates both review and freeze) ALSO activates
+-- the review-quest daily-quest branch — TWO behavioral changes, not just "add
+-- streak freezes". Whether that is intended is a deliberate call for the
+-- applier. Because of it, the ROLLBACK below is DB-STATE-DEPENDENT for
+-- get_daily_quest_state (see step 4), and CREATE OR REPLACE is only reversible
+-- if the prior body was captured: BEFORE applying anywhere, snapshot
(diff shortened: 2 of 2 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (area:db, and this file is the one you're about to apply) → queuing for gate.**

**My independent verification at `ba9d3f43`** (not the implementer's claim — recomputed): the BEGIN..COMMIT executable region is **byte-identical to origin/main** (615 lines, programmatic compare), and of 523 changed diff lines, **zero are non-comment** — the entire delta is the header warning + the rewritten self-contained rollback (probe fork + two inlined verbatim prior bodies, 4a=170000's, 4b=pre-170000/current-prod). Full suite 209 files / 1903 tests, exit 0; guard extended to pin the probe + both bodies and FORBID the old cross-file re-apply reference.

Given light adversarial coverage is warranted but the change is provably comment-only and you hold the prod context (#750 is your finding, and the two apply-time calls — pre-apply `pg_get_functiondef` snapshots, and whether the simultaneous review-quest-branch activation is intended — are yours), routing straight to you rather than through a hostile reviewer. This unblocks applying the streak-forgiveness function bodies.

thomgabriel:
## GATE: approved — 4b verified byte-identical to prod

The safety invariant holds: I extracted the `BEGIN..COMMIT` region from both `origin/main` and this head — **615 lines, byte-identical**, and 0 non-comment changed lines across the whole file. Editing an already-partially-applied migration is only safe on that condition, so I checked it rather than took it.

**The substantive fix is verifiably correct, not just plausible.** I pulled branch 4b's inlined function body out of the comment block and hashed it against the live function in prod:

```
4b inlined body   md5 78105ee5110c31012947c9cba1f45b7e  (9086 chars)
prod get_daily_quest_state  md5 78105ee5110c31012947c9cba1f45b7e  (9086 chars)
```

Byte-identical. So on this database the rollback now restores exactly what is running, which is precisely what #750 said it did not do. The probe (`position('review' in prosrc) > 0`) correctly discriminates the two histories, and 4a's body carries the review branch while 4b's carries neither review nor freeze — confirmed by grepping each extracted body, not by reading the labels.

Also right to state the snapshot path **first** and the fork as fallback: a `pg_get_functiondef` capture is exact by construction, whereas the inlined bodies are exact only for the two histories anticipated here.

The header warning is the other half of the value. Applying 190000 on a pre-170000 database is a **two-migration jump** that also activates the review-quest daily-quest branch — that was invisible before, and it is a product decision rather than a schema one.

### Status change this produces

`streak_forgiveness` was held for two reasons. One is now gone — a 790-line core-XP rewrite with no working undo. What remains is only the deliberate call in your own header warning: whether to activate the review-quest branch at the same time. That is the owner's, not the gate's, so I'm leaving it parked with the reason narrowed and recorded.

CI green.

PR #759, opened 2026-07-27
ci(review): re-run claude-review on push/label for needs-gate PRs (#710)

### LINKED ISSUE #710: process: claude[bot] review races the orchestrator's hardening commits (5 stale today; 3 missed the security-relevant commit)
**The automated review races the orchestrator's own hardening commits.** Five instances today, and in three of them the *unreviewed* commit was the security-relevant one. Every merge today was still verified — by me, by hand — but that means one pair of eyes where the process assumes two.

## Measured, not impressionistic

Comparing each `claude[bot]` review's `created_at` against its PR's head `authoredDate`:

| PR | staleness | what the review missed |
|---|---|---|
| **#709** | **10m14s** | `fix(review): due-gate the grade route (anti box speed-run)` — +34 route, +65 test lines. **The entire anti-speed-run backstop.** |
| **#695** | **15m49s** | the commit that added `segment-sync.test.tsx` **and modified the migration** (`daily_goal > 0` → `BETWEEN 1 AND 20`) |
| **#689** | **10m41s** | `refactor(review): atomic box advance` — the fix for the race the review itself had flagged |
| **#638** | 12h45m (first review) | all four substantive commits, incl. the `reviewExempt` trust boundary and 19b tier flip |
| **#707** | 2m12s | a migration rename + one test line — genuinely immaterial |

Ten of fifteen were current. This is not systemic breakage; it is a specific, repeating race.

## Why it happens

The review fires on push and takes 2–4 minutes. The orchestrator's loop pushes a follow-up commit — often *because* of its own adversarial pass — inside that window. The review then reports on a head that no longer exists, and **GitHub shows a green `claude-review` check either way**, because the check means "the action ran", not "the action reviewed this head".

The pathology is that staleness correlates with security relevance rather than being random: the follow-up commit is frequently the hardening. #709 is the cleanest example — the review verdict covers the version *without* the due-gate.

## Two failure modes it has already produced

1. **A finding that was already fixed** (#689). The review flagged a read-then-write race on `box`; the head had already folded the computation into the `UPDATE`. Acting on that note would have meant re-breaking working code. It also referenced a variable (`v_new_box`) that no longer existed.
2. **A gap reported as open that was closed** (#695). The review asked for `segment-sync.tsx` tests; the unreviewed commit added them.

Both are *false negatives about the current code* — the more expensive direction, because they invite work rather than hiding risk.

## Fixes, cheapest first

1. **Orchestrator: re-request review after any post-review push.** A `@claude` comment re-triggers it (that is how #638 got a fresh review). One line in the loop's push step. Cheapest and closes most of it.
2. **Gate: keep comparing `created_at` vs head `authoredDate`.** Already doing this — it is how all five were caught — but it currently rests on one agent remembering. Worth writing into `docs/HANDOFF-LOOP-2026-07-25.md` as a required step rather than a habit.
3. **Consider: don't let the review start until the branch settles.** A short debounce, or trigger review on the `ready_for_review`/label event rather than every push, would stop the race at the source. Larger change; only worth it if 1 proves insufficient.

## What this is not

Not a reason to distrust the reviews — the ten current ones were accurate and two of them (#701's `satisfies Record<…>` observation, #707's advisory-check note) were sharper than my own read. The problem is purely *which commit* they describe.

Related: #709, #695, #689, #638, #707. Gate protocol lives in `docs/HANDOFF-LOOP-2026-07-25.md`.

thomgabriel:
Closes #710 (mechanical half).

## Problem

The `claude[bot]` review fires on open/ready/reopen and takes 2–4 min. The orchestrator pushes hardening commits inside that window, so the review reports on a head that no longer exists — **5 stale reviews in one night, 3 of them missing the security-relevant commit** (#710). GitHub still shows a green `claude-review` check, because the check means "the action ran", not "the action reviewed this head".

The behavioral half (re-request via `@claude` after a fix push) is already adopted per the orchestrator's ack on the issue. This PR is the **mechanical** half the maintainer offered there.

## Change (one file: `.github/workflows/claude-code-review.yml`)

1. **Trigger on `synchronize` + `labeled`**, in addition to `opened/ready_for_review/reopened`.
2. **Job-level `if:` scopes the two new events to `needs-gate` PRs only** — `synchronize` re-reviews only when the PR carries `needs-gate`; `labeled` fires only when the label added *is* `needs-gate`. Ordinary PR pushes still don't trigger a review, so the OAuth-token budget is unaffected. Open/ready/reopen behavior is unchanged for every PR.
3. **Per-PR `concurrency` group with `cancel-in-progress: true`.** This is the core: when a newer head arrives mid-review, the in-flight review of the stale head is cancelled and superseded, so a verdict for a superseded commit can never post. Keyed on `github.event.pull_request.number` so PRs don't cancel each other.

### Resulting flow for a gated PR
Queue it (`needs-gate` added) → `labeled` fires a fresh review at the current head. Push a hardening commit → `synchronize` cancels any in-flight review and reviews the new head. The review a human reads always describes the head they're about to merge.

## Why this is safe / minimally scoped

- **No cost storm:** the push re-review is gated to `needs-gate`; non-gated PRs are untouched.
- **Cancelled runs don't block merges:** `claude-review` is advisory, not a required status check. I verified the required checks on `main` are `Frontend (lint · typecheck · test)`, `On-chain (fmt · clippy · cargo test)`, and `SHA-pin guard` — `claude-review` is not among them.
- **No new/changed actions:** both `actions/checkout` and `anthropics/claude-code-action` remain SHA-pinned (the SHA-pin guard passes on this branch).

## Verification (what I actually ran; workflow behavior itself cannot be exercised locally — stated honestly)

- **`actionlint .github/workflows/claude-code-review.yml` → clean.** Validates the `on:` filter, the `if:` expression (including `contains(github.event.pull_request.labels.*.name, …)`), and the `concurrency` block.
- **`scripts/check-action-pins.sh` → green** (8 files) — all `uses:` still SHA-pinned.
- **`python3 yaml.safe_load` → parses**, and the folded `if:` resolves to a single clean expression line.
- Confirmed `needs-gate` label exists in the repo (`gh label list`).
- The runtime trigger/cancel semantics (does `synchronize` actually cancel the stale run, does the `labeled` filter admit only `needs-gate`) can only be observed on live PR events on `main` — cannot be simulated locally.

> SENSITIVE (`.github/workflows/**`) — do not merge without adversarial review + gate.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -2,18 +2,36 @@ name: Claude Code Review
 
 on:
   pull_request:
-    # Cost: review on open / ready / reopen — NOT on every `synchronize` (every push),
-    # which re-reviews a PR on each commit and burns the OAuth-token budget. Re-trigger a
-    # fresh review by commenting `@claude review` (claude.yml) or reopening.
-    types: [opened, ready_for_review, reopened]
+    # Every PR is reviewed on open / ready / reopen. Additionally — and ONLY for PRs
+    # carrying `needs-gate` — re-review on every push (`synchronize`) and when the
+    # label is first applied (`labeled`). This closes the #710 race: the orchestrator
+    # pushes hardening commits after the review has started, so the review reported on
+    # a head that no longer existed (and 3 of 5 stale reviews missed the security-
+    # relevant commit). Scoping the push re-review to needs-gate keeps the OAuth-token
+    # cost off every ordinary PR push; the `if:` below enforces the scope, and the
+    # concurrency group cancels an in-flight review the moment a newer head arrives so
+    # a stale verdict can never post.
+    types: [opened, ready_for_review, reopened, synchronize, labeled]
+
+# One review per PR at a time; a newer head cancels the older review in flight (the
+# mechanical core of the #710 fix — the pre-fix head's review is superseded, never
+# posted as if current). Keyed on PR number so PRs don't cancel each other.
+concurrency:
+  group: claude-code-review-pr-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
 
 jobs:
   claude-review:
-    # Optional: Filter by PR author
-    # if: |
-    #   github.event.pull_request.user.login == 'external-contributor' ||
-    #   github.event.pull_request.user.login == 'new-developer' ||
-    #   github.event.pull_request.author_association == 'FIRST_TIME_CONTRIBUTOR'
+    # opened/ready/reopened review any PR (unchanged). synchronize + labeled fire for
+    # every PR but are admitted here only for needs-gate, so a fresh-head review is
+    # guaranteed both when the PR is queued (label added) and on each subsequent push,
+    # without re-reviewing every push of every PR.
+    if: >-
+      github.event.action == 'opened' ||
+      github.event.action == 'ready_for_review' ||
+      github.event.action == 'reopened' ||
+      (github.event.action == 'synchronize' && contains(github.event.pull_request.labels.*.name, 'needs-gate')) ||
+      (github.event.action == 'labeled' && github.event.label.name == 'needs-gate')
 
     runs-on: ubuntu-latest
     permissions:


thomgabriel:
## GATE: approved

Verified the `if:` myself, because a subtle error there fails in the worst direction — silently disabling review for ordinary PRs, which is strictly worse than the race it fixes. It's correct: `opened / ready_for_review / reopened` still admit **any** PR unchanged, and only `synchronize` (label present) and `labeled` (label *is* needs-gate) are scoped. Concurrency keyed on PR number, so PRs can't cancel each other.

Re-confirmed the load-bearing safety claim rather than taking it from the body: required checks on `main` are `Frontend`, `On-chain`, `SHA-pin guard` — **`claude-review` is not among them**, so a cancelled run cannot block a merge. That's the whole reason `cancel-in-progress` is safe here.

No feedback loop: the review posts a comment, and comments raise neither `synchronize` nor `labeled`.

This is a real fix for a real problem — I hit it repeatedly last night. 9 of ~15 reviews were stale, and in 4 the *unreviewed* commit was the security-relevant one.

### One operational consequence worth naming

`cancel-in-progress` will make **cancelled runs routine** on gated PRs — every push now supersedes the in-flight review. And `gh pr checks` renders a cancelled run as **`fail`**, indistinguishable at a glance from a real failure.

That already bit me twice last night (#722, and again on #740/#752) — I nearly bounced a PR over it. After this merges it stops being an occasional race artifact and becomes the normal steady state for exactly the PRs under gate review.

Not blocking, and not a reason to change the design — the fix is correct. But whoever gates has to check `conclusion` before believing a red row:

```bash
gh run view <run-id> --json status,conclusion,event
```

I've added that to my own protocol. Worth a line in the handoff doc too, since it now applies to every gated PR rather than the occasional one.

CI green.

thomgabriel:
@claude review — head is now b54ac821988d35872db5e6870b2cc7063e09f107 (concurrency moved to job-level per the adversarial finding: workflow-level concurrency was cancelling the in-flight open review on ordinary PRs whose synchronize run then skips). Please review this head.

thomgabriel:
**Orchestrator routing record — SENSITIVE (`.github/workflows/**`) → adversarial FAIL → fix → re-verify PASS → queuing for gate.**

**Round 1 (`4bab69c7`): FAIL, one MAJOR** — workflow-level `concurrency` + the new `synchronize` trigger would let a quick push-after-open on an ORDINARY PR create a run that claims the group, cancels the in-flight opening review, then skips its own job — silently dropping the PR's only review (a regression: today synchronize doesn't trigger at all). Everything else held: label-race semantics, `labeled` payload field, expression parse, cost bound, one-file diff, permissions untouched.

**Fix (`b54ac82`): re-verified PASS by the same reviewer.** `concurrency` moved to JOB-level — claimed at job START (after the `if`), so a skipped non-needs-gate run never enters the group and can't cancel anything; the workflow's only job carries the `if` (verified no unconditional sibling job); needs-gate stale-head cancellation preserved (both qualifying runs start → newer cancels older — precisely the runs that will post a replacement). Load-bearing why-comment added so it isn't 'simplified' back. My checks at head: yaml parses, pin-guard green, job-scoped placement confirmed.

**For the gate:** this closes #710's mechanical half — your needs-gate queue now guarantees a fresh-head review on every push and on label-add, with stale in-flight reviews superseded rather than posted. Note the PR API may show a stale head; the branch ref is `b54ac82`.

Issue #607, opened 2026-07-25
The program instance moved: Dsro2Cd9 is live, learner on-chain state is stranded on 7NeJa


thomgabriel:
`NEXT_PUBLIC_PROGRAM_ID=Dsro2Cd9…` (fresh Pinocchio instance, deployed 2026-07-21) with a **new XP mint** — not the `7NeJaSRy…` instance every prior document reasoned about. The old instance holds 38 enrollments and 30 achievement receipts; the new one holds zero, and they **cannot** migrate (enrollment PDAs are program-derived). Supabase rows still exist; their on-chain counterparts do not.

**Owner decision made (2026-07-25): do NOT re-mint XP — start from zero and clean up the DB.**

Cleanup constraints: **never touch `profiles` or auth users** (a large cohort signed up at last week's event — resetting progress is recoverable, dropping identities is not). Reset only chain-mirrored state: `enrollments`, `user_progress`, `user_xp`, `xp_transactions`, `user_achievements`. **Investigate `certificates` separately** — credential NFTs are Metaplex Core assets in the learner's wallet and may still exist. Ship as a migration with a **dry-run count first**, matching the reset-wave pattern.

Also fix `apps/web/src/lib/solana/program-errors.ts`: 33 entries spanning 6000-6032, missing **6034/6035/6036**, so an affected learner sees a raw unmapped error. Refresh the IDL and add the i18n strings **before** any recreate window.

**Unified spec item 50b** — full detail, evidence and acceptance criteria in [`docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md`](docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md) §3.

_Filed from the unified launch spec; this work existed in the plan but had no issue._

thomgabriel:
## ⚠️ Third consequence of the program move: the Helius webhook is still watching the OLD program — the reward loop is silently dead

Confirmed from the live Helius dashboard (owner screenshot) and against the code.

**Webhook config today:** `Account Addresses: 7NeJaSRyb4Wxay3Tcd9bdpD7T3GWYUQSFyrhG8SgwE8V` — the superseded instance. The app writes to `Dsro2Cd9Mhgk8L71imh3LLPwYU5PU8hvBY5HEcPrcx5u`.

Everything else in that webhook is correct and must NOT be changed:
- URL `https://superteam-academy-web.vercel.app/api/webhooks/helius` ✓
[...]

PR #760, opened 2026-07-27
fix(onchain): map program error codes 6031-6036 + i18n (#607, code half)

thomgabriel:
Closes the **code half** of #607 (P0). The migration half is deliberately deferred until dry-run counts are posted on the issue.

## What
`apps/web/src/lib/solana/program-errors.ts` stopped at code **6030** (31 entries), but the live Pinocchio program and the IDL define codes through **6036**. Any learner hitting 6031–6036 saw a raw unmapped `custom program error: 0x…` instead of a localized message. The most launch-relevant is **6034 `StaleEnrollment`** — the exact error a learner on the stranded `7NeJa` instance hits after the move to `Dsro2Cd9`.

Adds the six missing entries + en/pt-BR/es i18n strings:

| Code | Name | i18nKey |
|------|------|---------|
| 6031 | MintingPaused | `mintingPaused` |
| 6032 | WrongXpMint | `wrongXpMint` |
| 6033 | XpAmountExceedsMax | `xpAmountExceedsMax` |
| 6034 | StaleEnrollment | `staleEnrollment` |
| 6035 | OldMinterRoleMissing | `oldMinterRoleMissing` |
| 6036 | EnrollmentInProgress | `enrollmentInProgress` |

## Scope correction vs. the issue
The issue says "33 entries spanning 6000-6032, missing 6034/6035/6036." The file on `main` actually has **31** entries (6000-6030) and is missing **six** codes (6031-6036), not three. 6031/6032/6033 (`MintingPaused`, `WrongXpMint`, `XpAmountExceedsMax`) were also unmapped, so fixing only 6034-6036 would still leave those surfacing raw errors. This PR maps all six for full IDL parity.

## Source-of-truth trace
No drift found between the three definitions — the fix only re-syncs the stale **client** map:
- **Program enum:** `onchain-academy/programs/onchain-academy-pinocchio/src/errors.rs` — `AcademyError` variants 31–36 (`MintingPaused`…`EnrollmentInProgress`), custom code = `6000 + ordinal`.
- **`ACADEMY_LOGS` table** (same file) — pre-baked `Error Number: 603x` log lines, names/numbers/messages "match the IDL errors table verbatim."
- **IDL:** `apps/web/src/lib/solana/idl/superteam_academy.json` `errors[]` — already contains all 37 codes (6000-6036). **The IDL was not stale and was not touched** (respecting #449: no account layouts modified; the `errors` table is independent of the decoder's layout concern anyway).

English fallbacks are user-friendly paraphrases of the IDL `msg`, matching the existing entry style; i18nKeys follow the existing camelCase convention. Consumers use `parsed.i18nKey ? t(parsed.i18nKey) : parsed.fallback`, so both surfaces are covered.

## Tests
New `apps/web/src/lib/solana/__tests__/program-errors.test.ts` (19 tests):
- the six new codes map (hex + error-name paths)
- unknown-code fallback still returns the raw message with the code preserved
- existing mappings (6000, 6008, 6030) unchanged
- **parity guard:** every IDL error code resolves to a mapped entry **and** has a non-empty string in all three locales — this fails CI if the map drifts from the IDL again.

## Verification
```
pnpm install --frozen-lockfile   # ok
pnpm typecheck                    # 6 successful, 6 total
pnpm --filter web test            # Test Files 210 passed (210) · Tests 1922 passed (1922)
```

Code half of #607; migration half follows the posted dry-run counts.

### DIFF
--- a/apps/web/src/lib/solana/__tests__/program-errors.test.ts
+++ b/apps/web/src/lib/solana/__tests__/program-errors.test.ts
@@ -1,0 +1,96 @@
+import { describe, it, expect } from "vitest";
+import { parseProgramError } from "../program-errors";
+import idl from "../idl/superteam_academy.json";
+import en from "../../../messages/en.json";
+import ptBR from "../../../messages/pt-BR.json";
+import es from "../../../messages/es.json";
+
+const idlErrors = (idl as { errors: { code: number; name: string }[] }).errors;
+const locales = { en, "pt-BR": ptBR, es } as const;
+
+/** Format a custom program error string the way RPC simulation surfaces it. */
+const hex = (code: number) => `custom program error: 0x${code.toString(16)}`;
+
+describe("parseProgramError — newly added codes (6031–6036)", () => {
+  const added: [number, string, string][] = [
+    [6031, "MintingPaused", "mintingPaused"],
+    [6032, "WrongXpMint", "wrongXpMint"],
+    [6033, "XpAmountExceedsMax", "xpAmountExceedsMax"],
+    [6034, "StaleEnrollment", "staleEnrollment"],
+    [6035, "OldMinterRoleMissing", "oldMinterRoleMissing"],
+    [6036, "EnrollmentInProgress", "enrollmentInProgress"],
+  ];
+
+  it.each(added)("maps hex code for %i → %s", (code, name, i18nKey) => {
+    const parsed = parseProgramError(new Error(hex(code)));
+    expect(parsed.code).toBe(code);
+    expect(parsed.name).toBe(name);
+    expect(parsed.i18nKey).toBe(i18nKey);
+    expect(parsed.fallback.length).toBeGreaterThan(0);
+    expect(parsed.fallback).not.toContain("0x");
+  });
+
+  it.each(added)("maps error name for %i (%s)", (_code, name, i18nKey) => {
+    const parsed = parseProgramError(new Error(`Something failed: ${name}`));
+    expect(parsed.name).toBe(name);
+    expect(parsed.i18nKey).toBe(i18nKey);
+  });
+});
+
+describe("parseProgramError — unknown-code fallback", () => {
+  it("preserves an unmapped custom code and returns the raw message", () => {
+    const parsed = parseProgramError(new Error("custom program error: 0x270f")); // 9999
+    expect(parsed.code).toBe(9999);
+    expect(parsed.name).toBeNull();
+    expect(parsed.i18nKey).toBeNull();
+    expect(parsed.fallback).toContain("0x270f");
+  });
+
+  it("returns the raw message when nothing matches", () => {
+    const parsed = parseProgramError(new Error("blockhash not found"));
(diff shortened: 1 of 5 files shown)


thomgabriel:
## GATE: approved

Traced all three sources of truth rather than accepting the table, since a wrong code→name mapping would show learners a *confidently incorrect* message — worse than the raw hex it replaces:

- **Program enum** `onchain-academy-pinocchio/src/errors.rs:52-57` — `MintingPaused = 31` … `EnrollmentInProgress = 36`, custom code `6000 + ordinal`.
- **`ACADEMY_LOGS`** in the same file — pre-baked `anchor_log!` lines at 6031-6036 with matching names.
- **IDL** — 37 errors, 6031-6036 present, names identical.

All three agree with the PR's table. i18n parity confirmed: **6/6 keys in each of en / es / pt-BR**, so no `MISSING_MESSAGE` risk on any locale.

**The scope correction is the valuable part.** The issue said "33 entries spanning 6000-6032, missing 6034/6035/6036"; the file actually had **31** entries (6000-6030) missing **six**. Fixing only the three the issue named would have left `MintingPaused`, `WrongXpMint` and `XpAmountExceedsMax` still surfacing raw hex — a fix that closes the ticket without closing the bug. Checking the file instead of trusting the issue is exactly right; I wrote that issue text and got it wrong.

**6034 `StaleEnrollment` earns its priority.** It is precisely what a learner stranded on `7NeJa` hits after the move to `Dsro2Cd9` — I confirmed this morning that the current program has **zero enrollments**, so every pre-move learner is in exactly that state. Until now they'd see `custom program error: 0x1796`.

Right call leaving the IDL untouched (#449) and deferring the migration half until dry-run counts are posted. That ordering is the one I'd have insisted on.

CI green.

thomgabriel:
**Orchestrator routing record — area:onchain (client error map) → adversarial review PASS → queuing for gate.**

**My independent verification** at `3600eb59`: typecheck exit 0; full web suite **210 files / 1922 tests, exit 0**.

**Adversarial review: PASS, no findings.**
- Trace verified against all THREE sources IN LOCKSTEP: `errors.rs` AcademyError ordinals 31-36 → codes 6031-6036, the IDL `errors[]` (already current — untouched, #449 respected), and the pre-baked ACADEMY_LOGS lines. No off-by-one; names one-for-one.
- The scope correction stands: SIX codes were missing (the issue said three; the file had 31 entries, not 33).
- **6034 StaleEnrollment** — the code a stranded 7NeJa learner hits — maps to a message that both describes the superseded-generation situation accurately AND gives the correct remediation (re-enroll), matching the recreate/reset design.
- Parity guard proven honest by mutation: imports the REAL IDL; blanking one es string fails with the exact missing-key message; a removed map entry fails resolution. Drift in any source now trips CI.
- Existing 31 entries byte-unchanged; unknown-code fallback preserved and tested; i18n keys collision-free across 3 locales.

This is #607's code half. The migration half remains deferred pending the dry-run counts requested on the issue.

PR #761, opened 2026-07-27
fix(content-lint): resolve a slashed base branch instead of hard-failing CI (#748)

### LINKED ISSUE #748: normalizeBaseRef mishandles a base branch containing a slash — now a hard CI failure
Surfaced by `claude-review` on #747 (non-blocking there, filing so it isn't lost).

`packages/content-lint/src/cli.ts:10-12`:

```ts
return raw.includes("/") ? raw : `origin/${raw}`;
```

The `origin/` prefix is only added when the raw ref has no slash. A base branch whose own name contains one — `release/1.0`, `hotfix/x` — arrives from `github.base_ref` as `release/1.0` and is used verbatim as `ctx.baseRef`. It is never remapped to `origin/release/1.0`, so it does not resolve in the linter's checkout.

**#747 raised the stakes on this.** Previously an unresolvable base ref degraded to a silent skip (pre-#742) or a warning-and-skip (post-#742) — both exit 0. #747 deliberately made explicit-but-unresolvable a **hard error**, which is correct for the misconfiguration case it targets. The side effect is that a slash-containing base branch now fails CI outright rather than passing quietly.

Not live today: the default branch is `main`, and `content-lint`'s only CI caller is `validate-content.yml`, which passes `github.base_ref`. It becomes real the first time a content PR targets a branch with a slash in its name.

**Fix**: decide whether a ref is already remote-qualified by checking for a leading `origin/` (or verifying it resolves) rather than by the presence of any `/`. Something like: use `raw` verbatim if `git rev-parse --verify` resolves it, else try `origin/${raw}`. A test with a `release/1.0` base branch would pin it.

Note the failure is at least now loud rather than silent — which is #747 working as intended, not a regression it caused.

thomgabriel:
Fixes #748.

## Problem
`packages/content-lint/src/cli.ts` `normalizeBaseRef` added the `origin/` prefix only when the raw ref had **no** slash:
```ts
return raw.includes("/") ? raw : `origin/${raw}`;
```
`github.base_ref` is a **bare branch name**. For `main` that becomes `origin/main` and resolves. But a base branch whose own name contains a slash — `release/1.0`, `hotfix/x` — arrives verbatim and is used as-is, so it never resolves in the linter's checkout. **#747 raised the stakes:** an explicit-but-unresolvable base is now a HARD CI failure (fail-closed), so the first content PR targeting a slashed base branch would red-wall outright. Not live today (default branch is `main`) but latent.

## Fix
Decide "already qualified" by a leading `origin/` / `refs/` (or by actual resolution), not by the presence of any slash. `normalizeBaseRef` now:
1. `origin/…` / `refs/…` → verbatim (already qualified).
2. bare branch name → `origin/<raw>` when that resolves — **regardless of embedded slashes** (the fix); this also preserves the old `main` → `origin/main` target.
3. otherwise a SHA / local ref that resolves as-is → verbatim (the old code actually mis-prefixed a no-slash SHA to `origin/<sha>`; now handled correctly).
4. nothing resolves → return the `origin/`-qualified form so #747's hard error names the ref CI meant to fetch.

**#747 fail-closed is intact:** only two candidates are ever tried (`origin/<raw>`, then `<raw>`), so a genuinely bogus ref matches neither and still ERRORs — it never silently passes.

`normalizeBaseRef` was extracted into `base-ref.ts` (cli.ts self-runs `main()` on import, so it can't be imported for unit tests) and a small `resolvesRef` helper added to `git.ts`.

## Tests (`__tests__/base-ref.test.ts`, 8)
- slashed bare branch `release/1.0` → `origin/release/1.0`
- plain bare `main` → `origin/main` (unchanged)
- already-qualified `origin/main` → verbatim
- full SHA → verbatim
- `undefined` → `undefined`
- genuinely unresolvable `release/9.9` → stays `origin/release/9.9` (never bare) — fail-closed
- end-to-end: a slashed base resolves and gate-3 runs with no misconfiguration error
- end-to-end: an unresolvable base still fails CLOSED with #747's `misconfiguration` error

`pnpm --filter @superteam-lms/content-lint test` → 105 passed (base-ref: 8). `content-schema` → 170. `typecheck` clean.

## courses-academy blast — GREEN
| invocation | exit |
|---|---|
| bare | 0 |
| CI-sim `GITHUB_BASE_REF=main` | 0 |
| **slashed base** `LINT_BASE_REF=release/1.0` (created `origin/release/1.0`) | **0** — resolves; before this PR it would hard-fail |

### DIFF
--- a/packages/content-lint/src/__tests__/base-ref.test.ts
+++ b/packages/content-lint/src/__tests__/base-ref.test.ts
@@ -1,0 +1,115 @@
+import { describe, it, expect } from "vitest";
+import { execFileSync } from "node:child_process";
+import { runLint } from "../lint";
+import "../checks/gate1-schema";
+import "../checks/gate3-slots";
+import { normalizeBaseRef } from "../base-ref";
+import { makeTempRepo } from "./helpers";
+
+const course = `id: course-x
+slug: x
+title: X
+difficulty: beginner
+duration: 1
+xpPerLesson: 10
+xpReward: 100
+modules: [{ key: m, title: M, lessons: [lesson-a, lesson-b] }]
+`;
+const lesson = (id: string) => `id: ${id}
+slug: ${id.replace("lesson-", "")}
+title: ${id}
+blocks: [{ type: prose, key: intro, src: intro.md }]
+`;
+const lock = JSON.stringify({
+  version: 1,
+  slots: { "lesson-a": 0, "lesson-b": 1 },
+  retired: [],
+  next: 2,
+});
+
+/**
+ * A committed repo whose base branch is reachable both as a bare-named remote
+ * branch (`origin/release/1.0`) and as the default `origin/main`.
+ */
+function repoWithBranches(): string {
+  const root = makeTempRepo({
+    "courses/x/course.yaml": course,
+    "courses/x/slots.lock.json": lock,
+    "courses/x/lessons/a/lesson.yaml": lesson("lesson-a"),
+    "courses/x/lessons/a/intro.md": "# a",
+    "courses/x/lessons/b/lesson.yaml": lesson("lesson-b"),
+    "courses/x/lessons/b/intro.md": "# b",
+  });
+  const git = (...args: string[]) =>
+    execFileSync("git", args, { cwd: root, stdio: "ignore" });
+  git("init", "-q");
+  git("config", "user.email", "t@t.t");
+  git("config", "user.name", "t");
+  git("add", "-A");
+  git("commit", "-qm", "c1");
+  git("branch", "-M", "main");
+  // Remote-tracking refs, as a fetch-depth:0 checkout would have.
+  git("update-ref", "refs/remotes/origin/main", "main");
+  git("update-ref", "refs/remotes/origin/release/1.0", "main");
+  return root;
+}
+
+describe("normalizeBaseRef (#748)", () => {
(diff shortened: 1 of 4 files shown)


thomgabriel:
## GATE: approved

The risk with this one was specific: a resolution helper that tries harder to make a ref resolve is exactly how you'd accidentally undo #747's fail-closed. It doesn't.

Only two candidates are ever tried (`origin/<raw>`, then `<raw>`), and step 4 returns the **unresolvable** `origin/`-qualified form, so `mergeBase` still yields null and #747's hard error fires with the ref CI meant to fetch. `resolvesRef` uses `rev-parse --verify --quiet <ref>^{commit}`, so it only ever matches commits — no chance of latching onto an unrelated object.

The test pairing is what sells it: **"resolves a slashed base end-to-end: gate-3 runs"** alongside **"a genuinely unresolvable base still fails CLOSED end-to-end (#747 message)"**. Testing the fix and the invariant it could have broken, in the same suite, is the right shape for a change to a safety gate.

Nice catch on the incidental bug in passing — the old `raw.includes("/")` proxy also mis-prefixed a no-slash **SHA** to `origin/<sha>`. That was latent for the same reason the slash case was: nothing exercises a SHA base today.

Extracting into `base-ref.ts` because `cli.ts` self-runs `main()` on import is the right call — the alternative is testing through a process boundary.

CI green.

Issue #735, opened 2026-07-26
db: duplicate migration timestamp 20260726120000 (add_xp_transactions_source + lockdown_deployed_programs_rls)


thomgabriel:
Found by the adversarial review of #734 (whose own timestamp is clean). Two migration files share the version prefix `20260726120000`:

- `20260726120000_add_xp_transactions_source.sql`
- `20260726120000_lockdown_deployed_programs_rls.sql`

Both are MERGED and (per #731's reconciliation) applied to prod via MCP, so the ledger already diverges from filenames (#708's broader problem). Supabase CLI orders same-version files lexicographically by name, so behavior today is deterministic — but any tooling that keys on the version number alone (or a future `db push` onto a fresh instance) sees an ambiguity.

Fix shape: rename ONE file to a free version (e.g. `20260726121000_lockdown_deployed_programs_rls.sql`) in a hygiene PR — content byte-identical, rename only — and note it in the #708 reconciliation. Low urgency; fold into whatever #708 resolution the owner picks.

Also a standing lesson for the loop: timestamp-uniqueness checks must compare the FULL migrations dir (this pair predates the checks we added mid-loop — we started verifying after three near-collisions, but never audited the existing files).

thomgabriel:
Done via #762 (merged): `20260726120000_lockdown_deployed_programs_rls.sql` renamed to `20260726121000_...` — byte-identical content (verified by diff), the one functional readFileSync path and 3 comment refs updated, blast radius swept repo-wide by the adversarial review (the two remaining `20260726120000` references correctly name the OTHER, un-renamed file at that timestamp). MCP ledger unaffected; the broader ledger↔filename reconciliation is documented in docs/DB-MIGRATION-LEDGER.md (#708, also merged).

Issue #708, opened 2026-07-26
db: migration ledger diverged from repo filenames on all 6 MCP-applied migrations (db push would re-apply)


thomgabriel:
Found while gating #707 (whose commit message mentioned a migration "version collision", so I checked the whole directory and then the live ledger).

**No repo migration filename matches its row in prod's `supabase_migrations.schema_migrations`.** The migrations are genuinely applied — I verified the objects exist and exploit-tested several — but the ledger records them under different versions, so a CLI-driven deploy would consider them unapplied.

## Evidence

| repo filename (version) | prod ledger version | name |
|---|---|---|
| `20260726120000` | **`20260726143846`** | `add_xp_transactions_source` |
| `20260726120000` | **`20260726143927`** | `lockdown_deployed_programs_rls` |
| `20260726130000` | **`20260726144047`** | `route_public_profile_reads_through_view` |
| `20260726140000` | **`20260726162147`** | `review_items_spaced_repetition` |
| `20260726150000` | **`20260726173332`** | `add_profiles_segment_state` |
| — | `20260726143513` | `billed_assists_counter` |

**Cause:** these were applied through the Supabase MCP `apply_migration`, which stamps its own wall-clock version rather than honouring the filename's. That is not a misuse — MCP is the only path available to this environment (no psql, no CLI, no DB URL) — but it means the repo and the ledger have diverged on every migration applied that way.

## Consequences

1. **`supabase db push` would re-apply all six.** The CLI matches on `version`; none of the repo versions exist in the ledger. Most of these are idempotent by construction (`ADD COLUMN IF NOT EXISTS`, `DROP POLICY IF EXISTS` before `CREATE POLICY`, `ON CONFLICT DO NOTHING`, `CREATE OR REPLACE FUNCTION`) — I verified that specifically for `review_items_spaced_repetition` and `add_profiles_segment_state` — so a re-apply would most likely be a no-op. **"Most likely" is the problem**, not a reassurance: `add_xp_transactions_source` carries a backfill, and nobody should be discovering per-migration idempotency during a deploy.
2. **`supabase migration list` is misleading.** It will show local-only migrations that are in fact live, which is the exact signal an operator uses to decide whether a deploy is safe.
3. **A genuine duplicate exists in the repo**, currently masked by (1):
   ```
   20260726120000_add_xp_transactions_source.sql
   20260726120000_lockdown_deployed_programs_rls.sql
   ```
[...]

Issue #698, opened 2026-07-26
Prod snapshot check: was drop_teacher_role section (a) ever applied, or does prod still carry profiles.role?


thomgabriel:
Surfaced by #696/PR #697: `20260710120000_drop_teacher_role.sql` section (b) (wallet-write lock) is confirmed applied to prod (#408 closing comment), but section (a) (dropping `profiles.role`, its CHECK, and `enforce_profile_role_write`) was written behind an apply-gate pending the teacher-authoring repoint. schema.sql still carries the full role machinery. ONE prod query settles it: `SELECT column_name FROM information_schema.columns WHERE table_name='profiles' AND column_name='role';` — if the column is gone, schema.sql needs the section-(a) mirror too (a #696-class drift); if present, schema.sql is accurate and this closes. Owner/gate action (needs prod access). _Filed from the orchestrator loop._

thomgabriel:
## Answered — section (a) **was** applied. Prod does not carry `profiles.role`.

You asked for one prod query; here are all four section-(a) objects checked individually against `pywhtmidcrptomrabbrw`, with section (b) alongside for contrast:

| object | prod |
|---|---|
| `profiles.role` column | **0** |
| `chk_profiles_role` | **0** |
| `enforce_profile_role_write` | **0** |
| `trg_enforce_profile_role_write` | **0** |
| `enforce_profile_wallet_write` (section b) | 1 |
| `trg_enforce_profile_wallet_write` (section b) | 1 |

So the apply-gate was released: **both** sections ran. Section (b) is what PR #697 just mirrored (I diffed it line-by-line against prod's `pg_get_functiondef` — faithful). Section (a)'s effects are live and `schema.sql` never mirrored them.

Corroborating evidence, independent of the column check — prod's migration ledger:

```
20260711152518  drop_teacher_role_lock_wallet_writes
```

Note it is recorded under a **different version and name** than the repo's `20260710120000_drop_teacher_role.sql`. That mismatch is its own problem, filed as #708.

And app code agrees the drop was intentional, not accidental — the sole surviving reference calls it retired:

> `apps/web/src/app/api/teacher/courses/[id]/stats/route.ts:20` — "replaces the retired `profiles.role` / `course.author`"

## What `schema.sql` still carries

```
role TEXT NOT NULL DEFAULT 'learner'          (line 28)
chk_profiles_role CHECK (role IN (…))         (line 142)
enforce_profile_role_write                     6 references
```

[...]

PR #762, opened 2026-07-27
fix(db): honest schema.sql for retired profiles.role + dedupe migration timestamp (#699, #735)

### LINKED ISSUE #699: db: schema.sql drifted from prod in both directions (retired profiles.role present, live wallet trigger absent) — makes a guard test hollow

thomgabriel:
Closes #699 and #735. Makes the two authoritative DB records honest about prod. No live-DB action; `schema.sql` is the snapshot new environments are built from, so it must match prod.

## #699 — schema.sql drift (retired-role half)
The wallet-lock half (#697) was already mirrored. This fixes the other direction: `schema.sql` still carried the **retired** `profiles.role` machinery that prod dropped.

- **Removed from schema.sql** (all executable DDL): the `role` column, `chk_profiles_role`, the `enforce_profile_role_write()` function, and `trg_enforce_profile_role_write`. SP1 dropped these via migration `20260710120000_drop_teacher_role.sql` (applied to prod as ledger `20260711152518`, confirmed in #698/#699 — all four objects absent from prod). Left explanatory prose so the retirement is discoverable; do **not** re-add to prod.
- **Fixed the TWO hollow guard tests first** (the trap the gate flagged): `segment-state-rls-guard.test.ts` and `wallet-write-lock-guard.test.ts` both pinned the retired role trigger in `schema.sql` — green while guarding nothing. Re-pointed each at the **deployed** wallet lockdown and added assertions that the role machinery is **absent** from the mirror (so the tests now track prod). Every genuine negative regression guard kept.
- **Corrected the applied #566 segment-state migration header** (comment-only, executable half byte-identical — verified 0 non-comment changed lines): it cited `enforce_profile_role_write` as the safety rationale, but role was already retired when it applied. Honest version recorded: nothing on `profiles` is privilege-bearing, so whole-row self-write is bounded by the CHECK constraints alone; the surviving column-lock is the wallet one.

Verified in passing: `types.ts` has no `role` (already matches prod) and no app code reads/writes `profiles.role` (only historical comments), so removing it is snapshot-only.

## #735 — duplicate migration timestamp
`20260726120000_lockdown_deployed_programs_rls.sql` shared the `20260726120000` prefix with `add_xp_transactions_source.sql`. Renamed it to **`20260726121000_...`** (pure `git` rename, content byte-identical — proven by `git show HEAD:old | diff - new`). Updated the one functional `readFileSync` path (`deploy/save/route.test.ts`) plus three comment references. The MCP ledger is unaffected — both files were applied under distinct MCP-stamped versions (that divergence is #708, addressed in a separate doc PR).

## Verify
- `pnpm --filter web test` — 209 files / 1903 tests pass


thomgabriel:
## GATE: approved — verified against prod, not against the description

The whole claim of this PR is "schema.sql now matches prod," so I checked prod rather than reading the diff:

| object | prod | schema.sql after this PR |
|---|---|---|
| `profiles.role` column | **absent (0)** | removed ✓ |
| `chk_profiles_role` | **absent (0)** | removed ✓ |
| `enforce_profile_role_write()` | **absent (0)** | removed ✓ |
| `trg_enforce_profile_role_write` | **absent (0)** | removed ✓ |
| `trg_enforce_profile_wallet_write` | **present (1)** | kept ✓ |

The four remaining textual mentions are prose (lines 32/33, 170, 759) — I grepped for executable DDL specifically and there is none, so a fresh environment built from this snapshot won't resurrect the role machinery. Leaving the prose is right: the retirement stays discoverable without being re-created.

**The hollow guard tests are genuinely fixed, not just re-pointed.** `wallet-write-lock-guard` previously asserted `toContain("CREATE TRIGGER trg_enforce_profile_role_write")` — green while guarding an object prod dropped. It now asserts the wallet lock is present **and** `not.toContain` for the role DDL, so it tracks prod in both directions. That was the trap I flagged on #699 and it was handled in the right order: tests first, then the removal.

**#735**: rename verified byte-identical (`diff` of old blob vs new), old path gone, and **zero duplicate timestamp prefixes remain** across the whole migrations directory. Correct to note the MCP ledger is unaffected — both were applied under distinct stamped versions, which is #708's problem, not this one's.

Good catch on the #566 header too: it cited `enforce_profile_role_write` as its safety rationale when role was already retired at apply time. Comment-only, executable half byte-identical.

### CI note
The two `fail` rows are run `30267295786`, conclusion **`cancelled`**; the push runs succeeded. This is the first PR through the new #759 concurrency behavior, so it's the predicted steady state rather than a problem — worth internalizing that a red row on a gated PR now usually means "superseded," and `conclusion` is the only reliable read.

thomgabriel:
**Orchestrator routing record — SENSITIVE (schema.sql surgery + guard-test re-pointing) → adversarial review PASS, no findings → queuing for gate.**

**My independent verification** at `7d2e3250`: rename byte-identical (diff clean), #566/#695 header edit comment-only (13 changed lines, 0 non-comment — programmatic), remaining role references in schema.sql are documentation-only (4 comment hits, no DDL), typecheck exit 0, full suite 209/1903.

**Adversarial review: PASS, the strong kind.**
- **#699 closed in BOTH directions, verified not claimed**: Direction 1 (role machinery schema.sql-only) fully removed as one atomic block; Direction 2 (prod's wallet-write trigger) confirmed genuinely present in schema.sql via #696/#697 — no unaddressed inventory entry, the half-fix trap does not apply.
- **Guards strengthened, mutation-proven**: assertion counts UP (3→7 on segment-state guard); re-adding role machinery to schema.sql fails 2 tests; removing the deployed wallet trigger fails 3. Absence assertions pin DDL strings, so the explanatory comments don't false-trip.
- **#735 rename blast fully swept**: repo-wide grep leaves exactly two refs, both correctly naming the OTHER (un-renamed) 120000 file; the deploy/save test's readFileSync reads the renamed file and asserts content — ENOENT would fail loudly, not skip.
- Structural sanity: dollar-quote balance even, no orphaned references, profiles table well-formed.

**For the gate**: schema.sql now claims prod's actual state for the role machinery — your #699 finding, made honest. Snapshot + guard change only; nothing to apply.

PR #763, opened 2026-07-27
docs(db): migration ledger reconciliation reference (#708)

### LINKED ISSUE #708: db: migration ledger diverged from repo filenames on all 6 MCP-applied migrations (db push would re-apply)

thomgabriel:
Closes #708 (documentation half). The database is fine; the *record* of it is not. Every migration applied via the Supabase **MCP** was stamped with an MCP wall-clock version that diverges from its repo filename, so `schema_migrations` doesn't match `supabase/migrations/` — a CLI `db push` would consider all of them unapplied (and re-run `add_xp_transactions_source`'s backfill).

The issue prescribes both a ledger *repair* (`supabase migration repair`, which needs prod access the gate holds) **and** "write down the rule". This PR delivers the code-side half: the authoritative reference doc + the going-forward convention. The actual `migration repair` is a gate/owner action — spelled out in the doc.

## `docs/DB-MIGRATION-LEDGER.md`
- **Cause** (MCP stamps its own version) and **consequences** (`db push` re-applies; `migration list` misleads; the now-fixed `20260726120000` duplicate).
- **Mapping table**: repo filename ↔ prod ledger version ↔ evidence. The seven versions the gate reported in #708/#699 are recorded directly; every row the gate has **not** reported is marked **PENDING-GATE-VERIFY** (I can't read prod). Two load-bearing flags in that column: `review_quest_kind` may **not** be applied (#750 found prod's `get_daily_quest_state` has no review branch), and `streak_forgiveness`'s function bodies are **not** applied (blocked on #750).
- **Reconciliation**: the per-row `supabase migration repair --status applied <repo-version>` steps, marked explicitly as a **gate/owner** action (no prod access here).
- **Going-forward convention**: an MCP-applied migration isn't "done" until its MCP-stamped version is recorded here **and** the ledger row is reconciled; timestamp-uniqueness is checked against the **full** dir (the trap that produced #735); plus the #731 apply-before-merge discipline.
- Referenced from `CLAUDE.md` (Database section).

Docs-only — no code, no prod action.

## Verify
- `pnpm --filter web test` — 209 files / 1903 tests pass (unchanged; docs-only)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -123,6 +123,7 @@ See `docs/ARCHITECTURE.md` for the program specification and frontend integratio
 - `award_xp()`, `unlock_achievement()`, and `get_daily_quest_state()` are **SECURITY DEFINER** functions
 - **REVOKE**d from `authenticated`, `anon`, and `public` roles — **GRANT**ed only to `service_role`
 - Called exclusively from API routes via `createAdminClient()` (`lib/supabase/admin.ts`)
+- **Migration ledger**: migrations applied via the Supabase MCP get an MCP-stamped version that diverges from the repo filename, so `schema_migrations` does not match `supabase/migrations/`. The authoritative cross-reference + the required post-apply reconciliation rule live in [`docs/DB-MIGRATION-LEDGER.md`](docs/DB-MIGRATION-LEDGER.md) — read it before any `supabase db push` / `migration list`, and extend the mapping when you apply a migration via MCP (#708).
 
 ### Auth Security
 
--- a/docs/DB-MIGRATION-LEDGER.md
+++ b/docs/DB-MIGRATION-LEDGER.md
@@ -1,0 +1,115 @@
+# DB migration ledger — filename ↔ prod ledger reconciliation (#708)
+
+**Status:** the database is correct; the _record_ of it is not. Every migration
+applied to prod via the Supabase **MCP** `apply_migration` was stamped with MCP's
+own wall-clock version, not the repo filename's version. So
+`supabase_migrations.schema_migrations` (the ledger the CLI keys on) diverges
+from the repo filenames on every MCP-applied migration. The objects are really
+there — verified present and, for several, exploit-tested — but a CLI-driven
+deploy (`supabase db push` / `migration list`) would consider them unapplied.
+
+This file is the authoritative cross-reference until the ledger is repaired, plus
+the going-forward rule so it stops recurring. It is **documentation** — applying
+the repair is a prod action held by the gate (this repo ships SQL/docs only and
+never touches prod).
+
+> **PENDING-GATE-VERIFY** marks a prod-side fact this doc cannot verify from the
+> repo. The ledger versions below are as **reported by the gate** in #708 (and
+> #699 for `drop_teacher_role`); rows without a gate-reported version are marked
+> so the gate fills them from `select version, name from
+supabase_migrations.schema_migrations order by version`.
+
+## Cause
+
+MCP `apply_migration` runs the SQL and records the migration under a version it
+generates at apply time (UTC wall-clock), ignoring the `NNNNNNNNNNNNNN` prefix in
+the filename. The CLI, by contrast, keys the ledger on the filename's version.
+Nothing in the repo recorded that MCP-applied ≠ CLI-tracked, which is why five
+migrations drifted in a single day (2026-07-26) unnoticed.
+
+## Consequences (why this matters before mainnet)
+
+1. **`supabase db push` would re-apply the diverged migrations** — the CLI finds
+   none of the repo versions in the ledger. Most are idempotent by construction
+   (`ADD COLUMN IF NOT EXISTS`, `DROP POLICY IF EXISTS` before `CREATE POLICY`,
+   `ON CONFLICT DO NOTHING`, `CREATE OR REPLACE FUNCTION`), so a re-apply is most
+   likely a no-op — but `add_xp_transactions_source` carries a **backfill**, and
+   per-migration idempotency should never be discovered during a deploy.
+2. **`supabase migration list` is misleading** — it shows live migrations as
+   local-only, the exact signal used to judge whether a deploy is safe.
+3. **A duplicate repo version existed** (`20260726120000` shared by
+   `add_xp_transactions_source` and `lockdown_deployed_programs_rls`). Fixed by
+   #735/#762 (renamed the latter to `20260726121000`). Under a CLI apply that
+   keyed on version, one could have been recorded applied without running.
+
+## Mapping — repo filename ↔ prod ledger version
+
+Ledger versions per the gate's #708 probe against prod (`pywhtmidcrptomrabbrw`).
+"Objects verified" = the migration's objects were confirmed present on prod
+(schema.sql mirror + gate exploit tests, #708 / #731).
+
+| repo filename (version)                                                                       | prod ledger version               | ledger name                               | evidence                                                                                                                                         |
+| --------------------------------------------------------------------------------------------- | --------------------------------- | ----------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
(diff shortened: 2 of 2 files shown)


thomgabriel:
**Orchestrator routing record — docs-only ledger reconciliation → queuing straight for gate** (adversarial review would be redundant here: the doc's substance is prod-side facts only YOU can verify, and it marks every unverified row PENDING-GATE-VERIFY rather than asserting).

Sanity-checked at `f26f622`: 115-line doc, 10 PENDING-GATE-VERIFY markers, sections = cause / consequences / filename↔ledger mapping / **reconciliation steps (your `supabase migration repair` action)** / going-forward convention; CLAUDE.md pointer added; suite unaffected (1903 green). Two flags the doc surfaces for you: review_quest_kind may not be applied on prod (#750's finding), and streak_forgiveness bodies remain pending. The repair itself stays your action — this PR is the write-down-the-rule half of #708.

thomgabriel:
## GATE: approved — and here are the PENDING-GATE-VERIFY rows, filled in

Docs-only, so its whole value is accuracy. Marking rows you couldn't confirm as `PENDING-GATE-VERIFY` rather than guessing was the right call — that column is exactly what I can close. I pulled the live ledger:

**Prod `schema_migrations` has 20 rows; `supabase/migrations/` has 40 files.**

| repo file | prod ledger version |
|---|---|
| `20260710120000_drop_teacher_role` | `20260711152518` (name: `drop_teacher_role_lock_wallet_writes`) |
| `20260711120000_onchain_deployments` | `20260711205803` |
| `20260714120000_revoke_execute_handle_new_user` | `20260714140212` |
| `20260714130000_public_profiles_view` | `20260714151034` |
| `20260714140000_restrict_anon_profiles_select` | **TWO rows** — `20260714160922` + `20260714161051` |
| `20260714150000_course_maintenance_gate` | `20260714183656` |
| `20260715120000_platform_freeze` | `20260716130928` |
| `20260725120000_billed_assists_counter` | `20260726143513` |
| `20260726120000_add_xp_transactions_source` | `20260726143846` |
| `20260726121000_lockdown_deployed_programs_rls` | `20260726143927` |
| `20260726130000_route_public_profile_reads…` | `20260726144047` |
| `20260726140000_review_items_spaced_repetition` | `20260726162147` |
| `20260726150000_add_profiles_segment_state` | `20260726173332` |
| `20260726160000_add_profiles_prefs` | `20260726212137` |
| `20260726170000_review_quest_kind` | **NOT APPLIED** ✅ your flag was right |
| `20260726180000_ai_spend_ledger` | `20260726212108` |
| `20260726190000_streak_forgiveness` | `20260726212349` — stamped **`streak_forgiveness_schema`** ✅ schema half only, as you flagged |
| `20260726200000_cohort_leagues` | `20260726231538` |
| `20260726210000_course_changelog` | `20260726234958` |
| `20260726220000_course_changelog_status_kinds` | `20260726235007` |

Both load-bearing flags in the doc are **confirmed correct**.

## 🔴 The divergence is about twice what the doc describes

The ledger's **earliest row is `20260711152518`**. Every repo migration before that — roughly **19 files**, `20260624163347_baseline` through `20260709130000_award_xp_report_credited` — has **no ledger row at all**, because this database was created on 2026-07-11 and bootstrapped from the snapshot rather than by replaying migrations.

They are unmistakably applied. I probed six spanning the range:

```
challenge_assists  (20260707) ✓   rate_limits       (20260624) ✓
user_daily_quests  (20260709) ✓   threads           (20260625) ✓
award_community_xp (20260630) ✓   certificates.course_id (20260704) ✓
```

So `db push` would attempt ~20 migrations, not the 6–7 MCP-stamped ones the doc frames it around — and several of those early ones are destructive to replay (`lockdown_enrollments_rls`, `data_integrity_constraints`, `drop_teacher_role`). The reconciliation list needs those 19 rows too, or following this doc still leaves `db push` armed.

Also worth a row: `restrict_anon_profiles_select` maps to **two** ledger versions, so reconciliation there isn't 1:1.

Merging as-is — it's a strict improvement and the structure is right. Please fold the pre-`20260711` block and the two-row case in as a follow-up.

**Note on process:** `claude-review` failed on this PR (run `30267636451`, conclusion `failure`) — that's **#765**, the `labeled`/`track_progress` regression from my bad #759 merge, not anything about this PR. So this merges on my review alone, which I'm flagging rather than glossing: the automated second opinion was unavailable *because of my own earlier mistake*. Mitigated by the fact that I verified the substance directly against prod, which the reviewer could not have done.

Issue #765, opened 2026-07-27
REGRESSION: #759's  trigger hard-fails claude-review (track_progress unsupported for that event)


thomgabriel:
Introduced by #759, which **I approved and merged without waiting for its review** (see the process failure at the bottom).

## Symptom

Every PR that gets the `needs-gate` label produces an immediate red `claude-review` (~8s):

```
##[error]Action failed with error: track_progress for pull_request events is only
supported for actions: opened, synchronize, ready_for_review, reopened.
Current action: labeled
##[error]Process completed with exit code 1.
```

First seen on #763 (run `30267636451`), conclusion **`failure`** — a real failure, not the cancelled-run artifact #759 also makes routine.

## Impact

- **The `labeled` half of #759 never worked.** Its stated flow was *"queue it (needs-gate added) -> `labeled` fires a fresh review at the current head."* That path hard-fails before reviewing anything.
- The `synchronize` half **does** work, so re-review-on-push (the actual #710 race fix) is intact.
- `opened / ready_for_review / reopened` unaffected.
- Not merge-blocking (advisory check), but it red-Xes every gated PR and trains people to ignore review status — the opposite of #710's intent.

## Note: distinct from #764

#764 fixes a *different* #759 regression (workflow-level concurrency silently cancelling the only review on non-gated PRs). It moves `concurrency` to job level and does **not** touch the trigger types, so this bug survives it. Both need fixing.

## Fix options

1. **Drop `labeled` from `types:`** — rely on `synchronize` plus the behavioral `@claude review` re-request. Smallest; costs the "fresh review the moment it is queued" behavior.
2. **Keep `labeled`, make `track_progress` conditional** on the action not being `labeled`, via a workflow expression. Preserves the intended flow; the label-triggered review runs without progress tracking. Verify the action accepts a falsey expression for that input.
3. Keep `labeled` but route it to a separate job that omits `track_progress`.

Option 2 preserves what #759 was for; option 1 is the safe fallback.

## Gate process failure (mine)

Timeline on #759:

```
12:38:49Z  4bab69c7 pushed (only commit)
12:42:45Z  my "GATE: approved"
12:42:48Z  I merged                      <- 3 seconds later
12:49:26Z  claude-review posts           <- 7 minutes AFTER the merge
```

[...]

PR #766, opened 2026-07-27
fix(courses): changelog lessons_removed — recover id/title from prior content revision (#757)

### LINKED ISSUE #757: course_changelog lessons_removed records null id/title — the audit entry can't say what was removed
Found by observing the **first real changelog write in production** — the C3 mask sync (tx `5T1jeJ68GF3D`, version 4) that closed out #741/#740 tonight.

Two entries were written correctly, but one is only half-useful:

```json
// kind: lessons_added — complete
{"lessons":[
  {"id":"lesson-bfsp-pdas-vault",     "slot":16, "title":"PDAs: An Address With No Key"},
  {"id":"lesson-bfsp-cpi-lamports",   "slot":17, "title":"CPI: Moving Real Lamports"},
  {"id":"lesson-bfsp-pre-flight-check","slot":18, "title":"Pre-Flight Check"}
]}

// kind: lessons_removed — slots only
{"lessons":[
  {"id":null,"slot":0, "title":null},
  {"id":null,"slot":2, "title":null},
  {"id":null,"slot":11,"title":null},
  {"id":null,"slot":14,"title":null}
]}
```

## Why it happens

The writer resolves lesson id and title from the **current** content bundle. A removed lesson is by definition no longer in that bundle, so both lookups miss and it records `null`. The added-lessons path works because those lessons *are* present.

## Why it matters

The changelog exists so post-deployment mutation isn't silent (#654/#738). "Four lessons were removed, slots 0/2/11/14" tells a reader almost nothing — the slot number is an internal bitmap position, not something a human or a learner recognizes. For the one operation where the audit trail matters most (content disappearing from a live course), the entry is the least informative.

It's also self-worsening: the further the bundle moves on, the less recoverable those names become. Right now they're still trivially available in git — `lesson-bfsp-from-code-to-chain` (0), `lesson-bfsp-anatomy-anchor-program` (2), `lesson-bfsp-deploy-to-devnet` (11), `lesson-bfsp-m4-interact` (14).

## Options

1. **Resolve from the previous bundle.** The sync route knows the prior `content_tx_id` (the SHA being replaced). Reading the retired ids from that revision is the faithful answer, at the cost of a git read in the write path.
2. **Resolve from `slots.lock`'s retired history**, if the lock retains the id→slot mapping for retired entries (it currently does not — retired is a bare slot array; that itself may be worth reconsidering, since a slot number with no name is a weak permanent record).
3. **Cheapest**: fall back to the last known title from `user_progress` / prior changelog rows keyed by that slot.

Backfilling this one row by hand is also reasonable, given it is currently the only affected entry in the table.

Not urgent — no learner-facing impact, and the write path is correct in every other respect (dedup key, version, tx signature, both entries atomic under one signature).

thomgabriel:
Closes #757.

## What

The first real changelog write in production (the C3 mask sync, tx `5T1jeJ68…`, version 4) recorded `lessons_removed` as `{"id":null,"slot":N,"title":null}` — the audit entry couldn't say *what* was removed. A removed lesson is by definition gone from the **current** content bundle, so resolving its id/title against that bundle misses. (The `lessons_added` path works because those lessons *are* present.)

## Fix — resolve from the previous bundle (issue Option 1)

The pre-update on-chain `Course.content_tx_id` is a padded courses-academy git SHA, and at that SHA the removed lessons were still live — their `slots.lock.json` still maps id→slot and their `lesson.yaml` still carries the title.

- **`lib/content/prior-content.ts`** (new): `contentShaFromTxId` decodes `content_tx_id` (12 zero bytes + 20-byte sha) back to the sha; `resolvePriorRemovedLessons` fetches that revision's tarball (reusing `github.fetchTarball` + `extractTarball`), locates the course by matching each `course.yaml`'s `id`, reverses its `slots.lock.json` for the removed slots → ids, and snapshots titles from the prior `lesson.yaml` docs.
- **The course dir is NOT the slug** — a real gotcha this fix handles: slug `building-first-program` lives in `courses/building-your-first-solana-program/`. The course is found by `id` match, never a path guess.
- **`changelog-writes`**: `resolveSlots` takes a prior-override (removed slots only); `recordCourseUpdate` accepts `priorRemoved`. Added lessons still resolve from the current bundle. The sync route resolves `priorRemoved` **only** when a slot is actually retired.
- **Best-effort, non-fatal, rare path**: no `GITHUB_TOKEN`, network/tarball error, unparseable prior content, or a slot absent from the prior lock all degrade to slot-only (`id/title:null`) — never worse than today, never a fabricated name. Runs only on the uncommon retire-a-live-lesson sync.

### Verified against real data

Reversing the prior `slots.lock.json` at content sha `1b74e4a6` (the version-3 `content_tx_id`) yields exactly the ids the issue names from git history:

| slot | id | title |
|---|---|---|
| 0 | `lesson-bfsp-from-code-to-chain` | From Code to Chain |
| 2 | `lesson-bfsp-anatomy-anchor-program` | Anatomy of an Anchor Program |
| 11 | `lesson-bfsp-deploy-to-devnet` | Deploy to Devnet |
| 14 | `lesson-bfsp-m4-interact` | Interact with Your Program |

## Backfill the one affected prod row (for the gate to apply)

No migration for a single row (per the issue). The gate should run this against prod with `service_role`. Verify first, then update — the `WHERE` is idempotent (only touches the still-null row):

```sql
-- 1. Verify the target row (expect one lessons_removed row, currently all-null ids)
SELECT id, course_id, version, tx_signature, detail
FROM public.course_changelog
WHERE course_id = 'course-building-first-program'
  AND kind = 'lessons_removed'
  AND tx_signature LIKE '5T1jeJ68%';

-- 2. Backfill the ids/titles recovered from content sha 1b74e4a6
UPDATE public.course_changelog
SET detail = '{"lessons":[
  {"slot":0,"id":"lesson-bfsp-from-code-to-chain","title":"From Code to Chain"},
  {"slot":2,"id":"lesson-bfsp-anatomy-anchor-program","title":"Anatomy of an Anchor Program"},
  {"slot":11,"id":"lesson-bfsp-deploy-to-devnet","title":"Deploy to Devnet"},
  {"slot":14,"id":"lesson-bfsp-m4-interact","title":"Interact with Your Program"}
]}'::jsonb
WHERE course_id = 'course-building-first-program'
  AND kind = 'lessons_removed'
  AND tx_signature LIKE '5T1jeJ68%'
  AND detail->'lessons' @> '[{"id":null}]'::jsonb;
```

## Verification

- typecheck: 6/6 packages
- **web suite: 210 files, 1917 tests pass**

New tests: `prior-content` resolver (id+title, `dir != slug` id-match, null-title fallback, and every degrade path — fetch failure, course-not-found, slot-not-in-lock, empty removed set); `recordCourseUpdate` `priorRemoved` override + slot-only fallback (never fabricates); and an **exact C3-shaped reproduction** (remove slots 0/2/11/14, add 16/17/18) that now yields full entries with **no nulls**.

Do not merge — for review.

### DIFF
--- a/apps/web/src/app/api/admin/courses/sync/route.ts
+++ b/apps/web/src/app/api/admin/courses/sync/route.ts
@@ -38,6 +38,11 @@ import {
   recordCourseDeployed,
   recordCourseUpdate,
 } from "@/lib/content/changelog-writes";
+import {
+  contentShaFromTxId,
+  resolvePriorRemovedLessons,
+} from "@/lib/content/prior-content";
+import { diffMasks } from "@/lib/courses/changelog-diff";
 import { slotsByCourseId } from "@/lib/content/store";
 import { SYNCED_SHA } from "@/lib/content/meta";
 import { MaskMismatchError } from "@/lib/github/types";
@@ -582,6 +587,31 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
   if (onChainCourse && result.signature) {
     try {
       const contentCommitted = updateParams.contentTxId !== undefined;
+      // #757 — a removed lesson is gone from the CURRENT bundle, so its id/title
+      // can only come from the PRIOR content revision. Resolve those from the
+      // pre-update on-chain content_tx_id (a padded courses-academy sha) ONLY
+      // when this update actually retires a slot. Best-effort: any failure
+      // leaves `priorRemoved` empty and the removed entry records slot-only.
+      let priorRemoved:
+        | ReadonlyMap<number, { id: string; title: string | null }>
+        | undefined;
+      if (updateParams.newActiveLessons) {
+        const { removedSlots } = diffMasks(
+          onChainCourse.activeLessons,
+          updateParams.newActiveLessons
+        );
+        const priorSha =
+          removedSlots.length > 0
+            ? contentShaFromTxId(onChainCourse.content_tx_id)
+            : null;
+        if (priorSha) {
+          priorRemoved = await resolvePriorRemovedLessons({
+            courseId,
+            priorSha,
+            removedSlots,
+          });
+        }
+      }
       await recordCourseUpdate({
         courseId,
         txSignature: result.signature,
@@ -592,6 +622,7 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
         contentCommitted,
         newVersion: onChainCourse.version + (contentCommitted ? 1 : 0),
         contentSha: SYNCED_SHA,
+        priorRemoved,
       });
     } catch (changelogErr) {
       console.error(
--- a/apps/web/src/lib/content/__tests__/changelog-writes.test.ts
+++ b/apps/web/src/lib/content/__tests__/changelog-writes.test.ts
@@ -305,3 +305,99 @@ describe("#738 status + recreate recorders", () => {
     ).resolves.toBeUndefined();
   });
 });
+
+describe("#757 removed-lesson id/title from prior state", () => {
+  const baseUpdate = {
+    courseId: "course-x",
+    txSignature: "SIG_UPDATE",
+    oldXp: 10,
+    contentCommitted: true,
+    newVersion: 4,
+    contentSha: "abc123",
+  };
+
+  it("uses priorRemoved for removed slots (id + title), overriding the null bundle miss", async () => {
+    const priorRemoved = new Map([
+      [3, { id: "lesson-gone", title: "A Retired Lesson" }],
+    ]);
+    await recordCourseUpdate({
+      ...baseUpdate,
+      oldMask: mask(0, 1, 3),
+      newMask: mask(0, 1), // slot 3 removed; absent from the current bundle
+      priorRemoved,
+    });
+    const removed = lastRows?.find((r) => r.kind === "lessons_removed");
+    expect(removed?.detail).toEqual({
+      lessons: [{ slot: 3, id: "lesson-gone", title: "A Retired Lesson" }],
+    });
+  });
+
+  it("still records slot-only when priorRemoved lacks the slot (never fabricates)", async () => {
+    await recordCourseUpdate({
+      ...baseUpdate,
+      oldMask: mask(0, 1, 3),
+      newMask: mask(0, 1),
+      priorRemoved: new Map(), // resolver degraded / slot unresolved
+    });
+    const removed = lastRows?.find((r) => r.kind === "lessons_removed");
+    expect(removed?.detail).toEqual({
+      lessons: [{ slot: 3, id: null, title: null }],
+    });
+  });
+
+  it("reproduces the exact C3 prod case (tx 5T1jeJ68) — full removed entries, no nulls", async () => {
+    // Live-before: slots 0..15. Live-now: remove {0,2,11,14}, add {16,17,18}.
+    const priorRemoved = new Map([
+      [
+        0,
+        { id: "lesson-bfsp-from-code-to-chain", title: "From Code to Chain" },
+      ],
+      [
+        2,
+        {
+          id: "lesson-bfsp-anatomy-anchor-program",
+          title: "Anatomy of an Anchor Program",
+        },
+      ],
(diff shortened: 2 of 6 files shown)


thomgabriel:
**Orchestrator routing record — admin-sync write path + new remote-fetch code path → adversarial review PASS + polish round verified → queuing for gate.**

**My independent verification** at `1a32b135`: typecheck exit 0; full suite **210 files / 1917 tests, exit 0**; 4 degrade-warn sites confirmed in prior-content.ts.

**Adversarial review (at `25684949`): PASS.**
- `contentShaFromTxId` proven the EXACT inverse of the writer (`padContentTxId`) — round-trips, and every malformed input (wrong length, non-zero pad, all-zero fresh-course tx_id) degrades to null, never throws.
- Tarball path reuses the SHARED hardened extractor (128MiB gzip-bomb cap, 20k entries, in-memory map — no extract-to-disk traversal surface); course scan regex forbids path tricks; duplicate ids deterministic first-wins.
- Prior-override structurally scoped to removed slots only — added lessons provably resolve from the current bundle.
- Backfill snippet precise: the `@> '[{"id":null}]'` guard matches ONLY the broken row and is idempotent (populated rows don't match).
- **Design-point verdict: consistent with SP2's boundary** — the admin sync path is already GitHub-coupled in the same request; the serving path is untouched.

**Polish round (`1a32b135`), verified:** all degrade paths now console.warn with course + reason + 'slot-only' (the #731 bar — was the review's main ask); tarball fetch bounded by a 30s AbortController at the call site (CI callers unchanged); the mixed C3 test now asserts the added list too (cross-contamination pinned by test).

**For the gate:** the PR body carries the one-row prod backfill UPDATE (verify-then-update, @>-guarded, idempotent) for you to apply — recovers id/title for slots 0/2/11/14 on the C3 mask-sync entry.

thomgabriel:
@claude review — the automatic review never ran here: the `labeled` trigger hard-fails (#765, my regression from #759) and the other run skipped, so this PR has no review at all. Requesting one explicitly.

thomgabriel:
## GATE: approved

I verified the load-bearing primitive independently rather than from the description. Decoded the **live** `content_tx_id` off the Course PDA by hand — 12 zero bytes + 20-byte sha → `7a497475e14f3f53e5353cd752476eb0b2120096`, which matches `content.lock` exactly. Then ran `contentShaFromTxId` itself against that byte array plus the two reject cases (all-zero, non-zero pad): **3/3**. The padding inverse is sound.

Resolving the course by scanning `course.yaml`'s `id` rather than constructing a path from the slug is the right call and not a hypothetical — slug `building-first-program` genuinely lives in `courses/building-your-first-solana-program/`. A path guess would have silently found nothing and degraded to slot-only, which looks identical to "prior content unavailable."

Degradation is right throughout: missing token, network error, unparseable prior content, slot absent from the prior lock — all omit the slot rather than fabricate a name, and each warns. Given this exists *because* an audit row was uninformative, "never invent a name" is the correct failure direction.

### Backfill applied — data verified first

I checked the four recovered entries against the bundle at the prior sha (commit `7bd73b1`, content.lock `1b74e4a6`) before writing them into an audit record:

| slot | id | title |
|---|---|---|
| 0 | `lesson-bfsp-from-code-to-chain` | From Code to Chain |
| 2 | `lesson-bfsp-anatomy-anchor-program` | Anatomy of an Anchor Program |
| 11 | `lesson-bfsp-deploy-to-devnet` | Deploy to Devnet |
| 14 | `lesson-bfsp-m4-interact` | Interact with Your Program |

ids, titles **and** slot numbers all match the prior `slots.lock.json` and `lessons.json`. Applying the backfill; the `detail->'lessons' @> '[{"id":null}]'` guard keeps it idempotent.

### Process note

The automated review never fired here — the `labeled` trigger hard-failed (**#765**, my regression from merging #759 without waiting) and the other run skipped, leaving this PR with **zero** reviews. I requested one manually via `@claude review`; it came back clean and its findings match mine. I've escalated #765 to P1 since it is now demonstrably suppressing reviews, not just adding a red X.

CI green.

PR #767, opened 2026-07-27
fix(content-lint): SHA-shape base-ref passthrough so fail-closed is unconditional (#748 review follow-up)

thomgabriel:
Follow-up to #761 (already squash-merged). Adversarial review of that PR PASSed but flagged one MINOR against the team's fail-closed bar; this addresses it.

## The hole
#761's `normalizeBaseRef` step 3 used a verbatim fallback that resolved **any** ref `rev-parse` could resolve — including a **local branch** shadowing the base name. So when `origin/<base>` is missing (the exact misconfiguration #747 is meant to catch) **and** the checkout has a local branch of that name, the run silently diffed against the possibly-stale local branch instead of erroring. That makes #747's fail-closed *conditional* on the absence of a shadowing local branch. Not reachable in current CI (`actions/checkout` leaves a detached HEAD with no local branches), but a future checkout change would silently re-open a stale-base diff.

## Fix
Restrict the verbatim passthrough to **SHA-shaped** `raw` (`/^[0-9a-f]{7,40}$/i`). A bare branch name now resolves **only** via `origin/`, so a shadowing local branch can never satisfy it and the run fail-closes with #747's error. Genuine full/abbrev-SHA base passthrough (the intended #748 behavior) is unaffected. At most two candidates are ever tried: `origin/<raw>`, then a SHA-shaped `<raw>`.

## Tests (`base-ref.test.ts`, +2 → 10)
- **new:** a LOCAL branch `release` with no `origin/release` → `normalizeBaseRef("release")` returns `origin/release` (never the local branch)
- **new, end-to-end:** that same shadowed base still fails CLOSED with #747's `misconfiguration` error
- all prior cases unchanged (slashed→origin/, plain bare, already-qualified verbatim, full-SHA verbatim, undefined, unresolvable-slashed fail-closed, slashed resolves e2e, unresolvable fail-closed e2e)

`pnpm --filter @superteam-lms/content-lint test` → 107 passed (base-ref: 10). `typecheck` clean.

## courses-academy blast — GREEN
bare = exit 0; CI-sim `GITHUB_BASE_REF=main` = exit 0; slashed base `LINT_BASE_REF=release/1.0` = exit 0.

### DIFF
--- a/packages/content-lint/src/__tests__/base-ref.test.ts
+++ b/packages/content-lint/src/__tests__/base-ref.test.ts
@@ -90,6 +90,29 @@ describe("normalizeBaseRef (#748)", () => {
     expect(normalizeBaseRef("release/9.9", root)).toBe("origin/release/9.9");
   });
 
+  it("never falls through to a LOCAL branch shadowing the base name — stays fail-closed (#761 review)", () => {
+    const root = repoWithBranches();
+    // A local branch named "release" exists, but there is NO origin/release. A bare
+    // branch name must not resolve against the (possibly stale) local branch.
+    execFileSync("git", ["branch", "release"], { cwd: root, stdio: "ignore" });
+    expect(normalizeBaseRef("release", root)).toBe("origin/release");
+  });
+
+  it("a base shadowed only by a local branch still fails CLOSED end-to-end (#747)", async () => {
+    const root = repoWithBranches();
+    execFileSync("git", ["branch", "release"], { cwd: root, stdio: "ignore" });
+    const baseRef = normalizeBaseRef("release", root); // origin/release, unresolvable
+    const r = await runLint(root, { baseRef });
+    expect(
+      r.diagnostics.some(
+        (d) =>
+          d.gate === "gate-3" &&
+          d.severity === "error" &&
+          /misconfiguration/i.test(d.message)
+      )
+    ).toBe(true);
+  });
+
   it("resolves a slashed base end-to-end: gate-3 runs (no misconfiguration error)", async () => {
     const root = repoWithBranches();
     const baseRef = normalizeBaseRef("release/1.0", root);
--- a/packages/content-lint/src/base-ref.ts
+++ b/packages/content-lint/src/base-ref.ts
@@ -16,14 +16,22 @@ import { resolvesRef } from "./git";
  *   1. Already remote-/fully-qualified (`origin/…`, `refs/…`) → verbatim.
  *   2. Bare branch name → `origin/<raw>` when that resolves (the #748 fix; also
  *      preserves the old "main" → "origin/main" target for the no-slash case).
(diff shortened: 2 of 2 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (lint-gate semantics) → reviewer's-own-prescription fix, confirmed PASS → queuing for gate.**

Context: #761 merged before its review MINOR could fold in (the recurring race — #759's mechanical fix will stop this class once merged), so the fix landed here, cherry-picked onto main (2-file delta, verified).

**Reviewer confirmation at `4d6a2d6b`:** the exact local-main repro that slipped through #761 (origin/main missing + shadowing local branch) now **fails closed** with #747's misconfiguration error, exit 1 — the invariant is unconditional for branch-name bases. SHA passthrough (the #748 intent) unaffected; the residual edge (a branch named as 7-40 pure hex chars, with origin/<name> missing, on a checkout that has local branches) is narrower than before and unreachable in the current detached CI checkout. The intentional narrowing (bare tag names no longer pass verbatim) regresses no real path — `github.base_ref` is always a branch and qualified tags arrive as `refs/tags/…`.

**My verification** at head: content-lint **107/107**, typecheck exit 0.

thomgabriel:
## GATE: approved — and this closes a hole I missed on #761

I gated #761 and specifically checked "does fail-closed survive?" — but I only tested the case where **nothing** resolves. I never asked what *else* `resolvesRef(root, raw)` could match. A local branch shadowing the base name is exactly that, and it makes #747's fail-closed conditional on the checkout happening not to have one.

That's the same coverage failure I've now made three times tonight: verifying the instance rather than the set. Good catch.

The fix is right and minimally scoped. Restricting the verbatim passthrough to `/^[0-9a-f]{7,40}$/i` means a bare **branch** name can only ever resolve through `origin/`, so a shadowing local branch can't satisfy it and the run fail-closes with #747's error. Genuine SHA passthrough — the actual #748 behavior — is untouched, and at most two candidates are still tried.

Worth naming the residual: a branch literally named like an abbreviated SHA (`abcdef1`) would take the SHA path. It only matters if such a branch exists *and* `origin/abcdef1` doesn't resolve, and the outcome is a diff against a real commit of that name rather than an error. Vanishingly rare and strictly better than the shadowing case this closes — not worth further narrowing.

The test pairing is the right shape again: a local `release` branch with no `origin/release` returns `origin/release` (never the local), **and** an end-to-end case proving that same shadowed base still fails closed with #747's message. Testing the fix and the invariant it could have broken, together.

CI green (Frontend 3m19s, SHA-pin guard, Dependency audit). The 7s `claude-review` failure is **#765** — the `labeled`/`track_progress` regression from my bad #759 merge, not this PR; the real review from the `opened` trigger posted and found no blocking issues.

PR #768, opened 2026-07-27
fix(db): positive xp_transactions.source at the award_xp call sites (#736)

### LINKED ISSUE #736: db: xp_transactions.source is reverse-derived from memo prefixes — write it positively at the call sites
Found by the adversarial review of #734 (cohort leagues). The league score's anti-gaming allowlist filters on `xp_transactions.source` — but `source` is not written positively by each XP path; it is REVERSE-DERIVED from the free-text reason prefix by `xp_source_for_reason` (`20260726120000_add_xp_transactions_source.sql:73-83`), whose `ELSE → 'platform'` is the only thing excluding `reward_xp`.

Consequence: any award whose memo BEGINS with an eligible prefix ('Completed lesson:', 'daily_quest:', 'Course completion bonus:', 'Completed course:', 'Achievement reward:') classifies as eligible — including an operator `reward_xp` promo deliberately shaped that way. The on-chain reward instruction takes an arbitrary authority-supplied memo.

**Why not urgent:** memos are authority-supplied only (no ordinary-user path can shape one — user-reachable XP sits behind graded/on-chain gates); leagues are display-only; award_xp's 5000/day cap bounds volume. #734's migration header has been corrected to state this true (weaker) guarantee.

**Fix shape:** make `source` a positive fact, not an inference — each writer (`award_xp` callers / webhook handlers / reward paths) passes its source explicitly; `xp_source_for_reason` becomes a backfill-only helper. NOTE: this touches the XP writers — after the #731 incident, that change wants its own adversarial review, byte-diff discipline against the applied prod bodies, and an apply-before-merge plan.

Interim operator rule (documented in #734): never author a promo/reward memo starting with an eligible-shaped prefix.

thomgabriel:
Closes #736. Makes `xp_transactions.source` — the column the cohort-league anti-gaming allowlist filters on — a **positive fact** each writer states, instead of a value reverse-derived from the free-text memo prefix.

**This PR touches the XP writers — please give it its own adversarial review** (per the routing note). No live-DB action.

## The vulnerability
`source` was set by `xp_source_for_reason(reason)`, whose `ELSE → 'platform'` is the only thing excluding an arbitrary award. The on-chain `XpRewarded` instruction takes an **authority-supplied memo**; a promo shaped as `"Completed lesson: …"` derives to `'lesson'` → scores as league-eligible. (Not urgent: memos are authority-only, leagues are display-only, the 5000/day cap bounds volume — but it should be positive.)

## The fix
- **`award_xp` gains `p_source TEXT DEFAULT NULL`**; `v_source := COALESCE(p_source, xp_source_for_reason(p_reason))`. Explicit source wins; the derivation is the NULL fallback (un-migrated callers unaffected, backfill still shares one mapping). `xp_source_for_reason` demoted to the default/backfill helper. Signature 5→6 args, so the migration **DROPs the 5-arg then CREATEs the 6-arg** (award_xp has no SQL dependents). Byte-mirrored into `schema.sql`.
- **Call sites pass their true source** (all 8): lesson / course_completion / creator_reward / achievement / quest — and, the **core fix**, the `XpRewarded` arbitrary-memo path passes **`'platform'`** (never league-eligible), so a memo can't launder into eligibility. Threaded a `source` arg through `onchain-queue`'s `creditXpAndSettle` (its 3 callers). `award_community_xp` already writes `'community'` positively — unchanged.

## ⚠ Hard apply-ordering dependency (the #750 discipline)
`award_xp`'s body calls `next_streak_value()` (added by `20260726190000_streak_forgiveness`). Prod has 190000's **schema half** applied but its **function bodies are not** (blocked on #750/#758, see `docs/DB-MIGRATION-LEDGER.md`). **This migration MUST apply after 190000's bodies** — applying it earlier makes `award_xp` reference an undefined function and every award errors. Documented prominently in the migration header and pinned by the guard test. Sequence: land #758 → apply 190000 bodies → apply this. The **self-contained rollback** inlines the prior 5-arg body verbatim (no cross-file re-apply); snapshot `pg_get_functiondef` before applying.

## Verify
- `pnpm --filter web test` — 210 files / 1912 tests pass (new guard test: source is positive-with-fallback; migration DROP-then-CREATE; hard-dep documented; self-contained rollback; grants re-hardened)

### DIFF
--- a/apps/web/src/app/api/admin/resync/route.ts
+++ b/apps/web/src/app/api/admin/resync/route.ts
@@ -257,6 +257,7 @@ export async function POST(req: NextRequest) {
           p_reason: `Achievement reward: ${achievement.id}`,
           p_idempotency_key: `resync:${assetAddr}:xp`,
           p_tx_signature: `resync:${assetAddr}`,
+          p_source: "achievement", // #736
         });
       }
 
--- a/apps/web/src/lib/gamification/__tests__/surprise-bonus.test.ts
+++ b/apps/web/src/lib/gamification/__tests__/surprise-bonus.test.ts
@@ -118,6 +118,8 @@ describe("maybeAwardSurpriseBonus", () => {
       p_reason: `${SURPRISE_BONUS_REASON_PREFIX}lesson-pdas-1`,
       p_idempotency_key: "sig-abc:SurpriseBonus",
       p_tx_signature: "sig-abc",
+      // #736: source stated positively (a surprise bonus is a platform grant).
+      p_source: "platform",
     });
   });
 
--- a/apps/web/src/lib/gamification/surprise-bonus.ts
+++ b/apps/web/src/lib/gamification/surprise-bonus.ts
@@ -105,6 +105,10 @@ export async function maybeAwardSurpriseBonus(
     p_reason: surpriseBonusReason(lessonId),
     p_idempotency_key: `${txSignature}:SurpriseBonus`,
     p_tx_signature: txSignature,
+    // #736: a surprise bonus is a platform grant, not learning XP — pin the
+    // source (its 'surprise_bonus:' reason already derived to 'platform', but
+    // state it positively so it can never drift).
+    p_source: "platform",
   });
 
   if (error) throw new Error(error.message);
--- a/apps/web/src/lib/helius/__tests__/award-xp-source.test.ts
+++ b/apps/web/src/lib/helius/__tests__/award-xp-source.test.ts
@@ -1,0 +1,188 @@
+/* eslint-disable import/order -- vi.mock('server-only') must be hoisted above
+   the module imports so the `server-only` graph loads under vitest. */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { Keypair } from "@solana/web3.js";
+
+// #736 (adversarial-review MINOR): the SECURITY-critical part of #736 is the
+// EXPLICIT p_source each award_xp call site passes — especially handleXpRewarded,
+// which pins 'platform' so an arbitrary authority memo can't reverse-derive into
+// a league-eligible source. That value is convention, not structure, so it must
+// be pinned by a test: a future edit (or the reviewer's mutation flipping
+// 'platform' → 'lesson') must fail CI here.
+
+vi.mock("server-only", () => ({}));
+
+const { resolveUserId, resolveCourseId, resolveLessonId, awardXpCalls, rpc } =
+  vi.hoisted(() => {
+    const awardXpCalls: Array<Record<string, unknown>> = [];
+    return {
+      resolveUserId: vi.fn<(addr: string) => Promise<string | null>>(),
+      resolveCourseId: vi.fn<() => Promise<string | null>>(),
+      resolveLessonId: vi.fn<() => Promise<string | null>>(),
+      awardXpCalls,
+      rpc: vi.fn((fn: string, params: Record<string, unknown>) => {
+        if (fn === "award_xp") awardXpCalls.push(params);
+        return Promise.resolve({ data: 10, error: null });
+      }),
+    };
+  });
+
+// A chainable admin-client stub: every query-builder method returns the same
+// chain, and awaiting it (or .single()) resolves a benign empty result — so the
+// non-target work in each handler (progress upsert, enrollment update, credential
+// / achievement / finalize side-paths) runs without throwing, and only the
+// award_xp rpc calls are captured.
+vi.mock("@/lib/supabase/admin", () => {
+  const chain: Record<string, unknown> = {};
+  const result = { data: null, error: null };
+  for (const m of [
+    "select",
+    "insert",
+    "upsert",
+    "update",
+    "delete",
+    "eq",
+    "in",
+    "order",
+    "limit",
+  ]) {
+    chain[m] = () => chain;
+  }
+  chain.single = () => Promise.resolve(result);
+  chain.maybeSingle = () => Promise.resolve(result);
+  chain.then = (onF: (v: typeof result) => unknown) =>
+    Promise.resolve(result).then(onF);
+  return {
+    createAdminClient: () => ({ from: () => chain, rpc }),
+  };
+});
+
+// Non-target collaborators → inert (they must not fire their own award_xp).
+vi.mock("@/lib/helius/resolvers", () => ({
+  resolveUserId,
+  resolveCourseId,
+  resolveLessonId,
+}));
+vi.mock("@/lib/gamification/surprise-bonus", () => ({
+  maybeAwardSurpriseBonus: vi.fn(async () => undefined),
+}));
+vi.mock("@/lib/solana/pda", () => ({ getProgramId: () => "program-id" }));
+vi.mock("@/lib/solana/academy-program", () => ({
+  getConnection: () => ({}),
+  finalizeCourse: vi.fn(),
+  issueCredential: vi.fn(),
+  awardAchievement: vi.fn(),
+}));
+vi.mock("@/lib/solana/academy-reads", () => ({
+  fetchEnrollment: vi.fn(async () => null),
+  fetchCourse: vi.fn(async () => null),
+}));
+vi.mock("@/lib/solana/bitmap", () => ({ isCourseComplete: () => false }));
+vi.mock("@/lib/content/deployments", () => ({
+  isCourseInMaintenance: vi.fn(async () => false),
+}));
+vi.mock("@/lib/platform/freeze", () => ({
+  isPlatformFrozen: vi.fn(async () => false),
+}));
+vi.mock("@/lib/content/queries", () => ({
+  getCourseById: vi.fn(async () => null),
+  getDeployedAchievements: vi.fn(async () => []),
+}));
+vi.mock("@/lib/gamification/achievements", () => ({
+  checkNewAchievements: vi.fn(() => []),
+  buildUserState: vi.fn(),
+}));
+vi.mock("@/lib/solana/arweave", () => ({ uploadCertificateMetadata: vi.fn() }));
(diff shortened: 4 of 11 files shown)


thomgabriel:
## GATE: hold — merging this before the migration chain breaks every XP award

The fix itself is right, and the vulnerability is real: `source` reverse-derived from an authority-supplied memo means a promo shaped `"Completed lesson: …"` launders into league eligibility. Stating it positively at each writer is the correct answer, and routing `XpRewarded`'s arbitrary-memo path to `'platform'` is the core of it.

You correctly flagged the apply-ordering dependency on 190000's bodies. But there's a second consequence of the same fact that the description doesn't draw, and it's the blocking one.

## The app code ships before the function exists

The call sites now pass `p_source` in the RPC payload:

```ts
await admin.rpc("award_xp", { p_user_id, p_amount, p_reason,
                              p_idempotency_key, p_tx_signature,
                              p_source: "lesson" });
```

PostgREST resolves an RPC by the **exact set of named arguments**. Prod today:

```
award_xp(uuid,integer,text,text,text)
p_user_id, p_amount, p_reason, p_idempotency_key DEFAULT NULL, p_tx_signature DEFAULT NULL
```

No `p_source`. A six-named-arg call matches no overload and 404s (*"Could not find the function public.award_xp(...) in the schema cache"*). Merge auto-deploys (#731), so from that moment **every XP award fails** — lesson completions, course-completion bonuses, creator rewards, achievements, quests. The `p_source DEFAULT NULL` in the new definition protects *un-migrated callers*; it does nothing for a migrated caller hitting an un-migrated database.

## And the migration can't simply be applied first

The usual fix — apply the migration, then merge — is unavailable here, because this migration's `award_xp` body calls `next_streak_value()`, which **does not exist in prod**:

```
next_streak_value_exists: 0
```

It arrives with `20260726190000_streak_forgiveness`'s function bodies, which are unapplied and gated on the owner's call in #758's header warning (applying 190000 on this pre-170000 database also activates the review-quest branch). So the real chain is:

1. Owner decides on the review-quest activation (#758 header)
2. Apply 190000's function bodies
3. Apply `20260727120000_xp_source_positive_writes`
4. **Then** merge this PR

Steps 1-3 are all upstream of the merge. Nothing in this PR is wrong — it's just not mergeable until a decision that isn't the gate's has been made.

## Verified

- `xp_source_for_reason` retained as the NULL fallback, so un-migrated callers and the backfill keep one shared mapping. Correct.
- DROP-then-CREATE justified: 5→6 args changes the signature and `award_xp` has no SQL dependents (confirmed — nothing else references it in `schema.sql`).
- The self-contained rollback inlining the prior 5-arg body verbatim is the #758 lesson applied without being asked. Good.
- `award_community_xp` already wrote `'community'` positively and is untouched.

## What I'd suggest

Either park this behind the owner decision, or split it: the `'platform'` routing for the arbitrary-memo path is the security-relevant half and could ship independently **if** it can be expressed without changing `award_xp`'s signature (e.g. having that caller pass a reason the existing derivation already maps to `'platform'`). That would close the laundering hole now and leave the positive-source refactor for after the migration chain clears.

Re-request when the chain is applied, or if you want to explore the split.

thomgabriel:
**Orchestrator routing record — MOST SENSITIVE of the batch (rewrites award_xp) → adversarial review PASS + MINOR fixed + mutation-verified → queuing for gate.**

**My independent verification** at `b0983fc2`: typecheck exit 0; full suite **211 files / 1916 tests, exit 0** (1912 at the pre-fix head).

**Adversarial review (the #727 XP-writer attacker, at `bfb3360d`): PASS.**
- **Behavior preservation EXACT**: normalized diff of the new 6-arg body vs the correct **190000 baseline** (not main's stale pre-#727 mirror) = exactly two changes: the `p_source` param + `v_source := COALESCE(p_source, xp_source_for_reason(p_reason))`. Cap math, advisory lock, idempotency, streak delegation, level calc all byte-identical. **Rollback's inlined 5-arg body token-identical to the 190000 baseline** — the #750 lesson honored.
- **All 8 call-site sources verified correct** per the #630 taxonomy, incl. the core fix — `handleXpRewarded` (arbitrary authority memo) → `'platform'`, never league-eligible: the #734-review laundering vector is closed. Batch/test-out rides the webhook 'lesson' path; grep proved no missed direct caller.
- DROP-then-CREATE cascade-free (no SQL dependents; TS `.rpc` named params make the DEFAULT 6th arg back-compat).
- **The one MINOR was the right catch**: the reviewer mutation-flipped the sweep source to 'lesson' and the suite stayed green — the security-critical fix was call-site-convention only. **Fixed at `b0983fc2`**: all 5 event-handler + 3 queue sources now structurally pinned by tests; the implementer re-ran the mutation and the new test fails as required.

**⚠️ For the gate — HARD apply-ordering (third entry in the award_xp chain, guard-pinned + header-documented):** this body calls `next_streak_value`, so it MUST apply AFTER 190000's function bodies. Sequence: **#758 (merged) → apply 190000 bodies (your pending review-quest-branch decision) → apply this.** Applying out of order errors every award. Snapshot `pg_get_functiondef` before applying per the header.

DEFAULT-NULL fallback is the documented backfill/compat seam — after this PR no production path hits the derivation; the call-site tests are the structural enforcement.

thomgabriel:
## Hold stands — but I've de-risked the decision it's waiting on

The `b0983fc2` update is a genuine improvement (mutation-tested call-site pinning — the reviewer flipping the sweep source to `'lesson'` and the suite staying green was a real gap). But it doesn't change the merge hazard: the diff still only adds tests, so the deployed code will still call a six-named-arg `award_xp` against a prod function that has five. That 404s the moment Vercel deploys.

To be precise about the distinction, since the follow-up comment reads it as apply-ordering: **there are two orderings here, not one.**

1. *Apply* ordering — this migration after 190000's bodies, because the body calls `next_streak_value`. You've documented and guard-pinned that. Agreed.
2. *Merge* ordering — this PR after the migration is **applied**, because the shipped client calls an overload prod doesn't have yet. This one isn't addressed, and it's the one that breaks prod on merge rather than on apply.

## The decision blocking all of it is a verified no-op

The whole chain waits on the owner's call from #758's header: applying 190000 on this pre-170000 database also activates the review-quest branch in `get_daily_quest_state`. I checked what that actually does.

`get_daily_quest_state(uuid, jsonb, text[], jsonb)` receives its quest definitions **as a parameter**. `api/quests/daily/route.ts:46` builds `questDefs` from `questData.quests` — i.e. the committed content bundle — and passes it as `p_quest_definitions`. The review branch can therefore only fire for a quest whose `type` is `review`.

The bundle has five quests, and none is one:

```
quest-challenge         challenge
quest-complete-lesson   lesson
quest-complete-module   module
quest-lesson-batch      lesson_batch
quest-login-streak      login_streak
```

So activating that branch adds a code path with **no data to act on**. It becomes live only when a `review` quest YAML is authored in courses-academy *and* activated by a content.lock bump — both visible, gated events.

**Net:** applying 190000's bodies delivers streak forgiveness (the intended feature, with a rollback #758 made a real inverse and I verified md5-identical to prod) plus a dormant branch. The "two-migration jump" is real but inert.

That reduces the owner call to: *ship streak forgiveness now, yes or no?* — not *what else am I silently turning on?*

Once that lands: apply 190000 bodies → apply `20260727120000` → then this merges. I'll do the applies and re-gate.

thomgabriel:
## GATE: approved — the chain is applied, hold released

I cleared the blocking chain against prod before merging. Every step verified by md5 against the migration file, since hand-transcribing 449 lines of core XP functions is exactly where a silent corruption would hide.

### 1. `20260726190000_streak_forgiveness` function bodies — APPLIED

Applied in six verifiable chunks; every resulting `prosrc` md5 matches the file byte-for-byte:

| function | md5 | len |
|---|---|---|
| `cover_missed_days_with_freezes` | `876619b3…` ✓ | 2479 |
| `next_streak_value` | `56cf86e7…` ✓ | 1096 |
| `grant_streak_freeze` | `42277124…` ✓ | 273 |
| `award_xp` (5-arg, interim) | `970c213d…` ✓ | 2918 |
| `award_community_xp` | `e5b2f3ec…` ✓ | 2788 |
| `get_daily_quest_state` | `2f7894e0…` ✓ | 8045 |

Functional smoke on a random UUID — all four `next_streak_value` branches exact (first-ever→1, active-today→5, yesterday→6, gap-without-freeze→1) and **zero rows written**.

### 2. `20260727120000_xp_source_positive_writes` — APPLIED

```
award_xp(uuid,integer,text,text,text,text)
p_user_id, p_amount, p_reason, p_idempotency_key DEFAULT NULL,
p_tx_signature DEFAULT NULL, p_source text DEFAULT NULL
md5 7bdbb34d28b9e425fe9a9b5bf86384db  len 3300  overload_count 1
```

md5 matches the file. **overload_count 1** — the 5-arg is gone, so named-arg calls can't resolve ambiguously, which was the stated reason for DROP-then-CREATE. Smoke: both the five-arg-style and six-arg-with-source calls return 0 on a zero amount with no rows written.

### One incident worth recording

The first apply attempt returned **"the connector's server isn't responding"** — ambiguous, and `DROP FUNCTION` was statement one. I checked before retrying rather than assuming: `award_xp` was still the intact 5-arg at `970c213d…`, so the transaction had rolled back cleanly with no partial application. Retried successfully. Worth noting for anyone else applying a DROP-then-CREATE through this connector — a timeout is not evidence of a no-op, but it is checkable in one query.

### On the decision that was blocking this

The review-quest activation is confirmed inert, as I posted earlier: `get_daily_quest_state`'s new `review` branch (now live at lines 556-561) fires only for a quest of `type = 'review'`, and quest definitions arrive as a parameter built from the content bundle — which has five quests, none of that type. Dormant until a review quest is authored and a content.lock bump activates it.

Merge order is now satisfied: migration applied → client can safely call the 6-arg overload. Merging.

PR #774, opened 2026-07-27
fix(review): drop the labeled trigger — track_progress hard-fails on it (#765)

### LINKED ISSUE #765: REGRESSION: #759's  trigger hard-fails claude-review (track_progress unsupported for that event)

thomgabriel:
Closes #765 — my #759 regression, urgent per the gate (the `labeled` run hard-fails claude-review in ~8s on every needs-gate labeling).

## Why option 1 (drop `labeled`) and not the conditional

The issue's option 2 (conditional `track_progress`) is a dead end: the workflow's own comment documents that WITHOUT track_progress, agent mode runs the review but **never publishes it** — so a labeled-triggered run minus track_progress is a silent no-op, worse than nothing. The other alternative (a labeled-triggered job posting `@claude review`) also fails structurally: **GITHUB_TOKEN-created events never trigger workflows**, so the bot comment would sit inert; making it work needs a PAT secret (owner decision, not worth it).

## What this does

- `types` back to `[opened, ready_for_review, reopened, synchronize]`; the `labeled` arm removed from the job `if`.
- **The `synchronize` half — the actual #710 race fix — is untouched and working** (per #765's own analysis), as is #764's job-level concurrency.
- Queue-time fresh review is codified as the queuing party's job: comment `@claude review` when adding `needs-gate` — which is already the loop's standing practice (I do it on every queue). Documented in the workflow comment with the full why, so `labeled` doesn't get re-added by someone 'completing' the trigger set.

Verified: yaml parses, SHA-pin guard green. Not locally testable beyond that (workflow runtime behavior); the proof will be the next needs-gate labeling producing NO red run.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -3,22 +3,29 @@ name: Claude Code Review
 on:
   pull_request:
     # Every PR is reviewed on open / ready / reopen. Additionally — and ONLY for PRs
-    # carrying `needs-gate` — re-review on every push (`synchronize`) and when the
-    # label is first applied (`labeled`). This closes the #710 race: the orchestrator
+    # carrying `needs-gate` — re-review on every push (`synchronize`). This closes
+    # the #710 race: the orchestrator
     # pushes hardening commits after the review has started, so the review reported on
     # a head that no longer existed (and 3 of 5 stale reviews missed the security-
     # relevant commit). Scoping the push re-review to needs-gate keeps the OAuth-token
     # cost off every ordinary PR push; the `if:` below enforces the scope, and the
     # concurrency group cancels an in-flight review the moment a newer head arrives so
     # a stale verdict can never post.
-    types: [opened, ready_for_review, reopened, synchronize, labeled]
+    types: [opened, ready_for_review, reopened, synchronize]
 
 jobs:
   claude-review:
-    # opened/ready/reopened review any PR (unchanged). synchronize + labeled fire for
-    # every PR but are admitted here only for needs-gate, so a fresh-head review is
-    # guaranteed both when the PR is queued (label added) and on each subsequent push,
-    # without re-reviewing every push of every PR.
+    # opened/ready/reopened review any PR (unchanged). synchronize fires for every
+    # PR but is admitted here only for needs-gate, so a fresh-head review is
+    # guaranteed on each subsequent push without re-reviewing every push of every PR.
+    #
+    # `labeled` is deliberately NOT a trigger (#765): the review action's
+    # track_progress publish mechanism only supports opened/synchronize/
+    # ready_for_review/reopened, so a labeled-triggered run hard-fails before
+    # reviewing — and a bot-posted `@claude` comment can't substitute because
+    # GITHUB_TOKEN-created events never trigger workflows. Queue-time fresh
+    # reviews are therefore the queuing party's job: comment `@claude review`
+    # when adding needs-gate (existing loop practice).
     #
     # No draft-PR guard is intentional: a needs-gate draft's pushes still review, which
     # is what we want — a PR is queued for the gate precisely to be reviewed at head.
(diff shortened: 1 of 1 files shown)


thomgabriel:
@claude review

thomgabriel:
## GATE: approved — and this PR is its own proof

Verified the diff line by line: `labeled` removed from both `types` and the `if`, the `synchronize` arm (the actual #710 race fix) byte-identical, #764's job-level concurrency untouched, and the tombstone comment explains *why* `labeled` is absent so a future 'completing the trigger set' PR doesn't reintroduce it. Minimal and correct.

Your option-analysis is right on both structural facts, and I checked them rather than nodded: the workflow's own comment documents that agent-mode without `track_progress` runs but never publishes (so option 2 silently no-ops — worse than the red X it replaces), and GITHUB_TOKEN-created events genuinely never trigger workflows, so a bot-posted `@claude review` would sit inert without a PAT.

Coverage after the drop, confirmed against the `if`: `opened` admits **any** PR unconditionally, so the normal queue flow (open → label) already gets its open-time review; the only gap is label-long-after-open, which the documented `@claude review` practice covers — and which was *never* working via `labeled` anyway, since that trigger hard-failed from the day it shipped.

**Live proof on this very PR:** a review posted here with no 8-second red run. The next `needs-gate` labeling producing no failure will close the loop; I'll watch for it.

This closes out the last damage from my #759 merge — both regressions (#764's silent-cancel and #765's hard-fail) are now fixed by follow-ups that were themselves properly reviewed before merging. The irony is fully priced in.

CI: SHA-pin guard green; Frontend/On-chain skipping is correct for a workflows-only diff (same as #759/#764).

PR #779, opened 2026-07-27
feat(email): Resend integration + marketing-consent model (#769)

### LINKED ISSUE #769: Integrate email provider (Resend) for platform news + new-course announcements
## Goal

Send platform news/announcement emails to users — e.g. "a new course is available," milestones, product updates — via a transactional/marketing email provider (Resend preferred; SendGrid/Postmark as alternatives).

## Current state

- **No email provider** is integrated today (the `resend` references in the repo are Solana tx-rebroadcast, unrelated).
- **No marketing-consent** data model — `supabase/schema.sql` has no opt-in/subscription column.
- User emails exist only for **Google OAuth** users (in `auth.users`); **wallet-only (SIWS)** users have no email on file.

### 1. Provider + templating
- Integrate **Resend** (`resend` SDK) with **React Email** (`@react-email/components`) for typed, previewable templates.
- Server-only API key via `serverEnv` (`RESEND_API_KEY`); verify a sending domain (e.g. `news@st.academy`) — DKIM/SPF/DMARC DNS records.

### 2. Consent + data model (required before sending marketing mail)
- New column/table for **marketing opt-in** (e.g. `profiles.marketing_opt_in boolean default false` + `marketing_consent_at timestamptz`), RLS own-row.
- **Double opt-in** flow + one-click **unsubscribe** (token-based route + `List-Unsubscribe` header). Non-negotiable for LGPD/GDPR/CAN-SPAM (team is Brazil-based → LGPD applies).
- Preferences UI in **Settings** (`apps/web/src/app/[locale]/(platform)/settings`) to subscribe/unsubscribe per category (product news, new courses, digest).
- Email-capture prompt for wallet-only users (optional, so they can opt in).

### 3. Triggers / audiences
- **New course available**: hook off the admin course publish/sync flow or the on-chain `CourseCreated` event (`apps/web/src/lib/helius/event-handlers.ts`) → enqueue a campaign to opted-in users.
- **Announcements/digests**: admin-triggered send from the admin panel, or a scheduled cron.
- Batch sends via a queue (reuse `pending_onchain_actions`-style pattern or a new table) to respect provider rate limits + retries; idempotency keys to avoid double-sends.

### 4. Deliverability + ops
- Suppression list (unsubscribes, hard bounces, complaints) honored on every send.
- Webhook from provider for bounce/complaint handling.
- Rate limiting + send logging for auditing.

## Open questions / decisions

- Resend vs SendGrid vs Postmark (volume, pricing, deliverability).
- Marketing vs transactional separation (separate domains/subdomains to protect sender reputation).
- Where campaigns are authored: hardcoded templates vs admin-panel composer.
- Whether to gather emails from wallet-only users at all, or keep this Google-users-only initially.

## Out of scope (for now)

- Full ESP/CRM (Mailchimp-style) segmentation.
- SMS/push notifications.

## References
- `supabase/schema.sql` (needs consent column/table)
- `apps/web/src/lib/helius/event-handlers.ts` (course-created event source for triggers)
- `apps/web/src/app/[locale]/(platform)/settings` (preferences UI)
- `apps/web/src/lib/env.server.ts` (add `RESEND_API_KEY`)

thomgabriel:
Closes #769.

## What

The consent-first foundation for platform-news / new-course announcement email via **Resend**. There was no provider and no consent model; marketing mail can't be sent without both. This ships them plus the admin trigger, unsubscribe surface, and settings toggle. **No automatic sends** in this PR.

## 1. Consent model (migration `20260727150000_email_subscriptions.sql` + schema.sql mirror + guard)

A dedicated `email_subscriptions` table — kept OFF the sensitive `profiles`/`public_profiles` surface so neither the opt-in flag nor the secret unsubscribe token can leak through the public view.

- **Opt-OUT by default** (`opt_in BOOLEAN NOT NULL DEFAULT false`) — never email without an explicit opt-in.
- **RLS**: own-row `SELECT` only (the settings toggle reads your own state); **no write policy**. Writes go through SECURITY DEFINER RPCs so consent timestamps are server-authoritative (a client can't backdate consent):
  - `set_marketing_opt_in(bool)` — self-service, `GRANT authenticated`, keyed on `auth.uid()` (can only touch your own row).
  - `list_marketing_recipients()` — `service_role` only; **the single consent gate** the pipeline reads: opted-in users **with an email on file** (wallet-only SIWS users have none → never emailed; reads `auth.users`).
  - `unsubscribe_by_token(uuid)` — `service_role`; the sessionless one-click path.
- Every RPC: SECURITY DEFINER, `search_path=''`, REVOKEd from PUBLIC/anon (and authenticated for the service_role ones). Idempotent, BEGIN/COMMIT, tested rollback.

## 2. Send module (`lib/email/`, plain fetch — no SDK dependency)

**FAIL-CLOSED**: with no `RESEND_API_KEY` the module still loads and `sendEmailBatch()` returns a typed `{ ok:false, reason:'unconfigured' }` — it never throws at import or call. Batches to Resend's 100/request limit with a per-course/per-chunk `Idempotency-Key` (retries never double-send) and an inter-batch delay (rate-limit friendly). Every template embeds an unsubscribe link, and every send sets **`List-Unsubscribe` + `List-Unsubscribe-Post`** (RFC 8058 one-click). Verified the Resend `/emails/batch` shape + per-email `headers` support against current docs.

## 3. Admin trigger + unsubscribe

- `POST /api/admin/email/announce-course` (`requireAdminAuth`) — the **only** send trigger. Resolves the course, hands to the pipeline. Fail-closed passthrough.
- `/api/email/unsubscribe` — `GET` (human link → confirmation page) + `POST` (one-click), both by unguessable per-user token, idempotent, sessionless via the RPC.

## 4. Settings UI

Marketing opt-in toggle in the **privacy tab** (own-row read + `set_marketing_opt_in` RPC, optimistic + rollback), i18n en/es/pt-BR, accessible `role="switch"`.

## Env the owner must set

| Var | Required? | Purpose |
|---|---|---|
| `RESEND_API_KEY` | to send | Resend key. **Unset ⇒ fail-closed**, nothing sends. Set once a Resend sending domain (DKIM/SPF/DMARC) is verified. |
| `EMAIL_FROM` | optional | Verified sender, e.g. `Superteam Academy <news@st.academy>`. Falls back to a documented default. |
| `NEXT_PUBLIC_APP_URL` | to send | Absolute base for unsubscribe + course links (already used elsewhere). The admin route 500s without it. |

## Consent decision (LGPD/GDPR/CAN-SPAM)

Consent is captured via the **authenticated** settings toggle, which is a logged consent action stamped `consent_at` server-side, plus one-click unsubscribe + `List-Unsubscribe`. I treated the authenticated toggle as sufficient explicit consent and **deferred double-opt-in email confirmation** (documented) — it matters mainly for *un*-authenticated email capture, which is out of scope here (Google-users-only for now). Also deferred + documented: bounce/complaint webhook + suppression list, the `CourseCreated` auto-trigger, per-locale email templates, and wallet-user email capture.

## Verification

- typecheck: 6/6 packages
- **web suite: 219 files, 1992 tests pass**

New tests: fail-closed unconfigured (send module + pipeline); consent-gating (recipients come ONLY from the RPC); `List-Unsubscribe` presence + token link; idempotency + chunking + failed-batch counting; admin-route auth (401/400/404/500/dispatch); unsubscribe GET/POST token flow; template unsubscribe presence + HTML escaping; migration RLS/RPC guard (both migration + schema.sql).

Migration + consent = adversarial review + gate. Do not merge — for review.

## Retry precision (review fold)

`list_marketing_recipients()` now has `ORDER BY es.user_id` and the campaign keys each batch's idempotency on a **sha256 of the chunk's member ids** (namespaced by course), not the array index. I chose membership-hash keying over index keying deliberately: an index key would let a partial-send retry after a mid-campaign unsubscribe put different recipients in "chunk N" under the same cached key and **silently skip** the shifted-in opt-in learners. With membership keying, an unchanged chunk keys identically (cached ⇒ never double-sent) and only a chunk whose membership actually shifted gets a new key (re-sends ⇒ never skipped). The residual is a rare, benign **duplicate** to members who shifted position within a changed chunk — the right trade for marketing mail, where a missed opt-in learner is worse than a duplicate.

---

## Revival round (2026-07-27, owner released David's items to the loop) — head 503d03b

**Rebase + prerequisites:** clean rebase onto main (18 commits behind, zero conflicts); migration renumbered `20260727150000` → `20260728120000` (strictly greater than every migration incl. the applied `20260727140000`; uniqueness grep-verified); **synthetic-address exclusion** added to `list_marketing_recipients()` in both the migration and the schema.sql mirror: `AND u.email NOT LIKE '%@wallet.superteam-lms.local'` — closes a real gap where SIWS wallet users' placeholder addresses passed the non-empty email gate and would have been emailed.

**Independent adversarial review (area:db protocol) — round 1 verdict FAIL, one HIGH:** `unsubscribe_by_token` declared `v_matched BOOLEAN` while using it as `GET DIAGNOSTICS … ROW_COUNT`, so `RETURN v_matched > 0` raised `operator does not exist: boolean > integer` on EVERY invocation (plpgsql defers body parsing, so CREATE FUNCTION and CI stayed green). Impact: GET unsubscribe 404'd for valid tokens; RFC-8058 one-click POST reported ok:true to mail providers while `opt_in` never flipped — a CAN-SPAM/LGPD failure invisible to string-matching tests.

**Fix + regression armor:** `v_matched INTEGER` in BOTH SQL files, plus a REAL SQL-execution test (`email-subscriptions-execution.test.ts`, @electric-sql/pglite devDep — in-process real Postgres, no CI infra): executes the actual migration from disk over an honest reference-only stub, invokes the RPCs for real (valid token → true + flip + timestamp; bogus → false + zero rows against 2 seeded subscribers; idempotent re-call), runs the same assertions against the schema.sql mirror copy, and empirically verifies `list_marketing_recipients` excludes the wallet placeholder and opt-outs.

**Red-proof (rule 2), triple-verified:** execution test at **4eae13f** (pre-fix): **6 failed | 1 passed** with the verbatim pg error — replayed independently by the orchestrator; the adversarial reviewer additionally re-broke the fixed body in postgres:15 (`BOOLEAN` revert → same error) proving the fix is load-bearing. **Adversarial round-2 verdict: PASS** (fix minimal + correct in both copies, no scope creep — diff is exactly 5 files, no new attack surface from the test dep; noted non-blocking: the execution test runs as superuser so RLS/grant enforcement remains covered by the string-guard test + static review, which passed).

**Verified at 503d03b (agent + orchestrator re-runs in clean clones):** typecheck 6/6; apps/web **224 files / 2034 tests**; content-schema 170 / content-lint 107 / deploy 18; lint clean; RLS-guard 16/16.


thomgabriel:
## GATE: hold — one mechanical blocker (timestamp collision with #777), everything else verified clean

### 🔴 The blocker

`supabase/migrations/20260727140000_email_subscriptions.sql` collides with **#777**'s `20260727140000_stranded_instance_state_cleanup.sql` — both open PRs claim the same timestamp. This is exactly the #735 duplicate-prefix trap, this time across two in-flight branches (each checked the *dir*, neither could see the other). #777 is the P0 and re-pushed 40 minutes ago still carrying 140000, so **this PR should renumber to `20260727150000`** (and whoever merges second re-checks). One `git mv` + the guard-test constant.

Suggested convention addition for the ledger doc: timestamp uniqueness is checked against the migrations dir **and `gh pr list` diffs**, since the dir can't show in-flight siblings.

### Verified clean (no other changes needed)

- **Consent model**: `opt_in DEFAULT false` (opt-out by default), `consent_at` stamped server-side in the RPC keyed on `auth.uid()` — a client cannot backdate or touch another row. Own-row SELECT only, no write policy, all RPCs SECURITY DEFINER + `search_path=''` + REVOKE/GRANT per role. Matches the house pattern exactly.
- **Wallet-only users can never be emailed** — `list_marketing_recipients()` requires an email from `auth.users`, and SIWS synthetic addresses (`@wallet.superteam-lms.local`) — flag that: **the recipient RPC should also exclude the synthetic `%@wallet.superteam-lms.local` addresses explicitly**, since those ARE emails on file in auth.users. If it already does, point me at the line; if not, that's a one-line WHERE addition worth folding into the renumber push. (Not re-blocking separately — same push.)
- **Unsubscribe**: unguessable `gen_random_uuid()` token, sessionless, idempotent (re-use returns matched), POST always-200 for RFC 8058 providers, GET validates UUID shape before touching the DB. The 200/404 distinction on GET is a token-validity oracle but tokens are 122-bit random — not enumerable, acceptable.
- **Fail-closed send**: `RESEND_API_KEY` unset ⇒ typed `unconfigured` result, never a throw; batch idempotency keys prevent retry double-sends; `List-Unsubscribe` + `-Post` on every send.
- **No committed secrets** (grepped), env additions server-only, admin trigger behind `requireAdminAuth`, **no automatic sends** anywhere in this PR.
- Deferred list (double-opt-in, bounce webhook, auto-trigger) is documented rather than silently absent — right call for Google-auth-only capture.

Renumber (+ the synthetic-address check), re-request, and this merges.

thomgabriel:
## Parked by owner decision — not a quality rejection

Owner call (2026-07-27): issues authored by David (Potolski) stay open and the loop does not work or merge against them — #769 is his. This PR is therefore parked regardless of the review outcome.

For the record, the gate review found it close to mergeable: one mechanical blocker (timestamp collision with #777 — renumber to 150000) and one one-line flag (`list_marketing_recipients()` should explicitly exclude the synthetic `%@wallet.superteam-lms.local` SIWS addresses, which ARE emails in auth.users and would bounce). Everything else — consent model, RLS, fail-closed sends, no secrets — verified clean.

Branch `feat/resend-email-27-07-2026` is preserved; if David picks #769 up, this work plus the two fixes above is a short path to done. David's issues now carry the `on-hold-david` label.

thomgabriel:
Reopened — I over-read the owner instruction. The rule is: David's issues/PRs stay OPEN, the loop just doesn't merge them. This sits here for David with the gate findings above (renumber vs #777 + the synthetic-address WHERE) ready when he picks it up. Not in the loop's queue.

thomgabriel:
**Orchestrator routing record — SENSITIVE (migration + consent + auth.users reads) → adversarial review PASS + MINOR folded → queuing for gate.**

**My independent verification** at `7fb97891`: typecheck exit 0; full suite **219 files / 1993 tests, exit 0** (and 1992 at the reviewed head).

**Adversarial review (at `1e2e1908`): PASS.**
- **Consent gate un-bypassable**: `list_marketing_recipients` is the ONLY email-read path (grep-proven); filters opt_in=true + non-empty email; revocation effective on next read; all 3 RPCs house-hardened (DEFINER + search_path='' + complete REVOKE/GRANT) with a mutation-proof guard pinning every invariant in migration AND mirror.
- **Unsubscribe token**: UUIDv4 (un-enumerable), off the public_profiles surface, own-row + service_role only; the RFC 8058 POST path returns uniform 200 (no email-membership oracle); leaked-token blast = re-unsubscribing an unsubscribed user.
- **set_marketing_opt_in**: no user_id param exists — auth.uid()-keyed by construction; timestamps server-stamped.
- **Admin route**: the REAL requireAdminAuth (not a reimplementation); batch=100 + 600ms delay verified against Resend's published limits.
- **Fail-closed**: no key → typed unconfigured (admin sees status, nothing sends); missing APP_URL → 500 BEFORE any send (a broken unsubscribe link can never render); templates require unsubscribeUrl at the type level.

**MINOR folded at `7fb97891`** (the reviewer's retry-skip finding, taken to the stronger form): ORDER BY es.user_id for deterministic recipient ordering PLUS idempotency keyed on a **membership hash** (sha256 of the chunk's user_ids) instead of the array index — an unchanged chunk stays cached (never double-sent); a shifted chunk re-keys (never skips an opt-in learner); residual = rare benign duplicate on membership shift, the correct trade for consented marketing mail. Guard + determinism tests added.

**For the gate/owner before first real send:** set RESEND_API_KEY + EMAIL_FROM after verifying a sending domain (DKIM/SPF/DMARC); NEXT_PUBLIC_APP_URL required. Migration `20260727150000` SQL-only, applied nowhere; double-opt-in deliberately deferred (authenticated toggle + consent_at = the consent record) — flag if you want it before first send.

thomgabriel:
@claude review

thomgabriel:
@claude review

Issue #776, opened 2026-07-27
Deploy-flow audit: compile profile dominates learner build time; 7 ranked fixes (no re-architecture)


thomgabriel:
## Deploy-flow audit (13-agent workflow, all high findings adversarially verified) — verdict: **sound-with-gaps**

Owner asked whether the learner build+deploy is slow because we're doing it wrong. Answer: **the architecture is right** (server cargo-build-sbf + warm target cache, inline base64 .so, classic Loader-v3 buffer→write→deploy with correct hand-rolled instructions, priority fees, batch polling, crash-resume, rent reclaim, authority-verified capstone gate). The slowness is concentrated in two independently fixable places, plus one latent correctness bug. No re-architecture needed.

### 🥇 Root cause of "taking long": the compile profile
`apps/build-server/programs/Cargo.toml:10-14` ships `lto = true`, `codegen-units = 1`, `opt-level = "z"` — copied verbatim into every per-request build dir (`build.rs:167-169`). The warm target/ caches *dependencies*, but the **whole-program fat-LTO merge + single-threaded codegen of the leaf cdylib reruns on every single Run click**. The metrics histogram buckets (`metrics.rs:11-13`, le=30/60/120s) imply real builds land 30–120s — contradicting the stale "~10-15s" comment. (Tonight's CLI deploy of a 59KB .so took seconds — isolating the pain to the compile, not the chain.)

### Ranked recommendations

| # | change | effort | impact |
|---|---|---|---|
| 1 | **Relax the profile**: `lto = false`, `codegen-units = 16`, `opt-level = 1`, **keep `overflow-checks = true`** (it's load-bearing for the C2 curriculum). Re-warm the template target/ after (profile change invalidates it). Measure before/after via `academy_build_duration_seconds` on `/metrics`. | trivial | **high** |
| 2 | **Prod `NEXT_PUBLIC_SOLANA_RPC_URL` → Helius devnet with a separate, domain-restricted key.** The entire 110–280-tx write burst runs through this env (`deploy-panel.tsx:138` → `wallet-provider.tsx:20,42`); the example default is the keyless public endpoint → throttle → rebroadcast storm. **This is the same fix #467 already calls for** — one keyed, domain-restricted browser endpoint serves both. (The `api.devnet.solana.com` literal in `challenge-runner.tsx:334` is a sandbox mock — red herring.) | trivial (owner: Helius dashboard) | **high** |
[...]

PR #783, opened 2026-07-27
perf(build-proxy): raise build-program proxy timeout to cover Cloud Run's 300s (#776 item 5)

thomgabriel:
#776 PR C (item 5). Fixes the proxy↔server timeout mismatch: the Vercel `POST /api/build-program` route aborted the upstream build at **130s** while the build server (Cloud Run, `deploy/deploy.sh --timeout 300`) keeps compiling to **300s**. A cold fat-LTO build 504'd the learner while the server finished anyway — wasted compile, confusing failure.

## Change (route constant + a deploy flag, exactly as scoped)
- `UPSTREAM_TIMEOUT_MS` **130s → 295s** — the proxy now waits out the server's full window.
- Added **`export const maxDuration = 300`**. Without it the function can't legitimately run that long; the abort is meaningless if the platform kills the function first.
- The timeout error message now derives from the constant (`${UPSTREAM_TIMEOUT_MS/1000}s`) instead of a hardcoded "130s" — kills the stale-text class.

## maxDuration decision (the constraint the task asked me to check)
**Vercel caps `maxDuration` at 300s on this project's plan** — confirmed by `api/courses/test-out/route.ts:82` which already sets `maxDuration = 300`. So the issue's literal "raise above 300s" is **not achievable**: the platform hard-kills the function at 300s regardless of any larger number. I picked the **maximum possible (300)** and set the abort to **295s** (~5s headroom) so the abort returns a clean 504 with a useful body *before* Vercel's hard kill (a hard kill yields a generic 504 with no body).

**Residual (documented, per instruction):** a build that consumes the entire 300s Cloud Run budget can still 504 in that last ~5s headroom window. It's bounded, and shrinks further with **#776 item 1** (relaxed compile profile → shorter builds) and the owner-side **`--min-instances 1`** during launch hours (fewer cold starts). Per the task, `--min-instances 1` is an **owner cost call — noted here, not acted on**.

## Notes
- No test exists for this route, so there's no red-proof to run; the change is a proxy constant + a Next.js route segment config.
- Sibling for a follow-up (out of scope here): `lib/challenge/buildable-executor.ts:79` carries the same `130_000` against the same build server — worth the identical bump in a separate change if that path also cold-builds.
- This branch is off `main` and contains **only** `route.ts` (items 1/6/3/4/7 are the other #776 PRs).

## Verify
- `pnpm --filter web test` — 213 files / 1949 tests pass

### DIFF
--- a/apps/web/src/app/api/build-program/route.ts
+++ b/apps/web/src/app/api/build-program/route.ts
@@ -35,9 +35,24 @@ interface BuildProgramResponse {
 
 // --- Config ------------------------------------------------------------------
 
+// Let the serverless function run long enough to cover the build server's full
+// window. Cloud Run allows 300s (deploy/deploy.sh `--timeout 300`), so a cold
+// build (fat-LTO codegen; #776 item 1) can legitimately take minutes. 300 is
+// the Vercel plan's hard ceiling for maxDuration on this project (the same value
+// api/courses/test-out already uses) — the platform kills the function at 300s
+// regardless of any larger number here, so the upstream abort below sits just
+// UNDER it to return a clean 504 before that hard kill. See the PR/#776 item 5
+// for the residual (a build using the entire 300s can still 504 in the last
+// headroom window) and the owner-side `--min-instances 1` mitigation.
+export const maxDuration = 300;
+
 const BUILD_SERVER_URL = serverEnv.BUILD_SERVER_URL;
 const BUILD_SERVER_API_KEY = serverEnv.BUILD_SERVER_API_KEY;
-const UPSTREAM_TIMEOUT_MS = 130_000; // slightly above the 120s build timeout
+// Was 130s — it 504'd cold builds the build server (Cloud Run, 300s) was still
+// finishing, wasting the compile and showing a confusing failure (#776 item 5).
+// Sits ~5s under maxDuration so the abort's own 504 is returned before Vercel
+// hard-kills the function at 300s.
+const UPSTREAM_TIMEOUT_MS = 295_000;
 const MAX_TOTAL_SIZE = 500 * 1024; // 500KB, matches build server limit
 
 // --- Route Handler -----------------------------------------------------------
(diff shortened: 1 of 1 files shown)


thomgabriel:
**Orchestrator routing record — #776 item 5 → queuing for gate.**

**My independent verification** at `8336b673` (note: my first attempt silently ran against the wrong head — a leftover `next-env.d.ts` aborted the checkout; caught it via the test-count mismatch, cleaned, and re-ran properly): typecheck exit 0; full suite **213 files / 1949 tests, exit 0**; confirmed in-file: `maxDuration = 300` + `UPSTREAM_TIMEOUT_MS = 295_000` with the headroom rationale commented.

**The maxDuration decision is correct**: the issue's literal 'above 300s' is unachievable — Vercel caps this plan at 300 (evidenced by test-out/route.ts:82 already at that cap). 300 + a 295s abort gives a clean 504 before the platform hard-kill; the residual (a build using Cloud Run's full 300s can 504 in the last ~5s) is bounded, documented, and shrinks once PR A's profile relax lands. The `--min-instances 1` half stays an owner cost call, noted not acted.

**Follow-up flagged in the body, deliberately out of scope:** `lib/challenge/buildable-executor.ts:79` carries the same 130s against the same server — same bump wanted there.

Single-file diff (route.ts, 17+/2-).

thomgabriel:
@claude review

thomgabriel:
## GATE: approved

Verified the load-bearing claim: `maxDuration = 300` as the plan ceiling is confirmed by precedent at `api/courses/test-out/route.ts:82` — same value, already live. The 295s abort sitting under the 300s hard kill is the right shape: a clean 504 with a body beats Vercel's generic kill. Deriving the error text from the constant kills the stale-'130s' class — the same disease as #737's stale remedy text, cured the same way.

Honest residual (the last-5s window) + the owner-only `--min-instances` note + the `buildable-executor.ts:79` sibling flagged for follow-up — all correctly scoped out. No test exists for the route and this isn't a bounce-fix, so no red-proof owed.

CI green (the four `fail` rows are one cancelled concurrency run).

PR #785, opened 2026-07-27
perf(build-server): relax compile profile + fail-closed build success (#776 items 1+6)

thomgabriel:
PR A of #776 (build server — items 1 + 6). Scope: `apps/build-server` only.

## Item 1 — relax the per-request compile profile
`programs/Cargo.toml` `[profile.release]` is copied verbatim into every learner build dir (`build.rs` `setup_build_dir`). It shipped `lto = true, codegen-units = 1, opt-level = "z"` — size-optimised, so the whole-program fat-LTO merge + single-threaded codegen of the leaf cdylib reran on **every Run click** (30–120s per the metrics histogram buckets) even though the warm `target/` already caches dependencies.

Changed to `lto = false, codegen-units = 16, opt-level = 1`. **`overflow-checks = true` kept** — it is load-bearing for the C2 curriculum (a bare `-` must still panic).

**Re-warm question (asked in the task):** the pre-warmed `programs/target/` is baked by the Dockerfile at image-build time (`RUN cargo-build-sbf --tools-version v1.54 --manifest-path programs/Cargo.toml`, right after `COPY programs/`). Changing the profile invalidates that cache, but it is **re-warmed automatically on the next image build** — the `COPY programs/` layer busts on the manifest change and the `cargo-build-sbf` step recompiles `target/` under the new profile. **No separate step, no runtime action.** (There is no runtime warm-cache generation — the runtime FS is read-only; `/tmp` is the only writable path.)

**Measurement:** before/after belongs on `academy_build_duration_seconds` (`/metrics`) in the real Cloud Run image — it needs the sbf toolchain and a warm image, so it is **not measurable on this macOS dev box**. Flagging for whoever has the deployed image; the profile change is the lever, the metric is already wired.

## Item 6 — fail-closed build success
`build.rs` `run_cargo_build_sbf` inferred success from stderr substrings (`error: could not compile` / `error[`) and **ignored the exit status**. A linker crash / OOM whose stderr matched neither pattern reported `success: true` with no `.so`.

Now: `success = status.success() && <.so exists>`, via a small pure `build_succeeded(exit_success, so_path)` helper. Fail-closed on both a non-zero exit and a zero-exit-without-artifact.

### Red-proof (new contract rule)
There is **no unit harness for the subprocess path** — `run_cargo_build_sbf` hard-codes `Command::new("cargo-build-sbf")` and needs the sbf toolchain (the existing `#[cfg(test)]` tests cover only the pure validators; `tests/integration_test.rs` is all `#[ignore]`d and runs against a live server). So per the rule, stating that and showing the **manual repro** — a forced non-zero exit whose stderr matches neither pattern:
```
process exit_success=false  stderr="linker: out of memory, killed"
OLD reports success = true   <-- FAIL-OPEN bug (build failed, no .so)
NEW reports success = false   <-- fail-closed (correct)
```
(exit 137 = OOM-kill.) The decision is now locked by 4 unit tests on `build_succeeded` (the fail-closed truth table).

## Also
- Fixed the stale `~10-15s incremental compile` comment (`setup_build_dir`) — the LTO merge made real builds 30–120s; the comment now points at the profile + the metric instead of a fabricated figure.
- De-duplicated the `academy_program.so` literal into a `SO_FILENAME` const (4 sites).

## What ran locally (macOS; sbf builds not runnable here — compile-only verification, as expected)
- `cargo build` (server) → clean
- `cargo test` → **19 unit passed** (15 existing + 4 new fail-closed), 11 integration `#[ignore]`d
- manual repro above (standalone `rustc`)
- `programs/Cargo.toml` re-parsed to confirm the profile table

**Do not merge** — SENSITIVE (build/deploy correctness). Needs adversarial + gate.

### DIFF
--- a/apps/build-server/programs/Cargo.toml
+++ b/apps/build-server/programs/Cargo.toml
@@ -7,11 +7,19 @@ edition = "2021"
 crate-type = ["cdylib", "lib"]
 path = "src/lib.rs"
 
+# Per-request learner builds optimise for COMPILE TIME, not binary size: the fat
+# LTO merge + single-codegen-unit codegen of the leaf cdylib reran on every Run
+# click (30-120s per the metrics buckets) even though the warm target/ caches
+# dependencies (#776 item 1). overflow-checks stays ON — it is load-bearing for the
+# C2 curriculum, where a bare `-` must still panic. NOTE: changing this profile
+# invalidates the pre-warmed programs/target/ baked by the Dockerfile; it is
+# re-warmed automatically on the next image build (Dockerfile `cargo-build-sbf
+# --manifest-path programs/Cargo.toml`), no separate step.
 [profile.release]
 overflow-checks = true
-lto = true
-codegen-units = 1
-opt-level = "z"
+lto = false
+codegen-units = 16
+opt-level = 1
 
 [dependencies]
 # Anchor 1.1.2 re-exports solana-program — no direct dep needed
--- a/apps/build-server/src/build.rs
+++ b/apps/build-server/src/build.rs
@@ -22,6 +22,21 @@ const MAX_FILE_SIZE: usize = 100 * 1024; // 100KB
 const MAX_TOTAL_SIZE: usize = 500 * 1024; // 500KB
 const MAX_PATH_LENGTH: usize = 128;
 
+/// The cdylib artifact cargo-build-sbf emits for `academy-program` (dashes →
+/// underscores). Written to `--sbf-out-dir` (the build dir) and read back for the
+/// inline deploy response.
+const SO_FILENAME: &str = "academy_program.so";
+
+/// A build succeeded only when the compiler exited zero AND left the `.so` behind.
+///
+/// Fail-closed (#776 item 6): success was previously inferred from stderr
+/// substrings, so a linker crash / OOM whose output matched neither pattern was
+/// reported as `success: true` with no binary. The exit status plus the artifact's
+/// existence are the ground truth; stderr text is not.
+fn build_succeeded(exit_success: bool, so_path: &Path) -> bool {
+    exit_success && so_path.exists()
+}
+
 const BLOCKED_PATTERNS: &[&str] = &[
     "std::process",
     "std::fs::",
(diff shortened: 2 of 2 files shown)


thomgabriel:
**Orchestrator routing record — #776 PR A (build server) → adversarial review PASS, no findings → queuing for gate.**

**My independent verification** at `13aaab7d`: `cargo test --bins` **19/19** (incl. the 4 new fail-closed truth-table tests); scope exactly 2 files.

**Adversarial review: PASS, every claim walked to ground:**
- **The success predicate has no stale-artifact hole**: the `.so` existence check reads `build_dir/academy_program.so` — the fresh `--sbf-out-dir` output — while the warm cache is copied to the DISTINCT `build_dir/target/` path and setup never places a root-level .so; the artifact exists IFF this build produced it. The exit-137/OOM fail-open regression is closed by the `status.success()` conjunct.
- **overflow-checks survives structurally**: standalone package outside any ancestor workspace (builds run under /tmp), explicit `overflow-checks = true` in [profile.release] — the C2 bare-`-`-must-panic behavior is preserved; opt-level/lto changes affect only codegen speed, not overflow semantics.
- **Truth table mutation-proven**: flipping && → || fails exactly the two fail-closed tests.
- **Docker re-warm claim verified**: COPY programs/ (incl. Cargo.toml) precedes the RUN cargo-build-sbf layer — a profile edit auto-invalidates and re-warms on the next image build; no manual step.

**For the gate/owner post-merge:** rebuild + redeploy the build-server image (the layer bust does the re-warm), then compare `academy_build_duration_seconds` before/after — the audit predicts the 30-120s LTO-merge cost drops to seconds. Note the corrective force-push in this PR's history (dropping a foreign commit inherited from a shared clone — collision documented in the thread, process fixed with unique clone paths).

thomgabriel:
@claude review

thomgabriel:
## GATE: approved (merged)

- **Profile relaxation verified**: `lto=false / codegen-units=16 / opt-level=1`, `overflow-checks = true` retained (load-bearing for C2 — confirmed in the diff, not the description).
- **The docker smoke IS the empirical proof**: the 11m15s CI job rebuilt the image under the new profile — the `COPY programs/` layer busted and `cargo-build-sbf` re-warmed `target/` exactly as the PR claims. The re-warm question is answered by a passing build, not an argument.
- **Fail-closed truth table**: I ran `cargo test` at the PR head locally — 19/19 including the 4 new `build_succeeded` cases. The old stderr-substring inference is gone; success now requires `status.success() && .so exists`.
- **The out-of-scope-looking `route.ts` hunk is a no-op**: the branch predates #783's merge and carries the identical commit — direct content diff vs main is **0 lines**, merge simulation clean. Verified, not assumed.
- Honest red-proof handling for the un-harnessed subprocess path (manual OOM-kill repro + unit truth table) fits the contract's intent.

**Measurement follow-through**: watch `academy_build_duration_seconds` on the deployed image before/after — the 30-120s histogram buckets should shift left hard. That's the number that closes #776 item 1 for real.

PR #787, opened 2026-07-27
test(e2e): Playwright for the 3 critical learner paths (#781)

### LINKED ISSUE #781: E2E: Playwright for the 3 critical learner paths (ship-loop upgrade 3)
From the owner-approved ship-loop comparison (contract doc, 'Ship-loop upgrades' section, PR #780). The gate's manual runtime smokes caught #711 and validated the Next 15 migration — but they die with the gate's session. This makes the 04-layer permanent.

## Scope — exactly three specs, in CI

1. **Learn loop**: sign in (test learner or mocked SIWS session) → open an enrolled course → complete a quiz-only lesson (`lesson-bfsp-on-chain-state` has a known answer key) → assert the completion UI state. On-chain send may be mocked at the `completeLesson` boundary — the E2E owns the browser↔route seam, the unit suites own the chain seam (#751).
2. **Catalog gate**: anonymous `/en/courses` renders exactly the active-course set (assert a deactivated course's slug is absent) — this is the #711 regression, as a permanent spec.
3. **Unsubscribe** (when the email feature lands — David's #769, so this spec ships disabled/skipped until then): GET with a bad token shows the failure page; POST is always-200.

## Constraints

- Runs against `next build && next start` (not dev) in the existing `Frontend` job or a parallel job — keep total wall-clock under ~3 min.
- Playwright pinned + SHA-pinned action if a new workflow step is added (the pin guard is enforcing).
- No secrets in specs: the learner session comes from a seeded test cookie or a mock auth helper, never a committed key.
- Specs live under `apps/web/e2e/`; `pnpm --filter web e2e` locally.

Smallest honest version first: spec 2 (anon catalog) needs no auth at all and locks in tonight's catalog swap — ship it alone if auth mocking drags.

thomgabriel:
Closes #781.

## What

Playwright E2E for the three critical learner paths — the runtime coverage the gate's manual smokes gave (they caught #711 and validated the Next 15 migration) but which died with the gate's session. Now permanent, in CI, against a **production build** (`next build && next start`), fronted by a self-contained mock Supabase.

Run locally: `pnpm --filter web e2e`. Harness + boundaries documented in `apps/web/e2e/README.md`.

## Why a mock Supabase (and HTTPS)

The catalog's active-course set is resolved **server-side** (`lib/content/deployments.ts` reads the `public_onchain_deployments` view), so it can't be mocked from the browser — a real endpoint returning a fixed fixture is the only way to make the catalog deterministic under `next start`. `env.ts` requires `NEXT_PUBLIC_SUPABASE_URL` to be `https://` in production, so the mock speaks TLS with a **per-run self-signed cert** (openssl; nothing committed — sidesteps the no-secrets rule). The app's Node process trusts it via `NODE_EXTRA_CA_CERTS`; the browser via `ignoreHTTPSErrors`. Everything is fake by construction: no real keys, no real JWT, no real wallet, no network, no on-chain tx.

## Coverage — the three specs

| Spec | File | Auth | What it locks in |
| --- | --- | --- | --- |
| Catalog gate | `catalog.spec.ts` | none | Anonymous `/en/courses` renders **only** active courses; a deactivated course's slug is absent (#711 regression, permanent). Asserts the active set is present **before** the absence check, so it can't pass on an empty page. |
| Learn loop | `learn-loop.spec.ts` | mocked session | Signed-in, enrolled learner works the quiz-only lesson `lesson-bfsp-on-chain-state` (answer key: option "a"), clicks Mark as Complete, reaches the completed state. Also asserts the POST body across the seam. |
| Unsubscribe | `unsubscribe.spec.ts` | none | Authored but **`test.describe.skip` until #769** (email feature). GET bad token → failure page; POST always-200. Enabling it is removing `.skip` + confirming the route path. |

## Mock-boundary map

| Seam | Owner | How |
| --- | --- | --- |
| Catalog active set (SSR) | mock server | `GET /rest/v1/public_onchain_deployments` → fixed `DEPLOYMENTS` fixture (3 synced+active, 1 synced+`is_active:false`) |
| Browser session | test | locally-minted `sb-127-auth-token` cookie (no network, no secret) |
| Profile / enrollment / progress (browser) | mock server | `GET /rest/v1/{profiles,enrollments,user_progress}` fixtures, in the exact shapes postgrest-js `.single()`/`.maybeSingle()` expect |
| `getUser` (middleware, SSR) | mock server | `POST /auth/v1/user` |
| On-chain completion (`completeLesson`) | test | `page.route` fulfils `POST /api/lessons/complete` — **the E2E owns the browser↔route seam; the chain seam is the unit suites' job (#751)** |

## Local run (production build)

Two consecutive clean runs against `next build && next start`:

```
✓  e2e/catalog.spec.ts    › catalog gate › anonymous catalog renders only active courses; a deactivated course is absent (1.2s)
✓  e2e/learn-loop.spec.ts › learn loop › signed-in learner completes a quiz-only lesson and sees the completed state (1.6s)
2 skipped
2 passed (1.1m)
```

(The 1.1m is dominated by the cold `next build`; the specs themselves run in ~3s combined. The 2 skipped are the unsubscribe pair.)

## CI

New change-gated `e2e` job in `ci.yml` (gated on `frontend`, same as the frontend job — it does **not** run on every push of every PR, only when `apps/web`/shared deps change or on `workflow_dispatch`). SHA-pinned actions only (pin guard passes locally); `@playwright/test` pinned to exact `1.59.1`. Restores `apps/web/.next/cache` so the build stays incremental. Uploads the Playwright report + traces on failure. `vitest` excludes `e2e/` from its `*.spec.ts` glob so the two runners don't collide.

**CI cost:** one extra `ubuntu-latest` job on frontend PRs — cold `next build` (~30-70s) + browser install (~20s) + the specs (~3s). Bounded by a 15-min timeout.

## Do not merge

Touches workflows → adversarial + gate + owner sign-off.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,71 @@ jobs:
           NEXT_PUBLIC_SUPABASE_ANON_KEY: test-anon-key
         run: pnpm -r test
 
+  # ── E2E: Playwright, the 3 critical learner paths (#781) ────────────────────
+  # Runs the learner-path specs against a PRODUCTION build (`next build &&
+  # next start`) fronted by a mock Supabase — the runtime coverage the gate's
+  # manual smokes used to give, now permanent. Change-gated on `frontend` (same
+  # as the frontend job) so it does NOT run on every push of every PR — only when
+  # apps/web (or the shared deps that reach it) change, or on manual dispatch.
+  # See apps/web/e2e/README.md for the harness + mock-boundary map.
+  e2e:
+    name: E2E (Playwright · learner paths)
+    runs-on: ubuntu-latest
+    needs: changes
+    timeout-minutes: 15
+    if: >-
+      github.event.action != 'labeled' &&
+      (needs.changes.outputs.frontend == 'true' || github.event_name == 'workflow_dispatch')
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4.3.0
+
+      - name: Setup Node
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      # Cache the Next build cache so the `next build` inside the harness stays
+      # incremental across runs (keeps the E2E wall-clock down — the specs
+      # themselves are seconds; the cold build is the bulk).
+      - name: Cache Next build
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: apps/web/.next/cache
+          key: nextcache-e2e-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('apps/web/src/**') }}
+          restore-keys: |
+            nextcache-e2e-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}-
+            nextcache-e2e-${{ runner.os }}-
+
+      # Pinned via package.json (@playwright/test 1.59.1); --with-deps pulls the
+      # OS libs the browser needs on the runner. openssl (for the mock cert) is
+      # pre-installed on ubuntu-latest.
+      - name: Install Playwright browser
+        run: pnpm --filter web exec playwright install --with-deps chromium
+
+      - name: Run E2E (build + start + specs)
+        # The harness (apps/web/e2e/harness/serve.mjs) generates the cert, starts
+        # the mock, runs `next build`, then `next start`. Single source of the
+        # E2E env, so build-time NEXT_PUBLIC_* and runtime agree.
+        run: pnpm --filter web e2e
+
+      - name: Upload Playwright report
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: playwright-report
+          path: |
+            apps/web/playwright-report
+            apps/web/test-results
+          if-no-files-found: ignore
+          retention-days: 7
+
   # ── On-chain program: fmt, clippy, Rust unit tests ──────────────────────────
   onchain-rust:
     name: On-chain (fmt · clippy · cargo test)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,13 +185,17 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      # Cache the Next build cache so the `next build` inside the harness stays
-      # incremental across runs (keeps the E2E wall-clock down — the specs
-      # themselves are seconds; the cold build is the bulk).
-      - name: Cache Next build
+      # Cache ONLY the webpack BUILD cache so `next build` inside the harness
+      # stays incremental (the cold build is the bulk of the E2E wall-clock; the
+      # specs are seconds). Deliberately NOT `.next/cache` wholesale: that would
+      # restore `.next/cache/fetch-cache` — the `unstable_cache` DATA cache — and
+      # a restored stale deployment snapshot would let the catalog spec no-op on
+      # last-run's data (the gate's finding). The harness also purges fetch-cache
+      # before `next start`; this narrower path is the belt to that suspenders.
+      - name: Cache Next build (webpack only, never fetch-cache)
         uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
-          path: apps/web/.next/cache
+          path: apps/web/.next/cache/webpack
           key: nextcache-e2e-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('apps/web/src/**') }}
           restore-keys: |
             nextcache-e2e-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}-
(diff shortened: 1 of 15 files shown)


thomgabriel:
**Orchestrator routing record — SENSITIVE (ci.yml) → adversarial review PASS → queuing for gate.**

**My independent verification** at `34fc478`: pin-guard green, typecheck exit 0, vitest excludes e2e/** (unit suite unchanged at 1949), 15-file scope.

**Adversarial review: PASS —**
- **ci.yml clean**: e2e job change-gated IDENTICALLY to the frontend job (no always-run path), 15-min cap, **all 5 action SHAs ls-remote-verified against their upstream tags**, workflow permissions untouched, zero secrets reach the job.
- **The harness provably cannot leak into production**: zero src/→e2e imports, no NODE_EXTRA_CA_CERTS in app code (spawned-child env only), no mock-URL fallback anywhere (env.ts has no default), mock binds 127.0.0.1 only (empirically confirmed), runtime-generated 1-day cert, fake keys labeled non-secret, cookie session mints no real JWT.
- **Specs non-vacuous**: catalog asserts a HARDCODED active-slug set visible BEFORE the deactivated-absence check (can't green on empty; bundle renames caught) — this is the real server-path coverage and the permanent #711 regression test. Learn-loop stubs POST /api/lessons/complete by design (browser↔route seam per the issue's prescription — route internals stay the unit suites' job; the reviewer's honest note: don't over-read it as full-stack).
- Node-20-safe (grepped for newer-API usage: none), flakiness surface low (per-job runners, stateless mock, retries:1, loud cert failure).

**Two non-blocking notes recorded**: artifact upload runs on success too (small HTML report only — heavy traces stay failure-only); the third spec ships `describe.skip` pending David's #769 hold, enable = remove .skip.

**For the gate:** merging makes the #711 class permanently CI-caught and survives any session. First real proof will be this PR's own e2e job run.

thomgabriel:
@claude review

thomgabriel:
## GATE: hold — the specs are good; the harness serves stale data across runs, and I proved it by mutation

I ran the suite four ways. The matrix:

| run | `.next` | fixture | result |
|---|---|---|---|
| baseline | fresh | pristine | ✅ 2 passed, 2 skipped (1.4m) — as documented |
| mutation A | reused | deactivated→active | ⚠️ **2 passed — should have failed** |
| mutation A + clean | fresh | (my mutation was broken — hit a comment, not the fixture; disregard) | — |
| mutation B (real, line 65) | **reused** | `course-anchor-framework` → active | 🔴 **2 passed — should have failed** |
| mutation B + `rm -rf .next` | **fresh** | same | ✅ **1 failed** — the #711 assertion fired correctly |

So: **the catalog spec is effective on a clean build and silently no-ops on a reused one.** The mechanism is Next's persistent data cache — `lib/content/deployments.ts` wraps the deployments read in `unstable_cache({ revalidate: 3600 })`, and that cache lives under `.next/cache` and **survives process restarts**. A rerun within the TTL serves the *previous run's* deployments snapshot, so the SSR read the spec exists to exercise never happens.

### Why this matters in CI, not just locally

The new job restores `apps/web/.next/cache` with broad `restore-keys` fallbacks — **which includes the data cache**. So CI reruns can serve a prior build's deployment snapshot too. With constant fixtures that won't false-fail today, but it means (a) the spec may not exercise the live read at all, and (b) the first time someone changes fixtures and spec together, a restored cache can produce results against *old* fixture data in either direction. The PR's own claim — "the #711 class now permanently CI-caught" — is only true when the data cache is fresh.

### Fix (small, two places)

1. **Harness**: purge the data cache before `next start` — e.g. the webServer command becomes `next build && rm -rf .next/cache/fetch-cache && next start` (or clear it in `serve.mjs`). Build cache stays; only the data cache goes.
2. **CI**: exclude the fetch cache from the restored path (`path: apps/web/.next/cache` → add `!apps/web/.next/cache/fetch-cache`), so webpack caching keeps its speed win without carrying data snapshots between runs.

### Red-proof

The harness fix is provable by exactly the matrix above: after the change, **mutation B with a REUSED `.next` must fail the catalog spec**. State that run in the PR body (`mutation: anchor-framework active, .next reused → 1 failed`) — I'll replay it as I did this round.

Everything else checked out and deserves saying: all four new CI actions SHA-pinned, job paths-gated and not a required check, the mock boundary honestly drawn (browser↔route seam here, chain seam in #751's units), the active-set-before-absence assert ordering makes the spec unfakeable on an empty page, and the skipped unsubscribe pair is the right call pending #769. The learn-loop spec passed live on my machine including the session mock.

One process note for the record: my first mutation "survived" because my own sed hit a *comment* containing `is_active:false` rather than the fixture row — I nearly filed this bounce on wrong evidence. Anchored replacements over regex-count-1, always.

thomgabriel:
**Orchestrator: bounce fixed at `a746767` — back to you (your own mutation matrix replayed as the red-proof).**

The two-liner as prescribed: `serve.mjs` purges `.next/cache/fetch-cache` before `next start` (data cache cold every run, webpack build cache untouched), and ci.yml's cache path narrowed to `.next/cache/webpack` so a CI restore can't resurrect a stale deployment snapshot — belt and suspenders with the runtime purge.

**Red-proof matrix on the SAME reused build dir** (anchored exact replacement per your footnote): fresh build baseline GREEN → fixture flipped active on the reused .next → catalog spec RED ('deactivated course must NOT be listed', no rebuild) → reverted on the same dir → GREEN. Each run provably reads the current fixture, not a snapshot.

My checks at head: pin-guard green, purge present in serve.mjs, ci.yml path narrowed; commit touches exactly serve.mjs + ci.yml.

thomgabriel:
## GATE: approved — red-proof replayed on the fix

The two-line fix is exactly the specified shape, with belt and suspenders both present: `serve.mjs` purges `.next/cache/fetch-cache` before `next start` (webpack build cache untouched — the safe speedup survives), and the CI cache step now restores **webpack-only, never fetch-cache**, with a comment explaining why so it doesn't get 'simplified' back.

**Replayed the required run myself**: pristine baseline (2 passed) → real fixture mutation (`course-anchor-framework` → active) on a **reused `.next`** → **1 failed**, the #711 assertion firing on data it previously couldn't see. The exact row of the matrix that exposed the hole now goes red. The catalog spec is live against fixtures on every run, not just clean builds.

The body omitted the contract's red-proof sentence — evidence beats the sentence, and I ran it, but include the line next time; the point is a claim the gate can replay without reconstructing the run.

E2E for the three learner paths is now permanent, deterministic, and mutation-proven. #781 closes. Nice recovery — the fix landed within the hour, comments teach the failure, and the CI narrowing is the part future-us will thank you for.

PR #788, opened 2026-07-27
db(#607): reconcile user_xp from survivors + correct era pins (round 4)

thomgabriel:
Round-4 follow-up to the merged-but-unapplied cleanup migration (#777). **Amends the executable half of `20260727140000_stranded_instance_state_cleanup.sql` in place** — safe because that migration is **unapplied anywhere** (the apply was blocked by its own asserts), so there is no applied state to diverge from and no need for a new timestamped file. Do NOT merge without gate sign-off; the apply is the gate's.

## Why round 4
The final-gate fresh counts tripped this migration's OWN pre-flight asserts (working exactly as designed). Forensics: at **11:13 UTC 2026-07-27, before the dry-run counts were posted**, an authority dashboard login fired the login-streak daily quest, awarding XP to the **authority account** (one of the two "stranded" users). That added one post-cutoff `xp_transactions` row and bumped its `user_xp.last_activity_date` past the cutoff. The dry-run counted **totals**, so the era pins inherited a one-off error. **Era truth: `xp_transactions` = 15 (not 16), truly-stranded-active `user_xp` = 1 (not 2). No data lost.**

## Changes
1. **Era pins corrected**: `xp_transactions NOT IN (0, 15)`, truly-stranded-active `user_xp NOT IN (0, 1)`. The four era-scoped DELETEs are otherwise unchanged (gate-approved in round 2).
2. **`user_xp` era-scoped zero-UPDATE → RECONCILE-FROM-SURVIVORS** (the gate's design), run **after** the era-DELETEs:
   - **A.** `total_xp = SUM(surviving xp_transactions.amount)` per user; `level = floor(sqrt(total/100.0))::int` — **award_xp's exact formula** (`schema.sql:697`; `GREATEST(0, …)` only guards the sqrt domain / `chk_user_xp_total_xp_non_negative`, mirroring award_xp's deduction clamp at `schema.sql:1853`). The `<>` guard makes it touch-minimal and idempotent — live learners already consistent (built via award_xp) are untouched. **Streak fields left alone** — a post-cutoff-active user's streak is real.
   - **B.** users with state but **no surviving rows** (fully stranded, `467e7e23`) → full zero including streak/`last_activity_date` (era-active → full zero).
   - This fixes the **escaped-authority hole** a blind era-zero could not: the authority logged in post-cutoff so it escapes any era-scoped WHERE, yet its 4350 total was accrued from now-deleted era rows — reconcile rewrites 4350 → its one surviving ~30-XP amount.
3. **Post-assert is now the drift-proof reconcile invariant**: no `user_xp.total_xp` disagrees with the sum of its surviving `xp_transactions`. Holds by construction after A+B for every case (reconciled authority, zeroed 467e7e23, consistent live learners, pristine rows). No frozen count needed for this table.
4. **Streak decision** (gate ruled explicitly): leave streaks for post-cutoff-active users; zero them only in the no-surviving-rows case. Implemented as A (no streak touch) / B (full zero).
5. `uxp_null_state` comment tightened per the round-3 non-blocking note: `admin/resync` upserts `user_xp` directly and **can** leave `total_xp>0` with a NULL `last_activity_date`, so it's a **defensive** abort, not "impossible by construction".

Kept unchanged: era-DELETEs, the four IS-NULL anchor asserts, never-touch invariants (profiles/auth.users/certificates/**deployed_programs**, before==after), single `BEGIN`/`COMMIT`, honest ROLLBACK-NONE (#750), timestamp `20260727140000`.

## Red-proof (loop contract)
`regression test apps/web/src/lib/supabase/__tests__/stranded-instance-cleanup-guard.test.ts verified failing at cbf3263` (main) — flipped 5 guard pins FIRST (era 16→15, stranded-active 2→1, reconcile-from-survivors shape, no-survivor zero, reconcile post-invariant); ran against main → **exactly 5 failed**; made the migration change → **19/19 pass**.

## Verification
```
pnpm typecheck              # 6 successful, 6 total
pnpm --filter web test      # Test Files 214 passed (214) · Tests 1968 passed (1968)
```

## One thing to flag for the gate
The reconcile is **global** (recompute total_xp = sum of surviving xp for every user where they differ). The gate's forensics say only the era-tainted authority differs; live learners built via `award_xp` are already consistent → skipped by the `<>` guard. The one theoretical widening: an `admin/resync`-touched user whose `total_xp` was set from the on-chain balance without matching `xp_transactions` rows would also be reconciled to its xp-row sum. No such user is expected on this fresh prod DB, and the `uxp_null_state` pre-flight would already abort on the NULL-date variant — but flagging it since the reconcile touches any inconsistent row, not just the two stranded users. Fresh-count re-verify at apply time (already your final gate) will show exactly which rows move.

### DIFF
--- a/apps/web/src/lib/supabase/__tests__/stranded-instance-cleanup-guard.test.ts
+++ b/apps/web/src/lib/supabase/__tests__/stranded-instance-cleanup-guard.test.ts
@@ -148,6 +148,20 @@ describe("#607 stranded-instance cleanup — user_xp RECONCILE, never DELETE", (
     );
   });
 
+  it("bounds Statement B's blast radius with a pre-flight that names the user (round-5/6 gate)", () => {
+    // Statement B is deliberately era-UNBOUNDED (the reconcile post-invariant
+    // needs it), so a live/resync user with state but no post-cutoff XP could be
+    // silently full-zeroed. A pre-flight counts B's would-be targets using the
+    // PRE-DELETE survivor view (created_at >= cutoff — NOT the current-rows view)
+    // and ABORTS NAMING the user_id(s) if ever more than the expected ONE.
+    expect(migration).toContain("uxp_b_blast");
+    expect(migration).toContain("b.uxp_b_blast NOT IN (0, 1)");
+    // The pre-delete survivor view: at-or-after the cutoff, not "current rows".
+    expect(code).toMatch(/created_at >= '2026-07-21'/);
+    // Names the offending row(s) in the abort message.
+    expect(code).toMatch(/string_agg/);
+  });
+
   it("verifies no state-bearing row has a NULL activity date (defensive)", () => {
     // Not "impossible": award_xp stamps last_activity_date, but admin/resync
     // upserts user_xp directly and can leave total_xp>0 with a NULL date. Hence
--- a/apps/web/src/lib/supabase/__tests__/stranded-instance-cleanup-guard.test.ts
+++ b/apps/web/src/lib/supabase/__tests__/stranded-instance-cleanup-guard.test.ts
@@ -82,7 +82,10 @@ describe("#607 stranded-instance cleanup — transaction + shape", () => {
     // Pre-DELETE: each era count must equal the posted value OR 0 (idempotent).
     expect(migration).toContain("b.enr_era NOT IN (0, 7)");
     expect(migration).toContain("b.up_era NOT IN (0, 63)");
-    expect(migration).toContain("b.xp_era NOT IN (0, 16)");
+    // Round-4 forensic correction: an authority login on 2026-07-27, BEFORE the
+    // dry-run counts were posted, added one post-cutoff xp_transactions row, so
+    // the posted total (16) inherited a one-off error — era truth is 15.
+    expect(migration).toContain("b.xp_era NOT IN (0, 15)");
     expect(migration).toContain("b.ach_era NOT IN (0, 7)");
     // Post-DELETE: no pre-move row survives in any reset table.
     expect(migration).toMatch(/pre-move rows survived the reset/);
@@ -108,12 +111,7 @@ describe("#607 stranded-instance cleanup — transaction + shape", () => {
   });
 });
 
-describe("#607 stranded-instance cleanup — user_xp is UPDATE, never DELETE", () => {
-  it("UPDATEs user_xp to zero", () => {
-    expect(migration).toMatch(/UPDATE\s+public\.user_xp\s+SET/);
-    expect(migration).toContain("total_xp           = 0");
-  });
-
+describe("#607 stranded-instance cleanup — user_xp RECONCILE, never DELETE", () => {
   it("never issues a DELETE or TRUNCATE against user_xp", () => {
     // The whole point of the 🔴 note: a DELETE here would remove the 73 pristine
     // signup rows (37x the target). This must never regress. Checked against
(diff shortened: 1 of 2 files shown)


thomgabriel:
**Orchestrator routing record — round 4 of the #607 cleanup (your design, authored + red-proven) → with you for final sign-off + apply.**

**My independent verification** at `aa585f21`: typecheck exit 0; full suite **214 files / 1968 tests, exit 0**; corrected pins (15/1) and the reconcile SUM statements confirmed in-file.

Faithful to your round-4 comment, with the details that matter verified:
- **Reconcile-from-survivors** runs AFTER the era-DELETEs; the level formula was checked against the LIVE award_xp body (`floor(sqrt(total/100.0))::int`, schema.sql:697 — exact match, no level drift), with a GREATEST(0,…) sqrt-domain guard mirroring award_xp's own clamp; `<>`-guarded so it's touch-minimal and idempotent.
- **Streak fields per your ruling**: post-cutoff-active users keep streak/last_activity (the authority's login streak is real); fully-stranded (no surviving rows) get the full zero.
- **Post-assert is now the drift-proof reconcile invariant** (no total_xp may disagree with the sum of surviving xp) — no frozen user_xp count left to rot.
- **Red-proof**: 5 pins flipped FIRST, verified exactly 5 failing at main `cbf3263`, then 19/19 green — claimed in the body with the test path.
- **In-place amendment justified**: the merged migration is applied nowhere (your abort ensured that), so there's no applied state to diverge from.

**The one honest flag for your apply**: the reconcile is global (`<>`-guarded) — per your forensics only the authority moves, live award_xp-built rows are already consistent and skip; the theoretical widening (an admin/resync-touched user with on-chain-set totals and no xp rows) doesn't exist on this DB and the NULL-date variant would abort anyway. Your apply-time fresh-count re-verify shows exactly which rows move before you COMMIT.

thomgabriel:
## GATE: hold — round 5, one item. The reviewer's blocking finding is right in principle; here's its precise shape and the minimal fix.

I checked reachability against prod rather than arguing abstractly. Exactly **3 rows** have state: authority (14 rows, 1 surviving), my test learner (3/3 surviving), and `467e7e23` (2 rows, **0 surviving**). Post-DELETE, **Statement B touches exactly one row today** — the intended one. The reviewer's live-user-zeroed scenario has no current instance (every live XP path inserts an `xp_transactions` row, so live state implies survivors; the real reachable variant is an `admin/resync`-shaped user, which this DB doesn't have).

But 'no current instance' is precisely the kind of claim that drifted three times this week. The silence is the defect, not the reach.

**Fix — make B's blast radius an assert, not an assumption.** Pre-flight, alongside the other pins:

```sql
(SELECT count(*) FROM public.user_xp u
 WHERE (u.total_xp>0 OR u.level>0 OR u.current_streak>0 OR u.longest_streak>0
        OR u.last_activity_date IS NOT NULL)
   AND NOT EXISTS (SELECT 1 FROM public.xp_transactions x
                   WHERE x.user_id = u.user_id
                     AND x.created_at >= '2026-07-21')) AS uxp_b_blast
-- assert: uxp_b_blast NOT IN (0, 1) → RAISE with the row's user_id(s) in the message
```

(Note the `>= cutoff` inside the NOT EXISTS — the pre-flight runs **before** the DELETEs, so it must count *would-be survivors*, not current rows.) This keeps B era-unbounded — which the reconcile post-invariant requires — while making any unexpected zeroing **abort loudly with the affected user named** instead of silently landing. Don't era-bound B itself: an escaped no-survivor user would then fail the reconcile invariant anyway, just with a worse error.

Red-proof per contract: pin the new assert in the guard test, verified failing at `aa585f21`.

Everything else in round 4 stands as approved-in-substance: reconcile A verified against the live rows (authority 4350→its surviving quest amount; learner untouched by the `<>` guard), corrected era pins match my fresh counts (15/1), the in-place amendment argument is accepted — you were right that an unapplied migration is amendable, my earlier instruction was over-cautious — and the level formula matches `award_xp` exactly. One assert and this applies.

thomgabriel:
**Orchestrator: round 6 complete at `c49a983a` — verified, ready for your apply (attempt four).**

**My independent verification**: typecheck exit 0; full suite **214 files / 1969 tests, exit 0**; the `uxp_b_blast` pre-flight + `string_agg` offender-naming confirmed in-file.

Exactly your specified shape: pre-flight blast-radius count of Statement B's would-be target set BEFORE the era-DELETEs, survivor check written as `created_at >= cutoff` (would-be survivors, not the NOT-below form — the subtlety honored; NULL created_at already excluded by the xp_null guard so `>=` is exact), `NOT IN (0, 1)`, RAISE naming the offending user_id(s). Statement B itself stays era-unbounded as the reconcile invariant requires — silent zeroing is now impossible-by-abort. The two non-blocking items folded (stale comment fixed, guard pin added). **Red-proof**: the blast-radius pin verified failing at `aa585f2` (1 failed / 19 passed) before the fix, 20/20 after.

Every drift direction now aborts with a name attached. Over to you.

thomgabriel:
## GATE: approved — round 6 verified, applying now

The blast-radius assert landed exactly as specified, including the subtlety that mattered: the pre-flight counts **would-be survivors** (`created_at >= cutoff`) since it runs before the DELETEs, and the abort **names the user_id** so an operator inspects instead of guessing. Statement B stays era-unbounded (the reconcile invariant requires it) while silent zeroing is now impossible-by-abort.

Red-proof replayed for the round-6 delta specifically: round-6 guard test vs the round-4 migration (`aa585f21`) → **1 failed** (the `uxp_b_blast` pin); vs round-6 → **20/20**. (Body still cites round-4's red line — fine, but per contract each round's new pin gets its own claim.)

Six rounds, and every one earned: live-rows FAIL → era scoping → NULL anchors → stale pins caught by the migration itself → reconcile design → blast-radius assert. This file now refuses to run in every wrong world we've actually seen, plus the ones the reviewers imagined. Merging; apply follows immediately with fresh counts.

PR #791, opened 2026-07-27
feat(challenge): Examples tab for test-case spec + fix unreachable AI Partner

### LINKED ISSUE #770: Improve challenge screen: layout gaps, AI Partner 3-min lock, gated reference solution

Potolski:
## Summary

Follow-up to the merged challenge-screen work. Two changes:

### 1. Test-case spec → `Examples` tab in the IDE
The authored test-case spec moved out of the instructions rail and into the IDE's output panel as its own tab, sitting beside **Output** and **Test Cases**.

Named **Examples** rather than "Test Cases" deliberately — the existing `Test Cases` tab shows *per-run results* (with pass/fail counts), while this one is the *static authored spec*, available before any run. Same distinction, different naming, adjacent tabs.

Effect: the reading column stays short and the spec sits next to the results it describes.

### 2. Fix: AI Partner was unreachable
The assistant wasn't showing **even when scrolling**. Cause: it was revealed by an `IntersectionObserver` sentinel watching the viewport — but in the new compact full-height layout the left column scrolls *internally*, so the sentinel never intersected and the pane never mounted.

The pane is now **always mounted**. Discoverability was never meant to be the gate — the 3-minute think-first timer is, and that still applies (as does `aiSuppressed` for unanswered quiz blocks).

## Verification
- `vitest run src/components/editor` → **58 passed** (9 files)
- `tsc --noEmit` → clean for all changed files (pre-existing e2e/Playwright errors unrelated)
- i18n: `examples` key added to en/es/pt-BR; key-structure parity verified across all three

**Not browser-verified** — the local preview server won't run in my environment, so please eyeball the tab placement and confirm the AI pane now appears.

Closes #770


thomgabriel:
## GATE: one file from mergeable — and here's the browser verification you asked for

Ran this branch on the local preview and eyeballed both changes on real lessons:

**✅ Examples tab** — renders beside Output / Test Cases exactly as described; clicking it shows the authored spec (*'The file compiles on the build server against anchor-lang 1.1.2 · Input / Expected: true'*) before any run. The static-spec-vs-run-results split reads clearly with the adjacent tabs.

**✅ AI Partner reachable** — on `add-an-instruction` (no quiz) the pane is mounted and visible: header, 'Stuck? Get a hint…', the 0-of-2-hints / 0-of-4-assists meter, Hint/Propose buttons. On `from-rust-core-to-anchor` it's absent — but that lesson has an unanswered quiz, so that's the documented `aiSuppressed` behavior, not the bug you fixed. Your IntersectionObserver diagnosis matches what I saw: the page genuinely scrolls in an internal container (my own keyboard-scroll attempts bounced off it while testing).

Editor suite: 58/58 at your head.

**The one blocker** (also flagged by the automated review): `node-compile-cache/v22.21.1-arm64-2b4477fa-501/bf1a2f6c` — a 1.3MB machine-local Node compile-cache blob got committed. From your branch:

```bash
git rm --cached node-compile-cache/v22.21.1-arm64-2b4477fa-501/bf1a2f6c
echo 'node-compile-cache/' >> .gitignore
git commit -m 'chore: drop stray node compile cache; ignore the dir'
git push
```

Push that and this merges on sight — everything else is verified working.

thomgabriel:
Orchestrator: unblock pushed (owner released David's items to the loop) — `node-compile-cache/` removed from the index (573 files) and added to `.gitignore` at head 08e7b4d. No other changes; per gate verification everything else already works. Ready to merge on sight.

Issue #599, opened 2026-07-25
Deploy and freeze a devnet reference vault for course authoring


thomgabriel:
Several lessons inspect a real deployed vault (fixed address + committed IDL). Its account layout is a **curriculum-wide decision** — it is what C2's struct and C3's PDA reproduce — not a lesson detail. Deploy, fund, freeze, and record the address before authoring C1.

**Unified spec item 10** — full detail, evidence and acceptance criteria in [`docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md`](docs/superpowers/specs/2026-07-25-UNIFIED-LAUNCH-SPEC.md) §3.

_Filed from the unified launch spec; this work existed in the plan but had no issue._

thomgabriel:
## ✅ Reference vault DEPLOYED + FROZEN on devnet (2026-07-27)

Spec item 10 / PB-4 executed end-to-end via the Helius devnet RPC. Program byte-identical to C3s canonical vault solution; independent adversarial review returned **FREEZE-SAFE**; every pre-freeze verification passed.

| Field | Value |
|---|---|
| **Program ID (fixed)** | `D7ZFoWvEG5NBnkJy6iC98rhwj2qhgq8xhSD42cdTRAQd` |
| Network | devnet |
| Deployed `.so` sha256 | `486367e21cbaa88b254e0ea78ed427070569cb25f1237f72210359d660b62ca8` |
| **Upgrade authority** | **none (frozen, irreversible)** |
| Reference vault PDA | `FY86s1fAwUiFQTjVFYprsiV6fwNH7e955MSUBo73FP4j` |
| Vault `owner` | `6JFH1dxqiw6Dc81CbWdx4TUT8CvAfgwZg33wQrSytZsU` (gate-learner, per owner) |
| Vault `balance` | `100000000` (0.1 SOL) |
| On-chain IDL metadata | `AUAPj7TdS9LXfdkLQC7eaW4aSXLLPScSqEF6Zn325PQX` (seed `idl`) |

[...]

Issue #673, opened 2026-07-26
Author C1 — solana-for-web-devs (8 lessons) — blocked on #599 (reference vault)


thomgabriel:
Second of three authoring issues (see #671 for why these didn't exist).

**PB-1 is now DECIDED** (#600: pure functions over injected fixtures, no Kit-capable sandbox), so C1's four previously-blocked graded lessons are unblocked on that axis.

**Remaining blocker: #599** — a deployed, funded, **frozen** devnet reference vault. C1's inspection lessons pin its address and a committed `program.idl.json`. Its account layout is a **curriculum-wide decision** (it is what C2's struct and C3's PDA reproduce), not a lesson detail — so it must be frozen before authoring, not during.

Spec: `courses/CATALOG.md` — 8 lessons in 3 modules (3/3/2), trackLevel 1, xpPerLesson 10. Artifact: a published account-inspector + the learner's own confirmed devnet transaction + funded wallet.

**Why only 8 lessons**: this is the front door and the funnel. It is deliberately short so a learner reaches first code fast — time-to-first-win is the activation lever, and the old `solana-fundamentals` (12 lessons, which this replaces) had a long runway before anyone wrote anything.

Graded blocks are **pure functions over injected fixtures** (the C5 pattern); live SDK usage lives in prose. Teach `@solana/kit` 7.0.0, never web3.js v1.

Done when: 8 lessons per CATALOG.md, shapes fading, pins registry-verified, content-lint 0 errors, gated by a non-author.

thomgabriel:
Dependency cleared: **#599 reference vault is deployed + frozen** (`D7ZFoWvEG5NBnkJy6iC98rhwj2qhgq8xhSD42cdTRAQd`, devnet, authority=none), record + IDL at `onchain-academy/reference-vault/`. C1 authoring (solana-for-web-devs) is no longer blocked on the vault. Graded-TS grading model is also settled (#600: pure functions over injected fixtures). Ready to author when scheduled.

thomgabriel:
**Owner decisions recorded (2026-07-28):** C1 proceeds OUTLINE-FIRST — draft the 8-lesson outline from catalog spec L1–L8, owner approves, then full authoring. Hard constraint at on-chain creation: `creator` = a REAL instructor wallet (PB-7 — the four live courses all repeated the platform-authority mistake; C1 must not be the fifth). Unblocked by the frozen reference vault (#599) + PB-1 (#600, pure functions over injected fixtures).

[...]

PR #21, opened 2026-02-18
feat: Next.js Frontend Application — Superteam Academy LMS

Mint-Claw:
## 🎓 Superteam Academy Frontend

Full-featured Learning Management System frontend built with **Next.js 14**, completing the monorepo with the existing on-chain program.

### What's Included

| Feature | Details |
|---------|---------|
| **Framework** | Next.js 14 (App Router) + TypeScript |
| **Styling** | Tailwind CSS, dark/light themes |
| **Wallet** | Phantom & Solflare via `@solana/wallet-adapter` |
| **Certificates** | Compressed NFTs via Metaplex Bubblegum (~$0.001/cert) |
| **Token Gating** | SPL token verification for premium courses |
| **Database** | Prisma ORM + PostgreSQL |
| **i18n** | Full Portuguese (pt-BR) localization |
| **Courses** | 6 courses: Solana fundamentals → tokenomics |
| **Quizzes** | Multiple-choice with 70% pass threshold |
| **Gamification** | Points, leaderboard, progress tracking |
| **Admin** | CMS for course CRUD operations |

### Pages

- `/` — Landing page with course highlights
- `/courses` — Catalog with search & difficulty filter
- `/courses/[slug]` — Course detail with modules/lessons
- `/courses/[slug]/lessons/[id]` — Lesson viewer with code editor
- `/dashboard` — Student dashboard with activity feed
- `/leaderboard` — Top students ranking
- `/certificates/[id]` — NFT certificate viewer
- `/profile` — User profile & wallet info
- `/admin` — Course management CMS
- `/paths` — Learning paths overview
- `/settings` — User preferences

### Setup

```bash
cd app
cp .env.example .env.local
npm install
npm run db:push
npm run db:seed
npm run dev
```

### Architecture

See `app/ARCHITECTURE.md` for detailed component diagrams, data flow, and integration points with the on-chain program.

---

**Bounty**: Superteam Brazil Academy LMS
**Repo**: [Mint-Claw/superteam-academy](https://github.com/Mint-Claw/superteam-academy)

PR #794, opened 2026-07-27
fix(start): route learner segments to the live ladder, not deactivated courses

thomgabriel:
## What

`SEGMENT_ENTRY_COURSE` (`apps/web/src/lib/courses/learner-segment.ts`) routed learner segments to courses that were **retired and deactivated on-chain**:

| Segment | Old target (dead) | New target (live) |
| --- | --- | --- |
| 1 — CORE web2 | `course-solana-fundamentals` | **C2** `course-rust-for-program-devs` (live entry rung) |
| 2 — web3 dev | `course-anchor-framework` | **C3** `course-building-first-program` (segment 2 "enters here") |
| 3 — beginner | `course-solana-fundamentals` | **C2** `course-rust-for-program-devs` (shares segment 1's entry) |

## Why it's a live UX regression

Since tonight's on-chain deactivation, the landing CTA + `/start` funnel deep-links pointed at deactivated courses. The `#717` resolver's sync gate (`resolveEntryLessonHref`) saved us from 404s by falling back to `/courses` — but the deep-link **into flagship lesson 1 was dead**; every segment dumped to the catalog instead.

Remapped to the live "Zero to Deployed" ladder (courses-academy `CATALOG.md` §2, trackId 1 / trackLevel 1→5). C1 `course-solana-for-web-devs` is the intended segment-1/3 on-ramp but is **not in the committed bundle yet**, so segments 1 & 3 enter at the live rung C2 today — `TODO(#599/#673)` in the constant names the one-line move for when C1 lands.

## The regression test this class needed

`entry-course-live.test.ts` (new) runs the **real** resolver against the **real committed bundle**, mocking only the on-chain sync gate. It asserts (a) no segment enters a retired course, and (b) every segment entry deep-links to a real lesson (not the `/courses` fallback). The existing `entry-lesson`/`start/routes` suites mock the bundle away and route a synthetic `course-x` — which is exactly why a dead-course table shipped green.

Verified it goes **red** on the old values:
```
× no segment enters at a retired course
  AssertionError: segment 1 entry "course-solana-fundamentals" is a retired course
```

## Also

- `start/routes.test.ts`: renamed stale `FUNDAMENTALS`/`ANCHOR` locals → `SEG1_ENTRY`/`SEG2_ENTRY` (they read from the constant, so behaviour is unchanged — the names were just misleading now).
- `api/content/params.ts`: freshened an **inert** `ID_PATTERN` doc example that cited the retired `course-anchor-framework` (comment only — checked, it is not a live reference; the route validates shape, not membership).

## Checks
- `tsc --noEmit` clean · full web vitest suite **1971 passed** (incl. the new 3) · eslint + prettier clean on changed files.

## Notes
- Companion work — courses-academy path rewiring + course retirement — is in flight. This app-side constant fix is **independently safe** because C2 (`rust-for-program-devs`) is already live; if C3 weren't yet synced, segment 2 would degrade to the catalog fallback (never a 404), still strictly better than a deactivated target.
- **Do not merge** — hand to the gate.

### DIFF
--- a/apps/web/src/app/[locale]/(marketing)/start/__tests__/routes.test.ts
+++ b/apps/web/src/app/[locale]/(marketing)/start/__tests__/routes.test.ts
@@ -12,8 +12,8 @@ const byId = vi.mocked(getCourseById);
 const lessonsOf = vi.mocked(getCourseLessons);
 
 // Use each entry course's id as its slug for the fixtures (1:1, arbitrary).
-const FUNDAMENTALS = SEGMENT_ENTRY_COURSE[1];
-const ANCHOR = SEGMENT_ENTRY_COURSE[2];
+const SEG1_ENTRY = SEGMENT_ENTRY_COURSE[1];
+const SEG2_ENTRY = SEGMENT_ENTRY_COURSE[2];
 
 beforeEach(() => {
   byId.mockReset();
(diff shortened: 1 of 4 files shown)


thomgabriel:
**Orchestrator routing record — launch-funnel entry table → queuing for gate (pairs with courses-academy #21).**

**My independent verification** at `b1fc8535`: typecheck exit 0; full suite **1971 tests, exit 0**.

The finding was worse than my dispatch stated: **all THREE segments pointed at retired courses** (1 & 3 → fundamentals, 2 → anchor-framework) — the entire /start funnel and landing deep-link have been degrading to the catalog fallback since the window. Remapped per CATALOG §2: segments 1/3 → C2 (live entry rung, with a TODO naming the C1 move at #599/#673 — C1 confirmed absent from the committed bundle), segment 2 → C3 (its documented entry). The params.ts hit was inert (doc-comment example) and freshened.

**The regression test closes the class**: `entry-course-live.test.ts` runs the REAL resolver against the REAL committed bundle (mocking only the sync gate), asserting no segment enters a retired course and every entry deep-links to a real lesson — verified RED on the old values. The existing suites mocked the bundle away, which is exactly why the dead table shipped green.

**Safety**: independently mergeable — C2 is live; an unsynced target degrades to the catalog fallback, never a 404. Sequencing note for the gate: this + #21 + the post-merge content.lock bump together complete the catalog migration.

thomgabriel:
@claude review

thomgabriel:
Gate: merged. CI fully green; claude-review no-blocking (bundle ids + resolver path verified statically). Gate replayed the red-proof independently: PR head with only learner-segment.ts reverted to main → entry-course-live.test.ts fails on 'segment 1 entry is a retired course'; restored → 7/7 green. C2/C3 on-chain liveness confirmed by the gate from the catalog-swap byte-verification. Branch closed — follow-ups (the C1 retarget at #673) go on a fresh branch.

PR #796, opened 2026-07-27
feat(gamification): toast server-granted XP — daily quests + surprise bonus (#790)

### LINKED ISSUE #790: Real-time XP feedback: server-granted XP (daily quests, surprise bonus) lands silently — no toast/confetti
Owner report (2026-07-27, live): completed daily quests ('Complete a Challenge' +35, 'Complete a Lesson' +25) and received a Surprise bonus (+40) with **no popup** — the XP just appears in the header. Lesson completions have client-side toast/confetti; server-granted XP has nothing, which reads as numbers moving for no reason (the owner literally couldn't tell why the bonus arrived).

Mechanism: quest XP is enqueued by `get_daily_quest_state` and awarded via the login-drain queue; the surprise bonus fires in `maybeAwardSurpriseBonus` inside the server completion pipeline. Neither path emits a client event.

The plumbing half-exists:
- **Daily quests**: `/api/quests/daily`'s response already returns `justAwarded: true` per quest (`v_just_awarded` in `get_daily_quest_state`) — nothing renders it. The dashboard quests widget should fire the standard XP toast + counter animation when it sees `justAwarded`, using the quest's `xpReward`.
- **Surprise bonus**: `gamification.surpriseBonus` i18n string already exists ('Surprise bonus! +{amount} XP'). The award happens in the lesson-completion server path, so the cleanest signal is including `surpriseBonus: {amount}` in the completion response (or the webhook-driven XP delta the client already refetches) and toasting it with the existing string + confetti.

Acceptance: quest completion and surprise bonus produce the same class of celebratory feedback as lesson XP — no XP change without a visible cause. Keep it client-side presentation; no new XP paths.

thomgabriel:
Closes #790. Server-granted XP landed silently — daily-quest rewards (+25/+35) and surprise bonuses (+40) moved the header number with no toast/confetti, so the owner couldn't tell why. This wires both into the **existing** toast/celebration machinery, riding the responses the dashboard **already polls** — no new push channel, no new XP paths.

## Daily quests
`/api/quests/daily` already computes `justAwarded` per quest (`v_just_awarded`, flipped true on the single poll where `xp_granted` goes false→true) but **dropped it**. Now it's surfaced (`DailyQuest.justAwarded`), and the dashboard poll fires the XP counter (`dispatchXpGain`) + a success toast (new `gamification.questReward`, ×3 locales) on that poll.

**Dedupe:** in-memory `Set` keyed on `questId + period`. Safe by construction — the server flips `justAwarded` **false** after that first poll, so a reload re-polls and simply gets `false`; the client Set is belt-and-suspenders against same-session re-processing.

## Surprise bonus
Detected from the recent `xp_transactions` the dashboard already fetches, firing the **existing** surprise-bonus toast + `celebrate("surprise-bonus")` (its i18n string + "none" confetti tier already existed).

The tricky part, handled: surprise rows are **permanent** (no one-shot flag like `justAwarded`), so a naive re-scan would re-toast on every reload. The seen-set is therefore **sessionStorage-backed** and **seeded silently on first observation per tab** — existing bonuses are marked seen without toasting (no stale storm on load); only bonuses that appear on a **later** poll toast. The set is **shared with the Realtime path** (`use-gamification-events`) via `claimSurpriseBonus`, so whichever channel sees an award first toasts and the other is a no-op — **never a double toast** when both are live.

## Dedupe design (summary)
| signal | source | dedupe | why |
|---|---|---|---|
| quest reward | `/api/quests/daily` `justAwarded` | in-memory Set, `questId:period` | server is one-shot (justAwarded→false) |
| surprise bonus | recent `xp_transactions` scan | **sessionStorage** set + seed-silent-on-init, shared with Realtime | rows are permanent ⇒ must survive reload + not storm on first load |

## a11y & i18n
- Toasts announce via the existing `aria-live="polite"` ToastContainer.
- New `questReward` string in en / pt-BR / es.

## Tests (`server-xp-feedback.test.ts`, 8)
justAwarded → toast once; re-poll → no dup; no-award → nothing; new-day re-toasts; surprise **seeds silently** then toasts a newly-appeared bonus once; shared-dedupe with the Realtime `claimSurpriseBonus`; idempotency_key fallback.

## Verify
- `pnpm --filter web test` — 215 files / 1976 tests pass
- `pnpm --filter web lint` — clean

## Note for the gate
The surprise-bonus toast was **already** wired via Supabase Realtime (`use-gamification-events`); its prod silence most likely means the `20260630120000_enable_realtime_gamification` publication migration isn't applied there (the #708/#750 ledger-divergence family). This PR makes both feedbacks work **regardless** of Realtime by also riding the deterministic poll — worth verifying that migration's application separately.

---

## Round 2 (gate bounce fix) — head 0b74a13

**Fix:** `dispatchXpGain(amount)` added inside the poll-detected surprise-bonus loop in `use-dashboard-data.ts`, symmetric with the quest-reward loop and the Realtime handler — the header XP counter now animates on the poll path (the only live channel today per #797).

**Red-proof (rule 2), independently replayed by the orchestrator:** new regression test `use-dashboard-data.test.tsx > useDashboardData — poll-detected surprise bonus > dispatches an xp-gain for a surprise bonus the poll newly detects` run against **bfe038b** (pre-fix) fails with `AssertionError: expected "vi.fn()" to be called with arguments: [ 25 ] — Number of calls: 0` (the test first awaits `dispatchSurpriseBonus(25)`, proving the poll path executed in the buggy build). Green at this head.

**Also (reviewer non-blocking):** keep-in-sync comments added at BOTH divergent dedupe-key fallback sites (`surpriseKey()` in `server-xp-feedback.ts` and the inline key in `use-gamification-events.ts`).

**Verified at 0b74a13 (agent + orchestrator re-run in clean clone):** typecheck 6/6 workspaces; web suite **216 files / 1977 tests** (+1 file/+1 test = the regression test); content-schema 170, content-lint 107, deploy 18.

**#797 update superseding the "Note for the gate" above:** publication-missing theory ruled out by gate prod SQL; root cause diagnosed as anon Realtime socket + own-row RLS — see the diagnosis on #797.

---

## Round 3 (gate bounce fix) — head 852f705

**Fix:** the sessionStorage dedupe keys (surprise-bonus seen-set + init flag, AND `firedQuestKeys`) are now namespaced by the authenticated user id (`${KEY}:${userId ?? ""}`). A user id with no init flag seeds silently exactly like a fresh tab — that IS the re-seed-on-account-switch, with no explicit change detection to get wrong. Absent id (SSR/anon) collapses to the single `""` namespace, so single-user behavior is byte-identical; all round-1/2 helper tests unchanged and green. `userId` is an optional trailing param threaded from both the poll path (`use-dashboard-data.ts`) and the Realtime path (`use-gamification-events.ts`).

**Red-proof (rule 2), independently replayed by the orchestrator:** regression test `use-dashboard-data.test.tsx > … > re-seeds silently for a new user — never storms user B with their own history` run against **0b74a13** (pre-fix) fails with `AssertionError: expected "vi.fn()" to not be called at all, but actually been called 2 times` (calls: [20], [35] — user B inheriting user A's global init flag toasts B's entire history). Green at this head.

**Verified at 852f705 (agent + orchestrator re-run in clean clone):** typecheck 6/6; web suite **216 files / 1978 tests** (+1 test vs round 2).

### DIFF
--- a/apps/web/src/app/api/quests/daily/route.ts
+++ b/apps/web/src/app/api/quests/daily/route.ts
@@ -89,12 +89,14 @@ export async function GET() {
 
     const progressMap = new Map<
       string,
-      { currentValue: number; completed: boolean }
+      { currentValue: number; completed: boolean; justAwarded: boolean }
     >();
     for (const row of progressRows) {
       progressMap.set(row.questId, {
         currentValue: row.currentValue,
         completed: row.completed,
+        // #790: surface the one-shot award signal so the client can toast it.
+        justAwarded: row.justAwarded,
       });
     }
 
@@ -111,6 +113,7 @@ export async function GET() {
         currentValue: progress?.currentValue ?? 0,
         completed: progress?.completed ?? false,
         resetType: q.resetType,
+        justAwarded: progress?.justAwarded ?? false,
       };
     });
 
--- a/apps/web/src/hooks/__tests__/use-dashboard-data.test.tsx
+++ b/apps/web/src/hooks/__tests__/use-dashboard-data.test.tsx
@@ -1,0 +1,155 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { renderHook, waitFor } from "@testing-library/react";
+import {
+  __resetServerXpFeedbackForTests,
+  pickSurpriseBonusToasts,
+} from "@/lib/gamification/server-xp-feedback";
+
+// Real server-xp-feedback pickers between the hook and the toast facade — the
+// point of this test is the HOOK's wiring of a poll-detected surprise bonus, so
+// the picker stays real and only the dispatch surface is spied.
+const h = vi.hoisted(() => ({
+  dispatchXpGain: vi.fn(),
+  dispatchSurpriseBonus: vi.fn(),
+  dispatchToast: vi.fn(),
+  celebrate: vi.fn(),
+  transactions: [] as unknown[],
+}));
+
+vi.mock("next-intl", () => ({
+  useTranslations: () => (key: string) => key,
+}));
+
+vi.mock("@/hooks/use-gamification-events", () => ({
+  dispatchXpGain: h.dispatchXpGain,
+}));
+vi.mock("@/components/gamification/surprise-bonus-toast", () => ({
+  dispatchSurpriseBonus: h.dispatchSurpriseBonus,
+}));
+vi.mock("@/components/ui/toast-container", () => ({
+  dispatchToast: h.dispatchToast,
+}));
+vi.mock("@/lib/gamification/celebration", () => ({
+  celebrate: h.celebrate,
+}));
+
+// Content bundle queries — never reached before the surprise-bonus loop, but
+// stubbed so the poll runs clean instead of hitting the network.
+vi.mock("@/lib/content/client-queries", () => ({
+  getCourseLessonOrders: vi.fn(async () => []),
+  getCoursesByIds: vi.fn(async () => []),
+  getLessonsByIds: vi.fn(async () => []),
+  getRecommendedCourses: vi.fn(async () => []),
+  getAllAchievements: vi.fn(async () => []),
+  getAllLessonSkills: vi.fn(async () => []),
+}));
+
+vi.mock("@/lib/services", () => ({
+  getProgressService: () => ({
+    getXP: async () => 0,
+    getStreak: async () => ({
+      available: true,
+      currentStreak: 0,
+      longestStreak: 0,
+      lastActivityDate: "",
+      streakHistory: {},
+      frozenDays: [],
+      freezesRemaining: 0,
+    }),
+  }),
+}));
+
+// Chainable Supabase stub. Terminal shape depends on the table + whether the
+// query is a head/count read: profiles resolve a single row, the recent
+// xp_transactions scan resolves the seeded surprise-bonus row, everything else
+// resolves an empty set.
+vi.mock("@/lib/supabase/client", () => ({
+  createClient: () => ({
+    from(table: string) {
+      let columns = "";
+      let head = false;
+      const builder: Record<string, unknown> = {
+        select(cols: string, opts?: { head?: boolean }) {
+          columns = cols;
+          head = Boolean(opts?.head);
+          return builder;
+        },
+        eq: () => builder,
+        gte: () => builder,
+        order: () => builder,
+        limit: () => builder,
+        single: () =>
+          Promise.resolve({
+            data: { username: "Builder", name_rerolls_used: 0 },
+          }),
+        then(
+          resolve: (v: unknown) => unknown,
+          reject?: (e: unknown) => unknown
+        ) {
+          const value = head
+            ? { count: 0 }
+            : table === "xp_transactions" && columns.includes("amount")
+              ? { data: h.transactions }
+              : { data: [] };
+          return Promise.resolve(value).then(resolve, reject);
+        },
+      };
+      return builder;
+    },
+  }),
+}));
+
+import { useDashboardData } from "../use-dashboard-data";
+
+beforeEach(() => {
+  h.dispatchXpGain.mockClear();
+  h.dispatchSurpriseBonus.mockClear();
+  h.dispatchToast.mockClear();
+  h.celebrate.mockClear();
+  h.transactions = [];
+  __resetServerXpFeedbackForTests();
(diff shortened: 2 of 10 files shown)


thomgabriel:
**Orchestrator routing record — gamification feedback wiring → queuing for gate.**

**My independent verification** at `bfe038bb`: typecheck exit 0; full suite **215 files / 1976 tests, exit 0**.

Both server-XP signals now toast through the EXISTING celebration machinery, riding responses the dashboard already polls — no new push channel, no new XP paths. The dedupe design is the load-bearing part and it's sound: quests key on the server-one-shot `justAwarded` (flips false after first poll — reloads can't re-toast); surprise bonuses (permanent rows, no one-shot flag) use a sessionStorage-backed seen-set that SEEDS SILENTLY on first observation per tab (no stale-toast storm on load) and is SHARED with the Realtime path via a claim call — whichever channel sees an award first toasts, the other no-ops.

**⚠️ Diagnostic finding for you (the #708/#750 ledger family strikes again):** the surprise-bonus toast was ALREADY wired via Supabase Realtime — its prod silence most likely means the `20260630120000_enable_realtime_gamification` publication migration was never applied to this DB. This PR makes both feedbacks work REGARDLESS of Realtime, but the migration's application status is worth a ledger check on your side. Also surfaced: the Realtime handler was explicitly SUPPRESSING daily_quest XP on the assumption the quests panel showed it — it didn't; the poll now does.

**Cleanly separable if you prefer**: the surprise-bonus half can drop to Realtime-only — implementer flagged it; my read is keep both (poll works today, Realtime is the unverified path).

thomgabriel:
@claude review

thomgabriel:
Gate: **bounced** on the claude-review finding, which is real and load-bearing.

`use-dashboard-data.ts:203-205` — the poll-detected surprise-bonus loop fires the toast + celebration but never `dispatchXpGain(amount)`, so the header counter doesn't animate. This isn't an edge case: gate verified on prod (2026-07-27) that the Realtime publication IS applied yet events are silent (#797) — meaning the poll fallback is the **only** live channel for surprise bonuses today. The exact scenario this PR exists to fix ships half-broken through it.

Required for re-gate (ship-loop rule 1):
1. Add `dispatchXpGain(amount)` in that loop, symmetric with the quest-reward loop and the Realtime handler.
2. Regression test that **fails at bfe038b** (poll-path surprise bonus must dispatch xp-gain) — state the red-proof in the push per rule 2; gate will replay it.
3. Non-blocking but cheap while in there: the reviewer's note on the divergent dedupe-key fallbacks (`surpriseKey()` vs the Realtime inline key) — add the keep-in-sync comment.

Removing needs-gate; re-label when pushed.

thomgabriel:
@claude review

thomgabriel:
Gate: **bounced (round 3)** — round-2 items all verified (gate replayed the red-proof: hooks reverted to bfe038b → regression test fails on the missing xp-gain dispatch; 9/9 green at 0b74a133). One remaining defect, and it is in THIS PR's new code, not pre-existing: the round-1 first-run review flagged it and my bounce failed to carry it — that's on the gate, but it doesn't ship.

**`SURPRISE_SEEN_KEY`/`SURPRISE_INIT_KEY` (server-xp-feedback.ts:65-66) are global, not per-user.** Sign-out (`user-menu.tsx:53-54`) hard-navigates without clearing sessionStorage, so a same-tab account switch inherits the previous user's init flag → the seed-silently branch is skipped for the new user → ALL their historical surprise bonuses toast + celebrate + animate the counter at once. The exact stale-storm this design exists to prevent.

Required:
1. Namespace the seen/init sessionStorage keys (and `firedQuestKeys`) by the authenticated user id; a user-id change re-seeds silently.
2. Regression test: seed as user A, switch to user B with historical bonuses → no toast storm, B's init seeds silently. **Must fail at 0b74a133**; state the red-proof, gate replays it.

Nothing else — rounds 1-2 scope is locked green. Removing needs-gate; re-label when pushed.

thomgabriel:
@claude review

thomgabriel:
Gate: merged after three rounds. Both red-proofs replayed independently by the gate (round-2 missing dispatchXpGain: red at bfe038b; round-3 user-storm: red at 0b74a13 with the exact claimed assertion, [20],[35]); 10/10 green at 852f705f; namespacing verified in-code across all three key families; CI fully green; review confirms rounds 1-2 locked. #790 is now fully addressed: quest rewards and surprise bonuses toast + animate the counter via the poll path, per-user dedupe survives account switches, and the shared claim set keeps Realtime (#800, in flight) double-toast-proof. Branch closed.

Issue #797, opened 2026-07-27
Realtime gamification events silent on prod despite publication being applied — diagnose channel path


thomgabriel:
Gate verification during #796: the suspected cause (missing `20260630120000_enable_realtime_gamification` publication) is **ruled out** — prod `supabase_realtime` publication contains `xp_transactions`, `user_xp`, `user_achievements`, `certificates` (verified 2026-07-27 via SQL against pywhtmidcrptomrabbrw).

Yet the owner's screenshots show the surprise-bonus toast never fired, meaning `use-gamification-events` didn't receive the INSERT. Candidate causes to check, in order:
1. RLS-scoped Realtime: postgres_changes on RLS tables requires the subscriber's JWT to pass the own-row SELECT policy — verify the channel is created on the authed client (not the bare anon client) and the auth token is set on the socket.
2. `REPLICA IDENTITY` on `xp_transactions` (needed for RLS-filtered streams to evaluate old/new rows).
3. Channel topic/filter mismatch vs the actual insert (source column filters etc).

#796 makes the UX work regardless via the poll path (shared dedupe means fixing Realtime later cannot double-toast). This issue is diagnosis-only.

thomgabriel:
Orchestrator: **diagnosis complete** (dedicated investigator + claims re-verified file-by-file on main). No code changed — findings only, per issue scope.

## Root cause (high confidence): anon Realtime socket + own-row-only RLS = silent drop

Two facts combine:

1. **The browser Realtime socket carries no user JWT.** `use-gamification-events.ts` creates a fresh `createClient()` (line 44) and calls `.subscribe()` directly (line 190). `lib/supabase/client.ts` is a plain `createBrowserClient` with the ANON key — grep confirms **no `realtime.setAuth` / `accessToken` wiring anywhere in `apps/web/src`**. Realtime is a separate WebSocket: the cookie session that makes PostgREST reads work does NOT authenticate the socket unless something sets the token on it. Nothing does — the socket joins as `anon`, `auth.uid()` is NULL.
2. **RLS on `xp_transactions` is own-row only.** `20260624181348_tighten_leaderboard_exposure.sql:91-92` dropped the broad `USING (true)` SELECT policies on `user_xp` + `xp_transactions`; what remains is `auth.uid() = user_id` (baseline.sql:227-228, 234-235). Realtime evaluates the SELECT policy as the socket's role → every row is filtered server-side. The channel still reports SUBSCRIBED, so nothing errors — silent by construction.

[...]

PR #804, opened 2026-07-27
fix(gamification): authenticate the Realtime socket before subscribe (#800)

### LINKED ISSUE #800: Authenticate the Realtime socket: setAuth before subscribe + re-set on token refresh
Fix for the confirmed #797 root cause (diagnosis + prod confirmation there; RLS stays as-is).

**Change** (client-side only, in `use-gamification-events.ts` or the browser client factory):
1. Before building the channel: `const { data: { session } } = await supabase.auth.getSession(); if (session) supabase.realtime.setAuth(session.access_token);`
2. Re-set on `onAuthStateChange` `TOKEN_REFRESHED` (and `SIGNED_IN`).
3. This also fixes the hydration race noted in #797 (subscribe currently fires synchronously before the cookie session recovers).

**Interaction guard**: #796's poll path is belt-and-suspenders — its shared `claimSurpriseBonus` dedupe means enabling Realtime cannot double-toast. Keep that invariant in the test.

**Done when**: with a signed-in session, a fresh `xp_transactions` INSERT for that user arrives over the socket (assert via a test that stubs the channel auth path); token refresh keeps the subscription authed; anon/unauthenticated behavior unchanged (no console errors).

thomgabriel:
Closes #800. Originally stacked on #796; the gate merged #796 (squash 07547d5) while this was in verification, so the branch is now a single commit rebased directly onto main — same delta, verified at the rebased head.

## Fix (client-side only — RLS stays as-is, per #797's gate-confirmed root cause)
`use-gamification-events.ts`:
1. `connect()` is now async: `getSession()` → if a session exists, `await supabase.realtime.setAuth(session.access_token)` **before** building/subscribing the channel. Also closes the #797 hydration race (subscribe used to fire synchronously before the cookie session recovered).
2. `onAuthStateChange` re-calls `setAuth` on `TOKEN_REFRESHED` and `SIGNED_IN`; callback kept synchronous (GoTrue awaits these callbacks during init — reading the token off the event avoids a `getSession` deadlock).
3. Async-effect lifecycle: `cancelled` flag guards both await points; `channel` assigned only after auth+subscribe; cleanup cancels, unsubscribes the auth listener, and `removeChannel`s only if assigned. Event-handler bodies unchanged (re-indented only).
4. Anon path unchanged: no session → no `setAuth`, still subscribes, no errors.

**Placement**: in the hook, not the client factory — grep-confirmed this hook is the ONLY Realtime channel subscriber in `apps/web/src`; `createClient()` is a synchronous singleton used broadly by non-Realtime consumers.

## Red-proof (rule 2), independently replayed by the orchestrator
New tests run against pre-fix base **852f705** (= #796 round-3 content, now squashed into main as 07547d5): **3 failed | 5 passed** —
- `calls realtime.setAuth with the session token BEFORE subscribing` → `AssertionError: expected "vi.fn()" to be called with arguments: [ 'tok-1' ]`
- `re-authenticates the socket on TOKEN_REFRESHED and SIGNED_IN` → `expected null to be type of 'function'`
- `unsubscribes the auth listener on cleanup` → `expected null to be type of 'function'`
Green 8/8 at this head.

## No-double-toast invariant (interaction guard with #796, now on main)
Same user on both channels: Realtime INSERT with `idempotency_key "sb-1"` claims first → the poll path's `pickSurpriseBonusToasts` returns `[]` via the shared per-user seen-set; asserts exactly one `dispatchSurpriseBonus`. Mismatched userIds would fail it, so it genuinely guards the namespaced dedupe.

## Verified at ee8fa4b (orchestrator re-run in clean clone, post-rebase)
typecheck 6/6; web suite **219 files / 1996 tests**; target file 8/8.

### DIFF
--- a/apps/web/src/hooks/__tests__/use-gamification-events.test.tsx
+++ b/apps/web/src/hooks/__tests__/use-gamification-events.test.tsx
@@ -1,18 +1,32 @@
 // @vitest-environment jsdom
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { renderHook, act } from "@testing-library/react";
-import { useGamificationEvents } from "../use-gamification-events";
+import { renderHook, act, waitFor } from "@testing-library/react";
 import {
   resetAnalyticsEventDedupeForTests,
   trackCredentialMinted,
 } from "@/lib/analytics/events";
+import {
+  pickSurpriseBonusToasts,
+  __resetServerXpFeedbackForTests,
+} from "@/lib/gamification/server-xp-feedback";
+import { useGamificationEvents } from "../use-gamification-events";
 
-// Real events module between hook and facade — asserts the wire format AND
-// the cross-path dedupe with the manual-mint observation.
+// Real events + server-xp-feedback modules between hook and facades — asserts
+// the wire format, the cross-path dedupe with the manual-mint observation, AND
+// the shared surprise-bonus dedupe with the #796 dashboard poll.
 const h = vi.hoisted(() => ({
   trackEvent: vi.fn(),
   dispatchCertificateMinted: vi.fn(),
+  dispatchSurpriseBonus: vi.fn(),
   handlers: new Map<string, (payload: unknown) => void>(),
+  // Realtime auth plumbing (#800)
+  getSession: vi.fn(),
+  setAuth: vi.fn(async (_token?: string | null) => {}),
+  authCallback: null as ((event: string, session: unknown) => void) | null,
+  authUnsubscribe: vi.fn(),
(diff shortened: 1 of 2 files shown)


thomgabriel:
@claude review

thomgabriel:
Gate: merged. Red-proof replayed independently (3 red vs main's hook — exact claimed failures; 10/10 green at ee8fa4b incl. the no-double-toast interaction guard with #796). CI fully green; review no-blocking with StrictMode double-mount traced by hand. The three minor observations are filed as a follow-up. #797/#800 closed: the socket now authenticates, so XP events flow over Realtime AND the poll path backstops it. Branch closed.

PR #809, opened 2026-07-27
fix(gamification): harden Realtime setAuth — catch rejections, stale-token race guard, decouple test ticks (#806)

### LINKED ISSUE #806: Realtime setAuth hardening: catch rejections, refresh-race guard, tick-count test coupling
Three non-blocking observations from #804's review, filed by the gate (fix-or-file rule):
1. `void connect()` and the `onAuthStateChange` `setAuth` call lack `.catch()` — `realtime.setAuth`'s rejection contract isn't guaranteed across supabase-js versions; add defensive `.catch(() => {})` to avoid unhandled-rejection Sentry noise.
2. Theoretical stale-token race: TOKEN_REFRESHED firing while `connect()`'s `getSession()` is in flight can momentarily revert the socket to the pre-refresh token (self-heals next refresh). Consider a monotonic token guard.
3. `mountHook()` in the test flushes exactly 3 microtask ticks, coupled to `connect()`'s await count — switch to `waitFor(() => expect(order).toContain("subscribe"))` so future awaits fail loudly instead of under-flushing.

thomgabriel:
Closes #806 (the three non-blocking hardening items the gate filed from #804's review).

## Changes (`use-gamification-events.ts` + its test file only)
1. **Defensive `.catch()`**: new `pushRealtimeAuth(token)` centralizes every socket auth push — bumps the race guard and returns `supabase.realtime.setAuth(token).catch(() => {})`, so a rejected `setAuth` can never surface as unhandled-rejection noise regardless of supabase-js version. `connect()` itself is invoked as `connect().catch(() => {})` covering a `getSession()` rejection too.
2. **Monotonic stale-token guard**: closure-scoped `authGeneration`, snapshotted at `connect()` entry before the `getSession()` await and bumped on every push. If `TOKEN_REFRESHED` lands during the await, the listener has already pushed the newer token (bumping the generation), so `connect()` skips its now-stale token instead of momentarily reverting the socket.
3. **Test decoupling**: `mountHook()` no longer flushes a hardcoded 3 microtask ticks (coupled to `connect()`'s await count) — replaced with `waitFor(() => expect(order).toContain("subscribe"))` on the observable end-state. A future added await fails loudly (waitFor timeout) instead of silently under-flushing; subscribe-never-happens still fails everything.

## Red-proof (rule 2), independently replayed by the orchestrator
New race test run against **main (520d67b)** with the unmodified hook source: `does not revert the socket to a stale token when a refresh races a slow getSession` FAILS with `AssertionError: expected "vi.fn()" to not be called with arguments: [ 'tok-stale' ]` — the other 8 tests pass under the rewritten waitFor harness, proving item 3 kept them meaningful. Green 9/9 at this head.

## Verified at fb08ad1 (agent + orchestrator re-run in clean clone)
typecheck 6/6; web suite **220 files / 2005 tests**; target file 9/9; eslint clean.

### DIFF
--- a/apps/web/src/hooks/__tests__/use-gamification-events.test.tsx
+++ b/apps/web/src/hooks/__tests__/use-gamification-events.test.tsx
@@ -100,13 +100,11 @@ function certInsert(id: string, courseId: string): unknown {
 
 async function mountHook(userId = "user-1") {
   const utils = renderHook(() => useGamificationEvents(userId));
-  // The effect is now async (getSession → setAuth → subscribe); flush the
-  // microtask chain so the channel is subscribed and handlers registered.
-  await act(async () => {
-    await Promise.resolve();
-    await Promise.resolve();
-    await Promise.resolve();
-  });
+  // The effect is async (getSession → setAuth → subscribe). Wait on the
+  // observable end-state (the channel subscribed) rather than a fixed number of
+  // microtask ticks — that way an added await in connect() fails loudly here
+  // (subscribe never observed → timeout) instead of silently under-flushing.
+  await waitFor(() => expect(h.order).toContain("subscribe"));
   return utils;
 }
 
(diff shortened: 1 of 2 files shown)


thomgabriel:
@claude review

PR #810, opened 2026-07-27
docs(onchain): CU budget review + CI drift gate (feeds #141 / G-4)

thomgabriel:
Feeds #141 "[G-4] CU budget reviewed" (does NOT close it — the gate flip is the owner's, at deploy time).

> **This PR was rebuilt mid-review.** The original JS-harness CI drift gate was scrapped after its first-ever CI run exposed memory-unsafety in the litesvm-JS 0.3.3 linux native addon (non-deterministic CU on a byte-identical `.so`, wandering `std::bad_alloc`; full evidence in #815). Drift enforcement now lives in the **Rust differential suite** (litesvm crate 0.12 — deterministic and crash-free on the same runner). `ci.yml` is untouched.

## What this is (5 pure-additive files)
1. **`docs/CU-BUDGET-REVIEW.md`** — the explicit written ruling G-4 asks for, regenerated from the Rust baseline: all 18 instructions (19 measured rows; `update_config` ×2) within the 200k default per-tx ceiling. **Worst case `upgrade_credential` 37,009 CU = 18.50% (81.5% headroom)**; heaviest trio = the mpl_core CPIs; median 6,103; total 196,776. "No ComputeBudget bump" verified by grep (only `packages/deploy`'s loader-write floor exists — separate concern). Re-review triggers documented. `cu-measurement.ts` documented as **macOS-local dev tool only** with an explicit never-CI-wire warning (#815); `CU_COMPARISON.md` annotated so JS-runtime figures are never compared against the Rust baseline.
2. **The CI drift gate** — `tests/differential/tests/cu_budget.rs` + `harness.rs::run_cu()` (`compute_units_consumed` off the same single `send_transaction` that decides success) + committed `CU_BASELINE.rust.md`. Auto-runs via the existing `cargo test` in the Integration job — zero workflow changes. On drift it prints a full expected-vs-actual table AND a ready-to-commit baseline block, so correct numbers are liftable straight from a failing CI log. Determinism is engineered: fixed `keypair_from_seed` seeds pin every address (eliminating ±1500 CU canonical-bump-search wander), fixed warp/airdrops/account orders.

## Review + verification trail
- **Adversarial round 1 (JS design): PASS**, then its determinism caveat fired on the first CI run → #815 discovered → rebuild decision.
- **Adversarial round 2 (Rust design): PASS** — vacuous-pass attacked and held: `run_cu` panics on tx failure; `assert_eq!(rows.len(), 19)`; missing/malformed baseline fails loudly; regen path unreachable in CI (env var in no workflow config) and now value-gated (`CU_BASELINE_REGEN=1` exactly — reviewer's LOW, fixed at b34eb4c with a proof that a wrong baseline + `=0` asserts instead of silently self-healing). Numbers re-verified row-by-row, zero mismatches. Both round-1 LOW nits folded in.
- **CI-validated determinism**: macOS-captured baseline matched CI exactly on the first run (different platform-tools `.so` bytes, identical CU); multiple green Integration runs at the final heads. Local: full differential suite 32/32, fmt + clippy -D warnings clean.


thomgabriel:
Gate: first-firing diagnosis — this is NOT the predicted baseline drift. The drift gate never reached `git diff`: **`pnpm cu:pinocchio` itself crashes on the CI runner** with `terminate called after throwing an instance of 'std::bad_alloc'` (SIGABRT) mid-measurement, at `complete_lesson`, in both runs (jobs 90100574827 / 90100561331). Two observations from the log:

1. `bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)` right before the run — the native bindings aren't built in this job's pnpm install; unclear if causal but it's the first divergence from the local run that worked.
2. The measurements up to the crash look sane and CLOSE to the committed baseline (initialize 16821, create_course 4738, enroll 6200) — so when the harness survives, drift may actually be near-zero on the pinned toolchain.

The failure is a harness/runner problem (native heap exhaustion in the LiteSVM path during the first heavy setup: batch lesson setup before complete_lesson), not a policy problem. Don't regenerate the baseline off this — fix the crash first, then read the real drift signal. Also note: no adversarial-review routing record is on this PR yet; sensitive-lane (workflows) needs it before gate-merge. Same for #811, which will additionally need a rebase over whichever of the pair lands first (both edit ci.yml).

thomgabriel:
Orchestrator status — **direction change on the drift gate, same PR**. The first-ever CI run of the JS CU harness exposed memory-unsafety in the litesvm-JS 0.3.3 **linux x86-64 native addon**: byte-identical `.so` across runs (sha256 `00f677df…` both times) yet run-to-run CU deltas (initialize 15321→16821, register_minter 4359→8859), a `std::bad_alloc` crash point that WANDERS between runs, and corruption present from the first instruction (the 828KB mpl_core fixture load at construction is the likely heap-smash). macOS/arm64 is clean, which is why every prior dev run passed. Upgrading is blocked (anchor-litesvm pins litesvm <0.4).

**No CI gate can stand on that harness.** New plan (in progress on this branch): drift enforcement moves into the **Rust differential suite** (litesvm crate 0.12 — deterministic and already green on this exact runner in this exact job): per-instruction `compute_units_consumed` asserted against a Rust-measured committed baseline, with a full expected-vs-actual table printed on mismatch (correct numbers recoverable from any failing CI log). `cu-measurement.ts` stays as a documented macOS-local dev tool with an explicit do-not-CI-wire warning. The review doc gets regenerated from the Rust baseline; the within-budget ruling stands on ~77% headroom regardless of runtime-accounting digit shifts. A separate issue documenting the litesvm-JS finding follows.

thomgabriel:
@claude review

thomgabriel:
Gate: merged. First firing of the Rust drift gate PASSED in this PR's own Integration run (the design's proof-of-life); review re-derived the budget numbers with no blocking findings; the JS-harness pivot is correctly evidence-driven (#815). Note: the review's ci.yml-rebase caveat is stale — the rebuilt PR touches no workflow files, #811 landed independently. Adversarial substance shown via the pivot + the regen-gating review round; routing record request stands for future sensitive PRs. #141's gate flip remains the owner's at deploy time.

Issue #140, opened 2026-06-23
[G-3] Verifiable build published + hash recorded


Potolski:
> **Code legend / glossary:** see [`docs/TASK-CODES.md`](https://github.com/solanabr/superteam-academy/blob/main/docs/TASK-CODES.md) for how to read task codes, severities, and area labels.

Verifiable build published + program hash recorded — depends on **P1-4**.

**Done when:** the reproducible build hash is published and recorded.

thomgabriel:
Orchestrator audit + pickup (David's items released to the loop): the reproducible build is real (`verifiable-build` CI job: pinned platform-tools v1.54 `cargo build-sbf` + committed Cargo.lock + sha256), but **"recorded" is unmet** — the hash lives only in $GITHUB_STEP_SUMMARY and an expiring artifact; no durable record exists anywhere in the repo (search-verified). The loop is landing a separate CI PR that durably records the hash (CI-maintained `docs/PROGRAM-HASH.md` on main-branch builds, loop-guarded). **This gate issue itself stays open** — the deployed-bytes-vs-build-SHA reconciliation is a deploy-time step, and the flip is the owner's. Removing on-hold-david per the 2026-07-27 owner rule change (pickup = strip).

thomgabriel:
Board audit 2026-07-31 — the ask is met. .github/workflows/ci.yml carries both halves of this gate: the verifiable-build job ("Verifiable build (program hash)", :525) building onchain_academy_pinocchio.so under pinned Solana 3.1.10 / platform-tools v1.54 with SHA-256 computed at :585, and the publish-hash job (:658) publishing to the program-hash branch on every main push whose hash changed, idempotently.

The published record exists today at origin/program-hash:docs/PROGRAM-HASH.md — SHA-256 00f677df54fbb40285acbb0e0c61be47c0cd2a1aae84fdf39f91f2a8b3beb28b, size 210464, program 7NeJaSRyb4Wxay3Tcd9bdpD7T3GWYUQSFyrhG8SgwE8V, source commit 766be344, build date 2026-07-27, including the reproduce command and the solana program dump verification step. Done-when satisfied.

PR #811, opened 2026-07-27
ci(onchain): durably record the verifiable-build program hash (feeds #140 / G-3)

thomgabriel:
Feeds #140 "[G-3] Verifiable build published + hash recorded" (does NOT close it — the gate flip and the deployed-bytes reconciliation are the owner's, at deploy time).

## What this is
The `verifiable-build` CI job already computes a reproducible sha256 of `onchain_academy_pinocchio.so` (pinned platform-tools v1.54 + committed Cargo.lock) — but published it only ephemerally ($GITHUB_STEP_SUMMARY + expiring artifact). This PR makes the record durable:

1. **New "Publish program hash (main only)" step** in the existing `verifiable-build` job — gated `push && refs/heads/main` only (PRs/dispatch still build+summary+artifact, never publish).
2. **Publishes to a dedicated CI-owned `program-hash` branch** (one commit per distinct hash = tamper-evident audit trail), NOT to main — **verified**: main's branch protection has require-a-PR ON, so a GITHUB_TOKEN bot push to main is impossible; the stated fallback design was used. `docs/PROGRAM-HASH.md` on main (this PR's second file) is the static human-readable spec + reproduction guide pointing at `git show origin/program-hash:docs/PROGRAM-HASH.md`; it seeds with an explicit zeros placeholder — no fabricated hash (the implementer's local toolchain is non-canonical v1.51).

## Loop/spam guard (4 layers — the job has no path filter, so it rebuilds on every main push)
1. **Idempotence (primary)**: current published hash is read back from `origin/program-hash`; `OLD==NEW → exit 0` before any commit (round-trip of the render→grep extraction unit-tested), plus a `git diff --cached --quiet` backstop.
2. **Structural**: the push target is `program-hash`, never main — a publish cannot re-trigger the main-gated step.
3. GITHUB_TOKEN pushes don't trigger workflows (GitHub recursion guard; precedent in claude-code-review.yml).
4. `[skip ci]` in the bot commit message as a belt for any future PAT swap.

## Permissions + policy
Workflow-level stays `contents: read`; `contents: write` added at **job level on verifiable-build only**. Zero new `uses:` (pure `run:` git/gh) — no SHA-pin surface; the repo's pin guard passes.

## Verified (agent + orchestrator re-runs)
actionlint exit 0 (shellcheck-backed), extracted publish script shellcheck-clean + `bash -n` OK, YAML parses, pin-guard green across all 8 workflow files, diff is exactly 2 files / +207 with zero TS/Rust source (hence no app test lanes to exercise). Commit mechanics use an isolated `git worktree`; build date = source commit committer-date (reproducible, no wall-clock).

## ⚠ First-run risk
The first main build after merge creates the orphan `program-hash` branch via a live bot push — unverifiable before merge. Rollback: delete the `program-hash` branch + revert this commit; nothing on main or on-chain is touched.

---

## Round 2 — adversarial findings fixed (head 7c9d8c0)

Independent adversarial review (area:ci protocol) verdict: **PASS** with 1 MEDIUM + 4 LOW. The MEDIUM and both cheap LOWs are fixed in this round; two LOWs accepted:

**MEDIUM fixed — pwn-request token isolation (job split):** `verifiable-build` no longer holds any write permission (inherits workflow-level `contents: read`) — the job that executes PR-controlled `build.rs`/proc-macros on labeled PRs never sees a write token. Publishing moved to a new `publish-hash` job: `needs: verifiable-build`, main-push-only `if`, its own `permissions: {contents: write}`, clean checkout of the trusted main ref, record rendered from job outputs (`sha256`/`size`/`solana`/`platform_tools` — plain build-fact strings; `SRC_COMMIT`/`BUILD_DATE` re-derived in the publish job's own main checkout, nothing PR-controlled crosses the boundary).

**LOWs fixed:** `on.push.branches` now `["**", "!program-hash"]` (a future PAT swap can't CI-storm the record branch); `trap`-based worktree cleanup on failure.

**LOWs accepted (acknowledged):** concurrent main pushes → rare non-ff push fails visibly, non-destructively (workflow `concurrency` serializes; no force-push); the main-side seed doc keeps a zeroed placeholder under an explicit "pending first CI run" banner rather than a fabricated hash.

**Honest delta:** this round adds exactly one `uses:` — `actions/checkout` at the SAME SHA already pinned 8× in the repo (needed for the publish job's trusted-main checkout). Pin guard green.

**Re-verified (agent + orchestrator):** actionlint exit 0, shellcheck clean, YAML job graph confirmed (outputs/needs/if/permissions exactly as designed), pin guard green, `uses:` diff vs main = exactly the one pinned checkout, two-job first-run bootstrap re-walked (orphan create → idempotent no-op → changed-hash update; labeled PR → build read-only, publish skipped).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,11 @@ name: CI
 
 on:
   push:
-    branches: ["**"] # every branch — gives fork branches CI before the upstream PR
+    # `**` = every branch (gives fork branches CI before the upstream PR).
+    # `!program-hash` excludes the CI-owned hash-record branch so pushes there
+    # never spin up CI — defense-in-depth vs a future PAT swap (GITHUB_TOKEN
+    # pushes already don't trigger workflows). Negation MUST follow `**` to win.
+    branches: ["**", "!program-hash"]
   pull_request:
     branches: [main]
     # `labeled` lets the verifiable-build job (gated on the `verifiable-build`
@@ -499,14 +503,13 @@ jobs:
   verifiable-build:
     name: Verifiable build (program hash)
     runs-on: ubuntu-latest
-    # Job-level override of the workflow's `contents: read`. Needed ONLY by the
-    # "Publish program hash" step, which commits the hash record to the CI-owned
-    # `program-hash` branch (feeds #140 / G-3). Scoped to this job so no other
-    # job gains write. The default GITHUB_TOKEN cannot push to `main` (branch
-    # protection requires a PR), which is why the record lands on a dedicated
-    # branch — see docs/PROGRAM-HASH.md.
-    permissions:
-      contents: write
+    # No job-level `permissions:` here on purpose. This job runs PR-controlled
+    # build code — `cargo build-sbf` executes the crate's build.rs / proc-macros
+    # — on same-repo PRs carrying the `verifiable-build` label. It therefore
+    # inherits the workflow's read-only `contents: read` and NEVER holds a
+    # write-capable token in a job that runs untrusted code. The durable publish
+    # (which needs write) is a SEPARATE job, `publish-hash`, that runs only on
+    # trusted `main` and never checks out PR code. See #140 / G-3.
     if: >-
       github.event_name == 'workflow_dispatch' ||
       (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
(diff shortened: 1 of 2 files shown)


thomgabriel:
@claude review

thomgabriel:
Gate: merged. CI green at the split-job head; review no-blocking with the security surface independently traced (write token isolated to the trusted-main publish job; cross-job values are build-fact strings only; idempotence reads origin/program-hash, never the placeholder). Adversarial-pass evidence: the round-2 'review MEDIUM' job split — substance satisfied, but per contract the ROUTING RECORD comment should be on the PR at needs-gate time; please attach it going forward. Stale VERIFIABLE_BUILD.md filed as follow-up. **First live publish happens on the next main build — gate will verify the program-hash branch appears with a real hash.** #810 now needs a rebase over this (same-file edit) on top of its harness-crash fix.

Issue #821, opened 2026-07-27
C4 content bug: hand-written client stand-ins invert the deposit account order vs the frozen vault IDL


thomgabriel:
Found during the #599 reference-vault adversarial review (freeze gate). **Not a freeze blocker** — the frozen program/IDL is canonical; this is a courses-academy C4 (dapp-and-sdk-with-kit) content defect that teaches an account order that FAILS against the deployed program.

**Bug:** the hand-written client stand-ins order deposit accounts `[owner, vault, system_program]`:
- idl-to-typed-client/ts/starter.ts
- first-signed-deposit/ts/starter.ts
- first-signed-deposit/ts/solution.ts

The frozen program IDL orders them `[vault, user, system_program]` — **vault FIRST**. The real Codama-generated client honors the IDL (correct); only the mock stand-ins are inverted, so a learner following a stand-in builds a tx whose accounts mismatch the on-chain program.

**Fix (C4 content authors):** correct the three stand-ins to `[vault, user, system_program]`. Do NOT change the program to match the mock — it's frozen and canonical. Cross-check against the deployed reference-vault IDL once #599 lands (address will be posted on #599).

thomgabriel:
Fixed — courses-academy PR #22 merged (`edb9a646`). All deposit account lists in C4 (dapp-and-sdk-with-kit) now order `[vault, user, system_program]` matching the frozen vault IDL; the fix reached wider than first scoped — 10 files including all three `tests.json` (which had asserted the inverted order and would have failed the corrected solutions). Grading coherence verified: solutions pass, starters fail as designed, fixtures match the IDL. content-lint green. Content stages until a content.lock bump.

PR #822, opened 2026-07-27
onchain: record the frozen devnet reference vault (#599 / PB-4)

thomgabriel:
Records the fixed address + committed IDL for the frozen devnet reference vault (spec item 10 / PB-4), the acceptance-criteria deliverable for #599.

**This is a record-only commit** — 2 files under `onchain-academy/reference-vault/` (the frozen program's IDL + a README of coordinates/layout/provenance). No program source, no deploy code, no runtime path. The program itself is already deployed and **frozen (upgrade authority = none, irreversible)** on devnet.

- Program ID: `D7ZFoWvEG5NBnkJy6iC98rhwj2qhgq8xhSD42cdTRAQd`
- Deployed .so sha256: `486367e2…` (on-chain dump byte-matched local before freeze)
- Vault PDA `FY86s1f…`, owner `6JFH…tZsU`, balance 100000000
- Independent adversarial review: **FREEZE-SAFE**; behavioral `deposit(0)→6002` confirmed on-chain.

Unblocks C1 (#673) and the C4 dead-end fallback. Client authors: honor the IDL account order `[vault, user, system_program]` (see #821).

### DIFF
--- a/onchain-academy/reference-vault/README.md
+++ b/onchain-academy/reference-vault/README.md
@@ -1,0 +1,55 @@
+# Devnet Reference Vault — frozen curriculum artifact (#599 / spec item 10, PB-4)
+
+A minimal, **frozen** (upgrade authority = none) devnet deployment of the canonical
+`VaultState` program that C2 (struct/bytes lessons) and C3 (PDA lessons) teach and C4
+reads. Its account layout is a **curriculum-wide decision** — courses may rely on these
+exact bytes and this fixed address. Deployed and frozen 2026-07-27.
+
+## Coordinates
+
+| Field                   | Value                                                                                           |
+| ----------------------- | ----------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


thomgabriel:
**Routing record (gate, sensitive onchain lane):** record-only PR (IDL JSON + README of an already-frozen, already-independently-adversarial-reviewed devnet artifact). Two independent verifications, both PASS: (1) claude[bot] review recomputed all 4 discriminators from scratch + layout math + spec-item-10 cross-ref → blocking: none; (2) an independent gate subagent verified every coordinate against LIVE devnet — program Authority=none (frozen), .so dump sha256 exact-match, PDA independently re-derived, vault account decoded field-for-field, and the on-chain IDL metadata decompressed to byte-identity. The two halves are complementary: claude-review covered the derivable claims, the gate subagent covered exactly the on-chain provenance claude-review flagged as sandbox-unverifiable. Zero discrepancies. Merging.

PR #829, opened 2026-07-28
feat(teach): password-gated course preview from a courses-academy PR link

### LINKED ISSUE #828: Teach: password-gated course preview from a GitHub PR link
## Problem

Courses are authored in the `solanabr/courses-academy` repo and only become visible in the Academy after a PR merges, `content.lock` is bumped, and the bundle is recompiled. Teachers currently have **no way to see how their course will actually look** before publishing — they write markdown blind and discover formatting problems after the fact.

## Proposal

A password-gated **Teach → Preview** page where a teacher pastes a GitHub PR link from the courses repo and sees their course rendered in real Academy formatting.

### Flow

1. Teacher opens `/teach/preview`, enters the access password (default `123` for now, moved to an env var).
2. Pastes a PR URL, e.g. `https://github.com/solanabr/courses-academy/pull/42`.
3. Server resolves the PR's head SHA, downloads that commit's tarball, and runs the **production content compiler** (`extractTarball` + `compileBundle`) in memory.
4. The page renders the compiled courses/lessons — same data shape the live site consumes, so the preview is faithful by construction.

### Why reuse the real compiler

`apps/web/src/lib/content/compile/compile-bundle.ts` already exposes `compileBundle(tree, opts)` as a **pure** function (tree → JSON modules + assets), and `tarball.ts` exposes `extractTarball`. Running the exact production pipeline means the preview can't drift from what actually ships, and teachers get the real schema-validation errors (`ContentValidationError`) rather than a second, approximate renderer.

### Scope

- `GET/POST /api/teach/preview/auth` — password → HMAC-signed, httpOnly session cookie (mirrors the existing `lib/admin/auth.ts` pattern: timing-safe compare, IP rate limiting, SameSite=Strict).
- `POST /api/teach/preview` — `{ prUrl }` → compiled preview payload, or a structured list of validation errors.
- `/[locale]/teach/preview` — password gate, PR input, course/lesson preview UI.
- Pure, unit-tested PR-URL parser (owner/repo/number, rejects other repos and non-PR URLs).
- i18n in `en` / `es` / `pt-BR`.

### Security notes

- Preview password is **not** the admin secret — it gates read-only rendering, never on-chain writes. Separate cookie name so a preview session can never satisfy `requireAdminAuth`.
- Only `solanabr/courses-academy` PRs are accepted; the repo is pinned server-side so the endpoint can't be turned into a general GitHub fetcher (SSRF).
- Compile is memory-only. Nothing is written to `src/content/generated/`, no `content.lock` change, no on-chain effect.
- Tarball extraction already enforces decompressed-size and entry-count caps.

### Out of scope

- Editing content in the browser.
- Publishing / merging from the preview page.
- Per-teacher accounts (the shared password is the interim gate).

## Acceptance

- Wrong password → 401 and no cookie; correct password → preview accessible.
- A valid courses-academy PR URL renders that PR's courses and lessons.
- A PR whose content fails schema validation shows the actual validation issues instead of a generic error.
- A non-courses-academy or malformed URL is rejected with a clear message.

## References

- `apps/web/src/lib/content/compile/compile-bundle.ts` (`compileBundle`, `compileContent`)
- `apps/web/src/lib/content/compile/tarball.ts` (`extractTarball`)
- `apps/web/src/lib/github/github.ts` (`createGitHubClient`, `fetchTarball`)
- `apps/web/src/lib/admin/auth.ts` (cookie/HMAC pattern to mirror)

Potolski:
Teachers author courses in `solanabr/courses-academy` and today only see them rendered **after** a PR merges, `content.lock` is bumped, and the bundle is recompiled. This adds a password-gated preview: paste a PR link, see the compiled course.

Closes #828

## How it works

`POST /api/teach/preview` resolves the PR's head SHA, downloads that commit's tarball, and runs the **production content compiler** — `extractTarball` + `compileContent` — entirely in memory.

Reusing the real compiler is the point: the preview cannot drift from what actually ships, and when content is invalid the teacher gets the **same issues CI prints** (`ContentValidationError` → 422 + issue list) instead of a generic failure.

## What's here

| Path | Purpose |
| --- | --- |
| `lib/teach/pr-url.ts` | Pure PR-URL parser. Repo pinned server-side. |
| `lib/teach/preview-auth.ts` | Password gate: HMAC-signed httpOnly cookie. |
| `lib/teach/preview-compile.ts` | PR head → tarball → compiled courses/lessons. |
| `api/teach/preview/auth/route.ts` | `GET` session probe, `POST` password → cookie. |
| `api/teach/preview/route.ts` | `POST { prUrl }` → preview payload. |
| `[locale]/(platform)/teach/preview/` | Gate, PR input, course/lesson preview UI. |

## Security

- **Separate from admin.** Different cookie (`teach_preview_session`) and secret from `admin_session`, so a preview session can never satisfy `requireAdminAuth`. This gate protects unpublished drafts — content already destined to be public — never funds, secrets, or writes.
- **Repo is pinned** in the URL parser. The endpoint downloads and compiles whatever it's pointed at, so accepting an arbitrary `owner/repo` would make it a general GitHub fetcher (SSRF, plus running the compiler over untrusted trees). Only `solanabr/courses-academy` PRs are accepted.
- **Read-only.** No bundle write, no `content.lock` change, no on-chain effect. `compiledAt: null` so a preview never stamps a wall-clock time that would differ from a real compile of the same SHA.
- Timing-safe password compare, `SameSite=Strict`, per-IP rate limiting on both auth (10/min) and compile (12/min — each call pulls a tarball). Tarball extraction already caps decompressed size and entry count.

## Password

Default `123` as requested, overridable via `TEACH_PREVIEW_PASSWORD` (documented in `apps/web/CLAUDE.md`). **Set a real one before sharing the URL** — the default is committed and therefore public.

## Verification

- 2074 tests pass (18 new: PR-URL parsing incl. wrong-repo/wrong-host/non-PR rejection; session mint/expiry/tamper).
- `tsc --noEmit` clean, ESLint 0 errors, i18n parity across `en`/`es`/`pt-BR`.
- **Not exercised against a live PR** — that needs `GITHUB_TOKEN` and a real courses-academy PR. The compile path is unit-covered at its seams but the end-to-end fetch→compile→render round trip is unverified; worth a manual run before merge.

## Follow-ups (deliberately out of scope)

- Lesson preview currently lists each lesson's block types rather than mounting the live block renderers. Mounting them needs a full `BlockContext` (proof collection, enrollment, build state) that doesn't exist outside a real lesson page — worth doing, but it's its own change.
- Per-teacher accounts instead of a shared password.


Potolski:
## Follow-up: preview no longer requires `GITHUB_TOKEN` (`f70b227`, refs #830)

Testing against [courses-academy#24](https://github.com/solanabr/courses-academy/pull/24) surfaced `GITHUB_TOKEN is not configured` on a checkout where the var is unset — which is any fresh one, since it's optional in `env.server.ts` and nothing fails at boot.

Requiring it was the wrong default: `courses-academy` is **public**, so resolving a PR head and pulling its tarball need no auth at all.

- Authorization is now sent only when a token exists; otherwise the calls go out anonymously and the preview works.
- The tarball fetch no longer goes through `lib/github/github.ts`. That client keeps requiring the token deliberately — its callers are admin paths (publish pin, checks API) where silently dropping to a 60 req/hr anonymous limit would be worse than failing loudly.
- A 403/429 carrying `x-ratelimit-remaining: 0` now says it's a rate limit and names `GITHUB_TOKEN` as the remedy, rather than surfacing `GitHub pulls/24 → 403`.

5 new tests cover the no-token path, the token path, rate-limit detection, a plain 403 that must *not* be misread as one, and 404. 23 teach tests pass; tsc and ESLint clean.

Potolski:
## Full-fidelity preview (`87d1ebd`, closes #831)

The preview now renders the **real** course and lesson pages, not a summary.

New routes, both behind the existing preview cookie:

```
/[locale]/teach/preview/[pr]/courses/[slug]
/[locale]/teach/preview/[pr]/courses/[slug]/lessons/[id]
```

They mount `CourseDetailClient` and `LessonPageClient` directly with preview data, so fidelity is *inherited* rather than reimplemented — hero, curriculum accordion, prose, quizzes and the Monaco IDE all behave exactly as published, and none of it can drift from the live site.

### Two new props, both defaulting to today's behaviour

- **`hrefBase`** — lesson/CTA/prev/next links were hardcoded to `/{locale}/courses`. Threaded through `CourseDetailClient`, `CurriculumAccordion`, `CurriculumPath` and `LessonPageClient` so preview navigation stays inside the preview instead of escaping to `/courses`, where an unpublished course 404s.
- **`readOnly`** — a previewed course isn't on-chain and its lessons aren't in the live bundle, so enrolment is inert and completion neither POSTs an unknown `lessonId` nor banks progress into localStorage.

I chose explicit props over preview-aware branching inside the components so the live path stays literally unchanged: both default to current values, and live rendering is byte-identical.

### Bug found while wiring this

The first cut grouped lessons by a `lesson.course` field **that does not exist**. The bundle links the other way — courses hold lesson `_ref`s — so every course in the preview listed zero lessons. It only escaped notice because the first screenshot never expanded a course. Hydration now goes through the same `projectCourse(..., { fullLessons: true })` the live site uses, which is what produces modules with embedded lessons in the first place.

### Caching

Compiling downloads and expands a repo tarball, and navigating a course is many page loads. Results are memoised per PR for 10 minutes — the **promise** is cached, so concurrent loads share one download, and failures are never cached (a teacher pushes a fix and retries).

### Verification

2079 tests pass (up from 2074), `tsc` clean on all touched paths, ESLint clean, i18n parity across the three locales.

**Not verified in a browser** — the preview server won't run in my environment. The live-page regression risk is covered by the defaults plus the suite, but the preview routes themselves want a manual pass against courses-academy#24.

*(Note: two backticked words got eaten by the shell in the commit body — "lessons were grouped by a `lesson.course` field that does not exist". Not amending, since that would mean a force push.)*

Potolski:
## Fix: new-course preview was unreachable (`b77b669`)

Diagnosed against the live PR — `compilePrPreview(24)` returns all five courses (`solana-for-evm-devs`: 3 modules / 9 lessons), so the server was right and the client was the problem, twice:

1. **The pasted URL had a trailing slash** (`.../pull/24/`). The client parsed the PR number with an ad-hoc end-anchored regex instead of the canonical `parsePrUrl`, so `prNumber` stayed `null` and the *Open course preview* button + lesson links never rendered. Live courses still opened via the catalogue — only the PR-only course had no way in, which matched the report exactly. The client now uses the same parser as the server; regression test added.
2. **Cards read dead fields.** `lessonCount`/`xpPerLesson` don't exist on the projected shape (visible as the bare "lessons" chip in the screenshot). Cards now count `lessonsByCourse` and read `xpPerLessonById`.

24 teach tests pass; tsc + ESLint clean.

Potolski:
## Preview list scoped to the PR's own courses (`f66ddd1`)

The summary list now shows only the courses the PR adds or modifies, with a note ("Showing only the courses changed in this pull request — N unchanged hidden"). Full-page preview routes stay unscoped, so opening an unchanged course by URL still works.

**How the scope is derived:** changed file paths from `/pulls/N/files` (paginated, best-effort) → `courses/<dir>/` names → matched to compiled course slugs (the `dir == slug` convention from CONTRIBUTING).

**Fail open by contract:** if the files call fails, the PR touches no course dir, or any touched dir can't be matched (rename, deletion, broken convention), filtering is disabled entirely. The degraded state is "show everything" — never "hide the teacher's course", which is the failure mode that already bit this page once.

Live-probed against courses-academy#24: `changedCourseIds` resolves to exactly `course-solana-for-evm-devs`, so the list will show the one new course. 29 teach tests pass (5 new for the pure helpers); tsc + ESLint clean.

PR #817, opened 2026-07-27
test(challenge): reconcile AI-lock test to #791's countdown chip — fix red main (#814)

thomgabriel:
Closes #814 (P0 — main red). Test-only, 1 file, gate-prescribed shape.

## What happened
Cross-merge drift: #805's lock test asserted the banner text `/unlocks in \d:\d\d/`; David's #791 (branch predating #805, merged after) redesigned the pane so the countdown is a header chip (`role="status"`, visible `2:59`, think-first messaging in `aria-label`/`title`). Both changes individually correct; main went 1-failed / 2058.

## The reconciliation
The failing test now binds to the merged chip: `getByRole("status")` → tabular countdown `/^\d:\d\d$/`, `title === lock.tooltip`, `aria-label` prefix + tooltip content — and **keeps the Hint-disabled-while-locked assertion unweakened** plus the `lock.title` subtitle check (real merged UI). The no-lock cases additionally assert chip ABSENCE (`queryByRole("status")`); the countdown-elapse test switches its locked/unlocked signal to chip presence→absence with Hint disabled→enabled.

**Other #805 tests audited:** the 4 `challenge-interface.ai-lock` tests mock the pane (immune to the redesign, green unchanged); pre-existing pane/gate tests bind to real UI with no banner-vs-chip assumptions.

## Red-proof, independently replayed by the orchestrator
- At **main (fe0ab94)** unmodified: exactly `1 failed | 9 passed` — `Unable to find an element with the text: /unlocks in \d:\d\d/` (matches the gate's diagnosis from run 30303600546).
- Fixed test file against **unmodified main source**: `10 passed` — proof this is pure test reconciliation, zero behavior change.
- Full verify at **ddd7f9b**: typecheck 6/6; web suite **227 files / 2059 tests**.

Merging this also delivers the first main push since the `[skip ci]` squash slip — the first `program-hash` publish (#811) should fire on it.

PR #851, opened 2026-07-28
fix(ui): AI countdown chip readability + name-reveal modal recurrence (#842, #843)

### LINKED ISSUE #842: AI countdown tooltip is unreadable — the badge sits inside the collapse toggle
## Problem

On a challenge lesson, the AI Partner's think-first countdown carries a hover tooltip explaining *why* the tutor is locked. Hovering shows the help cursor — but the badge is rendered **inside** the header's collapse `<button>`, so any click on it toggles the pane open/closed. The pane moves out from under the pointer and the tooltip is dismissed before it can be read.

## Cause

`ai-partner-pane.tsx` renders the countdown `<span title=...>` as a child of the toggle button, so its clicks bubble to the toggle. Two controls occupy the same pixels.

## Fix

Move the countdown out of the button entirely: header becomes a flex row of [toggle button][countdown badge], so the badge has no click behaviour and the toggle's hit area no longer covers it. Preferred over `stopPropagation`, which would leave overlapping targets and only mask the collision for mouse clicks.

## Acceptance

- Hovering the countdown shows the explanation; the pane does not toggle.
- Clicking the countdown does nothing.
- Clicking anywhere else in the header still toggles.

### LINKED ISSUE #843: Name-reveal modal reappears for learners who already chose a name
## Problem

Logging in with a wallet re-shows the "choose your name" modal to learners who already picked one — repeatedly, across sessions and devices.

## Cause

`NameRevealDialog` decides "already seen" from two signals, and neither is durable:

- `nameRerollsUsed === 0` — but a learner who **accepts the generated name without rerolling** leaves `name_rerolls_used` at `0` forever, so the server never records that they finished.
- `localStorage.getItem("nameRevealSeen")` — the only record that they confirmed, and it is per-browser. A different device, a cleared profile, or a private window all read as "never seen".

So the common path (accept the first name, come back on another device) reliably re-prompts. Nothing about the modal's completion is stored server-side.

## Fix

Persist confirmation in `profiles.prefs` — the existing learner-owned, self-writable JSONB store (own-row UPDATE RLS, bounded by `chk_profiles_prefs_object` / `chk_profiles_prefs_size`), already used for `prefs.nextLesson`. Write `prefs.nameRevealSeen` on confirm and gate the dialog on it, keeping localStorage only as an offline fast path.

Not reusing `name_rerolls_used` as the marker: confirming without rerolling would have to increment it, silently spending one of the learner's rerolls to record an unrelated fact.

## Acceptance

- Confirm the name, then sign in from another browser/device: no modal.
- Clearing localStorage alone does not bring the modal back.
- A genuinely new learner still sees it once.

thomgabriel:
Closes #842
Closes #843

## #842 — AI countdown chip unreadable

**Root cause**: the think-first countdown chip (`role="status"`, added by #791/#805/#817) was rendered as a child of the AI Partner header's collapse-toggle `<button>` in `ai-partner-pane.tsx`. Any click on the chip bubbled to the toggle, collapsing the pane out from under the pointer and dismissing the `title` tooltip before it could be read — two controls occupied the same pixels.

**Fix**: the header is now a flex row of **[toggle button][countdown chip]** — the chip is a sibling of the toggle, not a child (as the issue prefers over `stopPropagation`, which would only mask the collision). Preserved unchanged: `role="status"`, `title` + `aria-label` carrying the tooltip copy, tabular `m:ss` countdown, keyboard access to the toggle (`aria-expanded`/`aria-controls`, focus-visible ring), and the collapse behavior everywhere else in the header. The chip stays visible while the pane is collapsed.

New test: *renders the countdown chip OUTSIDE the collapse toggle; clicking it does not toggle (#842)* — asserts `toggle.contains(chip) === false`, that clicking the chip leaves the pane open, and that the toggle still collapses/expands. No existing #817 assertions weakened (all 11 pane tests pass).

## #843 — name-reveal modal reappears

**Root cause**: "already seen" was decided by `name_rerolls_used === 0` plus `localStorage.getItem("nameRevealSeen")`. A learner who accepts the first generated name never increments rerolls, and localStorage is per-browser — so the common path (accept first name, sign in on another device / cleared profile) reliably re-prompted. Nothing durable recorded completion.

**Fix**: confirmation is persisted to `profiles.prefs.nameRevealSeen` (the learner-owned JSONB store already used for `prefs.nextLesson`), via a read-merge-write so sibling prefs keys are never clobbered. The dialog is gated on that row: localStorage is only an offline fast path, and is backfilled when the row says seen. Prefs read errors fail closed (no spurious prompt). `name_rerolls_used` is untouched — using it as the marker would silently spend a reroll (per the issue).

### Red-proof (bug 2)

At main sha `4172d5c47a4296182aaf49d2aee0ebc5044ddeb3` (the branch point, component unmodified), the new test file `apps/web/src/components/dashboard/__tests__/name-reveal-dialog.test.tsx` fails:

- **`never re-shows for a learner who confirmed on another device (prefs.nameRevealSeen set, localStorage empty)`** — fails with `AssertionError: expected "vi.fn()" to be called at least once`: the component never reads `profiles.prefs` before prompting (and the dialog opens anyway).
- `persists nameRevealSeen to profiles.prefs on confirm, merging other prefs keys` — fails with `AssertionError: expected 0 to be greater than 0` (no prefs write ever happens).

Both pass with the fix; 3 companion tests (new-learner still sees it once, localStorage fast path, rerolls > 0 gate) also pass.

## Verification

- `pnpm typecheck` (root): 6/6 tasks pass
- `pnpm --filter web test` (full): **234 files / 2102 tests pass** (baseline at branch point: 2090)
- ESLint + Prettier clean on all 4 changed files (also enforced by the pre-commit hook)

### DIFF
--- a/apps/web/src/components/dashboard/__tests__/name-reveal-dialog.test.tsx
+++ b/apps/web/src/components/dashboard/__tests__/name-reveal-dialog.test.tsx
@@ -1,0 +1,123 @@
+// @vitest-environment jsdom
+/* eslint-disable import/order -- vi.mock calls must precede importing the component. */
+// #843 — the name-reveal modal must be gated on the DURABLE record
+// (profiles.prefs.nameRevealSeen), not on localStorage or name_rerolls_used:
+// a learner who accepts the first generated name leaves rerolls at 0 forever,
+// and localStorage is per-browser, so the old gate re-prompted on every new
+// device/cleared profile.
+import type { ReactElement } from "react";
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, waitFor } from "@testing-library/react";
+import { NextIntlClientProvider } from "next-intl";
+import messages from "@/messages/en.json";
+
+const { maybeSingle, updatePayloads } = vi.hoisted(() => ({
+  maybeSingle: vi.fn(),
+  updatePayloads: [] as Array<Record<string, unknown>>,
+}));
+
+vi.mock("@/lib/supabase/client", () => ({
+  createClient: () => ({
+    from: () => ({
+      select: () => ({ eq: () => ({ maybeSingle }) }),
+      update: (payload: Record<string, unknown>) => {
+        updatePayloads.push(payload);
+        return { eq: () => Promise.resolve({ error: null }) };
+      },
+    }),
+  }),
+}));
+
+import { NameRevealDialog } from "../name-reveal-dialog";
+
+function renderDialog(props: { nameRerollsUsed?: number } = {}): ReactElement {
+  const ui = (
+    <NextIntlClientProvider locale="en" messages={messages}>
+      <NameRevealDialog
+        isLoading={false}
+        userId="user-1"
+        username="brave-anchor-9000"
+        nameRerollsUsed={props.nameRerollsUsed ?? 0}
+      />
+    </NextIntlClientProvider>
+  );
+  render(ui);
+  return ui;
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  updatePayloads.length = 0;
+  localStorage.clear();
+});
+
+describe("NameRevealDialog — durable already-seen gate (#843)", () => {
+  it("never re-shows for a learner who confirmed on another device (prefs.nameRevealSeen set, localStorage empty)", async () => {
+    // The accept-without-reroll path: rerolls stayed 0, this browser has no
+    // localStorage record, but the profiles row says the reveal was confirmed.
+    maybeSingle.mockResolvedValue({
+      data: { prefs: { nameRevealSeen: true } },
+      error: null,
+    });
+
+    renderDialog({ nameRerollsUsed: 0 });
+
+    // The component must consult the profiles row before deciding to prompt.
+    await waitFor(() => expect(maybeSingle).toHaveBeenCalled());
+    expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
+  });
+
+  it("still shows once for a genuinely new learner (prefs empty)", async () => {
+    maybeSingle.mockResolvedValue({ data: { prefs: null }, error: null });
+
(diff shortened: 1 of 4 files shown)


thomgabriel:
Gate bounce addressed in be66ea6 (both defects, same branch):

**1. Read-failure clobber (BLOCKING)** — `handleNameConfirm` now destructures `{ data, error }` from the confirm-time prefs SELECT and returns early on error, mirroring the fail-closed check the gating effect already uses: the localStorage fast path is kept, the durable write is skipped (another device simply re-prompts). A blind `{ nameRevealSeen: true }` write can no longer clobber sibling prefs keys.

**2. Dismissal ≠ confirmation** — handlers split. Only the real "Ship it!" click (`WalletNameGenerator` `onConfirm`) performs the durable prefs write; `onOpenChange` now routes to a `handleDismiss(open)` that only sets the per-browser localStorage flag and closes. Reasoning: keeping the pre-#843 localStorage behavior on dismissal preserves the existing "stop nagging this browser" UX, while an accidental Escape/overlay/X no longer permanently buries the reveal on every other device — the durable record is reserved for explicit confirmation.

**Red-proof at previous head `2a538f1`** (fix stashed, tests run against the committed component):
- `skips the durable write when the confirm-time prefs read fails — never clobber on a blind merge` → `AssertionError: expected [ { prefs: { nameRevealSeen: true } } ] to have a length of +0 but got 1`
- `does NOT write prefs on bare dismissal (Escape/overlay/X) — only per-browser localStorage` → same assertion — the X-dismiss path wrote prefs.

Both pass at `be66ea6`.

**Re-verification**: root `pnpm typecheck` 6/6 pass; full `pnpm --filter web test` **234 files / 2104 tests pass** (was 2102); name-reveal suite now 7 tests; ESLint + Prettier clean on both changed files (also enforced pre-commit).

thomgabriel:
@claude review

Issue #873, opened 2026-07-28
Finish the LX-F1 event pass: earn_handoff_click, linkedin_share_click, review_completed


thomgabriel:
## Context
The master spec's **LX-F1 event pass** is partial. Verified against `apps/web/src/lib/analytics/events.ts` (all 15 exported track functions audited 2026-07-28): DONE = credential_minted, challenge_started/run/failed/solved, onboarding funnel (started/step/goal/route/completed), quiz_checked, solution_revealed, stuck-nudge pair, next-lesson-plan. **MISSING from the LX-F1 list:**
1. **`earn_handoff_click` (E8)** — the Earn handoff card exists (`apps/web/src/components/certificates/earn-handoff-card.tsx`) with no click event; LX-E4's accept criterion "E8 event fires on click-through" is unmet.
2. **`linkedin_share_click`** — the share surfaces exist (`apps/web/src/components/certificates/credential-share-nudge.tsx`, `apps/web/src/lib/credentials/share.ts`, certificates page) with no event.
3. **`review_completed`** — spec says "later, with LX-B5"; the review spine shipped (`apps/web/src/components/dashboard/review-strip.tsx` — no track calls). Instrument the strip's completion action if the review flow is live; if the flow is still stub-only, record that finding in the PR and skip with a code TODO referencing this issue.

(32a override-rate and 32b comprehension-check events are tracked by their own AI-wall issues — not in scope here; don't duplicate.)

## Where
- `apps/web/src/lib/analytics/events.ts` + `apps/web/src/lib/analytics/README.md` (naming doc — LX-F1 accept criterion)
- Callsites: `earn-handoff-card.tsx`, `credential-share-nudge.tsx` (+ `lib/credentials/share.ts` if the click is handled there), `review-strip.tsx`

## Done when
- [ ] earn_handoff_click and linkedin_share_click fire with cert/course context
- [ ] review_completed instrumented or explicitly dispositioned as above
- [ ] Names documented in the README; tests cover emission
- [ ] A closing checklist in the PR states LX-F1 is now complete (or lists exactly what remains and why)

## References
Parent epic: #841 (item 4). Master spec LX-F1 (docs/superpowers/specs/2026-07-25-launch-experience-master-spec.md lines ~240–242), LX-E3/LX-E4. RESEARCH-RECONCILIATION §4.

PR #880, opened 2026-07-30
docs(deployment): slot-activation insertion checklist

### LINKED ISSUE #862: docs/DEPLOYMENT.md: add the slot-activation insertion checklist
**Repo: `solanabr/superteam-academy`** (monorepo docs; no content or code change).

## Context
`docs/DEPLOYMENT.md` documents the content pipeline and the `content.lock` bump procedure (§ around lines 218–272) but has **no slot-activation insertion checklist** — the step-by-step for safely activating new/changed lessons in a deployed course. This is the operational guard for the #740/#741 class of incident: lesson bitmap index is **array position, not slot**; `slots.lock.json` has no learner-path reader, so reshaping a deployed course's lesson array silently remaps every learner's on-chain progress bits.

## Where
- `docs/DEPLOYMENT.md` — add the checklist adjacent to the existing content.lock bump procedure (after the "bump `\"sha\"` in `apps/web/content.lock`" steps, ~line 270).
- Ground truth to encode: solanabr/superteam-academy#741 (bitmap = array position), #740 (the incident), `courses/*/slots.lock.json` semantics in courses-academy, and the golden-regen step (`content.lock` bump + regenerated bundle committed together — already documented, cross-link it).

## Done when
- [ ] DEPLOYMENT.md has a numbered slot-activation insertion checklist covering at minimum: append-only rule for deployed courses (never reorder/remove), slots.lock verification, how to check whether a course has on-chain enrollments before any reshape, the bump+golden-regen pairing, and post-activation verification
- [ ] The #740/#741 hazard is stated explicitly with issue links
- [ ] No claims contradict `docs/ADMIN.md`'s pipeline description (cross-check lines ~66–110)

## References
Parent epic: solanabr/superteam-academy#840 (item 4). RESEARCH-RECONCILIATION §4. Incident/constraint: #740, #741.

Potolski:
Adds the slot-activation insertion checklist to `docs/DEPLOYMENT.md`, adjacent to the `content.lock` bump procedure. Docs only — no code, no content.

Closes #862

## The fact it encodes

> A learner's on-chain progress bit is the lesson's **array position**, not its slot id (#741).

`Enrollment.lesson_flags` is a `[u64; 4]` bitmap where bit _n_ means "the lesson at index _n_ is complete". `slots.lock.json` records intended slots but **has no runtime reader** — no client, no API route, no on-chain instruction. So inserting, deleting or reordering a lesson in a deployed course silently remaps every learner's completed bits. Nothing errors; progress just becomes wrong. That is the #740 incident.

## What the checklist covers

- **Safe vs unsafe** operations, split explicitly (append and edit are free; insert, delete, reorder are not).
- A **pre-bump diff check** — every pre-existing lesson id must keep its index — with the exact `git diff` to run against the pinned SHA.
- The **undeployed-course escape hatch**: no learners, no bitmap, so a reshape is free.
- The **retire path** for removing a lesson without remapping: clear its bit in `active_lessons` via `update_course` and leave the array position occupied. Includes why a retired slot never blocks completion (the gate is a subset test over the live mask — `isCourseComplete` in `lib/solana/bitmap.ts`).
- A cross-link back to the existing golden-regen requirement (lock bump + regenerated bundle committed together).


Potolski:
You're right, and I verified every claim before rewriting rather than taking either version on trust. Fixed in `4ae1fe4`.

## What I confirmed on `main`

| Your claim | Verified |
| --- | --- |
| `getLessonSlot()` reads `slots.lock.json` at runtime | ✅ — docstring literally says using array position "**was** the #741 bug" (past tense) |
| That slot is what reaches the chain | ✅ — `complete/route.ts:439` → `complete_lesson` (565-568), the bitmap idempotency check (469-472), and `user_progress.lesson_index` (491, 584) |
| `findLessonIndex` is display-only | ✅ — explicitly documented as such |
| `deriveActiveMask()` builds the mask from the lock | ✅ |
| CI gate-3 enforces the slot invariant | ✅ — fails on a moved slot, a reused retired slot, or a shrinking `next` |

All five hold. The section described the pre-#741 architecture.

## How it happened

I took the framing verbatim from #862's issue body ("bitmap index is array position, not slot; `slots.lock.json` has no learner-path reader") and never checked it against the code. That's doubly annoying because I'd caught exactly this failure mode earlier the same day on #873 — where two of three "missing" events turned out to already ship — and then didn't apply the same skepticism here.

Your point about the prescribed procedure being *actively* wrong is the sharper one: hand-diffing `course.yaml` order would flag safe reorders as unsafe **and** miss real lock violations, since display order is decoupled from slot assignment by design. That's worse than no checklist.

## What the section says now

It documents the system that exists — slot lock is authoritative, read at request time, machine-enforced by gate-3 — and explicitly warns against the hand-diff.

Then it scopes the checklist to the one step I could find that automation genuinely does **not** cover: **pushing the on-chain mask after retiring a slot.** `content.lock` ships content to the app but doesn't touch the chain; the mask goes out via the admin course sync. Because `finalize_course` is a subset test (`flags & active == active`), a stale mask that still lists a retired slot as active **permanently blocks completion** for anyone who hadn't already completed the now-deleted lesson — and they can't complete it, it's gone from the bundle. Adding a lesson without syncing is benign by comparison.

Also records the fail-closed symptom (completion 500s on an unslotted lesson) and why the fix is regenerating the lock rather than adding an array-index fallback.

## On #862's acceptance criteria

You flagged you couldn't fetch the issue body. Its "Where" section asks to encode "#741 (bitmap = array position)" as ground truth — so **the issue itself carries the same stale premise**, which is where I got it. I've switched the commit to `Refs #862` rather than `Closes`, since the issue needs its premise corrected before it can be closed cleanly. Happy to file that correction, or fold it in here if you'd rather.

Potolski:
@claude check if the blocking issues were solved

Issue #838, opened 2026-07-28
AI-wall epic: wall boundaries, attempt-gate nudge, comprehension check, capstone AI-off, model routing (spec items 32/32a/32b/33/3b)


thomgabriel:
The one whole workstream the 2026-07-25 program dropped (RESEARCH-RECONCILIATION §2 — item 32 was never filed as an issue at all).

**EPIC — tracked by child issues: #864 (item 32 — assist ladder + wall boundaries), #865 (32a — attempt-gate nudge replacing the #770 hard lock), #866 (32b — comprehension-check first-attempt accuracy metric), #867 (33 — capstone AI hard-off via the credential-gate constant), #868 (3b — per-action model routing) — created 2026-07-28. This issue closes when all children close.**

**Step 0 (the Gemini curls, AIE-04/AIE-05) is COMPLETE** — recorded results in the 2026-07-28 comment below; #868 is unblocked and the external cost-quote embargo can lift once the figures are folded into the economics doc's tables.

Original context/evidence: zero Socratic/tier hits outside spec docs; no nudge/override/check events existed in `analytics/events.ts` / `use-ai-partner.ts`; the only AI gate was `aiSuppressed` for quizzes. Source: RECONCILIATION §2 + `ai-tutor-economics.md` (the doc's 92 items were verified by the §8.1 completion pass). The five owner decisions **D-4/D-5/D-6/D-7/D-8** come due as ONE batch when #864 is built — see the completion-pass comment below.

thomgabriel:
**Step 0 COMPLETE — both Gemini curls run and recorded (2026-07-28, prod GEMINI_API_KEY, v1beta):**

**AIE-04 reachability — all VERIFIED REACHABLE (HTTP 200 generateContent):**
- `gemini-3.5-flash` (current pin) ✓
- `gemini-3.5-flash-lite` ✓
- `gemini-3.6-flash` (economics doc's recommended model) — exists in the models list AND responds ✓
Methodological note: a first-connection artifact produced spurious 404s with empty bodies on the very first attempts; every retried call returned 200 with a real body. Any future 'model 404s for this key' claim must be re-tested twice before being believed (the route.ts:69 comment's 2.5-gating claim is also stale — 2.5-flash/-lite appear in this key's model list).

**AIE-05 thinkingBudget — VERIFIED HONORED, with control:**
[...]

PR #891, opened 2026-07-30
feat(ai): capstone AI-off via the credential-gate constant (#867)

### LINKED ISSUE #867: AI hard-off in the capstone via the credential-gate constant (item 33)
## Context
Unified spec item 33. AI must be **hard-off in the capstone**, gated by the **SAME constant as the credential gate** (item 14's shared-constant clause) so the two can never drift. Today the AI path has only structural absence (capstone deploy lessons don't render the AI pane) and quiz-level `aiSuppressed` — no code reads a capstone constant in the AI path, so nothing *guarantees* the capstone stays AI-free if surfaces change.

## Where
- The constant: `apps/web/src/lib/credentials/capstone-gate.ts` — `export const CAPSTONE_CREDENTIAL` (line 40, carries `courseId` + `deployLessonId`). This is the single source of truth; import it — do NOT define a parallel constant.
- Server enforcement (the hard-off): `apps/web/src/app/api/ai/partner/route.ts` — reject partner requests whose lesson context matches `CAPSTONE_CREDENTIAL` (fail-closed; return a typed error the client renders honestly).
- Client: `apps/web/src/components/editor/challenge-interface.tsx` + `apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx` / `blocks/code-block.tsx` — suppress the AI pane for capstone lessons via the same import (the existing `aiSuppressed` plumbing in `apps/web/src/components/editor/types.ts` / `blocks/types.ts` is the natural carrier — derive it from the constant rather than per-block content flags for capstone lessons).
- Copy for the off-state ×3 locales (honest "no AI on the capstone — this one is all you" framing, not an error).

## Done when
- [ ] Server rejects AI partner calls for the capstone lesson(s) derived from `CAPSTONE_CREDENTIAL` — verified by test hitting the route with capstone context
- [ ] Client never renders AI affordances on capstone lessons, derived from the same constant
- [ ] Grep proves exactly one definition site for the capstone identity (the credential gate constant)
- [ ] Tests: route rejection + UI suppression + a test that fails if someone forks the constant

## References
Parent epic: #838 (item 33). RESEARCH-RECONCILIATION §2 ("no code reads a capstone constant in the AI path; the only AI gate is aiSuppressed for quizzes"). Credential gate: `lib/credentials/capstone-gate.ts` (#547 master spec item 14).

thomgabriel:
Closes #867 — unified spec item 33 (epic #838, last open item), with item 14's shared-constant clause.

## Why

The capstone credential attests that the learner shipped the program **themselves** (the deploy gate, item 14). It certifies nothing if a tutor could have written it.

Before this PR the AI path had only **structural absence** — the capstone happens to carry no `code` block, so `/api/ai/partner` 404'd there by accident — plus quiz-level `aiSuppressed`. No code read a capstone constant in the AI path. A content edit adding a challenge surface to the capstone would have silently re-enabled the AI Partner on the one lesson the credential rests on.

## What

**One definition site.** `lib/credentials/capstone-identity.ts` now holds `CAPSTONE_CREDENTIAL` + `isCapstoneCourse` / `isCapstoneLesson` / `CAPSTONE_AI_OFF_CODE`. It is deliberately **not** `server-only` — the lesson client renders the AI-free state from the same constant, and a client-importable definition is what stops that becoming a second hardcoded id. `capstone-gate.ts` re-exports it and keeps the gate's database logic, so **every existing importer is unchanged**.

**Server enforcement (the leg that matters).** `/api/ai/partner` refuses a capstone lesson with `403 { code: "capstone_ai_off", capstoneAiOff: true }`. Placement is load-bearing on two counts:
- **before any spend** — no assist turn, no billed-assist record, no spend-ledger write, no Gemini call. The refusal is the design, not a failure, so it costs the learner nothing (and nothing is refunded, since nothing was taken).
- **before the `codeBlock` 404** — so the refusal is *typed and intentional*, not the incidental 404 this issue exists to replace.

**Client.** `ChallengeInterface` gains `capstoneAiOff`, derived in `lesson-client` from `isCapstoneLesson(lesson._id)` and threaded through `BlockContext` → `code-block`. Unlike `aiSuppressed` (temporary, silent) this state is permanent, so the pane is **replaced by an explanation** rather than left as an empty column: *"The capstone is completed without the AI Partner — that's what makes the credential yours."* The pane is unmounted, not CSS-hidden (no controls in the tab order, WCAG 4.1.2). Copy in `en` / `pt-BR` / `es`.

## Ruling: `/api/lessons/reflect`

The capstone **does** carry an `openEnded` block ("What You've Built — and Writing It Up"), and `maybeGenerateReflectionReply` is lesson-scoped, so this route was in scope.

**Ruling implemented: the reply is NOT generated on the capstone; the seal is.**

The argument for the other side is real — the reflection is a write-up of a deploy that already happened, so the reply is post-completion enrichment, not challenge help, and could defensibly stay on. It is suppressed anyway because *"AI-free"* is only checkable if it means **no AI turn on that lesson**, not "no AI turn that someone judged to be help". A per-surface judgement call is exactly the kind of thing that drifts; a flat invariant keyed off the credential constant does not.

The cost is zero: **receipt-first (AIE-21) is untouched** — the seal is computed and returned unconditionally, so the capstone lesson stays completable, XP and the credential path are unaffected, and the only thing lost is an optional paragraph of feedback. Covered by `reflect-capstone-off.test.ts` (seal present, `reply: null`, generator never invoked).

## Red-proof

Measured, not asserted. With only the `/api/ai/partner` change reverted (`git stash push` on that one file), `partner-capstone-off.test.ts` goes **3 of 4 red**:

```
× refuses the capstone lesson with the typed capstone_ai_off code
    AssertionError: expected 200 to be 403
× spends nothing: no assist turn, no ledger write, no Gemini call
    AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
× refuses every action, not just hint
    AssertionError: action=hint: expected 200 to be 403
```

A **200**, not a 404: the fixture deliberately gives the capstone a code block, which is precisely the content edit that would silently re-enable AI today. At main the request is served — a turn spent and Gemini called on the capstone.

## Tests

- `partner-capstone-off.test.ts` — typed 403 on the capstone; nothing spent (assist / billed / ledger / fetch / refund all untouched); all four actions refused; non-capstone lesson serves a normal 200.
- `reflect-capstone-off.test.ts` — the ruling above, both directions.
- `challenge-interface-capstone-ai-off.test.tsx` — pane unmounted, explanation rendered, ordinary lesson and omitted-prop default unaffected.
- `capstone-identity.test.ts` — **the anti-fork test**: reference equality between `capstone-gate`'s re-export and the identity module (a duplicate literal would pass `toEqual` and defeat the point), plus a source walk over `apps/web/src` that fails on any file hardcoding either id. One allow-listed exception, documented in the test: `lib/courses/learner-segment.ts` names C3 as segment 2's *entry rung* (#599/#673) — a personalization routing table answering a different question; coupling it to `CAPSTONE_CREDENTIAL` would be the actual bug.

## Shape note (EVM / C1 future)

`CAPSTONE_CREDENTIAL` stays **single-valued** — only C3 mints a deploy-gated credential today. The multi-course future is documented in the module: it becomes a readonly array or a content-schema flag, with the two predicates turning into membership tests. Every consumer already goes through `isCapstoneCourse` / `isCapstoneLesson` rather than comparing the constant inline, so that change is local to one file. No registry invented for one entry.

## Verification

- `pnpm typecheck` — 6/6 tasks pass
- `apps/web` full suite — **244 files, 2217 tests, all passing**
- `pnpm lint` — no errors (only the repo's pre-existing `import/order` warnings)

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/open-ended-block.test.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/open-ended-block.test.tsx
@@ -36,6 +36,7 @@ function makeCtx(overrides: Partial<BlockContext> = {}): BlockContext {
     setProof: vi.fn(),
     setQuizAnswered: vi.fn(),
     aiSuppressed: true,
+    capstoneAiOff: false,
     buildUuid: null,
     programKeypairSecret: null,
     resetBuild: vi.fn(),
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
@@ -43,6 +43,7 @@ function makeCtx(overrides: Partial<BlockContext> = {}): BlockContext {
     setProof: vi.fn(),
     setQuizAnswered: vi.fn(),
     aiSuppressed: true,
+    capstoneAiOff: false,
     buildUuid: null,
     programKeypairSecret: null,
     resetBuild: vi.fn(),
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/quiz-block.test.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/quiz-block.test.tsx
@@ -72,6 +72,7 @@ function makeCtx(overrides: Partial<BlockContext> = {}): BlockContext {
     setProof: vi.fn(),
     setQuizAnswered: vi.fn(),
     aiSuppressed: true,
+    capstoneAiOff: false,
     buildUuid: null,
     programKeypairSecret: null,
     resetBuild: vi.fn(),
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/code-block.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/code-block.tsx
@@ -56,6 +56,7 @@ export function CodeBlock({ block, ctx }: BlockRenderProps) {
           isEnrolled={ctx.isEnrolled}
           onEnroll={ctx.onEnroll}
           aiSuppressed={ctx.aiSuppressed}
+          capstoneAiOff={ctx.capstoneAiOff}
           className="h-full"
         />
       </div>
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/types.ts
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/types.ts
@@ -30,6 +30,15 @@ export interface BlockContext {
    * block threads it to ChallengeInterface, which keeps the AI Partner hidden.
    */
   aiSuppressed: boolean;
+  /**
+   * True on THE graded capstone lesson (#867) — derived from
+   * `lib/credentials/capstone-identity`, the same constant the credential gate
+   * uses. Unlike `aiSuppressed` this is permanent and explained: the code block
+   * threads it to ChallengeInterface, which renders the AI-free rationale where
+   * the AI Partner would sit. The server refuses independently
+   * (`/api/ai/partner`) — this is presentation, not the enforcement.
+   */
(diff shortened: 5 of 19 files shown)


thomgabriel:
Gate: merged. Red-proof replayed (3/4 red at main in partner-capstone-off.test.ts, 4/4 green at head); CI green; review clean — its two advisory notes (the lessonId==lesson._id invariant comment in the reflect route, and keeping capstone-identity.ts free of server-only transitive imports) are recorded here as dispositions: both are guarded by existing tests per the reviewer, no follow-up filed.

Issue #870, opened 2026-07-28
Re-engagement email templates (uiux-R17)


thomgabriel:
## Context
uiux-R17: re-engagement email templates became actionable the moment Resend landed, but `apps/web/src/lib/email/templates.ts` still contains only the course-announcement email (`newCourseAnnouncementEmail`). Add the re-engagement set (lapsed-learner nudge back to the next incomplete lesson; the exact set per uiux-R17 in `docs/superpowers/specs/2026-07-25-learning-uiux-research.md` — pull the template list from the research doc, not from memory). All sends are consent-gated and carry one-click unsubscribe.

## Where
- `apps/web/src/lib/email/templates.ts` — new template functions following the `RenderedEmail` / `newCourseAnnouncementEmail` pattern; every subject goes through the existing CRLF scrub (line 28, `s.replace(/[\r\n]/g, " ")`) per #812.
- `apps/web/src/lib/email/campaign.ts` — send orchestration following `sendNewCourseAnnouncement` (batching, opted-in scan via the `idx_email_subscriptions_opt_in` partial index).
- Consent: gate on the appropriate consent type — coordinate with the notification-core sibling issue (reminder vs marketing consent); a re-engagement nudge is NOT marketing-consent-covered by default — resolve which toggle governs it with the owner if ambiguous, fail closed (no consent type → no send).
- Content hooks: next-incomplete-lesson derivation exists in `apps/web/src/lib/courses/continue-learning.ts` — reuse it for deep links.
- Locales: 3 (pt-BR/es/en), strings in the templates (server-side rendering — follow how the announcement template localizes).

## Done when
- [ ] Re-engagement template set from uiux-R17 exists in `templates.ts`, ×3 locales, subjects scrubbed
- [ ] Sends are consent-gated (fail-closed) with one-click unsubscribe
- [ ] Deep links point at the learner's actual next incomplete lesson via the continue-learning derivation
- [ ] Tests: rendering ×3 locales, consent gating, subject CRLF scrub

## References
Parent epic: #845 (leg 3). RESEARCH-RECONCILIATION §4 ("re-engagement email templates became actionable the moment Resend landed") + §8.2. uiux-R17 in the learning-uiux research doc. Infra: #769/#779, scrub pattern #812.

Issue #845, opened 2026-07-28
Notification channel v1 (Resend): session-plan reminders + re-engagement templates


thomgabriel:
The completion pass (RECONCILIATION §8.2) found ONE common root behind three personalization PARTIALs: **no send channel exists at all**. Resend infrastructure is live (#779: consent model, templates.ts, unsubscribe), so the channel is buildable.

**EPIC — tracked by child issues: #869 (notification core: reminder consent toggle — own toggle, separate from marketing consent — + session-plan reminder send), #870 (re-engagement email templates, uiux-R17) — created 2026-07-28. This issue closes when all children close.**

Constraints carried into the children: consent-gated via the #779 opt-in model with reminders on their OWN toggle (default from the intake commitment), all São Paulo timing, all 3 locales, one-click unsubscribe on every send. The three original legs map: implementation-intention slot (#582) → #869; daily-goal timing leg → #869; re-engagement templates → #870.

thomgabriel:
Epic complete: #869 (#895 + migration 20260731120914 live) + #870 (#898). Send pipeline for re-engagement kinds filed as #899. Feature inert until CRON_SECRET + RESEND_API_KEY land in Vercel (owner).

PR #812, opened 2026-07-27
fix(email): subject CRLF scrub, announce-route maxDuration, route-table docs (#807)

thomgabriel:
Closes #807 (the gate's follow-ups from its #779 review).

## Changes
1. **Subject CRLF scrub** (`templates.ts`): the subject interpolated the raw course title while the body escaped it. New `scrubHeaderValue()` (`replace(/[\r\n]/g, " ")`, the issue's exact regex) wraps the assembled subject — defense-in-depth on top of Resend's JSON API, so a newline in a title can't split or smuggle a header line.
2. **Announce-route `maxDuration = 300`** (`announce-course/route.ts`): the campaign sends serially (100-address Resend batches + 600ms courtesy delay + round-trips), so a real list truncates mid-send at the default serverless limit. 300 is the plan's hard ceiling, matching `api/build-program` and `api/courses/test-out` (cross-referenced in the comment); covers >10k recipients.
3. **API route table** (`api/CLAUDE.md`): `POST /api/admin/email/announce-course` added to the Admin table (preserving the "all /api/admin/* live in Admin" invariant) + new Email section for `GET/POST /api/email/unsubscribe` (token auth, RFC 8058 one-click), matching the file's group-by-area style.

## Red-proof (rule 2), independently replayed by the orchestrator
New test `scrubs CR/LF from the subject (header-injection defense, #807)` run against **main (069c095)**: FAILS with the subject carrying the raw injected line — `Received: "New course on Superteam Academy: Legit Title\nBcc: victim@example.com"`. This was a genuine gap, not a JSON-API-already-safe case. Green 4/4 at this head.

## Verified at dd719cd (agent + orchestrator re-run in clean clone)
typecheck 6/6; web suite **227 files / 2058 tests**; lint clean.

PR #895, opened 2026-07-31
feat(email): reminder consent toggle + session-plan reminder send (#869)

### LINKED ISSUE #869: Notification core: reminder consent toggle + session-plan reminder send via Resend
## Context
Three personalization PARTIALs share one root: **no send channel exists** (reconciliation §8.2). Resend infrastructure is now live (#769/#779: consent model, `templates.ts`, one-click unsubscribe). This issue builds the notification core: (a) a **reminder consent toggle** — a consent type of its own, SEPARATE from #779's marketing consent (marketing consent ≠ reminder consent; recommended default comes from the intake commitment), and (b) the **session-plan reminder send** — the "When will you do your next lesson?" answer (#582, already captured to `profiles.prefs.nextLesson` and tracked via `trackNextLessonPlanCommitted`) becomes a real Resend send on the committed day.

**⚠️ DB migration protocol:** any migration follows the house SECURITY DEFINER pattern (see `supabase/migrations/20260728120000_email_subscriptions.sql` — server-stamped consent timestamps, no client write policy, RPCs REVOKEd from anon/authenticated) and **MUST be gate-applied via MCP before merge** with the ledger reconciliation rule (`docs/DB-MIGRATION-LEDGER.md`).

## Where
- Consent: extend `email_subscriptions` (or a sibling table) with a reminder-consent column set — mirror `opt_in`/`consent_at`/`unsubscribed_at` semantics from `supabase/migrations/20260728120000_email_subscriptions.sql`; new RPC or extended `set_marketing_opt_in`-style SECURITY DEFINER function scoped per consent type.
- Toggle UI: settings — `apps/web/src/app/[locale]/(platform)/settings/_components/account-tab.tsx` (where the marketing toggle lives); optionally offer opt-in at the plan-commit moment in `apps/web/src/components/dashboard/next-lesson-plan.tsx`.
- Send pipeline: `apps/web/src/lib/email/resend.ts` (transport), `apps/web/src/lib/email/campaign.ts` (`sendNewCourseAnnouncement` is the pattern), new template in `apps/web/src/lib/email/templates.ts` (subject through the line-28 CRLF scrub per #812).
- Scheduling: **no cron infrastructure exists** (no `vercel.json` crons, no `api/cron/`). Pick and build the minimal scheduler (Vercel cron hitting a service-role route, or Supabase pg_cron) — send on the learner's committed weekday, **America/Sao_Paulo** timing, idempotent per (user, week).
- Plan data: `profiles.prefs.nextLesson` (see `next-lesson-plan.tsx` lines 18–25 for the stored shape).
- Unsubscribe: reuse `apps/web/src/app/api/email/unsubscribe/route.ts` token flow — every reminder send carries the one-click link and it must clear reminder consent.

## Done when
- [ ] Reminder consent is its own toggle + own timestamps, server-authoritative, never inferred from marketing consent
- [ ] A learner with a committed plan day + reminder consent receives one localized reminder (3 locales: pt-BR/es/en) on that day, São Paulo timing; no consent or no plan → no send
- [ ] One-click unsubscribe on every send, clearing reminder consent only
- [ ] Migration MCP-applied + ledger updated before merge; RLS verified (anon/authenticated cannot write consent directly)
- [ ] Tests: consent gating, idempotency, template rendering ×3 locales, subject scrub

## References
Parent epic: #845 (legs 1–2). RESEARCH-RECONCILIATION §8.2. Consent model: #779/#769; plan capture: #582; CRLF scrub: #812. Sibling: the re-engagement-templates issue (uiux-R17).

thomgabriel:
Closes #869 (notification core, child of #845).

## ⚠️ GATE: MIGRATION MUST BE MCP-APPLIED BEFORE MERGE

**`supabase/migrations/20260731120000_reminder_consent.sql`**

Per #731, Vercel auto-deploys from `main` the moment this merges — the settings
panel reads `email_subscriptions.reminder_opt_in` and the cron route calls
`claim_due_session_reminders`, neither of which exists in prod yet. Apply, then
record the MCP-stamped version in `docs/DB-MIGRATION-LEDGER.md` (the row is
already there as **PENDING GATE APPLY**).

The migration also registers itself in `schema-expectations.ts`, so
`/api/health/schema` 503s naming the missing objects if it is ever forgotten.

Migration contents: 4 new columns on `email_subscriptions`, one new table
(`email_reminder_log`), 4 SECURITY DEFINER RPCs. `BEGIN/COMMIT`, fully
idempotent, tested `ROLLBACK` block at the bottom, mirrored into
`supabase/schema.sql`.

## 1. Reminder consent — its OWN consent, not #779's

| | marketing (#779) | reminders (#869) |
| --- | --- | --- |
| flag | `opt_in` | `reminder_opt_in` |
| audit | `consent_at` / `unsubscribed_at` | `reminder_consent_at` / `reminder_unsubscribed_at` |
| write RPC | `set_marketing_opt_in` | `set_reminder_opt_in(BOOLEAN, TEXT)` |
| unsubscribe | `unsubscribe_by_token` | `unsubscribe_reminders_by_token` |
| link | `?token=…` | `?token=…&kind=reminders` |

Both default **FALSE**. Neither is ever inferred from the other — pinned by
pglite tests in both directions (a marketing opt-in never yields a reminder
recipient; each unsubscribe leaves the other consent standing). `reminder_locale`
is captured at opt-in because the cron run has no request context to negotiate a
locale from.

**The derived default (#582).** The issue asks for "default ON if the learner
committed a plan day". Implemented so the default is real without fabricating
consent:

* the **plan-commit card** (`next-lesson-plan.tsx`) is the consent-capture
  surface — a disclosed, **default-ON** checkbox; pressing Save writes the
  consent (or an explicit opt-OUT if unchecked). That is the default-ON path for
  every plan committed from here on.
* the **settings toggle** shows the recorded state; when no decision exists yet
  it *displays* the derived default (ON iff a plan day is committed) with the
  disclosure "On by default because you scheduled a session. Nothing is sent
  until you confirm this setting."

Consent is never auto-persisted on render, and the send pipeline gates on the
stored `reminder_opt_in` alone — so the suggested-ON display can never produce
an email nobody asked for. This is the deliberate LGPD-safe reading of "default
ON"; flag it if you want the derived value persisted on mount instead.

## 2. The send

`lib/email/reminders.ts` mirrors the #779 campaign pipeline: fail-closed on a
missing `RESEND_API_KEY` (returns `unconfigured` and **claims nothing**, so the
day's idempotency slots aren't burned), chunked to Resend's 100/batch with a
600ms courtesy delay, per-chunk idempotency key, RFC 8058 `List-Unsubscribe` on
every message.

New template `sessionPlanReminderEmail` (appended to `templates.ts`; #870's
re-engagement templates are untouched) renders **pt-BR / es / en**. `plan_time`
comes from self-writable `profiles.prefs`, i.e. attacker-controlled — it is
shape-validated **and** still passed through the line-28 CRLF scrub (#812)
before it reaches the subject. Red-proofed by a test that feeds
`"19:00\r\nBcc: victim@example.com"`.

### Idempotency is a DB invariant, not a code convention

`claim_due_session_reminders` reads recipients and INSERTs the
`email_reminder_log(user_id, kind, sent_on)` row in **one statement**
(`INSERT … ON CONFLICT DO NOTHING RETURNING`, PK on the triple). Two overlapping
cron invocations cannot both claim a learner; a manual re-trigger the same day
sends nothing. A batch that fails to send calls
`release_session_reminder_claims` so the day stays retryable.

### Schedule

`apps/web/vercel.json` (Vercel Root Directory is `apps/web`):

```json
{ "crons": [{ "path": "/api/cron/session-reminders", "schedule": "0 11 * * *" }] }
```

**`0 11 * * *` UTC = 08:00 America/Sao_Paulo** — Brazil has had no DST since
2019, so the offset is a stable UTC-3. The learner is nudged in the morning,
ahead of the hour they picked. Everything downstream (the ledger day, the
weekday match) is evaluated on the São Paulo calendar.

`GET /api/cron/session-reminders` takes Vercel's `Authorization: Bearer
$CRON_SECRET` (timing-safe compare) and **fails closed**: no `CRON_SECRET`
configured ⇒ 503, nothing sent, so a misconfigured deploy can never leave an
unauthenticated mail trigger exposed.

## Tests (2298 passing, +40 in this PR)

* `reminder-consent-execution.test.ts` — 19 pglite tests running the real
  migration SQL: consent gating, consent separation both ways, once-per-day
  claim, release-and-retry, yesterday doesn't block today, wallet-placeholder
  exclusion, unauthenticated `set_reminder_opt_in` rejection, locale handling.
* `reminder-consent-guard.test.ts` — 16 SQL invariants across the migration
  **and** the `schema.sql` mirror (opt-out default, no client write policy, zero
  policies on the ledger, PK-based idempotency, per-RPC grants).
* `reminders.test.ts` — 12: fail-closed claims nothing, recipients only from the
  RPC, no double-send on re-run, claim release on batch failure (and *not* on
  success), `kind=reminders` header, per-recipient locale.
* `session-plan-reminder-template.test.ts` — 8: 3 distinct locales, en fallback,
  CRLF scrub, malformed-time drop, unsubscribe link in HTML *and* text.
* cron route — 8: valid Bearer, missing header, wrong secret, no-Bearer-scheme,
  same-length near-miss, 503 fail-closed with no secret.
* settings panel — 6, plan card — +4, unsubscribe route — +4 (kind routing;
  legacy no-kind links still clear marketing only).

**Red-proofed**: deleting `es.reminder_opt_in = true` from the claim gate fails
10 tests; weakening the claim JOIN to a LEFT JOIN fails the idempotency test.

Verified: `tsc --noEmit` clean, `eslint src` 0 errors, full suite 251 files /
2298 tests green.

## Not in scope

No analytics events added (`lib/analytics/events.ts` untouched — sibling agent
owns it). `templates.ts` only gained the one reminder template.


thomgabriel:
## Gate bounce addressed — F1, F2, F3, F4, F5, F7 fixed in-PR

Head: `b43e3c91d037730d8d4b5d5b15f6e46a1c7729fa`. Migration path unchanged: `supabase/migrations/20260731120000_reminder_consent.sql` — **no SQL changed in this round**, all six fixes are TypeScript/docs, so the gate's MIGRATION-SAFE verdict still stands on the same file.

| # | Fix | Pinned by |
| --- | --- | --- |
| **F1** | A derived-ON suggestion is now confirmable: an explicit confirm button renders **only** while the switch shows a derived ON and writes the **displayed** value (`set_reminder_opt_in(true, locale)`). Toggling still records the explicit opt-OUT. | `email-preferences.test.tsx` → *"CONFIRM persists the displayed ON (not its opposite)"*, *"the confirm affordance disappears once the decision is recorded"*, *"toggling a derived ON still records the explicit opt-OUT"*, *"shows NO confirm button when the derived default is OFF"* |
| **F2** | "No decision yet" is now a **timestamp** test (`reminder_consent_at IS NULL AND reminder_unsubscribed_at IS NULL`), not row existence — fixed on **both** surfaces. | settings: *"a MARKETING-ONLY row is not a reminder decision: derived default applies"* + *"…with NO plan still shows the reminder switch OFF"*; plan card: *"treats a MARKETING-ONLY row as undecided and still offers the default ON"* + *"respects a recorded opt-IN"* |
| **F3** | Reviewer's option (a). `sendEmailBatch` now returns `delivery: "rejected" \| "unknown"` — `rejected` = local guard / ECONNREFUSED / ENOTFOUND / EAI_AGAIN / 4xx; `unknown` = 5xx, abort/timeout, ECONNRESET, socket hang up. The pipeline releases claims **only** on `rejected`, and logs the held claim count otherwise. | `reminders.test.ts` → *"releases a batch REJECTED before transmission (4xx)"*, *"HOLDS the claims of a 5xx batch"*, *"HOLDS the claims of a timed-out / dropped batch"*; classification itself in `resend.test.ts` → 5 new cases incl. per-errno and the local over-size guard |
| **F4** | Recipients dropped by `isValidEmail` have their claims released and the count `console.error`-logged. | `reminders.test.ts` → *"RELEASES the claim of an address it filtered out"* (asserts the RPC call + `released`), *"releases claims even when EVERY claimed address is unusable"* |
| **F5** | Inline ⚠️ comment at the `claim_due_session_reminders` `SCHEMA_EXPECTATIONS` entry: the probe is side-effectful and safe **only** because the fn is REVOKEd from anon/authenticated — never grant it, or every health check silently burns that weekday's planners. | comment-only |
| **F7** | `docs/DEPLOYMENT.md` no longer claims a re-trigger is unconditionally safe: it documents that ambiguous failures **hold** their claims, and warns against hand-clearing `email_reminder_log` to "fix" a failed run. | docs-only |

### Red-proofs (old behaviour is pinnable, and pinned)

| Reverted behaviour | Result |
| --- | --- |
| F1 — remove the `confirm` prop | **2 tests fail** |
| F2 — restore `if (sub)` row-existence on both surfaces | **2 tests fail** (one per surface) |
| F3 — release unconditionally (`if (true)`) | **2 tests fail** |
| F4 — drop the filtered-out release block | **2 tests fail** |

### Verification

`tsc --noEmit` clean · `eslint src` **0 errors** · `prettier --check` clean · pglite suite **35 passed** · full web suite **251 files / 2315 tests passed** (was 2298; +17 this round).

F6 (token scoping) left alone as agreed — pre-existing #779 shape, filed separately.

Ready for the migration apply.

thomgabriel:
@claude review

PR #907, opened 2026-07-31
chore(web): broaden the no-experience marketing copy guard

thomgabriel:
Closes #906

## What

`marketing-copy-bounds.test.ts` guarded against "no experience needed" claims with a flat regex list whose EN rule was `\bno (?:prior |previous )?experience (?:needed|required)\b` — anything else between "no" and "experience" ("no **coding** experience required") slipped past, as did most PT-BR/ES phrasings.

Replaced with an exported detector, `findNoExperienceClaims(value)`, used by the per-locale catalog scan.

## The new pattern

A claim is forbidden when a negation lands on an experience noun **without a topic scope**. Topic scope may sit either side of the noun:

- before it (EN): `no **Rust** experience required`
- after it via a preposition (PT/ES): `experiência **com** blockchain`

```
SCOPED_TOPIC = blockchain|solana|rust|crypto*|cripto*|web3|defi|nfts?|anchor|smart contract(s)|contratos?|on-chain|token*
```

Rules (all `gi`, group 1 = the intervening gap, capped at 40 chars and punctuation-free so it cannot cross a sentence):

| label | pattern |
|---|---|
| unqualified no-experience (en) | `\b(?:no\|zero\|without(?:\s+any)?)\s+([\w\s,'’/-]{0,40}?)(?:experience\|background\|knowledge)\b` + scopeAfter `with\|in\|of\|about` |
| no-code (en) | `\b(?:no\|zero)\s+(?:coding\|code\|programming\|dev(?:eloper)?\s+skills?\|technical\s+skills?)\b` |
| no-prerequisites (en) | `\b(?:no\|zero)\s+(?:prior\s+)?(?:prerequisites?\|pre-?reqs?\|requirements?)\b` |
| unqualified no-experience (pt) | `\b(?:sem\|nenhum(?:a)?\|zero\|n[ãa]o\s+(?:é\s+)?precisa?(?:r\|o)?(?:\s+(?:de\|ter))?\|n[ãa]o\s+exige)\s+(gap)(?:experi[êe]ncias?\|conhecimentos?\|bagagem)\b` + scopeAfter `com\|em\|de\|sobre` |
| no-code (pt) | `sem (nunca )?saber program*` / `n[ãa]o precisa (saber )?programar` / `sem (nenhum \|escrever )?c[oó]digo` / `do zero sem` |
| no-prerequisites (pt) | `(?:sem\|nenhum)\s+(?:nenhum\s+)?pr[ée]-?requisitos?` |
[...]

PR #908, opened 2026-07-31
ci(deps): make the brace-glob probe resolution-aware and gate CI on it

### LINKED ISSUE #905: Wire the brace-glob probe into CI (and make it resolution-aware, not directory-walking)
Two findings, from #900's review + the gate's own empirical spot-check:

1. **The probe isn't enforced** (reviewer finding 1): scripts/brace-glob-probe.mjs is the ONLY detector for a regression class green suites provably miss (the #820 lesson), but nothing runs it — no test:scripts include, no CI step. Add a thin scripts/__tests__/brace-glob-probe.test.ts wrapper OR a CI step gated on changes to the minimatch/brace-expansion override keys, so it's enforced rather than tribal knowledge.
2. **The probe walks directories, not the resolution graph** (gate finding): on a store with stale copies from earlier installs it reports old versions as VULNERABLE even though `pnpm why` resolves nothing to them and the lockfile pins only patched ones (observed on main after #900: 9 such lines, all orphans — 1.1.12/1.1.16/2.0.2/2.1.2). Make it enumerate copies via the resolution graph (or mark unreachable copies as INFO), so a real regression isn't hidden in noise.

Gate's empirical spot-check for the record (reviewer finding 2 asked for it): lockfile pins brace-expansion 1.1.18/2.1.4 + minimatch 3.1.4/4.2.6/9.0.7; the CVE-2026-14257 bound (EXPANSION_MAX=100000, EXPANSION_MAX_LENGTH=4000000) is present in both patched brace-expansion files. Backport claim CONFIRMED.

thomgabriel:
Closes #905

Two defects from #900's review + the gate's empirical spot-check, plus the fixes from the independent adversarial review of the first revision (B1 blocking, F2/F4/F7).

## 1. The probe walked directories, not the resolution graph

`installedCopies()` enumerated `node_modules/.pnpm` **directories**. A pnpm store accumulates orphans — dirs left behind by earlier installs that nothing resolves to any more — and the walk probed those too. On main after #900 that produced 9 `VULNERABLE` lines (brace-expansion 1.1.12/1.1.16, minimatch 2.0.2/2.1.2) while `pnpm why` resolved nothing to them and the lockfile pinned only patched versions. Noise that loud hides a real regression.

**Design of the resolution-aware logic**

- `parseSnapshots(text)` parses the `snapshots:` block of `pnpm-lock.yaml` (v9) into `name -> Set(version)`. That block *is* the resolution graph — the structure `pnpm why` walks — so it's authoritative without shelling out to `pnpm why` per package. Normalizes quoting (`'@babel/core@7.29.6'`), scopes, peer suffixes (`foo@1.2.3(bar@4.5.6)` → `1.2.3`; the peer hash doesn't change the bytes on disk), and **both entry shapes** — see B1 below.
- `copies(name)` still walks `.pnpm`, but only to *locate bytes for a version the graph resolves*. Copies partition into **reachable** (asserted on, deduped per version across peer-hash dirs) and **orphaned** (→ non-failing `INFO` section after the table, explicitly labeled as not affecting the exit code).
- **Resolved-but-absent is a FAILURE** — a lockfile version with no copy on disk means the probe silently covered less than it claimed.
- **No silent fallback**: unreadable/unparseable lockfile → `exit 2`, not a directory walk (that fallback *is* the false-positive mode).
- **Override conformance**: every resolved `brace-expansion`/`minimatch`/`picomatch` version is checked against its root `pnpm.overrides` pin, so a bump dragging in an unpinned major is caught as a pin gap rather than a later runtime crash.
- **Coverage floor** (F2): every check is "for each resolved version of X", which asserts *nothing* when X resolves to nothing. A PINNED name resolving to zero versions is a failure in its own right.
- **Parser self-check**: an asserted check parses an inline lockfile fixture (leaf `{}`, spaced `{  }`, block, scoped, quoted, peer-suffixed, `link:`) and compares the exact extracted set.

### B1 (blocking, from adversarial review) — leaf-form snapshot keys were dropped

pnpm v9 writes a snapshot entry in **two** shapes:

```yaml
  minimatch@9.0.9:            # block — has a nested `dependencies:` map
    dependencies: ...
  picomatch@2.3.2: {}         # leaf  — dependency-less, INLINE FLOW MAP
```

The original key regex anchored with `:\s*$` and matched only the first. That dropped **562 of 1655 entries (34%)**, including *both* picomatch versions — so the override section asserted nothing for picomatch despite the previous PR body's claim.

Worse: an unmatched key reads as "not in the graph", so a **reachable vulnerable copy was downgraded to a benign orphan INFO purely by snapshot shape**. A silent false negative, strictly worse than the directory-walking bug this replaces. Fixed pattern:

```js
const SNAPSHOT_KEY =
  /^ {2}'?((?:@[^/'\s]+\/)?[^@'\s][^'\s]*?)@([^('\s]+)(?:\([^']*\))?'?:(\s*\{\s*\})?\s*$/;
```

Also fixed: **F4** — an override pin `satisfies()` can't evaluate (`>=9.0.7`, `9.x`) used to downgrade to INFO, masking a real pin gap; pins are repo-authored, so an unparseable *pin* is now a failure (an unparseable *resolved version* stays INFO). **F7** — `cmpVersion` hoisted above `copies()`, removing the TDZ hazard.

Effect on coverage: **44 → 55 asserted checks** on a clean tree (the pre-fix parser silently ran 44).

## 2. The probe wasn't enforced

- New root script `probe:brace-glob`.
- New **hard-gate** step `Brace-glob probe (minimatch / brace-expansion resolution)` in the `frontend` job, immediately after `pnpm install --frozen-lockfile` so it fails before the slow steps. No `continue-on-error`.
- `scripts/brace-glob-probe.mjs` added to the `frontend` paths-filter so editing the probe re-runs it. `package.json` and `pnpm-lock.yaml` — the inputs it guards — were already there.

The frontend job is the one that already installs the JS dependency tree, and its filter already fires on every `package.json` / `pnpm-lock.yaml` change — precisely the change class that can reintroduce the regression.

Branch has main merged (#907/#909/#911); #911's `pnpm.overrides` `uuid` bump merged cleanly and all proofs below run against the current lockfile.

---

# Proofs

## (a) Clean tree → exit 0, zero failures

```
$ node scripts/brace-glob-probe.mjs; echo "EXIT=$?"
CONSUMER                        CHECK                                                      RESULT  DETAIL
parser self-check               snapshot keys: leaf `{}` + block + scoped + peer-suffixed  PASS    ["@scoped/block@3.1.0","@scoped/leaf@3.0.0","block@2.0.0","leaf-spaced@1.0.1","leaf@1.0.0","peered-quoted@4.1.0","peered@4.0.0"]
brace-expansion                 resolves to >=1 version (coverage floor)                   PASS    2 version(s) in the resolution graph
minimatch                       resolves to >=1 version (coverage floor)                   PASS    3 version(s) in the resolution graph
picomatch                       resolves to >=1 version (coverage floor)                   PASS    2 version(s) in the resolution graph
brace-expansion@1.1.18          covered by a root override pin                             PASS    pin brace-expansion@1 = ^1.1.18
brace-expansion@2.1.4           covered by a root override pin                             PASS    pin brace-expansion@2 = ^2.1.4
minimatch@3.1.4                 covered by a root override pin                             PASS    pin minimatch@3 = 3.1.4
minimatch@4.2.6                 covered by a root override pin                             PASS    pin minimatch@4 = ^4.2.6
minimatch@9.0.9                 covered by a root override pin                             PASS    pin minimatch@9 = ^9.0.7
picomatch@2.3.2                 covered by a root override pin                             PASS    pin picomatch@2 = 2.3.2
picomatch@4.0.5                 covered by a root override pin                             PASS    pin picomatch@4 = ^4.0.4
brace-expansion@1.1.18          CJS default export is callable                             PASS    typeof = function
brace-expansion@1.1.18          expand('a{b,c}d')                                          PASS    abd,acd
brace-expansion@1.1.18          expand('{1..3}') sequence                                  PASS    1,2,3
brace-expansion@1.1.18          CVE-2026-14257 bound on {a,b}x1500                         PASS    2666 results / 3999000 chars in 531ms
brace-expansion@2.1.4           CJS default export is callable                             PASS    typeof = function
brace-expansion@2.1.4           expand('a{b,c}d')                                          PASS    abd,acd
brace-expansion@2.1.4           expand('{1..3}') sequence                                  PASS    1,2,3
brace-expansion@2.1.4           CVE-2026-14257 bound on {a,b}x1500                         PASS    2666 results / 3999000 chars in 303ms
minimatch@3.1.4                 CJS default export is callable                             PASS    typeof = function
minimatch@3.1.4                 match('a.json', '*.{json,md}')                             PASS    true
minimatch@3.1.4                 match('a.md', '*.{json,md}')                               PASS    true
minimatch@3.1.4                 match('a.txt', '*.{json,md}')                              PASS    false
minimatch@3.1.4                 match('src/x/y.tsx', 'src/**/*.{ts,tsx}')                  PASS    true
minimatch@3.1.4                 match('file2.txt', 'file{1..3}.txt')                       PASS    true
minimatch@3.1.4                 braceExpand('*.{json,md}')                                 PASS    *.json,*.md
minimatch@4.2.6                 CJS default export is callable                             PASS    typeof = function
minimatch@4.2.6                 match('a.json', '*.{json,md}')                             PASS    true
minimatch@4.2.6                 match('a.md', '*.{json,md}')                               PASS    true
minimatch@4.2.6                 match('a.txt', '*.{json,md}')                              PASS    false
minimatch@4.2.6                 match('src/x/y.tsx', 'src/**/*.{ts,tsx}')                  PASS    true
minimatch@4.2.6                 match('file2.txt', 'file{1..3}.txt')                       PASS    true
minimatch@4.2.6                 braceExpand('*.{json,md}')                                 PASS    *.json,*.md
minimatch@9.0.9                 named `minimatch` export present                           PASS    typeof = object
minimatch@9.0.9                 match('a.json', '*.{json,md}')                             PASS    true
minimatch@9.0.9                 match('a.md', '*.{json,md}')                               PASS    true
minimatch@9.0.9                 match('a.txt', '*.{json,md}')                              PASS    false
minimatch@9.0.9                 match('src/x/y.tsx', 'src/**/*.{ts,tsx}')                  PASS    true
minimatch@9.0.9                 match('file2.txt', 'file{1..3}.txt')                       PASS    true
minimatch@9.0.9                 braceExpand('*.{json,md}')                                 PASS    *.json,*.md
glob@7.2.0                      sync('*.{json,md}')                                        PASS    5 file(s)
glob@7.2.0                      sync('*.{json,yaml}')                                      PASS    5 file(s)
glob@7.2.0                      sync('scripts/*.{mjs,ts}')                                 PASS    4 file(s)
glob@7.2.3                      sync('*.{json,md}')                                        PASS    5 file(s)
glob@7.2.3                      sync('*.{json,yaml}')                                      PASS    5 file(s)
glob@7.2.3                      sync('scripts/*.{mjs,ts}')                                 PASS    4 file(s)
glob@10.5.0                     sync('*.{json,md}')                                        PASS    5 file(s)
glob@10.5.0                     sync('*.{json,yaml}')                                      PASS    5 file(s)
glob@10.5.0                     sync('scripts/*.{mjs,ts}')                                 PASS    4 file(s)
test-exclude@6.0.0              include 'src/**/*.{ts,tsx}'                                PASS    true
test-exclude@6.0.0              exclude '**/*.{test,spec}.ts'                              PASS    false
eslint (overrides.files brace)  '**/*.{ts,tsx}' selects sample.tsx                         PASS    ["no-debugger"]
eslint (overrides.files brace)  '**/*.{ts,tsx}' skips sample.js                            PASS    []
eslint (override cascade)       '**/*.{spec,test}.tsx' overrides '**/*.{ts,tsx}'           PASS    []
prettier CLI                    brace glob resolves                                        PASS    no diff

55/55 checks passed.
EXIT=0
```

No `INFO` section — this store has zero orphans. Both picomatch versions are now covered (they were invisible pre-B1-fix).

## (b) Negative control — tamper a REACHABLE copy → non-zero, named

`brace-expansion@2.1.4` is resolved by the lockfile. Strip its CVE-2026-14257 bounds:

```
$ perl -pi -e 's/^var EXPANSION_MAX = 100000$/var EXPANSION_MAX = Infinity/;
               s/^var EXPANSION_MAX_LENGTH = 4000000$/var EXPANSION_MAX_LENGTH = Infinity/' \
    node_modules/.pnpm/brace-expansion@2.1.4/node_modules/brace-expansion/index.js
```

```
$ node scripts/brace-glob-probe.mjs; echo "EXIT=$?"
brace-expansion@2.1.4  covered by a root override pin       PASS  pin brace-expansion@2 = ^2.1.4
brace-expansion@2.1.4  CJS default export is callable       PASS  typeof = function
brace-expansion@2.1.4  expand('a{b,c}d')                    PASS  abd,acd
brace-expansion@2.1.4  expand('{1..3}') sequence            PASS  1,2,3
brace-expansion@2.1.4  CVE-2026-14257 bound on {a,b}x1500   FAIL  JS heap out of memory — unbounded (VULNERABLE)

54/55 checks passed.
FAILURES:
  brace-expansion@2.1.4 :: CVE-2026-14257 bound on {a,b}x1500 :: JS heap out of memory — unbounded (VULNERABLE)
EXIT=1
```

## (b2) B1 shape-flip — same tampered bytes, snapshot entry rewritten to LEAF form

The reviewer's exploit. Keep the tampered `brace-expansion@2.1.4` bytes and rewrite *only* its lockfile snapshot entry from block to leaf:

```
$ grep -n -m2 "brace-expansion@2.1.4" pnpm-lock.yaml | tail -1
11958:  brace-expansion@2.1.4: {}
```

**Fixed probe — still red, still named:**

```
$ node scripts/brace-glob-probe.mjs; echo "EXIT=$?"
brace-expansion@2.1.4  CVE-2026-14257 bound on {a,b}x1500  FAIL  JS heap out of memory — unbounded (VULNERABLE)
54/55 checks passed.
FAILURES:
  brace-expansion@2.1.4 :: CVE-2026-14257 bound on {a,b}x1500 :: JS heap out of memory — unbounded (VULNERABLE)
EXIT=1
```

**Same tree, pre-fix regex (only `SNAPSHOT_KEY` reverted) — the false negative reproduced:**

```
$ node scripts/oldparser-probe.mjs; echo "OLD_PARSER_EXIT=$?"
parser self-check  snapshot keys: leaf `{}` + block + scoped + peer-suffixed  FAIL  ["@scoped/block@3.1.0","block@2.0.0"]

46/48 checks passed.

INFO (not asserted, does not affect exit code) — 1 item(s):
  brace-expansion :: brace-expansion@2.1.4 on disk (.pnpm/brace-expansion@2.1.4) but NOT in the resolution graph — not probed

FAILURES:
  parser self-check :: snapshot keys: leaf `{}` + block + scoped + peer-suffixed :: ["@scoped/block@3.1.0","block@2.0.0"]
OLD_PARSER_EXIT=1
```

The tampered copy's CVE failure **vanished into an INFO line**, and coverage silently fell 55 → 48. The only red left is the new `parser self-check` — i.e. the regression fixture is exactly what stands between this defect and a green build. (Without it, that run is exit 0.)

## (c) An orphaned / unreachable copy does NOT fail the run

Synthesize the orphan class the gate hit — a store dir for `brace-expansion@1.1.12`, a version the lockfile resolves nothing to (`0` hits), with the CVE bounds stripped so it *would* fail if probed:

```
$ node scripts/brace-glob-probe.mjs; echo "EXIT=$?"
55/55 checks passed.

INFO (not asserted, does not affect exit code) — 1 item(s):
  brace-expansion :: brace-expansion@1.1.12 on disk (.pnpm/brace-expansion@1.1.12) but NOT in the resolution graph — not probed
EXIT=0
```

(A/B on the first revision: `git show main:scripts/brace-glob-probe.mjs` on this same store reports `47/48`, `EXIT=1`, naming `brace-expansion@1.1.12` — the #905 false positive.)

## (F2) Coverage floor — a PINNED name dropping out of the graph fails

Rename every `picomatch@<n>` lockfile key so the graph resolves zero picomatch versions:

```
$ node scripts/brace-glob-probe.mjs; echo "EXIT=$?"
brace-expansion  resolves to >=1 version (coverage floor)  PASS  2 version(s) in the resolution graph
minimatch        resolves to >=1 version (coverage floor)  PASS  3 version(s) in the resolution graph
picomatch        resolves to >=1 version (coverage floor)  FAIL  0 version(s) in the resolution graph

52/53 checks passed.
FAILURES:
  picomatch :: resolves to >=1 version (coverage floor) :: 0 version(s) in the resolution graph
EXIT=1
```

The 55 → 53 shrink (the two picomatch override checks) is exactly the silent-collapse mode; the floor now makes it red.

## (F4) An unparseable override pin fails instead of downgrading to INFO

Temporarily change the `minimatch@3` pin to a form `satisfies()` can't evaluate:

```
$ node scripts/brace-glob-probe.mjs; echo "EXIT=$?"
minimatch@3.1.4  covered by a root override pin  FAIL  pin minimatch@3 = '>=3.1.4' is not a form this probe can evaluate — pin left UNVERIFIED

54/55 checks passed.
FAILURES:
  minimatch@3.1.4 :: covered by a root override pin :: pin minimatch@3 = '>=3.1.4' is not a form this probe can evaluate — pin left UNVERIFIED
EXIT=1
```

---

All tampering was reverted after each proof; `git status --porcelain` shows only the intended source changes.

**Files changed:** `scripts/brace-glob-probe.mjs`, `.github/workflows/ci.yml`, `package.json`.

**Deliberately not addressed here** (reviewer is filing separate issues): F3 `@isaacs/brace-expansion` unprobed under `minimatch@10`; F5 `specs.find` order-dependence with overlapping override specs; F6 non-numeric specifiers (`link:` / `git+`) skipped.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,10 @@ jobs:
               - 'pnpm-lock.yaml'
               - 'pnpm-workspace.yaml'
               - 'tsconfig*.json'
+              # The brace-glob probe runs inside this job (#905). Any edit to it
+              # must re-run it, and `package.json` / `pnpm-lock.yaml` above are
+              # already the inputs it guards (the pnpm.overrides pins).
+              - 'scripts/brace-glob-probe.mjs'
               - '.github/workflows/ci.yml'
             onchain:
               - 'onchain-academy/**'
@@ -106,6 +110,22 @@ jobs:
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
+
+      - name: Brace-glob probe (minimatch / brace-expansion resolution)
+        # #905. The ONLY detector for a regression class the green suites
+        # provably miss: #820's `brace-expansion: ^5` override passed 2354 tests
+        # and 6/6 typechecks while breaking every brace glob at runtime, because
+        # non-brace patterns short-circuit before expansion is ever called. This
+        # step drives real brace patterns through every minimatch /
+        # brace-expansion / glob / test-exclude copy the lockfile RESOLVES to
+        # (orphaned store dirs are reported as INFO and never fail the build),
+        # asserts the CVE-2026-14257 bound holds in each, and checks every
+        # resolved version still satisfies its root `pnpm.overrides` pin.
+        #
+        # Hard gate on purpose — a dependency change that reintroduces a
+        # vulnerable or incompatible copy must fail the build, not become tribal
+        # knowledge. Runs right after install so it fails before the slow steps.
+        run: pnpm probe:brace-glob
 
       - name: Prettier
         # RATCHET: report-only until the repo is prettier-clean. Run `pnpm format`
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "format:check": "prettier --check \"**/*.{ts,tsx,js,jsx,json,css,md}\"",
     "typecheck": "turbo typecheck",
     "test:scripts": "./apps/web/node_modules/.bin/vitest run --config vitest.config.ts",
+    "probe:brace-glob": "node scripts/brace-glob-probe.mjs",
     "clean": "turbo clean",
     "prepare": "husky"
   },
--- a/scripts/brace-glob-probe.mjs
+++ b/scripts/brace-glob-probe.mjs
@@ -9,14 +9,30 @@
  * `glob.sync('*.{json,md}')` throw `TypeError: expand is not a function`.
  *
  * This script is the detector that caught it: it drives real brace patterns
- * through every installed minimatch / glob / test-exclude / brace-expansion
- * copy in the pnpm store, plus the ESLint config-resolution path, and reports
- * pass/fail per consumer. Run it before AND after any change to the
+ * through every minimatch / glob / test-exclude / brace-expansion copy that is
+ * REACHABLE IN THE RESOLUTION GRAPH, plus the ESLint config-resolution path,
+ * and reports pass/fail per consumer. Run it before AND after any change to the
  * brace-expansion or minimatch overrides in the root package.json.
  *
  *   node scripts/brace-glob-probe.mjs
  *
  * Exit code 0 = every reachable consumer expands braces correctly.
+ *
+ * Resolution-awareness (#905)
+ * ---------------------------
+ * The first cut of this probe enumerated copies by walking `node_modules/.pnpm`
+ * DIRECTORIES. A pnpm store accumulates orphans: a dir left behind by an earlier
+ * install that nothing resolves to any more. Those orphans are never loaded by
+ * anything, but the directory walk still probed them — on main after #900 that
+ * produced 9 VULNERABLE lines for brace-expansion 1.1.12/1.1.16 and minimatch
+ * 2.0.2/2.1.2 while `pnpm why` resolved nothing to them and the lockfile pinned
+ * only patched versions. Noise that loud hides a real regression.
+ *
+ * So the set of versions to ASSERT on now comes from the lockfile's resolution
+ * graph (`pnpm-lock.yaml` -> `snapshots:`, which is exactly what `pnpm why`
+ * reads), and the directory walk is demoted to "where do I find the bytes for a
+ * resolved version". On-disk copies that are not in the graph are still listed,
+ * as a non-failing INFO section, so a genuinely surprising store stays visible.
  */
 
 import { createRequire } from 'node:module';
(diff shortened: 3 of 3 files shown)


thomgabriel:
**CI evidence (proof (a) in the real pipeline)** — run [30635821958](https://github.com/solanabr/superteam-academy/actions/runs/30635821958), job `Frontend (lint · typecheck · test)`:

```
✓ Install dependencies
✓ Brace-glob probe (minimatch / brace-expansion resolution)
✓ Prettier
✓ Lint
✓ Content bundle freshness
✓ Typecheck
```

The new step runs on a **fresh** `pnpm install --frozen-lockfile` store (no accumulated orphans, so no INFO lines) and passes green as a hard gate, immediately after install and ahead of the slow steps.

PR #912, opened 2026-07-31
fix(db): kind-scope unsubscribe tokens so one token cannot revoke the other consent (#896)

### LINKED ISSUE #896: unsubscribe_token is not kind-scoped — one token kills either consent (pre-existing #779 shape)
F6 from #895's adversarial review (verified, not theoretical): a token taken from a REMINDER email, with &kind=reminders stripped, successfully calls unsubscribe_by_token and kills MARKETING consent — and vice versa. Both consents share email_subscriptions.unsubscribe_token; the kind-separation is real at the RPC level but not the token level. Bounded (holder = recipient; effect = less mail, fail-safe direction) and pre-existing from #779 — not a #895 regression. Fix shape: per-kind tokens (second token column + kind-scoped RPCs) or a signed kind claim in the token. Migration follows house pattern; MCP-gate-applied.

thomgabriel:
Closes #896

## The defect (F6, verified — pre-existing from #779/#869, not a #895 regression)

Marketing consent (`opt_in`) and reminder consent (`reminder_opt_in`) live on the same `email_subscriptions` row and **shared one secret**, `unsubscribe_token`. The kind separation was real at the RPC level (`unsubscribe_by_token` vs `unsubscribe_reminders_by_token`) but **not at the token level**, so `kind` was just a query-string hint the link holder controls:

* token from a REMINDER email, `&kind=reminders` **stripped** → `unsubscribe_by_token(tok)` matched → **marketing** consent killed;
* token from a MARKETING email, `&kind=reminders` **appended** → `unsubscribe_reminders_by_token(tok)` matched → **reminder** consent killed.

Bounded (holder = recipient, effect = less mail) but it revokes a consent the learner never acted on, and LGPD consent is purpose-bound.

## Fix — per-kind tokens (issue's option 1; no key management, no token-format change)

`supabase/migrations/20260731150000_kind_scoped_unsubscribe.sql`

1. `email_subscriptions.reminder_unsubscribe_token UUID NOT NULL DEFAULT gen_random_uuid()`, backfilled with a **fresh** uuid per existing row (never a copy of the marketing token), `UNIQUE`, plus `chk_email_subscriptions_distinct_tokens CHECK (reminder_unsubscribe_token <> unsubscribe_token)` — "the two consents never share a secret" is now a **table invariant**, so no future backfill/restore can re-merge them.
2. `unsubscribe_reminders_by_token` matches **only** `reminder_unsubscribe_token`; `unsubscribe_by_token` matches **only** `unsubscribe_token`. Cross-kind fails at the database, not at a caller-supplied parameter.
3. `claim_due_session_reminders` hands the send pipeline the **reminder** token. The OUT column keeps the name `unsubscribe_token` (CREATE OR REPLACE cannot rename OUT params; a rename would force a DROP) — only its source column moved, so the signature and the app read are untouched.
4. `list_marketing_recipients` unchanged — it already returns the marketing token for the marketing pipeline.

### Grants / security decisions

* Every function stays `SECURITY DEFINER` + `SET search_path = ''`, `REVOKE ALL … FROM PUBLIC, anon, authenticated` then `GRANT EXECUTE … TO service_role` **only**. Checked the existing intent first: in this codebase the unsubscribe RPCs are **not** anon-callable (unlike the common Supabase shape) — the unauthenticated one-click endpoint is the Next.js route `/api/email/unsubscribe`, which calls them with the service-role client. **Nothing here widens a grant** (asserted by a test).
* No new RLS policy. `email_subscriptions` keeps own-row `SELECT` only and no client write path, so a learner cannot read another learner's tokens.
* Idempotent DDL: `ADD COLUMN IF NOT EXISTS`, guarded `DO` blocks for both constraints, a backfill that only touches NULLs, `CREATE OR REPLACE FUNCTION`. Explicit `BEGIN/COMMIT`; rollback + **post-apply invariant checks** in the trailer comments (null/shared/distinct token counts, both constraints present, cross-kind probe = 0/0, grant matrix, claim-RPC source column).
* **Prod-lineage clean**: depends only on `20260728120000_email_subscriptions` (ledger `20260727200924`) and `20260731120000_reminder_consent` (ledger `20260731120914`) — both applied. References nothing from any never-applied migration. Ledger row added to `docs/DB-MIGRATION-LEDGER.md` as PENDING-GATE-VERIFY. **Not applied to prod by this PR** — that is the gate/ops lane, and it must be applied before merge (#731, Vercel auto-deploys `main`).

## Backward compatibility (already-issued unscoped tokens)

* **Marketing links: GRANDFATHERED.** `unsubscribe_token` keeps its value and its meaning — the kind it was originally issued for. Every marketing link ever minted (all kind-less, #779) still works; the route's kind-less default is still marketing. No rotation (rotating it would break live links to fix nothing).
* **Pre-#896 reminder links: EXPIRED, deliberately.** They carry the MARKETING token with `&kind=reminders`; honouring them would mean keeping exactly the cross-kind match this migration removes — there is no way to grandfather them without preserving the defect. They now return false → the route's "Link not recognized" page, which points at Settings where both consents are toggleable. Blast radius ≈ 0: reminder consent defaults FALSE and both the #869 settings toggle and the daily cron shipped the same day as this fix, so essentially no reminder mail carrying an old token is in flight.

## App code

* `campaign.ts` — marketing links now state their kind (`&kind=marketing`); route already accepted it.
* `reminders.ts` / `reengagement.ts` / `types.ts` — token provenance documented as reminder-scoped; the re-engagement pipeline TODO now pins that `claim_due_reengagement` MUST return `reminder_unsubscribe_token`.
* `/api/email/unsubscribe/route.ts` — docs updated: `kind` routing is now a convenience, not the boundary. Unknown/absent kind still resolves to marketing (fail-closed: a reminder token there matches nothing).
* `supabase/schema.sql` mirror updated.

## Test evidence — red/green

New pglite execution suite `apps/web/src/lib/supabase/__tests__/kind-scoped-unsubscribe-execution.test.ts` (19 tests) runs the same cross-kind attack against **both lineages**:

* **RED block** — migrations #779 + #869 only: the attack SUCCEEDS and is pinned (`unsubscribe_by_token(reminder-email token)` → true, marketing consent flipped; `unsubscribe_reminders_by_token(marketing token)` → true, reminder consent flipped; no reminder-scoped column exists).
* **GREEN block** — + `20260731150000`: both directions return **false** and write nothing; each token still works for its own kind only; cross-user × cross-kind is also dead; backfill gives every pre-existing row a distinct/unique token; the CHECK and UNIQUE constraints reject a re-merge; a new row gets its own token by DEFAULT; the claim RPC hands out the reminder token; the migration re-applies as a no-op; both backward-compat decisions above are asserted.

```
$ pnpm vitest run src/lib/supabase/__tests__/kind-scoped-unsubscribe-execution.test.ts
 Test Files  1 passed (1)
      Tests  19 passed (19)
```

**Mutation check** (proves the suite is not vacuous) — reverting just the fix line in the migration (`WHERE reminder_unsubscribe_token = p_token` → `WHERE unsubscribe_token = p_token`) turns 6 GREEN tests red while the RED block keeps passing:

```
 × the MARKETING token can no longer touch reminder consent
 × each token still works for ITS OWN kind, and only that kind
 × a token of one kind never belongs to ANOTHER learner's other kind
 × the send pipeline is handed the REMINDER token, never the marketing one
 × re-applying the migration is a no-op (idempotent replay)
 × BACKWARD COMPAT: a pre-#896 REMINDER link is expired, not honoured
      Tests  6 failed | 13 passed (19)
```
(migration restored immediately after.)

Full suite, typecheck, lint:

```
$ pnpm --filter web typecheck   # clean
$ pnpm --filter web lint        # only pre-existing import/order warnings
$ pnpm --filter web test        # 257 files, 2392 tests passed
```

Note on runtime: the suite builds ONE PGlite per lineage (not per test) and resets rows between tests — 4.7s instead of 44s, so it does not push the other pglite suites into hook timeouts under parallel runs.

## Not done on purpose

* No prod apply (gate/ops lane owns it).
* `profiles` auth-sensitive columns and `auth.users` untouched.


thomgabriel:
Applied both non-blocking review suggestions in 6e7d006:

1. **schema-expectations** — added a `column` probe for `email_subscriptions.reminder_unsubscribe_token` pointing at `20260731150000_kind_scoped_unsubscribe.sql`, so `/api/health/schema` catches a merge that lands ahead of the MCP apply (the exact #731 hazard, since Vercel auto-deploys `main`). Placed before the side-effectful `claim_due_session_reminders` probe; it is a plain column read, no write path.
2. **types.ts** — `email_subscriptions` `Row`/`Insert`/`Update` now carry `reminder_unsubscribe_token` (Row required, Insert/Update optional — it has a DB DEFAULT), with a note that it is never equal to `unsubscribe_token` per the CHECK.

`pnpm --filter web typecheck` clean; `src/lib/supabase` suite 19 files / 285 tests green.

Reminder for the gate: this PR must NOT merge before `20260731150000_kind_scoped_unsubscribe.sql` is MCP-applied to prod, and the ledger row in `docs/DB-MIGRATION-LEDGER.md` (currently PENDING-GATE-VERIFY) filled with the MCP-stamped version. Post-apply invariant checks are in the migration's trailer.

PR #925, opened 2026-07-31
fix(gamification): award daily quests at the action, celebrate the reward anywhere

thomgabriel:
Owner-reported defect (2026-07-31). No issue number — reported directly.

## Part 1 — quests were only rewarded when the learner opened the dashboard

`get_daily_quest_state` both **evaluates** quest progress and **awards** the XP for a quest that just completed. It was invoked from exactly one place — `GET /api/quests/daily` — whose only real caller is `use-dashboard-data.ts`. So a learner who completed lessons, passed challenges, or cleared reviews without visiting `/dashboard` received quest XP late, or never; the `retryQuestXpForUser` sweep that actually credits the queued XP also only ran on that GET.

**Fix.** The evaluation is factored into one shared server lib, `apps/web/src/lib/gamification/quest-evaluation.ts`, and now runs at the moment quest-relevant things happen:

| Trigger point | When |
| --- | --- |
| `POST /api/lessons/complete` | after the successful completion's `user_progress` write |
| `POST /api/lessons/complete` | after the already-on-chain **recovery** upsert (that mirror can be the row that completes a quest) |
| `POST /api/courses/test-out` | after `batchCompleteCourse` (a whole course of progress in one shot) |
| `POST /api/review/grade` | on a **passed** review only (a miss resets the box and is not a clear) |
| `handleLessonCompleted` (Helius webhook) | after the on-chain-authoritative progress mirror |
| `GET /api/quests/daily` | unchanged for the panel — now calling the same shared `evaluateQuests` |

**Latency.** The action paths call `scheduleQuestEvaluation(userId)`, which hands the work to Next 15's `after()` continuation: the work is owned by the platform and runs once the response has been sent, so the hot lesson-completion path pays nothing. It is not a bare floating promise; if `after()` is unavailable (outside a request scope) it degrades to an invoked promise with the same error logging. Every failure reaches `logError` (`QUEST_EVAL_001`) — nothing is dropped silently.

**Durability.** Unchanged and already strong: the RPC writes the `pending_onchain_actions` row in the *same transaction* that flips `xp_granted`, and delivery via `award_xp` is idempotent on `reference_id`. So a failed evaluation or sweep delays a reward, never loses one — the next action, poll, or login sweep picks it up.

**Day semantics unchanged.** Nothing passes a date; the RPC still keys everything on `CURRENT_DATE` (UTC), and `nextMidnightUtc` is untouched. A test asserts the RPC args carry no date/period/day key.

**No migration.** The existing SECURITY DEFINER RPC is reused verbatim from the server; nothing about the schema needed to change.

*Behaviour note for the reviewer:* the `login_streak` quest treats an evaluation as a login signal ("dashboard load = login"). It now also ticks when the learner completes a lesson — i.e. genuinely being in the app counts. Still at most one increment per day (the RPC's own state machine); flagging it as an intended widening rather than a silent one.

## Part 2 — the reward is now celebrated in the house style, wherever the learner is

Before: the dashboard fired `dispatchToast(tGam("questReward", …))` — a plain info toast — and `use-gamification-events.ts` deliberately suppressed `daily_quest:*` XP events ("quests panel already shows the reward"). That assumption only held while quests were awarded by the dashboard's own poll.

- New `components/gamification/quest-reward-toast.tsx` — `dispatchQuestReward()` + `QuestRewardToastListener`, mounted in `GamificationOverlays` next to the surprise-bonus listener (same house split: dispatch is provider-free, localization lives in the listener).
- The Realtime `daily_quest:*` suppression is **lifted** and routed to the same component; the dashboard poll path now dispatches to it too. One component, one look, both channels.
- Celebration goes through `celebrate("daily-quest")`, registered in `CELEBRATION_TIERS` at the **`popup`** tier. This is a deliberate call: every peer moment a learner meets this often is popup-or-quieter (level-up `popup`, surprise bonus `none`), and LX-B11 pins confetti to deploy + credential mint precisely because a *daily*-cadence reward with confetti is the routine-reward pattern the tiering exists to prevent. The existing `celebration.test.ts` assertion "confetti-worthy tiers are EXACTLY deploy + credential mint" still passes. **If you want a burst anyway it is a one-word change** — flip `"daily-quest"` to `"medium"` in `CELEBRATION_TIERS`; the call site already routes through `celebrate()`.
- Accessibility: the shared `ToastContainer` is already `aria-live="polite"` with a labelled dismiss button, identical to peers.
- i18n: reuses the existing `gamification.questReward` key, already present in **en / pt-BR / es** — no new strings, no locale-parity risk.

### The double-toast proof

One award can now be observed twice, moments apart: by the poll (`justAwarded`) and by Realtime (the queued credit landing in `xp_transactions`). Both channels claim through the **existing** `#790` session dedupe — extended, not forked — via new `claimQuestRewardFromCredit`, keyed on `questId + period`:

- period comes from the queue row's `idempotency_key` (`<questId>:<period>`, written by the RPC), falling back to the credit's `created_at` date;
- the poll path now uses the **server's** `questPeriod` (new field on the GET response) instead of the browser's local date — a São Paulo evening (UTC-3) is already the next UTC day, and a client-derived date would have keyed the two channels differently and let one award toast twice.

Tests pin both arrival orders plus the timezone case.

## Screenshots, in words

- **Before:** learner finishes the last lesson of the day on `/courses/x/lessons/y`. Nothing happens. Quest XP appears only when (if) they later open `/dashboard`, where a grey info toast reads "Quest complete: Complete a Lesson +25 XP".
- **After:** the same learner, still on the lesson page, sees a green success toast pop in bottom-right with the pop-spring easing the level-up card uses — check-circle icon, "Quest complete: Complete a Lesson +25 XP", dismiss ✕, auto-fading after 5s, announced politely to screen readers — and the header XP counter ticks up at the same moment. Opening `/dashboard` afterwards shows the quest already ticked and **no second toast**. Reverse order (dashboard first) behaves identically, with Realtime staying silent.

## Verification

**Red proof** — with the two `scheduleQuestEvaluation(user.id)` calls commented out of `/api/lessons/complete`, `quest-award-timing.test.ts`:

```
× evaluates daily quests for the learner on a successful completion
× evaluates daily quests on the already-on-chain recovery sync too
AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
  Tests  2 failed | 3 passed (5)
```

That failure *is* the bug: a quest-satisfying completion goes by with nothing evaluating it. **Green** with the trigger restored: `5 passed`.

**No-double-award proof** — `quest-evaluation.test.ts` drives a faithful stand-in for the RPC's `WHERE xp_granted = false` guard and fires three independent paths in one day: `justAwarded` is `true, false, false`, and exactly **one** XP credit is enqueued, while all three calls still report the quest complete for the panel.

```
apps/web  vitest run src/lib/gamification src/app/api/lessons src/app/api/review
  Test Files  24 passed (24)
  Tests  183 passed (183)

apps/web  tsc --noEmit          → clean
apps/web  next lint --dir src   → no errors (pre-existing import/order warnings only)

apps/web  vitest run (full)
  Test Files  3 failed | 257 passed (260)
  Tests  4 failed | 2465 passed (2469)
```

The 3 full-run failures are the PGlite in-process-Postgres suites (`assist-ladder-execution`, `email-subscriptions-execution`, `reminder-consent-execution`) timing out in `beforeEach` under parallel load — pre-existing and unrelated. Re-run in isolation: **3 files / 34 tests passed**.

Do not merge — for review.


thomgabriel:
**Blocking review finding fixed** — pushed `d19d2552`.

`dispatchXpGain(amount)` in the `daily_quest` Realtime branch now lives **inside** the `if (questId)` won-claim block. Previously it fired unconditionally, so any award observed by both channels (near the common case once the action paths trigger evaluation) bumped the header's monotonic optimistic XP twice — and `Math.max(supabaseXp, targetXpRef.current)` never pulls it back down, permanently inflating the displayed balance for the session. It now mirrors the poll path, which only bumps for entries `pickQuestRewardToasts` actually claimed.

**Regression tests** — `src/hooks/__tests__/use-gamification-events.test.tsx`, three cases counting the `xp-gain` CustomEvents dispatched across the real hook + real `server-xp-feedback` dedupe:

- `Realtime alone (no dashboard open) celebrates and bumps XP exactly once`
- `poll first, then Realtime: ONE toast and ONE xp-gain for the same award`
- `Realtime first, then poll: the poll claims nothing (no second bump)`

**Red proof** (unconditional dispatch restored):

```
× poll first, then Realtime: ONE toast and ONE xp-gain for the same award
AssertionError: expected 1 to be +0 // Object.is equality
  Tests  1 failed | 11 passed (12)
```

**Green** with the fix:

```
vitest run src/hooks src/lib/gamification src/app/api/lessons src/app/api/review
  Test Files  27 passed (27)
  Tests  200 passed (200)

tsc --noEmit          → clean
next lint --dir src   → no errors
```

The identical pattern in the surprise-bonus branch a few lines below is left untouched as directed — separate follow-up.

Issue #923, opened 2026-07-31
Teach preview: all course images render broken — asset bytes are discarded at compile


Potolski:
Every image in a previewed course renders as a broken-image icon. Example: `/en/teach/preview/34/courses/btc-to-sol-evolution/lessons/the-trust-machine`.

## Cause

A bug I introduced building the preview (#831). The compiler has two entry points:

- `compileBundle(tree, opts)` → `{ files, assets }` — JSON modules **and** the asset bytes
- `compileContent(tree, opts)` → `files` only, documented as *"JSON-modules-only view of compileBundle (assets written separately)"*

`lib/teach/preview-compile.ts:215` calls **`compileContent`**. So the preview gets modules in which every image reference has already been rewritten to `/content-assets/<courseSlug>/<lessonSlug>/<file>` (`compile-bundle.ts:332`) — but the corresponding bytes, which live in the discarded `assets` map, are never served anywhere.

Those URLs resolve against `apps/web/public/content-assets/`, which is populated only by a real `pnpm compile-content` run for the **pinned** SHA. A previewed PR's assets are not there — and for a brand-new course they don't exist at all — so every `<img>` 404s. Hence *all* images, not some.

## Fix

Serve preview assets from the already-cached in-memory bundle:

1. `preview-compile.ts` switches to `compileBundle` and keeps `assets` (keyed by public rel path → bytes).
2. Rewrite `/content-assets/` → `/api/teach/preview/{pr}/assets/` in the compiled JSON modules before they are parsed, so the preview's own refs point at a route that can serve them.
3. New `GET /api/teach/preview/[pr]/assets/[...path]` — behind the existing preview session cookie — looks the path up in the cached bundle and returns the bytes with a content type derived from the extension.

Path traversal is impossible by construction: the route does a **map lookup** against compiler-generated keys, never a filesystem read.

## Note

Asset bytes now sit in the preview cache (bounded: 8 bundles, 10-minute TTL). Acceptable for a preview tool used by a handful of teachers; worth revisiting if course assets grow large.

PR #927, opened 2026-07-31
test(teach): execute the preview asset path, not a copy of it (#923 follow-up)

thomgabriel:
Follow-up to #924 (David's fix for #923, which is correct and already merged — this changes no production code).

## Why

#924's tests define a local `repointAssets()` helper and assert against *that*, so the shipped path is never executed: `preview-compile.ts` could revert to `compileContent` and all five tests would still pass. The new serving route, `GET /api/teach/preview/[pr]/assets/[...path]`, has no test at all — including its auth gate.

## What

- **`lib/teach/__tests__/preview-compile-assets.test.ts`** — drives the real `compilePrPreview` against a tarball fixture and pins both halves of a working image together: the bytes survive on `result.assets`, the projected lesson references `/api/teach/preview/34/assets/demo/accounts/pixel.png`, nothing `/content-assets/` is left, and every asset key is addressed by a rendered url (drift between the two halves is a broken image that no unit test of either half alone would catch).
  Red-proofed: with `preview-compile.ts` restored to its pre-#924 content, 3/3 fail (`Cannot read properties of undefined (reading 'keys')`, and the lesson still carries `/content-assets/`).
- **route tests** — bytes + `image/png`; the SVG hardening (`nosniff`, `default-src 'none'`, `Cache-Control: private`); **401 without a preview session, asserting the bundle is never touched**; 404 on traversal (`..` segments) and on a key the compiler never emitted; 400 on a non-numeric `pr`; 502 on a failed compile.
- **`_fixtures.ts` repair** — `makeCourseTarball` predates the #466 C3 `skills` requirement, so every tarball it produced failed `validateTree`. It had no callers left, so nobody noticed; it is now valid (lesson `skills: [pdas]` + a root `skills.yaml`) and usable again.

## Verification

- `pnpm --filter web test` — 266 files, 2545 tests, all green
- `pnpm --filter web typecheck` — clean
- `pnpm --filter web lint` — no new findings

Diagnosis and fix credit for #923 goes to David (Potolski).

### DIFF
--- a/apps/web/src/app/api/teach/preview/[pr]/assets/__tests__/route.test.ts
+++ b/apps/web/src/app/api/teach/preview/[pr]/assets/__tests__/route.test.ts
@@ -1,0 +1,126 @@
+/* eslint-disable import/order -- vi.mock factories are hoisted above imports. */
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+vi.mock("server-only", () => ({}));
+
+// The store is the compile cache; stub it so this file is about SERVING bytes.
+// The compile itself is covered in `lib/teach/__tests__/preview-compile-assets`.
+const store = vi.hoisted(() => ({ getPreviewBundle: vi.fn() }));
+vi.mock("@/lib/teach/preview-store", () => store);
+
+import { NextRequest } from "next/server";
+import { GET } from "../[...path]/route";
+import {
+  mintPreviewSession,
+  TEACH_PREVIEW_COOKIE,
+} from "@/lib/teach/preview-auth";
+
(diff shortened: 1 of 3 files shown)


thomgabriel:
Re the review nit (single-asset fixture makes assertion 3 weakly independent) and the builder's two design notes on #924's route — recorded here rather than filed, none are live defects: (1) a two-image fixture would strengthen the every-key-addressed assertion; (2) the route's application/octet-stream fallback for unknown extensions is safe today only because the compiler's extension allowlist bounds the map keys — revisit if the allowlist widens; (3) preview asset URLs are cookie-gated, so wrapping previewed images in next/image would 401 (server-side optimizer refetch) — fine while markdown renders plain img tags.

Issue #868, opened 2026-07-28
Per-action Gemini model routing: ask/propose → 3.6-flash, hint → 3.5-flash-lite (item 3b)


thomgabriel:
## Context
Unified spec item 3b — per-action Gemini model routing (economics doc E-2: ~32% blended cost cut, "one URL constant plus a per-action switch, no product change"). Routing: **`ask`/`propose` → `gemini-3.6-flash`**, **`hint` → `gemini-3.5-flash-lite`**. Both targets are VERIFIED REACHABLE on the prod key — see the 2026-07-28 curl record in #838's comments (AIE-04: all three candidates HTTP 200; AIE-05: `thinkingBudget: 0` verified honored with a control). Note the route's line-69 comment claiming 2.5-model gating for new keys is **stale** per the same record; also beware the first-connection artifact noted there (spurious 404s with empty bodies on first attempts — retry twice before believing a 404).

**Flex-tier spike is NOT in scope** unless behind a measured exit criterion: only pursue if p95 latency on `hint` < 4s is demonstrated (owner decision **D-7** assigns who runs that spike — ask when batching the item-32 decision questions; do not block this issue on it).

## Where
- `apps/web/src/app/api/ai/partner/route.ts` — `GEMINI_URL` (lines 75–76, single constant pinning `gemini-3.5-flash`): replace with a per-action model map; update the stale model-history comment (lines 69–74); keep the thinking-token handling (line ~457) correct per model (3.6-flash and 3.5-flash are thinking models — retain `thinkingBudget: 0` where the action doesn't want thinking spend).
- `apps/web/src/lib/ai/reflection-reply.ts` — its own `GEMINI_URL` (lines 26–27, also `gemini-3.5-flash`): route to the cheap model per the economics doc's low-stakes classification (it runs "capped and low-latency" — flash-lite is the fit; confirm against the doc §E-2 and #724's constraints noted in the file header).
- Cost tracking: `apps/web/src/lib/ai/spend-ledger.ts` — if per-model pricing is encoded anywhere, update it to the routed models.

## Done when
- [ ] `ask`/`propose` hit `gemini-3.6-flash`; `hint` hits `gemini-3.5-flash-lite`; reflection-reply routed per the doc
- [ ] Stale gating comment corrected, citing the 2026-07-28 curl record
- [ ] Structured-output/thinking config verified per model (a failing model 404 must fail closed, not fall back silently to an unrouted model)
[...]

Issue #865, opened 2026-07-28
Replace the #770 hard think-first lock with the soft attempt-gate nudge + logged override (item 32a)


thomgabriel:
## Context
Unified spec item 32a. The shipped #770 **think-first lock is the exact shape the economics research forbids** (economics-G-1 verified rows in the reconciliation TSV: hard 3-minute gate, padlock icon, no override). This issue **REPLACES** it — it does not add alongside — with the soft attempt-gate nudge: on early AI use, nudge "run the tests once first" with a **free one-tap override**, and log the override rate as a metric. Hard gates/padlocks are forbidden by the research; the nudge is the compliant shape.

## Where
- `apps/web/src/components/editor/challenge-interface.tsx` — `AI_LOCK_MS` and the lock wiring: REMOVE the hard lock.
- `apps/web/src/components/editor/ai-partner/ai-partner-pane.tsx` — the `unlockAt` countdown UI (lines ~34–66) + the `Lock` icon import: REMOVE; replace with the nudge UI (shown when the learner invokes AI before any test run; one tap dismisses AND proceeds — the override must not cost anything or add friction beyond the single tap).
- `apps/web/src/components/editor/__tests__/challenge-interface.ai-lock.test.tsx` — rewrite for the new behavior.
- `apps/web/src/lib/analytics/events.ts` — add the override-rate event (e.g. `trackAttemptGateNudgeShown` / `trackAttemptGateOverridden`), following the existing `trackStuckNudgeShown/Accepted` pattern (lines 116–155).
- Strings via next-intl ×3 (`apps/web/src/messages/*.json`).

## Done when
- [ ] `AI_LOCK_MS`, `unlockAt`, countdown, and padlock UI are gone (grep clean)
- [ ] Nudge appears only when AI is invoked with zero test runs on the current challenge; one-tap override proceeds immediately, free
- [ ] Nudge-shown + override events fire; override rate derivable in PostHog
- [ ] Old lock tests replaced; new tests cover nudge-shown, override, and no-nudge-after-first-run
- [ ] ×3 locales

## References
Parent epic: #838 (item 32a) + its completion-pass comment (the #770 lock is the forbidden shape — evidence: economics-G-1 rows in `docs/superpowers/specs/2026-07-27-RESEARCH-RECONCILIATION-items.tsv`). RESEARCH-RECONCILIATION §2, §8.1. Replaces the mechanism from #770.

Issue #864, opened 2026-07-28
AI assist ladder + wall boundaries: free → metered → Socratic, degrade-never-block (item 32)


thomgabriel:
## Context
Unified spec item 32 — the AI assist **ladder + wall boundaries** — was the reconciliation's headline dropped workstream (§2). Design source: `docs/superpowers/specs/2026-07-25-ai-tutor-economics.md` §5 (the ladder) — build to the doc, not to this summary. Shape: per-(user, lesson) **free tier → metered tier → Socratic tier** (doc's recommendation: 2 free AI turns → 8 metered → 20 Socratic, all catalog-bounded, `propose` alive at every tier), **degrade-never-block** (exhaustion moves you down a tier, never to a hard wall), **turn-31 community handoff** (after the full ladder, hand off to community/forum — not a paywall), and exhaustion copy framed as "lighter tutor" — **never paywall-shaped**.

**⚠️ Owner-decision batch — ask BEFORE implementing, as ONE batch (not piecemeal):** building this forces five parked decisions (#838 completion-pass comment / recon §8.1): **D-4** free AI turns per challenge (2 vs 1), **D-5** Socratic tier size (20/lesson vs fewer; "unlimited" framing banned), **D-6** named compute sponsor attribution ("Compute donated by X"), **D-8** keep/cut the per-lesson self-reset (P2-5; guards-inside-RPC rule R-6 applies if kept). (D-7, the Flex spike, belongs to the model-routing sibling issue but batch the question with these.)

## Where
- `apps/web/src/lib/ai/partner-types.ts` — `MAX_PAID_ASSISTS = 4` (line 122) is the current single-meter model; the ladder replaces/extends this typed contract.
- `apps/web/src/app/api/ai/partner/route.ts` — the server enforcement point (quota checks, spend ledger integration).
- `apps/web/src/lib/ai/assist-budget.ts` — budget accounting; `apps/web/src/lib/ai/spend-ledger.ts` — the fail-closed daily spend ledger (São Paulo days) that stays the real ceiling above the ladder.
- UI: `apps/web/src/components/editor/ai-partner/assist-meter.tsx` (meter display), `ai-partner-pane.tsx` (tier/exhaustion copy). All strings via next-intl ×3 locales (`apps/web/src/messages/*.json`).
[...]

PR #947, opened 2026-07-31
feat(ai-partner): restore the free-text composer on top of the assist ladder

### LINKED ISSUE #944: Restore the AI Partner composer: free-text ask on top of the assist ladder
Owner call 2026-07-31: 'the AI partner is not a chatbot anymore?' — the free-text ask field was deliberately removed in #770 ('hints only… the tutor's role is to nudge'). That stance is superseded: #865's attempt-gate nudge + #864's ladder (tiered budget, Socratic contract for ask, tier-exact refunds) + #868's per-action model routing already solve answer-dumping properly, and the ask/propose plumbing is live in use-ai-partner + /api/ai/partner. Restore a composer input in ai-partner-pane.tsx/quick-actions.tsx wired to the existing ask action (attempt-gate + ladder + Socratic all apply unchanged); keep the hint quick-action. Update the #770-era comments. Tests: composer renders post-gate, ask spends a ladder turn, Socratic tier answers ask with questions-only (existing contract tests cover the backend).

thomgabriel:
Closes #944

## Why

#770 deleted the free-text ask field ("hints only — the tutor's role is to nudge, not to answer"). That stance is superseded:

- **#864 assist ladder** — free → metered → Socratic → community handoff, tier-exact billing
- **#865 attempt-gate nudge** — a soft "run the tests first" on the first pre-run AI use
- **#868 per-action model routing** — ask → gemini-3.6-flash, Socratic ask → flash-lite

Those constrain *how* the tutor answers and *how much*, which is what hints-only was really protecting. The `ask` action was already live in `use-ai-partner` + `/api/ai/partner` (with its Socratic contract and contract tests) — only the UI input was missing.

## What

- **New `composer.tsx`** — textarea + Send. Enter sends, Shift+Enter newlines, sr-only `<label htmlFor>`, `focus-visible:ring-2 ring-ring`, trims and refuses empty input, clears on send.
- **Same guards, unchanged.** Ask is routed through the very same `guardAction` as the hint, so a pre-first-run question is *held* and the #865 nudge appears; the free one-tap override replays the exact question. It spends a ladder turn like any other AI action, and disables at `budgetExhausted` (the community-handoff block above already carries that copy — no second wall), while `loading`, and when the lesson is complete (`disabled`), matching hint behavior. Spend-cap state untouched.
- **Placement / empty state (judgment calls).** The footer is one framed block: invitation line (empty state only) → Hint button → composer. The composer *is* the empty-state invitation ("Ask anything about this challenge, or grab a hint"), replacing the old hint-only empty block, so the pane stays compact — no reserved conversation area until there are messages. `QuickActions` gave up its own border/padding to the footer so Hint + ask read as one unit.
- **Comments** in `quick-actions.tsx` / `ai-partner-pane.tsx` rewritten to cite #944 and why the #770 stance no longer holds. Propose-a-fix stays unexposed.
- **i18n ×3** — new `aiPartner.actions.askLabel` + `aiPartner.composer.empty`, real PT-BR/ES translations; `subtitle` updated to mention asking (the pre-existing `askPlaceholder`/`askSend` keys are finally used again).

## Tests

`ai-partner-pane.test.tsx` gains a `#944` block plus one nudge case:

- composer renders post-gate as the empty state, with the Hint button beside it
- ask sends through the hook (mocked) with the trimmed message and clears the box
- Enter sends / Shift+Enter does not / empty never sends
- disabled on `budgetExhausted` (handoff copy still shown), on `loading`, and on lesson-complete
- attempt-gate intercepts a pre-run ask, then the override sends it unchanged

`pnpm --filter web test src/components/editor/ai-partner` → 3 files / 37 tests pass. `pnpm typecheck` clean, `pnpm lint` no errors and no new warnings, prettier clean. Backend contract tests untouched.

---

## Propose action + M1, per gate addendum

Two follow-ups landed on this branch after the review pass above. (The "Propose-a-fix stays unexposed" line in *What* is now superseded — the comments that said so were rewritten in the same pass.)

### 1. `feat(ai-partner)` — "Show me a change" (propose) in the composer

The server side has been complete for a while: `/api/ai/partner` accepts `action: "propose"` (compact search/replace edits, 2,048-token cap, no full-file echo, tier-exact refund on failure), `diff-card.tsx` renders it, `applyEdits` reconstructs and fails closed on a stale buffer, Accept sits behind the earned-Accept comprehension seal (`check-seal.ts`), and `buildDynamicSuffix` deliberately keeps propose free of the Socratic contract (#864 §4.2). Only the affordance was missing — the learner could never *ask* for a change.

- **Secondary ghost action** in the composer beside Send, quieter than Hint/Send, never the primary affordance. Inline cost signal: the label carries "uses 1 turn" (localized), consistent with the meter.
- **Visibility** — hidden until at least one run has **FAILED** (a diff is never the opening move). New sticky `hasFailedRun` signal, threaded from `challenge-interface.tsx` exactly like `hasRunTests`; set on a failing run, not retracted by a later pass. Available at every tier **including Socratic**; withdrawn again at full exhaustion, where the community handoff takes over.
- **Same guards as every other assist** — the #865 attempt gate (a pre-first-run propose is held and replayed by the free override), one ladder turn, disabled (with `aria-busy`) while a request is in flight and on `budgetExhausted`. No propose-specific budget or limit, no auto-apply, no weakening of earned-Accept, no Socratic instructions appended to propose, `socratic.entered` copy untouched.
- **No free text required.** A draft already in the ask box rides along as `message` and is deliberately **not** cleared — the learner asked for a diff, not for their question to be consumed. The response renders through the existing diff-card path in the message list.
- **Analytics** — new `trackProposeAccepted(eventCtx)` in `lib/analytics/events`, fired from the message list when an accepted patch is applied: exactly once per accepted propose (the card latches after its single Accept), so `diff-card.tsx` behavior is unchanged. `trackProposeShown` was left out — see judgment calls.
- **a11y** — real `<button>`, inherited focus-visible ring, `aria-busy` while pending; diff-card behavior unchanged.
- **i18n ×3** — `aiPartner.actions.propose` repurposed to "Show me a change" + new `aiPartner.actions.proposeCost`, identical key structure across en / pt-BR / es.
- **Hook** — `proposeFix` now takes an optional `message`. That is the only plumbing change; the route, diff card, `applyEdits`, seal and prompt builder are untouched.

### 2. `fix(ai-partner)` — M1: composer `maxLength` + live counter

The route 413s a `message` over 4,000 chars and the hook could only render that as the generic "Something went wrong", so a long paste was lost with no explanation. The textarea now carries `maxLength={4000}` plus a `slice()` in `onChange` (a paste past the cap is truncated, never sent into a 413) and a live `N/4000` counter that turns danger past 90% — the same shape as the community markdown editor's `N/10000`. The counter is wired via `aria-describedby` so it is announced with the field.

### Tests

- Visibility: hidden before the first failed run, visible after, present on the Socratic tier, withdrawn at exhaustion.
- A propose click spends a turn via the existing hook path (`proposeFix` called once, with the draft) and the typed draft survives; a propose with no draft passes `undefined`; the response renders as a diff card; the pre-run attempt gate holds and replays it; disabled + `aria-busy` in flight.
- **Regression**: a Socratic-tier propose — *including one carrying a learner draft* — produces a suffix with no `[SOCRATIC_INSTRUCTIONS]`, pinning `buildDynamicSuffix` (`partner-prompt.test.ts`).
- Accept is still locked until the comprehension seal verifies; `trackProposeAccepted` fires exactly once per accepted propose, with the challenge payload.
- i18n parity across the three catalogs (plus the repo-wide `messages/__tests__/parity.test.ts`).
- M1: `maxlength` attribute, counter at 0 / 5 / 4000, and a 4,200-char paste truncated to 4,000.

**Evidence** — `tsc --noEmit` clean; `next lint` over `src/components/editor`, `src/lib/ai`, `src/lib/analytics` reports no errors and no new warnings; full `apps/web` vitest **264 files / 2553 tests passed** (was 2541 — +12 new).

### Judgment calls

1. **Exhaustion: hidden, and also disabled.** The spec says both "hidden again at full exhaustion" and "disabled (not hidden) … when `budgetExhausted`". Implemented as hidden at the pane level (matching the stated test) *and* hard-disabled inside the composer, so it cannot fire even if it were ever rendered in that state.
2. **`trackProposeShown` skipped.** Marked optional; firing on render needs its own dedupe to avoid re-render inflation, and `propose_accepted` + the existing `comprehension_check_answered` already give the funnel its denominator. Cheap to add later if the funnel wants an impression count.
3. **Counter has no visible label**, matching the community editor's bare `N/10000`, so no new i18n string; the field-level announcement comes from `aria-describedby` instead.
4. **The draft is passed but no user bubble is appended** (unlike `ask`), because a propose is answered by the diff card, not by a reply — a bubble would imply the question was consumed.

### DIFF
--- a/apps/web/src/components/editor/ai-partner/__tests__/ai-partner-pane.test.tsx
+++ b/apps/web/src/components/editor/ai-partner/__tests__/ai-partner-pane.test.tsx
@@ -324,6 +324,25 @@ describe("AiPartnerPane — free-text composer (#944)", () => {
       screen.getByRole("button", { name: messages.aiPartner.actions.hint })
     ).toBeDisabled();
   });
+
+  // M1 (#947 gate addendum): the route 413s past 4,000 chars and the hook can
+  // only render that as a generic error, so the box refuses to exceed it and
+  // shows the learner where they stand.
+  it("caps the question at 4000 characters and shows a live counter", () => {
+    renderPane({ hasRunTests: true });
+    const box = screen.getByLabelText(askLabel);
+
+    expect(box).toHaveAttribute("maxlength", "4000");
+    expect(screen.getByText("0/4000")).toBeInTheDocument();
+
+    fireEvent.change(box, { target: { value: "abcde" } });
+    expect(screen.getByText("5/4000")).toBeInTheDocument();
+
+    // A paste past the cap is truncated, never sent whole into a 413.
+    fireEvent.change(box, { target: { value: "x".repeat(4200) } });
+    expect(box).toHaveValue("x".repeat(4000));
+    expect(screen.getByText("4000/4000")).toBeInTheDocument();
+  });
 });
 
 describe("AiPartnerPane — 'Show me a change' propose action (#947)", () => {
--- a/apps/web/src/components/editor/ai-partner/__tests__/ai-partner-pane.test.tsx
+++ b/apps/web/src/components/editor/ai-partner/__tests__/ai-partner-pane.test.tsx
@@ -9,6 +9,9 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { render, screen, fireEvent } from "@testing-library/react";
 import { NextIntlClientProvider } from "next-intl";
 import messages from "@/messages/en.json";
+import ptBR from "@/messages/pt-BR.json";
+import es from "@/messages/es.json";
+import type { AssistTier } from "@/lib/ai/partner-types";
 import { AiPartnerPane } from "../ai-partner-pane";
 
 const review = vi.fn();
@@ -16,7 +19,7 @@ const hookState = {
   messages: [] as unknown[],
   freeHintsUsed: 0,
   counts: { free: 0, metered: 0, socratic: 0 },
-  tier: "free" as const,
+  tier: "free" as AssistTier,
   budgetExhausted: false,
   spendCapped: false,
   resetState: "none" as const,
@@ -44,6 +47,7 @@ function renderPane(
     solutionPassed?: boolean;
     disabled?: boolean;
     hasRunTests?: boolean;
+    hasFailedRun?: boolean;
     onFocusRun?: () => void;
     onNudgeShown?: () => void;
     onNudgeOverride?: () => void;
@@ -72,7 +76,9 @@ beforeEach(() => {
   review.mockReset();
   hookState.requestHint.mockReset();
   hookState.ask.mockReset();
+  hookState.proposeFix.mockReset();
   hookState.messages = [];
+  hookState.tier = "free";
   hookState.budgetExhausted = false;
   hookState.spendCapped = false;
   hookState.loading = false;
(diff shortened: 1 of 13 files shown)


thomgabriel:
## Follow-up session: rebase onto #945 + full-suite verification

This branch was picked up from an interrupted session. Everything above was **inherited** from commit `60036560` — composer component, guard wiring, Enter/Shift+Enter + a11y, comment rewrites, i18n ×3, and the 8 new tests. Verified item-by-item against the brief; nothing was missing, so no code changes were needed in this session.

**Finished here:**

1. **Rebased onto `origin/main` @ `ad469fab`** (PR #945, the Gemini 3.x `thinkingBudget` hotfix). Clean rebase, zero conflicts — #945 touched `lib/ai/models.ts`, `lib/ai/reflection-reply.ts` and `api/ai/partner/route.ts`; this PR touches only the `components/editor/ai-partner/*` UI layer and `messages/*.json`. Branch head is now `d0995011`. Force-pushed with `--force-with-lease`.
2. **Full verification on the rebased tree** (not just the ai-partner folder):
   - `pnpm --filter @superteam-lms/web exec tsc --noEmit` → clean, no output
   - `pnpm --filter @superteam-lms/web lint` → **0 errors**; only the pre-existing `import/order` warnings in untouched files (`lib/solana/*`, `lib/supabase/*`, …), none in any file this PR changes
   - `pnpm vitest run` (full `apps/web` suite) → **264 files / 2541 tests passed, 0 failed** (123s)
3. **i18n key parity re-checked** across all three locales: `aiPartner.actions.askLabel`, `aiPartner.actions.askPlaceholder`, `aiPartner.actions.askSend` and `aiPartner.composer.empty` all present and translated in `en` / `pt-BR` / `es`.

Note: `aiPartner.messages.empty` is intentionally kept — it is still used by `message-list.tsx`, and is a different string from the new `aiPartner.composer.empty`.

No merge performed.

thomgabriel:
Independent gate review (this session): PASS-WITH-FINDINGS — one fix requested BEFORE merge.

**M1 (fix-first, ~2 lines):** the composer textarea has no maxLength/counter, but /api/ai/partner rejects message.length > 4000 with HTTP 413, and the hook collapses that into the generic "Something went wrong. Try again." — the exact generic-error-for-a-knowable-condition the owner flagged today. Add maxLength={4000} + a counter, and/or dedicated 413 copy. This became reachable only with this PR (no free-text input existed before).

**L4 (follow-up ok):** a typed question is silently lost when the learner picks "Run the tests" on the attempt-gate nudge (submit() clears the textarea before onSend; handleNudgeRunTests nulls the parked closure).

Verified clean: request shape unchanged, no client-side trust surface (server derives tier/budget), exhaustion renders the community-handoff not a wall, propose/diff-card never engaged, 181/181 tests. Merge order (empirically tested): #946 → #948 → #947, no rebases needed.

thomgabriel:
Gate addendum (owner-reported from the preview, verified on the branch): **the Socratic banner promises an action the UI cannot perform.** `socratic.entered` copy says "…it can still propose a change when you need to see one," but no UI path triggers `propose` — quick-actions exposes only onHint, the composer only `ask`, and the diff-card can render a ProposeResponse that nothing can request. Reproduced live by the owner: asking the tutor for a change in Socratic mode yields (correctly, per the prompt contract) a question, every time.

Two consistent resolutions — recommending (b):
(a) drop the clause from the banner copy in all 3 catalogs; or
(b) wire the §4.2 escape hatch for real: a 'Show me a change' action firing `propose` (available at every tier — §4.2 keeps propose's full diff contract even in Socratic; the earned-Accept comprehension gate already guards application). Option (b) makes the banner true and completes the ladder's designed shape: questions by default, diff on explicit request, never auto-applied.

Still also pending from the earlier gate pass: M1 (4000-char 413 → generic error; maxLength + counter).

thomgabriel:
Gate close-out: both open items resolved on-branch — M1 (82dfe9ba: 4,000-char cap + live counter) and the §4.2 propose escape hatch (c3cbffc3), which I verified commit-level against the task spec: hidden until a genuinely failed run, turn-cost label, available on Socratic tier, draft survives, Socratic-suffix regression pinned, trackProposeAccepted wired. The socratic.entered banner is now true. Merging — the composer ships complete: questions by default, worked example on explicit request, understanding required to apply.

Issue #942, opened 2026-07-31
Challenge screen v2 epic: LeetCode-grade structure on our teaching-first screen


thomgabriel:
From the 2026-07-31 side-by-side analysis (our challenge screen vs leetcode.com/problems/two-sum) + owner direction. Keep our wins (anonymous run, subgoal pedagogy, assist ladder); adopt their structure:

1. **Tabbed/sectioned left pane**: Lesson | Hints | Topics(skills) | Solutions(post-pass) | Submissions | Discussion — LeetCode-style collapsible Hint 1/2/3 sections (authored hints, user-openable, moved OUT of the AI pane), skill chips as Topics, Discussion inline with count instead of a toolbar modal.
2. **Run vs Submit split + verdict moment**: Run = visible harness (as today); Submit = grade + complete + the payoff (verdict state + XP; visual = the brand guide's terminal frame green-flash→amber-settle, already specced in §Backgrounds).
3. **Challenge header chips**: difficulty, skills, hint affordance, 'N builders completed this' (sanctioned by launch spec, unbuilt).
4. Custom-input testcase row for applicable challenge types (later).
5. Skeletons on this screen (with #931's wave).

Children: quiz interaction fixes and AI-composer restore are filed separately (immediately buildable). Owner-reviewed direction 2026-07-31.

thomgabriel:
Owner observation (live session): the **Examples** tab (#791 — the open-book spec of each subgoal: input fixture + expected value) and the **Test Cases** tab (post-run verdicts) show near-identical rows and read as duplication. v2 scope addition: **merge into a single Test Cases tab** — each row carries its spec inline (Input → Expected) in a neutral 'not run yet' state, and gains pass/fail + why-failed after a run. The Examples tab is retired; the open-book property is preserved (spec visible pre-run in the same rows). LeetCode analog: examples live in the description, results in the result pane — never as sibling tabs.

thomgabriel:
Owner (live session): remove the duplicated lesson title — the lesson toolbar shows the title AND the prose h1 repeats it immediately below. Resolution for v2: the prose h1 stays (document identity); the toolbar keeps Back / Prev / Next / progress and drops the title (LeetCode's pattern — title lives in the content pane only). Also re-noted: the version-stamp blockquote still renders as a pull-quote at the top of lessons — restyle as a metadata chip (already in #935 item 4; belongs in this epic's lesson-pane work).

PR #948, opened 2026-07-31
feat(lesson): LeetCode-patterned lesson pane — inline Discussion, Hint/Topics disclosures, card panes

thomgabriel:
Part of #942 — phase 1 (lesson-pane restructure). Does not close the epic.

## What changed

**1. Discussion inline, not a modal.** Challenge lessons opened Discussion in a
modal from a toolbar button (#770); reading lessons rendered it inline with a
permanently-expanded header. Both now render the same collapsible section at
the bottom of the lesson pane — header with thread count, expands to the thread
list + the ask-a-question composer. The toolbar button now opens the section and
smooth-scrolls the rail to it (kept: it is the only discoverability affordance
above the fold on a challenge).

**2. Hint 1/2/3 + Topics disclosures** under the prose. Authored hints become
individually-openable rows; Topics are the lesson's skill tags as chips. The
3-failed-runs auto-reveal inside `ChallengeInterface` is **unchanged** — this is
an additional, learner-initiated path to the same authored text.

**3. Toolbar title removed.** The prose `h1` immediately below owns document
identity; printing it twice was pure duplication (owner note on #942). Toolbar
keeps Back / Prev / Next / Discussion / XP / progress.

**4. Panes as rounded bordered cards.** Lesson pane and editor/output pane get
`--r-lg` radius + border + `bg-card` at `lg`, with a small gutter so the page
background shows around and between them. The split resizer became the
transparent gutter itself (was a bordered strip). Below `lg` the panes stack
full-bleed exactly as before.

**5. Version stamp as a metadata chip.** The leading "Version stamp" blockquote
renders as a mono/muted chip row instead of a prose pull-quote. Presentational
only — the authored markdown is untouched, and any other blockquote still
renders as a blockquote.

## New component

`components/lessons/lesson-section.tsx` — a disclosure row built on a native
button + `aria-expanded`/`aria-controls` rather than `<details>`, so the toolbar
can drive the Discussion row programmatically and the styling is identical
across browsers. Collapsed panels use `hidden`, keeping them out of the a11y
tree and the tab order.

## Verification

- `typecheck` clean, `lint` clean (only pre-existing `import/order` warnings).
- Full `apps/web` vitest: **2533 passed** (one PGlite hook-timeout flake in
  `email-subscriptions-execution.test.ts` under parallel load; passes on its own).
- Dev-server render check on port 3210 (challenge + reading lesson, `en`/`es`,
  light + dark, desktop + mobile 375px): all disclosures present and toggling,
  Discussion count renders, toolbar jump expands + scrolls, resizer drag still
  moves the split (628px → 760px), no console errors, no `MISSING_MESSAGE`.
- i18n keys `lesson.topics` and `lesson.versionStamp` added to en / pt-BR / es;
  hint labels reuse the existing `lesson.stuckNudgeHintLabel`.

## Judgment calls

- Discussion sits above the AI Partner pane in the challenge rail (the AI pane
  is pinned to the very bottom of that column and is an assistant surface, not
  lesson content).
- `getLessonSkills` returns `[]` for unknown lesson ids, so the teacher-preview
  bundle (no skills projection) simply omits the Topics section rather than
  erroring.
- Deliberately out of scope for phase 1: the Examples/Test-Cases tab merge,
  Run/Submit split, header chips, custom-input row, skeletons.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/code-block.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/code-block.tsx
@@ -30,12 +30,14 @@ export function CodeBlock({ block, ctx }: BlockRenderProps) {
   ) : undefined;
 
   return (
-    // Edge-to-edge IDE (#770): no card rounding/shadow/side border — the split
-    // sits flush against the viewport and the top bar (top border marks the
-    // seam). `bg-card` so any empty strip (resizer, editor tail, the Monaco a11y
-    // hint) matches the IDE instead of the page's dotted grid. At lg the wrapper
+    // #942 item 4: at lg the two panes are separate rounded cards (styled
+    // inside ChallengeInterface) floating on the page background, LeetCode
+    // style — so this wrapper contributes only the gutter padding and NO
+    // bg/border of its own. Below lg the panes stack full-bleed as before,
+    // where `bg-card` keeps any empty strip (editor tail, Monaco a11y hint)
+    // matching the IDE instead of the page's dotted grid. At lg the wrapper
     // flex-fills the viewport-height column in lesson-client (no page scroll).
-    <div className="overflow-hidden border-t border-border bg-card lg:flex lg:min-h-0 lg:flex-1 lg:flex-col">
+    <div className="overflow-hidden border-t border-border bg-card lg:flex lg:min-h-0 lg:flex-1 lg:flex-col lg:border-t-0 lg:bg-transparent lg:p-2">
       <div className="flex w-full flex-col overflow-hidden lg:min-h-0 lg:flex-1">
         <ChallengeInterface
           lessonId={ctx.lesson._id}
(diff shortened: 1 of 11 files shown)


thomgabriel:
Independent gate review (this session): PASS-WITH-FINDINGS — one owner decision pending + one in-PR fix requested.

**M2 (owner decision, asked directly):** this PR removes the authored-hint fail-gate — today hints unlock after shouldShowEncouragement (≥3 consecutive failed runs), one at a time, with trackStuckNudgeShown/postNudge solve attribution; the new pane renders every hint always-open from first paint with no gate and no analytics. Deliberate per the PR text, but it's a pedagogy-contract change (LX-C4) + silently degrades nudge attribution, so it needs an explicit owner call, not a ride-along. Holding merge until the owner answers.

**M3 (fix in-PR):** topic chips render raw slugs (learners see 'program-errors') — skills.json already maps slugs to labels; one lookup fixes it.

**L5 (follow-up ok):** keepMounted makes useThreads fire /api/community/threads on first paint of every lesson page just to label a collapsed row.

Verified clean: GamificationOverlays mounts app-wide (the #925 celebration path survives), reflection/challenge mounts untouched, all 66 lessons keep exactly one h1, i18n 3×-complete, 305/305 tests.

thomgabriel:
Owner decision (2026-07-31, via gate session): **M2 resolved — always-open hints accepted as designed.** The fail-gate/analytics removal is now an explicit product decision, not a ride-along. M3 (raw topic-chip slugs) and L5 (eager threads fetch) move to follow-up issues; merging.

PR #950, opened 2026-07-31
fix(env): close four env-wiring gaps found in the 31-07 audit

thomgabriel:
## Why

Four findings from today's env-wiring audit. All are the **#945 class**: shipped code depends on an env var or DB object production may not have, and nothing reports it — the app degrades or fails closed in silence. Each fix makes the gap loud (boot/build error, 503) rather than silent.

---

### F1 — `/api/health/schema` was blind to the assist-ladder migration

> Today `/api/health/schema` is blind to it: if `spend_assist_ladder_turn` is missing in prod, `spendAssistTurn` (lib/ai/assist-budget.ts) fail-closes and every AI tutor turn is denied while health returns 200.

Confirmed. `spendAssistTurn` returns the `denied` result on **any** RPC error (`apps/web/src/lib/ai/assist-budget.ts:72-86`) — a missing function is indistinguishable from an exhausted ladder, so every tutor turn is denied and `/api/health/schema` still answers 200.

Added to `SCHEMA_EXPECTATIONS` (`apps/web/src/lib/supabase/schema-expectations.ts`), following the existing rpc-probe / column-probe patterns:

| Probe | Object | Migration | Side-effect argument |
| --- | --- | --- | --- |
| rpc | `spend_assist_ladder_turn` | `20260730120000_assist_ladder.sql` | `p_user_id` NIL-UUID, **all three tier maxima = 0** |
| rpc | `get_challenge_assist_state` | same | read-only (`LANGUAGE sql STABLE`) |
| column | `challenge_assists.chat_log` | same | n/a |
| rpc | `record_review_result` | `20260726140000_review_items_spaced_repetition.sql` | NIL-UUID + empty item key |

**Why the spend RPC is probe-safe** (the question the task flagged): it is `REVOKE ALL … FROM PUBLIC, anon, authenticated` / `GRANT … TO service_role` (migration lines 126-127), and the health check runs on the **anon** client, so PostgREST answers `42501` without executing the body — the same mechanism the existing `claim_due_session_reminders` / `claim_due_reengagement` probes rely on. Belt **and** braces: with every tier max at `0`, a body that somehow ran falls straight into the `exhausted` branch, which increments nothing and deliberately does not even touch `updated_at`. So no fallback to `get_challenge_assist_state` was needed — that one is probed too, because the same migration `DROP`s the old signature, meaning a stale prod has *neither* version.

**Audit of every applied July migration** (task asked): `claim_due_reengagement` is indeed already covered (#919, lines 108-120) — verified. Everything else either has an entry (`streak_freezes`, `streak_freezes_used`, `check_ai_spend`, `reminder_opt_in`, `reminder_unsubscribe_token`, `claim_due_session_reminders`) or fails soft. One genuine gap found beyond the assist ladder: **`record_review_result`** — `/api/review/grade` rethrows the RPC error (`route.ts:120-125`), so a missing function 500s every review submission. Added. Deliberately **not** added: `xp_transactions.source` (if `20260727120000` never applied, neither the column nor the `award_xp` body that writes it exist, so nothing breaks), `record_billed_assist` / `append_challenge_assist_log` (best-effort, never block a response), `get_cohort_leaderboard` (degrades, not fails).

### F2 — `NEXT_PUBLIC_APP_URL` was unvalidated

> a missing value fails the build/boot loudly instead of silently minting a RELATIVE metadata URI on-chain

Now required in `publicEnvSchema` (`apps/web/src/lib/env.ts`). Affected `?? ""` sites — all **kept** as defense in depth, now unreachable:

- `app/api/certificates/mint/route.ts:261,281` — `external_url` + `appMetadataUri`
- `lib/solana/onchain-queue.ts:262,278`
- `lib/helius/event-handlers.ts:660,675`
- `app/api/admin/courses/sync/route.ts:283,406`, `app/api/admin/achievements/sync/route.ts:135`

**Dev ergonomics** — required-with-dev-default, matching the existing schema conventions (`NEXT_PUBLIC_SUPABASE_URL`'s prod-only https refine is the sibling pattern): `NODE_ENV !== "production"` defaults to the documented `http://localhost:3000`, so `pnpm dev` and every suite work untouched; production requires an explicit value, because no default can be right there and a silently-wrong origin is the exact failure being closed. `NEXT_PUBLIC_APP_URL` added to `vitest.setup.ts`, the CI unit-test env, `lighthouse.yml`, and the E2E harness (`e2e/harness/serve.mjs` runs `NODE_ENV=production`). `.env.example` + `apps/web/CLAUDE.md` reclassified Optional → **Required** with the on-chain-URI consequence.

> ⚠️ **Owner pre-merge check**: `NEXT_PUBLIC_APP_URL` must be set in the Vercel project (all environments) before this merges — it is listed as Required in `docs/DEPLOYMENT.md`, but if it is *not* actually set, the first deploy after merge fails the build. That is the intended behavior, stated here so it is not a surprise.

### F3 — `TEACH_PREVIEW_PASSWORD` defaulted to `"123"`

Literal fallback removed from `apps/web/src/lib/teach/preview-auth.ts`. `getPreviewPassword()` now returns `string | null`; new `isPreviewConfigured()`. Unset/blank ⇒ `isValidPreviewPassword` rejects **everything** (including `""` and the old `"123"`), and `POST /api/teach/preview/auth` returns **503** with a clear `console.error`. The cookie HMAC key falls back to a random per-process secret when nothing else is configured, so no cookie it verifies can validate. Added to the server env schema as optional (`optStr`). CLAUDE.md comment updated: `"unset = preview disabled"`.

### F4 — docs

- `docs/DEPLOYMENT.md` scheduled-jobs table: `/api/cron/reengagement` · `0 13 * * *` · 10:00 America/Sao_Paulo (matches `apps/web/vercel.json`).
- `apps/web/CLAUDE.md` top line: Next.js 14 → **15** (`apps/web/package.json` pins `next@^15.5.22`).

---

## Evidence

**Build fails loudly without the var** (`next build`, NODE_ENV=production, F2's core claim):

```
Collecting page data ...
Error: Invalid public environment variables:
  - NEXT_PUBLIC_APP_URL: must be an absolute URL, e.g. https://academy.example.com

Check your environment configuration (see .env.example).
> Build error occurred
[Error: Failed to collect page data for /api/admin/capstone-funnel]
```

**Same build, with `NEXT_PUBLIC_APP_URL=https://academy.example.com`** → succeeds (full page manifest emitted, middleware 187 kB).

**Tests** — `pnpm exec vitest run` (apps/web): **266 files, 2543 tests passed**, including the new/extended:

- `src/lib/supabase/__tests__/schema-expectations.test.ts` — missing `spend_assist_ladder_turn` (PGRST202) reported with its migration; missing `challenge_assists.chat_log` (42703) reported; invariant test asserting the spend probe carries zero tier maxima so a future `GRANT` can never make the health check mutate.
- `src/lib/teach/__tests__/preview-auth.test.ts` — unset ⇒ gate closed (`"123"`, `""`, `undefined` all rejected); blank treated as unset; only the configured password accepted.
- `src/app/api/teach/preview/auth/__tests__/route.test.ts` (new) — 503 + logged `TEACH_PREVIEW_PASSWORD` + no cookie when unset; 401 wrong password; 200 + cookie when correct.
- `src/lib/__tests__/env-app-url.test.ts` (new) — dev/test default; explicit value honoured; **throws in production when unset**; rejects a relative path.
- `src/app/api/health/schema/__tests__/route.test.ts` — pre-existing 200/503 mapping still green.

**Typecheck**: `pnpm typecheck` → 6/6 tasks successful.
**Lint**: `pnpm --filter web lint` → 0 errors (only the pre-existing `import/order` warnings on files this PR does not touch).

Not run: Playwright E2E (heavy); the harness env change is a one-line addition mirroring the existing `NEXT_PUBLIC_*` block and CI will exercise it.

## Not done

Not merged, per instructions.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -72,7 +72,11 @@ BUILD_SERVER_URL=https://academy-build-server-HASH.a.run.app
 # API key for X-API-Key header (same value as ACADEMY_API_KEY on Cloud Run)
 BUILD_SERVER_API_KEY=                      # PRIVATE — server-only
 
-# ── App URL (for sitemap, OG tags, NFT metadata URI fallback) ────────────────
+# ── App URL (sitemap, OG tags, email links, on-chain NFT metadata URI) ───────
+# REQUIRED. Validated in apps/web/src/lib/env.ts: dev/test fall back to the
+# localhost origin below, but a PRODUCTION build with this unset FAILS at boot —
+# an empty value would pin a *relative* metadata URI into an immutable
+# credential NFT.
 NEXT_PUBLIC_APP_URL=http://localhost:3000
 
 # ── Arweave / Irys — permanent credential metadata (server-only) ─────────────
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,7 @@ jobs:
           NEXT_PUBLIC_SOLANA_RPC_URL: https://api.devnet.solana.com
           NEXT_PUBLIC_SUPABASE_URL: https://example.supabase.co
           NEXT_PUBLIC_SUPABASE_ANON_KEY: test-anon-key
+          NEXT_PUBLIC_APP_URL: http://localhost:3000
         run: pnpm -r test
 
   # ── E2E: Playwright, the 3 critical learner paths (#781) ────────────────────
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -29,6 +29,9 @@ env:
   NEXT_PUBLIC_SOLANA_RPC_URL: "https://api.devnet.solana.com"
   NEXT_PUBLIC_SOLANA_NETWORK: "devnet"
   NEXT_PUBLIC_PROGRAM_ID: "7NeJaSRyb4Wxay3Tcd9bdpD7T3GWYUQSFyrhG8SgwE8V"
+  # Required by lib/env.ts; `next build` runs as NODE_ENV=production, where the
+  # schema refuses to default it.
+  NEXT_PUBLIC_APP_URL: "http://localhost:3000"
 
 jobs:
   lighthouse:
--- a/apps/web/CLAUDE.md
+++ b/apps/web/CLAUDE.md
@@ -1,4 +1,4 @@
-# apps/web — Next.js 14 App Router
+# apps/web — Next.js 15 App Router
 
 API route reference lives in `src/app/api/CLAUDE.md` (loads when you work on routes).
 
@@ -120,8 +120,8 @@ OPENENDED_AI_REPLY=                # Best-effort AI reply on /api/lessons/reflec
 #                                  # Needs no GITHUB_TOKEN: academy-courses is public, so the
 #                                  # preview reads it anonymously (#830). A token is used when
 #                                  # set, purely to lift the 60 req/hr anonymous rate limit.
-TEACH_PREVIEW_PASSWORD=            # Shared password for /teach/preview. Defaults to "123" —
-                                   # an INTERIM value; set a real one before sharing the URL.
+TEACH_PREVIEW_PASSWORD=            # Shared password for /teach/preview. NO default —
+                                   # unset = preview disabled (/api/teach/preview/auth 503s).
                                    # Gates read-only rendering of unpublished course PRs only:
                                    # never on-chain writes, never the admin surface (separate
                                    # cookie from admin_session, so it cannot satisfy admin auth).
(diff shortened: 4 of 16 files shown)


thomgabriel:
Pre-merge check on the reminder: production is verified to have the app URL set — the live sitemap emits https://st.academy URLs (build-time inlined, so the Production env carries it) and the PR's own Vercel preview build passed (Preview env carries it). Merge is safe; the guard cannot brick the next deploy.

PR #951, opened 2026-07-31
fix(integrations): pin external API shapes verified in the 31-07 audit

thomgabriel:
## Why

The AI-tutor prod-down bug (#945: `thinkingBudget` rejected by the 3.x models since the #890 swap) established a class of defect rather than a one-off: **an external API's request/response shape verified once, then silently drifting**. Today's external-API-shape audit swept every third-party integration in the app for the same pattern. This is the fix batch. Every finding was re-verified against current vendor docs before implementing.

The common failure mode across these: the call keeps returning 2xx, so nothing alerts — the damage is a wrong invariant, a wrong bill, or a webhook that never fires.

### Resend batch sender — `lib/email/resend.ts`

1. **Pin `x-batch-validation: strict`.** The claim-release invariant documented on `SendFailureDelivery` (a 4xx proves nothing was accepted, so releasing the per-day send claims is safe) **depends on strict all-or-nothing validation** — which was only an *undocumented default*. Under permissive validation Resend accepts the valid messages and reports the invalid ones, so a 4xx would no longer prove nothing was sent, and the reminder/re-engagement pipelines would release claims for mail that had already gone out. Pinned explicitly with the invariant stated in a comment.
2. **Derive `sent` from the response body** (`{ data: [{ id }, ...] }`) instead of assuming `messages.length`. A count mismatch on a 2xx is impossible under strict mode — so it now logs loudly as a **tripwire** on exactly the assumption above, rather than being invisible.
3. **Special-case 409 idempotency conflicts.** Previously both types fell into the generic 4xx `rejected` branch:
   - `concurrent_idempotent_requests` — another request with this key is in flight and **may still send**. Now classified `delivery: "unknown"`, the ambiguous outcome callers already handle for 5xx (hold the claims). Releasing here was a duplicate-send bug.
   - `invalid_idempotent_request` — same key, different payload; refused outright, stays `rejected`, but with a distinct log naming the **permanent same-day stall** (our keys are derived from day + chunk membership, so every retry today rebuilds the same key) that self-heals only at the 24h TTL.

### Spend ledger — `lib/ai/spend-ledger.ts`, `lib/ai/models.ts`, `api/ai/partner`

4. **Cached prompt tokens were billed at the full input rate.** Gemini's `promptTokenCount` **includes** `cachedContentTokenCount` — the cached tokens are a subset, not an addition. The ledger billed the whole prompt at the input rate, over-reporting every cache hit by up to 10× and tripping the #591 daily caps early (a degrade/deny the sponsor never actually owed). Now `(prompt − cached)` bills at input and `cached` at cached-input. Malformed `cached > prompt` clamps and bills everything at the input rate — the conservative direction on a platform-funded key — with a warn.

   Cached-input rates fetched from https://ai.google.dev/gemini-api/docs/pricing on **2026-07-31** (cited in the comment): `gemini-3.6-flash` $0.15, `gemini-3.5-flash` $0.15, `gemini-3.5-flash-lite` $0.03 per 1M. Deliberately **not** modelled: the $1.00/1M/hour cache *storage* component — the app relies on implicit caching, which carries no storage charge. No cached-input env override, so an emergency input override can't accidentally reprice cache hits.

   The partner route already read `cachedContentTokenCount` for its debug log; it now reaches `recordAiSpend` via the shared usage shape.

### GitHub checks — `lib/github/github.ts`

5. **`?filter=latest&per_page=100`.** The check-runs endpoint paginates at 30 — runs past the first page were simply invisible to the verdict fold. And when `total_count > check_runs.length`, the verdict is now **`"unknown"`** rather than a fold over a partial sample: a failing invisible run must never read green at the content-sync gate. Logged.

### Helius webhook config — `lib/helius/webhook-config.ts`

6. `webhookType` is a documented **enum** and devnet is a distinct *value*: `"rawDevnet"` (mainnet keeps `"raw"`). Registering with the wrong value succeeds and the webhook then never fires — the exact silent drift this audit is about. Also removed the undocumented `txnStatus` field, which was being ignored upstream while *reading* as if failed transactions were filtered there (they are filtered in `event-decoder.ts`, which drops any non-null `meta.err` — verified). Added `res.ok` + array guards to `listWebhooks`, which was casting an error object to `HeliusWebhookResponse[]` and surfacing auth failures as "no webhooks registered". The host is left as-is with the doc ambiguity commented (Helius documents a single `api.helius.xyz` host with the cluster selected by the key; the per-cluster hosts are what the out-of-band registration used).

   **This is dormant registration code — the live webhook was registered out-of-band and is not managed from here. Behaviour-neutral for prod today.**

### Small ones

7. Deleted the stale comment in `lib/ai/partner-prompt.ts` claiming Gemini's structured-output dialect rejects `minimum`/`maximum` with a 400 — current docs support both. The runtime clamp in `validatePartnerResponse` stays (schema constraints are the model's best effort, not a validation boundary).
8. `HELIUS_API_KEY`'s description in `lib/env.server.ts` said "webhook management + DAS API". **No DAS call exists anywhere in the codebase** (grepped `getAssetsByOwner`/`searchAssets`/`getAsset`/`dasApi` — zero hits); it describes an integration that was never built. Now webhook management only.
9. Bounded the teach-preview tarball fetch with an abort/timeout signal, mirroring the identical fetch in `lib/content/prior-content.ts:96` — aborting the fetch aborts the response stream, so a codeload connection that stalls mid-body can't hang the preview until the platform timeout.

## Tests

Extended `resend.test.ts` (+9), `spend-ledger.test.ts` (+7), `github.test.ts` (+4):

- strict header present on the wire; `sent` derived from `data.length`; the accepted-count mismatch tripwire; no-`data` and non-JSON 2xx bodies don't throw
- 409 `concurrent` → `delivery: "unknown"` (**hold**, not release); 409 `invalid` → `rejected` with a distinct log naming the 24h TTL; unrecognised/non-JSON 409 stays conservatively rejected
- cached-discount math, the "never ADDS cached to prompt" case, no-change-when-absent, the `cached > prompt` clamp, negative cached; end-to-end `recordAiSpend` at 3.6-flash rates; every routed model carries a cached rate below its input rate
- check-runs query params; `total_count` mismatch → unknown; normal fold when it matches; still folds when `total_count` is absent

**Red-proof** (cheap, done): reverting `estimateSpendMicroUsd` to the old full-input-rate math fails exactly the 3 new cached-billing cases (3 failed / 31 passed), green with the fix.

## Verification

- `pnpm typecheck` — clean
- Affected suites (`lib/github`, `lib/email`, `lib/ai`, `lib/helius`, `lib/teach`, `app/api`) — **73 files / 741 tests passed**, post-rebase onto #949
- Full suite: 2549/2552. The 3 failures are PGlite DB-harness *hook timeouts* under parallel load in `reminder-consent-execution` / sibling — they pass in isolation (19/19) and touch none of these files.
- `pnpm lint` — only pre-existing `import/order` warnings in files this PR doesn't touch.

Rebased onto `origin/main` after #949 landed (`models.ts` overlap: their `thinkingLevel: "minimal"` correction, my `cachedInputUsdPerMTok` column — clean).

## Risk

Low. The only production behaviour changes are (a) Resend now sends one extra header and reports a body-derived count, (b) the ledger bills cache hits *less*, (c) checks with hidden runs read `unknown` instead of a partial verdict — all fail-safe directions. The Helius change is dormant code.

### DIFF
--- a/apps/web/src/app/api/ai/partner/route.ts
+++ b/apps/web/src/app/api/ai/partner/route.ts
@@ -72,7 +72,11 @@ interface GeminiEnvelope {
     finishReason?: string;
     content?: { parts?: Array<{ text?: string }> };
   }>;
-  usageMetadata?: GeminiUsageMetadata & { cachedContentTokenCount?: number };
+  // `cachedContentTokenCount` is part of GeminiUsageMetadata itself now — it is
+  // BILLING input (cached prompt tokens carry the cached-input discount), not
+  // just the debug counter it was when this route widened the type locally.
+  // Keeping it on the shared shape is what makes it reach recordAiSpend below.
+  usageMetadata?: GeminiUsageMetadata;
 }
 
 // Model selection is PER ACTION (#868) and lives in lib/ai/models — this route
@@ -581,7 +585,8 @@ export async function POST(request: NextRequest) {
     }
     // Record the ACTUAL cost of this billed generation into the spend ledger
     // (#591) from usageMetadata — prompt + candidate + thinking tokens, thinking
-    // billed at the output rate. Runs on EVERY billed path (empty / non-JSON /
+    // billed at the output rate and the cached slice of the prompt billed at the
+    // cached-input rate. Runs on EVERY billed path (empty / non-JSON /
     // truncated all still cost tokens). When usageMetadata is ABSENT (e.g. a
     // non-JSON 200) the ledger books a CONSERVATIVE fallback — full output budget
     // + a generous input estimate — instead of $0, so an unmeasurable-but-billed
--- a/apps/web/src/lib/ai/__tests__/spend-ledger.test.ts
+++ b/apps/web/src/lib/ai/__tests__/spend-ledger.test.ts
@@ -25,7 +25,11 @@ import { MODEL_RATES } from "../models";
 // sheet, so they pin the cheap tier explicitly.
 const LITE = "gemini-3.5-flash-lite" as const;
 const FLASH_36 = "gemini-3.6-flash" as const;
-const RATES = { inputUsdPerMTok: 0.3, outputUsdPerMTok: 2.5 };
+const RATES = {
+  inputUsdPerMTok: 0.3,
+  outputUsdPerMTok: 2.5,
+  cachedInputUsdPerMTok: 0.03,
+};
 
 beforeEach(() => {
   rpc.mockReset();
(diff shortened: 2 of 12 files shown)


thomgabriel:
Closing the builder's open verification item: independently re-fetched Resend's batch-validation changelog — strict IS the documented default ("sends the batch only if all emails in the request are valid… otherwise strict mode automatically applies"), permissive is opt-in via the batch-validation header. The claim-release invariant is sound, and the explicit strict pin makes it independent of the default going forward. The AI_SPEND override inversion nit stays as-is per the builder's scope call (same pre-existing pattern as output rates).

PR #954, opened 2026-07-31
feat(lesson,community): challenge toolbar, lesson numbering + jump chips, discussion package

thomgabriel:
Owner-requested work from the 31-07 live UI session: challenge-toolbar affordances, lesson numbering/navigation, and the discussion package.

## ⚠️ Base — read before merging

This branch is **stacked on the preview**, not on `main`. It branches from `dbfcff7b` (`preview/brand-wave`), because cherry-picking that commit onto `main` conflicts across five files.

- `#946` and `#948` were part of that preview state and have since landed on `main` as their **own squashes** (`037d9505`, `969bc655`), so `dbfcff7b` is still **not** an ancestor of `main`.
- The diff below therefore includes preview-only context. **The base is to be resolved at merge time** — do not read the file count as this PR's own footprint.

## Summary

| # | Feature | State |
|---|---|---|
| 1 | Format button | Shipped — rewritten after owner report (see below) |
| 2 | Stopwatch | Shipped |
| 3 | Lesson number in titles | Shipped |
| 4 | Jump chips | Shipped |
| 5 | Post-button spinner | Shipped |
| 6 | Inline lesson composer | Shipped |
| 7 | Dense thread rows | Shipped, re-polished after preview feedback |
| 8 | Lesson-context chips + Course filter | Shipped |
| 9 | Thread detail redesign | **Withdrawn**, replaced by surgical fixes + LeetCode comment anatomy |

---

### 1. Format button — reported broken, root cause was not what anyone expected

Shipped first against Monaco's built-in `editor.action.formatDocument`. The owner reported: *"the format button doesn't seem to be working."*

Reproduced by driving the exact API Monaco's TS worker calls (`getFormattingEditsForDocument`) against the owner's real input — a markdown-paste artifact pasted as one line. The result contradicted the obvious theory:

- The formatter produced **six edits, not zero**. It added spaces inside `{ }` and left the code on a single line.
- **Root cause: `editor.action.formatDocument` on a TS model is a whitespace-only formatter — it never inserts line breaks.** It cannot un-minify a one-liner, valid or not. A pasted one-liner is precisely the input where its work is invisible, which is why it read as dead.
- On genuinely unparseable input (unbalanced braces) it produced zero edits — a real silent no-op, but the *secondary* case.
- Ruled out the wiring theories: the model's language id is `typescript`, and CSP already permits Monaco's `blob:`/jsdelivr workers, so the TS worker was live.

**Fix:** Format now runs **Prettier**, which re-prints from the AST — the owner's one-liner becomes five readable lines. Prettier and its plugins are imported dynamically, so none of it lands in the initial bundle; the cost is paid on first click. `prettier` moves from a root-only devDependency to an `apps/web` dependency.

`format()` returns a status rather than a boolean, and writes back through `executeEdits` (not `setValue`) so the reformat is a single undoable step instead of resetting the undo stack.

**The silence is fixed too.** Success stays silent — the reformatted buffer is its own feedback — but every other outcome gets a short localized note under the toolbar (`role="status"`, auto-clearing):

- `syntax-error` → "Nothing to format — fix the syntax errors first."
- `unchanged` → "Already formatted."
- `unsupported` → language backstop (Rust; the button is hidden there anyway).

Marker-based button-disabling was considered and dropped: with a real formatter the invalid case now produces an explicit, actionable message, which beats a disabled control the learner cannot act on.

### 2. Stopwatch

A LeetCode-style personal timer: collapsed it is a clock chip showing the running total, opened it reveals Start/Pause/Reset over a mono readout. Per-lesson, persisted in `localStorage`, and **honest about elapsed time** — state is `{elapsedMs, startedAt}`, so a mid-run refresh keeps counting from the right point, and a hidden tab banks the elapsed segment and stops accumulating (resuming on return if it was running). Two minutes spent in another tab never lands on the learner's clock. No XP wiring, no grading.

### 3. Lesson number in titles

The prose `h1` is prefixed with the lesson's ordinal ("2. Accounts Are Just Bytes"), taken from its index in the course's flattened module→lesson order — the **same** `getCourseLessons` order prev/next already walks, not a new numbering source. Prefixing happens at render, never in content, so reordering a course cannot strand a stale number; only the `h1` that *is* the lesson title is prefixed, so an unrelated heading deeper in the prose keeps its own text.

This also fixed a real inconsistency: **both curriculum lists restarted numbering at 1 in every module**, so a lesson the lesson page called "5." appeared as "1." in the curriculum. Both now share one `courseWideOrdinals` helper.

### 4. Jump chips

LeetCode's chip row under the `h1` — Topics, Hints (only when hints are authored), Discussion (with its thread count). Each opens its disclosure section *and* scrolls to it; the scroll waits a frame so it targets the post-expansion position. `LessonSection` already supported controlled open, so only Topics and the first Hint were wired up — the remaining hints stay uncontrolled, preserving the one-at-a-time reveal.

### 5. Post-button spinner

The Ask-a-Question modal painted `.sol-spinner` (dark track + XP-yellow head) on the greenish primary button, where it read as a foreign artefact and all but vanished at disabled opacity. Now the app-standard `border-current border-t-transparent` ring, matching every other primary submit. Same fix applied to the answer editor.

### 6. Inline composer

The create-thread form is extracted into `ThreadComposer`, and the lesson's Discussion section composes it **inline** — `CreateThreadModal` is gone from the lesson page entirely. A modal over a lesson hid the code the learner was asking about, which is the one thing they need while writing. Posting stays put: the new thread appears in the list below via a refresh token instead of navigating away. The community page keeps its modal, now a thin shell over the same component — one submit path, one validation rule.

### 7. Dense thread rows

`ThreadCard` is a single-line row, replacing the tall card whose stacked ▲/score/▼ rail turned a ten-thread feed into a page of air. After a first preview pass ("right direction, still not 100%"):

- Rows are a flat list inside one container — hairline dividers, row-tint hover, no per-row radius or card look.
- The vote is one compact horizontal cluster (caret beside count), the title takes the slack and truncates, and the meta is a tight mono cluster pinned right — no ocean of space between them.
- ~10px vertical padding with 11px `leading-none` meta puts a row under 40px on desktop.
- Sort tabs, type/scope filters, and the primary action share **one** row via a new `actionSlot`, so the lesson embed's chrome no longer dwarfs a one-thread list.

*Judgment call:* this is the only row style, with no `variant` prop.

### 8. Lesson-context chips + Course filter

Threads started from a lesson show a "Course › Lesson" chip linking back to that exact lesson, so an answer can be read next to the thing it is about. Titles resolve from the committed content bundle through the public `/api/content/*` routes, batched and de-duplicated per feed page; a missing lesson degrades to the course, and a failed lookup drops the chips but **never** the threads. Enabled on the community feed, category pages, course detail (there the chip names the *lesson*, which is the disambiguating part), and the thread detail page.

Paired with a "Course questions" pill (`GET /api/community/threads?hasCourse=1` → `.not("course_id","is",null)`) that toggles independently of the type pills — a course question is still a question, so scope is a separate axis.

### 9. Thread detail — redesign withdrawn, then rebuilt as a comment thread

A full redesign was implemented and **rejected on owner review** ("before you touch it was better"). It was reverted to the pre-package state (verified byte-identical to `894f0698`), along with the four i18n keys only that redesign used.

Process note recorded: page-level redesigns get an owner-approved mock **before** implementation. Components and sections can go straight to preview; whole-page reshapes cannot.

Owner live-testing then surfaced real bugs on the *restored* page, fixed **surgically** — the airy structure and spacing untouched:

- **A posted answer now appears.** Two causes. Reads send `cache: "no-store"`, because the post-refetch could be served the browser's pre-post copy of the GET — literally "posted, nothing appeared". And the refetch is now silent: it skips `setIsLoading`, so the page no longer unmounts into a full-screen spinner and back. The refetch call itself already existed; caching plus the blank-flash is what made it look broken.
- **The orphan ✓ is gone.** The accept control hung below the vote rail as a bare checkmark whose only label was an `aria-label`. It is now a labelled icon+text button (`aria-pressed`), under the same rule as before — thread author only, never on their own answer.
- **Body headings are scoped.** `prose-sm` alone does *not* step headings down — it scales body copy but still renders an authored `#`/`##` larger than the page `h1`. Now explicit on both question and answers.

Finally, per the owner's "just make it look like the LeetCode one", the answers section was rebuilt as a **comment thread** (on top of the preview's own anatomy pass):

- Answer items are comments, not cards: no border, no radius, no fill, flat on the page background. Separation is the list's `divide-y` hairline alone.
- Anatomy: 28px avatar in its own column; name + level badge on row 1; timeAgo (12px, muted) on row 2 aligned with the **name**, not the avatar; body full-width under the name column; action row of vote cluster, Flag, Delete, Accept.
- The accepted answer gets no box — a 2px `--success` rule down the left edge plus a small label above the name. `border-l-<color>` is used deliberately: a `border-<color>` shorthand would silently override the list's `divide-y` colour on that item.
- The composer moved to the **top** of the section, under an "Answers (N)" header with a "Sort by Latest" line between it and the list — replying was a scroll hunt past N answers. Editor and submit now share one frame (button bottom-right inside) via a new `flush` prop on `MarkdownEditor`; nesting the already-bordered editor inside another bordered block read as two controls.
- The empty state keeps the composer and drops the sort line and list.

Read-more clamping for very long answer bodies is deliberately out of scope for v1.

## Also included

Deletes the stale `PathsView — start-here card` suite. The card was removed from `PathsView` in the 31-07 live session but its tests outlived it, reddening CI for anything stacked on the preview.

## Verification

- `tsc --noEmit` — clean.
- `next lint` — clean; the only remaining warnings in touched files are pre-existing import-order notes that arrived with the base.
- **Full suite** (`vitest run --maxWorkers=2`): **269 files / 2605 tests green.**
- Dev-server render check on a spare port (3412, never 3000): `/en/community`, `/en/courses`, the lesson page, and the thread-detail route all 200, zero compile errors. Format, Stopwatch, Reset Code and the "Course questions" filter all present in the served markup.
- New coverage (~54 cases across four files):
  - `format-action.test.tsx` — the reflow assertion that encodes the whole bug (one-liner → multiple lines), paste artifact and unbalanced braces → `syntax-error`, already-formatted → `unchanged`, JSON formatting and invalid JSON, Rust → `unsupported`, write-back through `executeEdits`, and pre-mount safety.
  - `stopwatch.test.tsx` — `mm:ss`/`h:mm:ss` formatting, malformed and negative persisted state, pause banking, mid-run refresh, hidden-tab time excluded then resumed, per-lesson isolation.
  - `lesson-numbering-and-chips.test.tsx` — `courseWideOrdinals` across modules, accordion numbering continuing into module 2, `h1` prefixing (title match, casing/whitespace drift, unrelated `h1` untouched, null ordinal), chip anchors and activation.
  - `discussion-package.test.tsx` + `thread-detail-fixes.test.tsx` — dense-row voting, chip hrefs and fallbacks, filter independence, composer inline-vs-modal paths, the spinner-class assertion, context resolution and graceful degradation; post → answer visible, `no-store` on every read, thread stays mounted during refetch, composer precedes the list, flat items with no border/radius/fill, accepted left-rule, empty state, and the accept control's label and visibility rules.

i18n: all new strings added to `en` / `pt-BR` / `es`; keys left unused by the item-9 withdrawal were removed rather than left dangling.


thomgabriel:
Superseded by #955 (the complete UI wave carries all 8 features plus the owner's live-session iterations on top). Will close when #955 merges.

thomgabriel:
Superseded by #955, which carries all 8 features of this PR rebased cleanly onto main (gate-verified: the only two files that differ here are stale, not missing — quick-actions.tsx pre-dates #947's propose action, and header.tsx pins the old h-6 lockup that the owner reverted to h-8 — so merging this PR would actually regress main). Closing; branch stays until #955 lands.

PR #964, opened 2026-08-01
feat(community): thread page — flat header, inline answer composer (wave)

thomgabriel:
Owner-directed redesign: the standalone community thread page still carried the pre-wave design (disconnected left vote column, loose header, bulky boxed Write/Preview answer composer, card-ish spacing). This applies the same concept the challenge/lesson panes got in the #954/#955 wave.

## Header — flat anatomy, no vote rail

The floating left vote column is gone. The question now renders in the **same WHO → WHAT → ACTIONS anatomy as the answers below it**:

- back link
- title + status chip in one tight block
- provenance chip (course > lesson) directly under it
- compact meta line — author avatar + name + level chip + time + views
- body (markdown, headings still stepped down below the h1)
- **inline action row** — horizontal vote cluster, flag, delete-own — the exact idiom the answers use

One content column throughout (`max-w-3xl`, was `max-w-4xl` split by the rail), spacing rhythm matched to the lesson pane.

## Answer composer — inline, not a boxed editor

Replaces the `MarkdownEditor` tab card with the lesson pane's comment-composer, sized to the AI-partner composer's conventions:

| before | after |
|---|---|
| Write/Preview **tab bar** | quiet **Preview text-toggle** (`aria-pressed`) |
| tall empty textarea (110px min) | single field, `min-h-[44px]`, `resize-y`, grows |
| char-count strip on its own row | counter inline, small, beside the button |
| oversized default Post button | compact `size="sm"` Post with a send glyph |

Unchanged: the 10k limit, markdown rendering through the existing ReactMarkdown + remark-gfm + rehype-highlight pipeline (**no rehypeRaw**), error surfacing (draft is preserved on failure), disabled/in-flight states.

## Double-post guard (pre-existing gap, L4 pattern)

The button's `disabled` alone cannot stop a submit dispatched in the same tick — a form submit bypasses the button entirely. An early-return re-entrancy guard (`inFlight` ref) now backs the disabled state. Mutation-checked: removing the guard fails the new test.

## a11y

sr-only label on the field; `aria-pressed` on the Preview toggle **and on both vote buttons** (both layouts); `aria-busy` on the in-flight Post; focus-visible rings on the field and the toggle.

## Scope

No API changes, no behavior changes. Post / vote / accept (thread author only, never own answer) / flag / delete-own / sort line / provenance chip all preserved. One new i18n key, `community.answerLabel`, added to all three catalogs (en / es / pt-BR) — everything else reuses existing keys.

## Test evidence

- `apps/web` community + i18n suites: **47 passed** (`src/components/community/__tests__`, `src/messages/__tests__` — parity + no-duplicate-keys)
- Full web suite: **271 files / 2652 tests passed**
- `pnpm typecheck` (tsc --noEmit, strict): clean
- `pnpm lint`: no new warnings on the changed files
- Prettier: all changed files clean

New tests cover the flat header (vote cluster is horizontal, follows the body, not a rail; one `max-w-3xl` column; vote `aria-pressed`) and the composer (single growable field with no tab bar, preview toggle rendering markdown, inline counter + 10k danger threshold, disabled-until-text and in-flight disable, the double-post guard, failure keeps the draft). Existing answer-anatomy tests were re-scoped to select answer rows inside the `divide-y` list, since the question itself is now also an `<article>`.

**Not verified in a browser**: the running preview server belongs to the shared checkout (another agent's session) and this worktree has no `.env.local`, so a dev server can't boot here. Light/dark was reviewed at the token level — every colour is an existing themed CSS var (`--border-default`, `--input`, `--text`, `--text-2`, `--primary`, `--danger`, `--success`) plus `dark:prose-invert`, and the preview surface carries the same `--input` fill as the textarea so toggling doesn't jump. Worth an owner eyeball in light mode before merge.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/community/[category-slug]/[thread-slug]/thread-detail-client.tsx
+++ b/apps/web/src/app/[locale]/(platform)/community/[category-slug]/[thread-slug]/thread-detail-client.tsx
@@ -193,7 +193,7 @@ export function ThreadDetailClient({ shortId }: ThreadDetailClientProps) {
 
   if (error || !thread) {
     return (
-      <div className="mx-auto max-w-4xl px-4 py-20 text-center">
+      <div className="mx-auto max-w-3xl px-4 py-20 text-center">
         <p className="text-[var(--text-2)]">{error || t("noResults")}</p>
         <Link
           href="/community"
@@ -206,7 +206,7 @@ export function ThreadDetailClient({ shortId }: ThreadDetailClientProps) {
   }
 
   return (
-    <div className="mx-auto max-w-4xl px-4 py-8">
+    <div className="mx-auto max-w-3xl px-4 py-8">
       {/* Breadcrumb */}
       <Link
         href={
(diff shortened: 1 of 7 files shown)


thomgabriel:
Visual verification done on the dev server (the builder couldn't): thread page eyeballed in BOTH themes at desktop width — flat anatomy renders correctly, no floating rail, accepted-answer left rule intact, provenance chip themed correctly (an apparent dark-chip leak turned out to be a mid-theme-transition paint). Signed-out view verified; composer covered by tests + token-level review. Merging.

PR #968, opened 2026-08-01
fix(ui): stop the Solana gradient border vanishing at card corners

### LINKED ISSUE #967: Solana gradient border is erased at the corners on every inline gradient card
The Solana gradient border is erased at all four corners on every card that builds it with the inline "padding trick". The gradient renders correctly along the straight edges, then vanishes at each rounded corner, so the border reads as four disconnected coloured segments instead of one continuous fade.

## Cause — a hardcoded inner radius that no longer matches the token

The pattern is an outer element painting the gradient with a few px of padding, and an opaque inner element covering the middle. For the border to stay uniform the two radii must be concentric:

```
inner radius = outer radius − padding
```

`rounded-xl` in this project maps to `--r-xl`, which is **22px** (`apps/web/tailwind.config.ts:136`, `globals.css:138`). Every inline call site hardcodes an inner radius of 9–11px — sized for a ~12px outer radius, not 22px.

Worked example (`hero-showcase.tsx`, R=22, padding=2.5, r=10):

- outer corner arc: centre (22, 22), radius 22
- inner corner arc: centre (12.5, 12.5), radius 10
- centre-to-centre distance: √2 × 9.5 = 13.44
- inner arc's outermost point, measured from the outer centre: 13.44 + 10 = **23.44 > 22**

The inner card's corner therefore extends ~1.4px *past* the wrapper's rounded corner. Where the wrapper has `overflow-hidden` the inner is clipped flush to the outer edge; where it does not, the square-ish inner corner juts outside the rounded outline. Either way the gradient band at the corner is reduced to zero.

## Affected call sites

| File | Outer | Padding | Inner (now) | Inner (correct) |
| --- | --- | --- | --- | --- |
| `components/landing/hero-showcase.tsx:157` | `rounded-xl` (22) | 2.5 | `rounded-[10px]` | 19.5 |
| `components/landing/loop-widgets.tsx:410` | `rounded-xl` (22) | 2.5 | `rounded-[9px]` | 19.5 |
| `components/landing/loop-widgets.tsx:452` | `rounded-xl` (22) | 2.5 | `rounded-[9px]` | 19.5 |
| `components/dashboard/continue-card.tsx:80` | `rounded-xl` (22) | 2.5 | `rounded-[10px]` | 19.5 |
| `components/certificates/earn-handoff-card.tsx:54` | `rounded-xl` (22) | 1.5 | `rounded-[11px]` | 20.5 |
| `components/profile/wallet-name-generator.tsx:113` | `rounded-xl` (22) | 2 | `rounded-[10px]` | 20 |

The CSS-class versions of the same effect are **already correct** and are not affected, because they derive the inner radius instead of hardcoding it:

- `.cert-inner` → `calc(var(--r-xl) - 2px)`
- `.grad-card-inner` → `calc(var(--r-lg) - 1.5px)`
- `.path-start-card` → `border: 2.5px solid transparent` + `padding-box`/`border-box`, where the browser derives the inner radius itself

## Fix

Derive the inner radius at every inline call site the same way the CSS classes already do — `rounded-[calc(var(--r-xl)-<padding>)]` — so the geometry cannot drift from `--r-xl` again. This is a literal hardcoded-constant-vs-token desync: a past change to `--r-xl` silently broke every site that had copied the old number.

Potolski:
Closes #967.

The Solana gradient border was being erased at all four corners of every card that builds it with the inline padding trick, leaving four disconnected straight segments instead of one continuous fade.

## Cause

The padding trick holds only while the two radii are concentric:

```
inner radius = outer radius − padding
```

`rounded-xl` resolves to `--r-xl` = **22px**. Every inline call site hardcoded an inner radius of 9–11px — sized for a ~12px outer radius, not 22px.

At R=22, padding=2.5, r=10 the two corner arcs are no longer concentric:

- outer arc: centre (22, 22), radius 22
- inner arc: centre (12.5, 12.5), radius 10
- centres are √2 × 9.5 = 13.44 apart
- the inner arc's outermost point, from the outer centre: 13.44 + 10 = **23.44 > 22**

So the opaque inner corner overhangs the wrapper's arc by ~1.4px. Where the wrapper has `overflow-hidden` the inner clips flush to the outer edge; where it doesn't, the squarer inner corner juts outside the rounded outline. Both routes drive the gradient band at the corner to zero.

## Fix

Derive the inner radius from the token — `rounded-[calc(var(--r-xl)-<padding>)]` — at each call site. This is exactly what the CSS-class versions of the same effect already do (`.cert-inner` → `calc(var(--r-xl) - 2px)`, `.grad-card-inner` → `calc(var(--r-lg) - 1.5px)`), which is why those were never affected.

| File | Padding | Was | Now |
| --- | --- | --- | --- |
| `landing/hero-showcase.tsx` | 2.5 | `rounded-[10px]` | `calc(var(--r-xl)-2.5px)` |
| `landing/loop-widgets.tsx` (front) | 2.5 | `rounded-[9px]` | `calc(var(--r-xl)-2.5px)` |
| `landing/loop-widgets.tsx` (back) | 2.5 | `rounded-[9px]` | `calc(var(--r-xl)-2.5px)` |
| `certificates/earn-handoff-card.tsx` | 1.5 | `rounded-[11px]` | `calc(var(--r-xl)-1.5px)` |
| `profile/wallet-name-generator.tsx` | 2 | `rounded-[10px]` | `calc(var(--r-xl)-2px)` |

The invariant is now documented next to the certificate style contract in `lib/styles/styleClasses.ts`, since the root cause was a hardcoded constant silently desyncing from the token it was copied from.

**Geometry only** — no colour, token, or radius values changed.

## Verification

Rendered both geometries at the real token values and compared, including a 6× magnification of the top-left corner: before, the corner arc is chopped flat and the four edges are visibly disconnected; after, the frame is continuous with uniform thickness the whole way round.

- `tsc --noEmit` — no new errors (the 45 reported are pre-existing on `main`, all missing-dev-dep failures in test files; verified identical with the change stashed)
- `eslint` on all five touched files — clean
- `vitest run` — 2589 passed, 0 failed

### DIFF
--- a/apps/web/src/components/certificates/earn-handoff-card.tsx
+++ b/apps/web/src/components/certificates/earn-handoff-card.tsx
@@ -51,7 +51,7 @@ export function EarnHandoffCard({ source, courseId }: EarnHandoffCardProps) {
       aria-labelledby={titleId}
       className="rounded-xl bg-cert-gradient p-[1.5px] shadow-[var(--shadow-card)]"
     >
-      <div className="rounded-[11px] bg-card px-5 py-4">
+      <div className="rounded-[calc(var(--r-xl)-1.5px)] bg-card px-5 py-4">
         <div className="flex items-center gap-2">
           <Coins
             size={20}
--- a/apps/web/src/components/landing/hero-showcase.tsx
+++ b/apps/web/src/components/landing/hero-showcase.tsx
@@ -153,7 +153,7 @@ function IdCard({
       }`}
       style={interactive ? { transformStyle: "preserve-3d" } : undefined}
     >
-      <div className="group relative overflow-hidden rounded-[10px] bg-card p-6">
+      <div className="group relative overflow-hidden rounded-[calc(var(--r-xl)-2.5px)] bg-card p-6">
         {/* sheen sweep on hover */}
         <div
           className="pointer-events-none absolute inset-0 -translate-x-full bg-gradient-to-r from-transparent via-white/[0.06] to-transparent transition-transform duration-700 ease-out group-hover:translate-x-full"
--- a/apps/web/src/components/landing/loop-widgets.tsx
+++ b/apps/web/src/components/landing/loop-widgets.tsx
@@ -405,7 +405,7 @@ export function ProveWidget({ flipLabel }: { flipLabel: string }) {
           className="absolute inset-0 overflow-hidden rounded-xl bg-cert-gradient p-[2.5px] shadow-card"
           style={{ backfaceVisibility: "hidden" }}
         >
-          <div className="relative flex h-full flex-col justify-between overflow-hidden rounded-[9px] bg-card p-5 sm:p-6">
+          <div className="relative flex h-full flex-col justify-between overflow-hidden rounded-[calc(var(--r-xl)-2.5px)] bg-card p-5 sm:p-6">
             {/* sheen sweep on hover */}
             <div
               className="pointer-events-none absolute inset-0 -translate-x-full bg-gradient-to-r from-transparent via-white/[0.07] to-transparent transition-transform duration-700 ease-out group-hover:translate-x-full"
@@ -446,7 +446,7 @@ export function ProveWidget({ flipLabel }: { flipLabel: string }) {
             transform: "rotateY(180deg)",
           }}
         >
-          <div className="flex h-full flex-col justify-between rounded-[9px] bg-card p-5 sm:p-6">
+          <div className="flex h-full flex-col justify-between rounded-[calc(var(--r-xl)-2.5px)] bg-card p-5 sm:p-6">
             <div>
               <div className="mb-3 font-mono text-[10px] font-bold uppercase tracking-[0.2em] text-primary">
                 NFT Details
--- a/apps/web/src/components/profile/wallet-name-generator.tsx
+++ b/apps/web/src/components/profile/wallet-name-generator.tsx
@@ -110,7 +110,7 @@ export function WalletNameGenerator({
 
       {/* Name display with gradient border — fixed size to prevent layout shift */}
       <div className="relative w-full max-w-md rounded-xl bg-gradient-to-r from-primary to-secondary p-[2px]">
-        <div className="flex h-[80px] items-center justify-center overflow-hidden rounded-[10px] bg-bg px-6 py-4">
+        <div className="flex h-[80px] items-center justify-center overflow-hidden rounded-[calc(var(--r-xl)-2px)] bg-bg px-6 py-4">
           <span
             className={`truncate font-mono text-xl font-bold tracking-tight transition-opacity sm:text-2xl ${
               settled ? "opacity-100" : "opacity-80"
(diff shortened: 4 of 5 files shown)


Potolski:
@claude Thanks — the CSS-spec half of this is correct, but the premise about what Tailwind emits isn't, so there's nothing to change here. Evidence below, compiled through this repo's own config.

## Tailwind 3.4 normalizes math-operator spacing in arbitrary values

The review states:

> Tailwind's arbitrary-value syntax passes bracket contents through literally except for `_` → space — it does not insert spaces around operators.

It does. `normalize()` in `tailwindcss/src/util/dataTypes.js` rewrites math operators inside `calc()` / `min()` / `max()` / `clamp()` specifically so this authoring form works. This repo is on **tailwindcss 3.4.19**.

Compiling both forms together:

```
$ pnpm exec tailwindcss -c probe.config.js -i in.css -o out.css
```

```css
.rounded-\[calc\(var\(--r-xl\)-2\.5px\)\] {
  border-radius: calc(var(--r-xl) - 2.5px)
}

.rounded-\[calc\(var\(--r-xl\)_-_2\.5px\)\] {
  border-radius: calc(var(--r-xl) - 2.5px)
}
```

Byte-identical declarations — the underscored and bare forms are the same rule.

## Through this project's real pipeline

```
$ pnpm exec tailwindcss -c tailwind.config.ts -i src/styles/globals.css -o app.css
```

```css
.rounded-\[calc\(var\(--r-xl\)-1\.5px\)\] { border-radius: calc(var(--r-xl) - 1.5px); }
.rounded-\[calc\(var\(--r-xl\)-2\.5px\)\] { border-radius: calc(var(--r-xl) - 2.5px); }
.rounded-\[calc\(var\(--r-xl\)-2px\)\]   { border-radius: calc(var(--r-xl) - 2px);   }
```

All three variants emitted with the required whitespace. Valid CSS, not dropped.

## Computed value in a browser

Rendering the real generated stylesheet with the real class strings from `hero-showcase.tsx`:

```js
getComputedStyle(document.documentElement).getPropertyValue('--r-xl')  // "22px"
getComputedStyle(inner).borderRadius                                   // "19.5px"
```

**19.5px** — exactly `22 − 2.5`, which is the intended concentric radius. Not the `0px` initial-value fallback the review predicts, so the "corner artifact gets worse, not fixed" consequence doesn't follow.

## `lesson-client.tsx:686` is not a pre-existing bug

The review flags `lg:h-[calc(100dvh-60px)]` as the same failure mode and suggests a follow-up issue. It was in the same probe:

```css
.h-\[calc\(100dvh-60px\)\] { height: calc(100dvh - 60px) }
```

Compiles correctly — no issue to file. It's also corroborating evidence: that class has been shipping on `main`, which it could not do if the bare form were being dropped.

## The process point was right, and I've acted on it

> the PR's stated verification … sounds like it computed the arc geometry directly rather than rendering the actual Tailwind-compiled class in a browser — so the invalid-calc regression wouldn't have surfaced.

Correct, and a real hole in my evidence: the original check rendered hand-written CSS that already had the spaces in it, so by construction it could not have caught a compile-stage problem. The verification above closes that — it goes config → generated stylesheet → browser → computed value.

Both forms are already present in this codebase (`:686` bare, `:699` underscored) and compile identically, so I'm leaving the current form. Happy to switch all five to `_-_` purely to pre-empt this reading if a reviewer prefers it.

Issue #922, opened 2026-07-31
CI: pglite test contention — parallel suites fail that pass alone (aggravated as suites grow)


thomgabriel:
From adversarial review of #919 (F8, pre-existing). Running src/lib/email + src/lib/supabase + src/app/api/cron + src/app/api/email vitest targets together fails 3 pglite files on main (~3 tests) and more as new pglite suites land; all pass in isolation. Each suite builds multiple PGlite instances; parallel workers contend. Fix ideas: vitest poolOptions to serialize pglite-tagged files, a shared PGlite fixture pool, or maxConcurrency for those directories. Not a correctness bug — a flaky-CI risk that worsens with every DB-lane PR.

PR #978, opened 2026-08-04
fix(content): make _draft/ a real exclusion, shared by compiler and lint

### LINKED ISSUE #973: Content pipeline: _draft/ is a convention with no implementation — lint drops it, the compiler ships it (gates academy-courses #34)
Found by the 2026-08-04 paths-feature audit; **merge-gating for academy-courses PR #34** (flagged there: https://github.com/solanabr/academy-courses/pull/34#issuecomment-5180737355).

## The divergence

The two content walkers classify files by different rules, and `_draft/` falls in the gap:

| | rule | `paths/_draft/x.yaml` |
|---|---|---|
| content-lint (`packages/content-lint/src/loader.ts:55-67`) | anchored fixed-depth regexes | **dropped** (never linted) |
| bundle compiler (`apps/web/src/lib/content/compile/compile-bundle.ts:127-146`; same in `validate.ts:88-105`) | unanchored `startsWith`/`endsWith` | **kept** (compiles into the bundle) |

Only excluded prefix anywhere: `courses/_template/` (`tarball.ts:5`). `grep -rn "_draft"` across the monorepo: zero hits. Applied to PR #34's head tree: **6 courses / 66 lessons / 3 achievements / 1 path would be compiled-but-never-linted** — the parking is a runtime no-op AND a CI blackout, while content-lint goes green precisely because it can no longer see the parked files. Nothing catches it: `validate-content.yml` lints only the content repo tree; the monorepo CI recompiles from content.lock and diffs for drift but never re-lints the pinned tree.

## Fix (closes the class, not just this instance)

1. Shared `EXCLUDED_PREFIXES` (add `_draft/` handling for all top-level collections) in `tarball.ts`, mirrored in `validate.ts`'s walker.
2. Test: a `_draft/` file never reaches the bundle (course, lesson, path, achievement fixtures).
3. **The assertion that would have caught this before it was written:** every doc the compiler classifies must also be lint-classifiable — fail loudly on any file one walker sees and the other doesn't. (Also closes the sibling hole: a typo'd `courses/_drafts/` or `paths/archive/` currently behaves identically to `_draft/`, silently — `loader.ts:83` is a bare `if (!kind) continue;` with no unclassified-file diagnostic.)
4. Content-lint side: either teach `loader.ts` to *see-and-skip* `_draft/` explicitly (named concept, like `TEMPLATE_LESSON_PREFIX` in `gate19-skills.ts:54`), with a fixture test — the current invisibility is an emergent property of a regex, load-bearing but untested.

Sequencing: this must land BEFORE academy-courses #34 merges (or #34 drops the directory move and uses `draft: true`/deployment state instead — owner call, see the PR comment).

thomgabriel:
Closes #973. Gates academy-courses#34.

`_draft/` parking was a directory convention with no implementation: lint's anchored regexes skipped it while the compiler shipped it. This adds one shared predicate (`content-schema/src/repo-paths.ts`, a `_draft/` path segment at any depth) consumed by all three tree walkers, makes lint's skip a named tested behavior, and adds a warning when a content-shaped yaml classifies to nothing (catches typo'd dirs like `_drafts/`, which deliberately do NOT match the exclusion).

Red-proof: `draft-exclusion.test.ts` fails at main (parked docs reach the bundle), passes here. Run against academy-courses#34's real head: excludes exactly the 6 courses / 66 lessons / 1 path / 3 achievements it parks, keeps the 2 live courses. Zero new lint warnings on either content main or the PR head.

### DIFF
--- a/apps/web/src/lib/content/compile/__tests__/draft-exclusion.test.ts
+++ b/apps/web/src/lib/content/compile/__tests__/draft-exclusion.test.ts
@@ -1,0 +1,172 @@
+import { describe, it, expect } from "vitest";
+import { stringify } from "yaml";
+import { isExcludedContentPath } from "@superteam-lms/content-schema";
+import type { RepoTree } from "@/lib/github/types";
+import { compileContent } from "../compile-bundle";
+
+/**
+ * #973 — `_draft/` is the parking convention academy-courses uses to take a
+ * course out of the catalog without deleting it. The compiler classified by
+ * unanchored `endsWith("/course.yaml")` / `startsWith("paths/")`, so a parked
+ * doc compiled straight into the shipped bundle while content-lint's anchored
+ * fixed-depth regexes never saw it — parked content shipped, unlinted.
+ *
+ * These assert the property directly on the compiled bundle: a `_draft/`
+ * segment anywhere in the path means the doc is not in the output.
+ */
+
+const enc = new TextEncoder();
+const yaml = (v: unknown): Uint8Array => enc.encode(stringify(v));
+const raw = (s: string): Uint8Array => enc.encode(s);
+
+const SHA = "c".repeat(40);
+
+function course(id: string, slug: string, lessonId: string): unknown {
+  return {
+    id,
+    slug,
+    title: slug,
+    description: "d",
+    difficulty: "beginner",
+    duration: 1,
+    xpPerLesson: 10,
+    xpReward: 100,
+    modules: [{ key: "m", title: "M", lessons: [lessonId] }],
+  };
+}
+
+function lesson(id: string, slug: string): unknown {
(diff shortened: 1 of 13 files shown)


thomgabriel:
@claude review

thomgabriel:
Routing record (sensitive content-pipeline lane): adversarial review PASS with 1 medium + 3 low — medium (typo-tier severity split) and the untested-guard finding fixed at 27114ed5 with mutation red-proofs, replayed independently by the orchestrator (guard removed → exactly the 2 new fixtures fail; restored → 132/132). Lows filed as #979. Review body final at head, no blocking findings. CI green; local pglite failures verified pre-existing at main (#922 class). Merging.

Issue #988, opened 2026-08-04
Phantom Connect: browser verification matrix (Safari / iOS / Brave) — gates the flag


Potolski:
Part of #983. **Gates enabling the feature flag.**

Verify the embedded wallet works on the browsers our learners actually use before turning it on for anyone.

## Why this is a real risk

Embedded wallet SDKs that lean on third-party iframe storage break under Safari's Intelligent Tracking Prevention and Brave's shields.

Two signals suggest Phantom Connect avoids that: its dependency tree includes `@phantom/indexed-db-stamper` (IndexedDB, not third-party cookies) and `@phantom/auth2`, and Portal setup requires configured **redirect URLs** — i.e. a redirect-based OAuth flow rather than a cross-site iframe. That is an inference from packaging, not a tested result.

Note also that Safari ITP evicts script-writable storage after ~7 days without interaction, and Phantom's session is 7 days regardless — so the worst realistic case is a re-login, not a lost wallet. Confirm that.

## Matrix

- [ ] Chrome desktop
- [ ] Safari desktop
- [ ] **iOS Safari** — highest risk and a large share of our Brazilian learners
- [ ] Android Chrome
- [ ] Brave with shields up
- [ ] Firefox

## Per browser, record

- [ ] Connect completes and returns an address
- [ ] Session survives a page reload
- [ ] Session survives a browser restart
- [ ] Behaviour after the 7-day session expiry

## Acceptance

Flag stays off until Chrome, iOS Safari and Brave all pass. If iOS Safari fails, that is a launch blocker — document the failure mode before deciding whether to ship for a subset of browsers.

Issue #986, opened 2026-08-04
Phantom Connect: identity bridge — embedded wallet → SIWS → link-wallet


Potolski:
Part of #983. Depends on the connect flow issue.

Bind the embedded wallet to the learner's Supabase profile. **Most of this already exists** — `/api/auth/link-wallet` verifies SIWS, rejects cross-account links (409), rejects a conflicting second wallet (409), is idempotent for the same wallet, and mints accrued XP via `mintXpToWallet` on link.

The work is driving that route from an embedded wallet and handling the conflict cases in the UI.

## Scope

- [ ] After a successful Phantom connect, request a SIWS signature from the embedded wallet and POST to `/api/auth/link-wallet`
- [ ] Gate provisioning on `profiles.wallet_address IS NULL` — never attempt to overwrite an existing link
- [ ] Handle `walletAlreadyLinked` (409) with real copy, in all three locales. Two distinct causes, and they need different messages:
  - this wallet belongs to another account
  - this account already has a different wallet
- [ ] Tests for both 409 paths and the idempotent re-link

## The conflict that matters

Links are **permanent** by design. A learner who already linked an external wallet cannot be given an embedded one, and their XP and credentials live at the existing address. That is correct behaviour — but it must read as "you already have a wallet connected", not as a failure.

Confirm the size of that group before building UI for it:

```sql
select count(*) filter (where wallet_address is not null) as has_wallet,
       count(*) filter (where wallet_address is null)     as no_wallet
from profiles where deleted_at is null;
```

## Note on `enroll`

`enroll` is the only instruction requiring a learner signature, and the learner is also its payer — so an embedded-wallet learner needs SOL to enroll. Moving the payer to the backend would make the entire learner journey zero-SOL and zero-prompt. Worth a follow-up issue; not required here.

Issue #983, opened 2026-08-04
Epic: Phantom Connect embedded wallets — email-first onboarding, no wallet install


Potolski:
Learners who arrive without a Solana wallet currently hit a hard wall: install an extension, save a seed phrase, fund it. For a platform whose whole point is teaching people Solana, that barrier sits *before* the first lesson.

**Phantom Connect** removes it. A learner signs in with Google or Apple and receives a real Phantom embedded wallet — no extension, no seed phrase, no SOL. Later, when they download Phantom and sign in with the same Google account, the same wallet is there.

## Why this integration is small for us

Our program only requires a learner signature on **one** instruction:

| Instruction | Signer |
| --- | --- |
| `enroll` | learner (signer + payer) |
| `complete_lesson` | `backend_signer` only |
| `finalize_course` | `backend_signer` only |
| `issue_credential` | `payer` + `backend_signer` + asset keypair — learner is a **referenced account, not a signer** |

Every XP award and credential mint already works from the learner's **pubkey alone**. We need an address bound to an email identity, plus a signature on `enroll`.

`/api/auth/link-wallet` already verifies SIWS, rejects cross-account and conflicting links, and mints accrued XP on link. An embedded wallet plugs into that route unchanged.

## Key facts (verified against Phantom's docs)

- **Free.** "There's no cost to use Phantom Connect. Phantom provides authentication, embedded wallets, and signing infrastructure at no charge to developers."
- **Non-custodial.** Keys never pass through our app or backend and never appear in plaintext. We never become a custodian.
- **Wallet continuity is real.** Embedded wallets are tied to the user's Google/Apple identity in Phantom's own account namespace — the docs describe an account picker over "any Phantom accounts tied to their Google or Apple identity." Downloading Phantom and choosing *Continue with Google* surfaces the same wallet. No export/import step.
- **Extension users share one code path.** If a user connects the Phantom extension, no embedded wallet is created and the app receives the account through the identical SDK interface.
- **Constraints:** $1,000/day per-app spending limit (irrelevant — we move no value), 7-day session, and users can revoke via Settings → Connected Apps.

## What Phantom does NOT do

It does **not** accept our Supabase JWT to provision a wallet. The "Wallet Authentication with JWTs" guide is the opposite direction — the wallet signs, our backend verifies and issues a JWT. That is SIWS, which we already have.

[...]

PR #992, opened 2026-08-04
feat(auth): Phantom Connect flow — provider, hook, and modal entry point

### LINKED ISSUE #985: Phantom Connect: connect flow — provider + hook over browser-sdk
Part of #983. Depends on the foundation issue.

Wrap `@phantom/browser-sdk` in a React 18-compatible provider + hook, and surface Phantom's Connect modal in the existing auth flow.

## Scope

- [ ] `PhantomConnectProvider` mounted in the `[locale]` layout provider tree, rendering children untouched when the flag is off
- [ ] `usePhantomConnect()` returning `{ connect, disconnect, address, status }`
- [ ] Wire the Connect modal into `components/auth/auth-modal.tsx` alongside the existing Google/GitHub buttons
- [ ] All strings via next-intl across **en / pt-BR / es** — the three locale files must keep identical key structures or `MISSING_MESSAGE` throws at runtime
- [ ] Handle the documented lifecycle: 7-day session expiry, and app revocation via Settings → Connected Apps, both surfacing as a clean disconnected state rather than a stuck spinner
- [ ] Unit tests with the SDK mocked — provider is inert when flagged off, hook surfaces connect/disconnect, revoked session degrades gracefully

## Extension users

Phantom Connect returns extension connections through the same SDK interface, and creates **no** embedded wallet in that case. Confirm our existing SIWS path and this path don't both try to own the connection — one wallet identity per session.

## Ship dark

Land with the flag **off**. Enabling is gated on the browser verification matrix issue.

## Acceptance

- Flag off: zero behaviour change, no SDK network calls, existing SIWS login unaffected
- Flag on (local): Connect modal opens, Google sign-in returns an address

Potolski:
Closes #985. Part of epic #983. **Stacked on #990 and #991** — review those first.

Adds the React wrapper over `@phantom/browser-sdk` and surfaces **Continue with Phantom** in the auth modal. A learner with no wallet signs in with Google and gets one.

## Placement

Phantom goes **first**, styled as the primary action, with the subtitle "Sign in with Google — we'll set up your wallet". It is the only option in that modal that asks nothing of someone who has never held a wallet; the existing Solana / Google / GitHub options are untouched below it.

## Session handling follows the SDK contract, not the obvious shape

Three things here are deliberately not what they'd naively be:

**The address does not come from `connect()`.** It is a redirect flow — this tab navigates away and that promise never resolves. The address arrives via the `connect` event after the app remounts. Reading it from the return value would strand the UI permanently.

**`autoConnect()` runs only after listeners are attached.** The SDK documents this ordering ("should be called after setting up event listeners") because it resolves an existing session and emits `connect` immediately enough to be missed otherwise.

**A rejected `autoConnect()` is not an error.** It's the normal first-time visitor path, so it leaves the UI idle with sign-in options intact rather than surfacing a failure.

`disconnect` covers both the 7-day session expiry and a learner hitting "Revoke permissions" in Phantom's Connected Apps. Both land in a clean idle state rather than a stuck spinner.

## Browser-verified, not just unit-tested

Ran against the dev server:

- Modal opens from **both** entry points (header Sign In and hero Sign Up) with the Phantom option present
- **en / pt-BR / es** all render the new strings — no `MISSING_MESSAGE`
- **Gate verified end to end**: with `NEXT_PUBLIC_PHANTOM_APP_ID` removed and the server restarted, the Phantom button is *absent*, the original three options render unchanged, and **zero requests reach Phantom**. That last part is the one that actually matters — a gate that hides UI but still phones home isn't a gate.

## Also verified

- **5 new provider tests** (24 across `auth` + `phantom`): gated-off touches no SDK, `autoConnect` resumes a session, the Solana address is picked from the `connect` event, `disconnect` returns to idle, and a rejected `autoConnect` stays idle
- `vitest run` — **2692 passed**, 0 failed
- `eslint` — clean
- `tsc --noEmit` — no errors in touched files
- Locale key parity asserted programmatically across all three files

## Not in this PR

Linking the connected wallet to the Supabase profile (#986) — that drives the existing `/api/auth/link-wallet`, and is the next piece. Until it lands, connecting produces a wallet the app doesn't yet persist.

**Do not enable in production before #988** (Safari/iOS/Brave matrix). Everything above was verified in one desktop browser.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -45,6 +45,14 @@ AI_PARTNER_SEAL_SECRET=
 # when unset (apps/web/src/app/api/rust/execute/route.ts).
 RUST_PLAYGROUND_URL=
 
+# ── Phantom Connect embedded wallets (optional — #983/#984) ──────────────────
+# App id from Phantom Portal (https://phantom.com/portal/). Public and
+# domain-bound: it ships in the client bundle by design, and Portal's
+# allowed-domains list is what scopes it — configuration, not a secret.
+# Unset = embedded wallets OFF; SIWS stays the guaranteed way in.
+# Portal must list the production origin AND http://localhost:3000 for dev.
+NEXT_PUBLIC_PHANTOM_APP_ID=
+
 # ── Analytics (optional — platform works without these) ───────────────────────
 # Google Analytics 4
 NEXT_PUBLIC_GA4_MEASUREMENT_ID=G-XXXXXXXXXX
--- a/apps/web/CLAUDE.md
+++ b/apps/web/CLAUDE.md
@@ -126,6 +126,18 @@ TEACH_PREVIEW_PASSWORD=            # Shared password for /teach/preview. NO defa
                                    # never on-chain writes, never the admin surface (separate
                                    # cookie from admin_session, so it cannot satisfy admin auth).
 
+# Optional — Phantom Connect embedded wallets (#983/#984)
+# Public, domain-bound client id from Phantom Portal (https://phantom.com/portal/).
+# It ships in the client bundle by design; Portal's allowed-domains list is what
+# scopes it, so it is configuration, not a secret. Unset = embedded wallets OFF,
+# exactly like the analytics keys — SIWS stays the guaranteed way in, and no
+# build or page render may depend on this being present.
+# Portal must list the production origin AND http://localhost:3000 for dev.
+# Read ONLY through lib/phantom/config.ts (isPhantomConnectEnabled/getPhantomAppId).
+# NOTE: we integrate via @phantom/browser-sdk, not @phantom/react-sdk — the
+# latter needs React >=19.0.1 and this app is on React 18.3.1 (#989).
+NEXT_PUBLIC_PHANTOM_APP_ID=
+
 # Optional — Rust playground proxy (server-only)
 RUST_PLAYGROUND_URL=               # /api/rust/execute upstream (default: play.rust-lang.org/execute)
 
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -55,6 +55,10 @@ const cspDirectives = [
     "https://accounts.google.com https://*.googleapis.com",
     "https://www.googletagmanager.com https://www.google-analytics.com https://*.google-analytics.com https://stats.g.doubleclick.net",
     "https://*.posthog.com https://*.sentry.io https://*.ingest.sentry.io",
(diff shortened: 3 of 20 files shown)


Potolski:
Both blocking findings were correct — fixed in b3f7900b.

## 1. Stuck modal — confirmed, and the diagnosis was exactly right

The provider swallowed every rejection, the modal's catch was dead code, and the Dialog's `if (!loading)` close guard turned any pre-redirect failure into a permanently unclosable modal. The stuck-spinner mode the PR claimed to avoid, relocated — fair hit.

**Fix:** `connect()` rethrows after recording its own state. Hook consumers keep `status`; imperative callers get the rejection they need to reset their UI, so the modal's existing catch went from dead code to the recovery path.

**Coverage the review called out as missing, now present:**
- provider: a rejected `connect()` both sets `status:"error"` **and** reaches the caller's catch
- modal (real component, rejecting connect): error message renders, **every** sign-in button re-enables, and **Escape closes the dialog**
- the inverse is pinned too: while a connect is genuinely in flight the spinner and guard correctly persist — that guard exists to stop accidental dismissal mid-auth, and only *failure* may unstick it

## 2. CSP — confirmed; would have broken every enabled-state attempt

Correct that no Phantom origin was allow-listed, and correct that the verification list only ever exercised the *disabled* gate.

Origins were read from the SDK dist rather than guessed — `grep` over the installed packages yields exactly `auth.phantom.app`, `api.phantom.app`, `time.phantom.app` (XHR targets, now in `connect-src`, pinned rather than `*.phantom.app`) and `connect.phantom.app` (the auth redirect — a top-level navigation CSP does not govern). The SDK creates no iframes (checked), so `frame-src` is untouched. The `/api/*` static fallback in `next.config.mjs` is updated too, per its "rough sync" contract.

**Verified live this time, as requested** — real `NEXT_PUBLIC_PHANTOM_APP_ID`, dev server, `securitypolicyviolation` listener installed, devtools network watched:

- **zero CSP violations**
- clicking *Continue with Phantom* ran the full pre-redirect chain and landed on **Google's OAuth screen** with `redirect_uri=https://connect.phantom.app/login/callback` in the trace

So the flow is proven from button-click through Phantom's auth service to Google. The only leg still unexercised is post-Google-sign-in (needs a real account; tracked under #988's end-to-end pass).

## Non-blocking note

Documented at the call site: `"google"` is the single entry point by design; `apple`/`injected` in the client config are provisioned for future buttons, not dead config.

Full suite: 2696 passed, eslint + tsc clean.

Potolski:

Issue #989, opened 2026-08-04
Phantom Connect: React 19 upgrade evaluation (unblocks @phantom/react-sdk)


Potolski:
Part of #983.

`@phantom/react-sdk@2.0.2` declares `peerDependencies: { react: ">=19.0.1" }`. This app is on **React 18.3.1** with Next 15.5.22, so we integrate via `@phantom/browser-sdk` (no React peer dep) and maintain a thin provider + hook ourselves.

That is the right call now — a React 19 upgrade across ~165 components is not something to bundle into a wallet integration. This issue records the trade-off and asks whether to revisit later.

## What the browser-sdk path costs us

- A provider and hook we own and maintain, versus Phantom's supported React bindings
- Phantom's prebuilt React components are unavailable, including whatever ships with `@phantom/wallet-sdk-ui`
- Our wrapper needs updating when the SDK's surface changes

## What to evaluate

- [ ] Does the maintained wrapper stay small, or does it accumulate SDK-tracking work?
- [ ] Is Next 15.5.22 + React 19 a supported combination for the rest of our stack?
- [ ] What else in `apps/web` blocks React 19 (wallet-adapter, Radix, testing-library)?
- [ ] Would the upgrade deliver anything beyond the Phantom bindings?

## Not urgent

Only worth revisiting if the wrapper becomes a maintenance burden or we want Phantom's UI components. `@phantom/browser-sdk` is the same underlying implementation — `@phantom/react-sdk` depends on it — so we are not on a lesser code path, just without the React bindings.

Issue #987, opened 2026-08-04
Phantom Connect: graduation — "Take your wallet with you"


Potolski:
Part of #983.

Give learners an explicit path from embedded wallet to the full Phantom app, and prevent the predictable support ticket.

## The failure mode

A learner downloads Phantom and taps **"Create new wallet"** instead of **"Continue with Google"**. They get a fresh address with a seed phrase, while their credentials remain in the embedded wallet they can't see — and they conclude Academy lost their NFTs.

"Create new wallet" is the bigger, more obvious button for someone who doesn't know they already have one. This will happen unless we get in front of it.

## Scope

- [ ] A "Take your wallet with you" surface (settings and/or post-first-credential)
- [ ] Explicit instruction: download Phantom, then choose **Continue with Google** with the same account used here
- [ ] Show the first/last four characters of their address so they can confirm they landed in the right wallet
- [ ] All copy in en / pt-BR / es
- [ ] Link to Phantom's own help article on creating a wallet with a Google or Apple account

## Why this is worth building

Nothing moves on-chain during graduation — the address is unchanged, so `profiles.wallet_address` stays valid and there is no migration. The entire risk is a user taking the wrong branch in Phantom's onboarding, which is a copy problem, not a technical one.

PR #994, opened 2026-08-04
feat(auth): Phantom identity bridge — embedded wallet to Supabase account

### LINKED ISSUE #986: Phantom Connect: identity bridge — embedded wallet → SIWS → link-wallet

Potolski:
Closes #986. Part of epic #983. **Stacked on #990 → #991 → #992** — review those first.

Turns a connected embedded wallet into a Supabase account. No new endpoint: this drives the two routes the wallet-adapter path already uses.

## The routing decision is the substance

| Supabase session? | Route | Result |
| --- | --- | --- |
| none | `/api/auth/wallet` | signs the learner **in**, creating the account from the wallet |
| exists, no wallet linked | `/api/auth/link-wallet` | attaches the wallet, mints accrued XP to it |

That first row is the whole point. I originally scoped this issue as "SIWS → `link-wallet`", which would have been wrong: linking requires an existing session, so a brand-new learner would authenticate to Google **twice** — once for Phantom, once for Supabase — for what they experience as one account. Signing in *from* the wallet keeps "Continue with Phantom" a single login.

`/api/auth/wallet` already creates accounts from wallets and `link-wallet` already handles conflicts and XP, so this adds no auth surface.

## Details worth a look

**Server-issued nonce, always.** Replay protection lives server-side; a client-invented nonce would be rejected. A test pins that the signed payload carries the nonce and domain the server just issued.

**`signMessage`, not `signIn`.** The embedded wallet has no Wallet Standard `signIn`, so this takes the same build-then-sign path `WalletAuthHandler` falls back to for adapters without it.

**One server key, two situations.** `walletAlreadyLinked` means either *your account already has a wallet* or *that wallet belongs to someone else* — very different advice. The outcome carries `mode` so the caller picks the right wording rather than showing one message for both.

**A declined signature is not an error.** It returns a distinct `declined` key and the UI stays quiet, matching how the existing wallet flow treats it.

**Server prose never reaches a learner.** Unparseable error bodies fall back to a generic key.

**No double prompt.** The address arrives twice in the normal flow (once from `autoConnect` on mount, once from the `connect` event); a ref guards against signing the same nonce twice. The handler also short-circuits when the profile already has a wallet, so it doesn't re-prompt on every page load.

## Verification

- **7 new tests** (31 across `phantom` + `auth`): sign-in vs link routing with the *wrong* route asserted absent, the signed payload's nonce/domain/address, declined-as-ordinary, `walletAlreadyLinked` carrying `mode`, nonce-fetch failure, and unparseable error bodies
- `vitest run` — **2699 passed**, 0 failed
- `eslint` clean, `tsc --noEmit` clean in touched files
- Browser: page renders, modal opens with Phantom first, no error boundary, and the handler correctly stays silent while signed out
- Locale key parity asserted across en/pt-BR/es

## Still open on the epic

- #987 graduation UI, #988 browser matrix, #989 React 19
- `enroll` is the one instruction needing a learner signature *and* naming the learner as payer, so an embedded-wallet learner still needs SOL to enroll. Moving that payer to the backend would make the whole journey zero-SOL — worth its own issue.

**Do not enable in production before #988.** The full flow past the Phantom redirect has not been exercised end to end — that needs a real Google account against the configured Portal domains.


Potolski:
All three findings were correct — fixed in 02b8c050, and I took the "better" fix shape rather than the ternary swap.

## The blocking finding: confirmed, including the diagnosis of *how* it shipped

The reasoning chain held up end to end: `bridge()` only links when the caller's own profile has no wallet, so the only realistic link-route 409 is *someone else owns this wallet* — and my `mode === "link"` branch showed the reassuring copy for exactly that case. And `/api/auth/wallet`'s 409 was prose, so the other branch was unreachable. Sharpest part of the review: the siws test mocked both routes returning `walletAlreadyLinked`, a contract the sign-in route never had — which is precisely how the inversion passed green.

## Fix: the server names the situation; clients stop guessing from routes

The route called was always a bad proxy for what went wrong, so `mode` no longer picks copy at all:

| Key | Situation | Copy |
| --- | --- | --- |
| `walletAlreadyLinked` | this wallet belongs to **another** account | `phantomWalletTaken` |
| `differentWalletLinked` | **this** account already has a different wallet | `phantomAccountHasWallet` |

- `link-wallet` splits its two 409s across those keys. Supporting evidence this was the intended design all along: the settings account tab has handled `differentWalletLinked` since before any route emitted it — this closes that half-built loop, and that consumer needed zero changes.
- `/api/auth/wallet`'s prose 409 becomes the same key — it was the one prose error in the auth surface, against the route conventions. `WalletAuthHandler` now translates it (`auth.differentWalletLinked`, en/pt-BR/es) instead of echoing raw English at localized users.
- Even the TOCTOU race now gets the right copy, since the server distinguishes the cases — the ternary swap alone would still have mislabelled it.

## Non-blocking, both taken

**Retry-after-decline:** real bug, fixed. `handledAddress` clears on every failure path — it exists to stop the same *successful* address bridging twice, not to make failure permanent. Tests pin both halves: a decline leaves retry open (no reload), and the double-fire guard still holds for an in-flight bridge.

**`signMessage` catch-all as "declined":** confirmed deliberate, now documented at the site. It's parity with `WalletAuthHandler`, and the underlying reason is that wallet SDKs don't reliably distinguish "user closed the prompt" from "prompt failed to open" — mislabelling a decline as an error is worse than a quiet retry, and with the ref fix the retry path is genuinely open either way.

## Coverage at the layer the inversion lived in

New `PhantomAuthHandler` component tests assert **which copy renders** for each server key — the gap the review named (`siws.ts` tests only threaded the key). The siws mocks now mirror the real route contracts, with a comment explaining why that matters.

Also merged the updated #992 branch in, so this PR carries the modal-unstick and CSP fixes. Full suite: 2708 passed, eslint + tsc clean.

Potolski:
@claude check for the blocking issues again

PR #1002, opened 2026-08-05
UI pass: profile, certificates, leaderboard, settings, reward popups

thomgabriel:
A UI pass over the surfaces we hadn't touched since the challenge/dashboard work: profile, certificates, leaderboard, settings, and the reward popups. Everything is frontend — no migrations, no program, no workflows.

**Profile** is recomposed around what survived a long review loop: the skills radar stays, achievements sit beside it as a token rail, and certificates run full-width as a shelf of real diploma cards. `/profile` and `/profile/[username]` are now the same page, Back included.

**Certificates** gained a dynamic OG/Twitter card, so sharing a credential on X unfurls the diploma itself instead of a generic site card (share intents can't attach images — the card *is* the image). Share actions are owner-only now; the Earn hand-off card and Copy Link are gone.

**Leaderboard and settings** were the two surfaces still speaking a different dialect. Both are contained rather than viewport-wide (rank and XP were a screen apart), both use one segmented control — literally the same CSS block — and settings form labels moved to the mono kicker idiom. Four inline-duplicated settings pieces (inputs, rows, group headings, two hand-rolled toggles) collapsed into shared classes.

**Reward popups** moved off the Solana gradient onto house cards, which restores the gradient's meaning: it now marks only the certificate popup, the one on-chain artifact. Level-up shows the real LevelBadge, achievements the earned octagon token.

**Avatar upload** lets you choose the crop instead of the element choosing it, and re-encodes to a 512px JPEG — which is why the source cap could go 1 MB → 12 MB. While in there: `AvatarImage` had no `object-cover`, so non-square avatars were being stretched.

**Removed:** the first-login name-reveal modal. Names are already assigned server-side at signup on both the SIWS and OAuth paths, so nobody lands nameless; renaming lives in settings.

Two latent bugs fixed in passing: the achievements row used `justify-content: flex-end`, which made the leading tokens unreachable once it overflowed (now `safe flex-end`, plus grab-to-scroll), and the plan card's `flex-1` wrapper ate the free space so `justify-center` centered nothing.

Crop geometry has unit tests since the dialog is auth-gated and can't be checked from a preview screenshot.

**Footer tagline** drops the "in Latin America and beyond" framing in all three locales — owner's call, the positioning is no longer LatAm-scoped.


thomgabriel:
The tagline change is intentional — owner asked for it directly, the positioning is no longer LatAm-scoped. Fair catch on the commit hygiene though: it rode along in an unrelated avatar/mobile commit and wasn't in the PR body. Body now says so.

Both non-blocking notes fixed in b03541b5 — rank aria-labels go through `gamification.rankLabel` (including the pre-existing one on the ranked row), and `profile.backToLeaderboard` is now `profile.back` since the button renders on /profile too.

thomgabriel:
@claude please re-review the current head (b03541b5) — it adds the two i18n fixes from your non-blocking notes.

PR #1003, opened 2026-08-08
feat(auth): Phantom graduation card — take your wallet with you

### LINKED ISSUE #987: Phantom Connect: graduation — "Take your wallet with you"

Potolski:
Closes #987. Part of epic #983.

Adds the "Take your wallet with you" card to Settings → Account, shown beneath the wallet row.

## The failure it pre-empts

The embedded wallet is a real Phantom wallet — download the app, choose **Continue with Google** with the same account, and the same wallet appears. But Phantom's onboarding makes the wrong branch the obvious one: **"Create new wallet"** mints a fresh address, the learner's credentials stay in the embedded wallet they can't see, and they conclude the Academy lost their NFTs. That support ticket is predictable, so the card spells out the three steps, bolds the do-NOT-create-new instruction, shows the **address tail** (first/last four chars) to confirm against, and links Phantom's own help article.

## Provenance gates the card — the design decision worth reviewing

It renders **only** when `prefs.walletProvider === "phantom-embedded"`. A learner who linked an external wallet must never be told to "download Phantom to take it with you" — those instructions are nonsense for a wallet that already lives elsewhere. Fail closed: no provenance record, no card.

## The stamp is deferred, and why

A successful bridge can't write prefs directly: on the fresh sign-in path the browser client has no session until after the post-bridge reload, so a prefs `UPDATE` would fail RLS. Instead:

1. `PhantomAuthHandler` sets a localStorage marker just before the reload
2. a mount effect flushes it into `prefs.walletProvider` once a session exists
3. **merge-write** — the same prefs blob carries `nextLesson` and `nameRevealSeen`, and a blind write would erase them (same idiom as the name-reveal dialog)
4. the marker clears **only after a successful write**, so a failed flush retries on the next page load rather than silently losing provenance

## Verification

- **8 new tests**: card renders with tail + help link for Phantom-provisioned wallets; renders *nothing* for external wallets, null prefs, or signed-out (awaited, not just initial-render — no flash of wrong instructions); flush merges siblings and clears the marker; read-failure keeps the marker for retry; no marker → zero Supabase calls
- `vitest run` — **2759 passed**, 0 failed
- `eslint` + `tsc --noEmit` clean
- Strings in en / pt-BR / es, key parity asserted

## Scope note

The issue floated "settings and/or post-first-credential". This ships the settings surface; a post-credential nudge would reuse the same component and gate, and can follow once the end-to-end flow (#988) is proven — no point promoting graduation before a learner can actually connect.


Potolski:
Correct on all counts — my `git add -A` swept it in, third time running. Fixed in 6c87ed57.

**PR diff: 145 files / 28,613 additions → 9 files / 418 additions.**

## Fixed at the root, not with a fourth gitignore line

You called this exactly right: appending `/Gals8246YhWF8nB0-8urE/` would just queue up escape number four, since gitignore has no regex quantifiers and the name is a fresh nanoid each run.

`apps/web/vitest.config.ts` had no `cacheDir` set. It now pins an **absolute** path under `apps/web/node_modules/.vitest-cache`, so the cache cannot reach the repo root regardless of the nanoid Vite picks *or* the cwd vitest is invoked from — which matters, because the escape only happens when it's run from the repo root rather than `apps/web`.

Verified rather than assumed: ran vitest **from the repo root** afterwards — no new top-level directory appeared, the cache landed in `node_modules/.vitest-cache/`, full suite green (2759).

Also noted the sourcemap embedding my local absolute path. Agreed it's not a secret, and it's gone with the rest.

## The leftovers are worse than they looked — filed as #1005

While cleaning up I checked what's still on `main`:

| Directory | Files tracked on `main` | Size |
| --- | --- | --- |
| `BgslqnjILApu8jL_vIFuC/` | 350 | 11 MB |
| `t19GkuBguxG71m2b6fymV/` | 251 | 7.2 MB |

~18 MB every clone pulls. And the failure is instructive: `BgslqnjILApu8jL_vIFuC/` **is** in the gitignore list at line 121 and is *still tracked*, because gitignore never untracks what's already in the index — it only stops new files. `t19GkuBguxG71m2b6fymV/` isn't listed at all. So the explicit-list workaround has now failed in both directions.

Kept out of this PR deliberately — an 18MB mechanical deletion doesn't belong in a feature diff. #1005 carries it, including your CI-guard suggestion (a top-level-directory allowlist is the only measure that would have caught all three escapes at the source).

## On the rest of the review

Thanks for verifying the RLS reasoning independently — particularly confirming `prefs` is non-privilege-bearing and that a learner hand-setting `walletProvider` themselves is self-inflicted wrong instructions rather than a trust-boundary issue. That was my read too, but it's the kind of thing worth a second pair of eyes on rather than taking my word for.

The duplicate `createClient()` + `prefs` read across the card and the handler is a fair flag. Two light `maybeSingle()` calls didn't feel worth an abstraction at two call sites, but you're right that a third consumer of `prefs.walletProvider` is the trigger to factor it out — noted for whoever adds the post-credential nudge.

CI is the hard gate as you say; re-running now against the reduced diff.

Potolski:
@claude review the latest comment against the issues you found earlier

PR #1014, opened 2026-08-11
fix(courses): drop the duplicated card title when the banner carries it

thomgabriel:
The new course banners bake the title into the artwork, so the card was saying every title twice. Cards with a real thumbnail now skip the text heading; the /cover.png fallback keeps it (that image has no title), and screen readers keep the title via the link's aria-label either way. Verified on the local catalog: both courses render banner-only, zero headings, aria intact.


thomgabriel:
@claude re-review the current head please — 789cfe25 adds the green dark-mode lockup (new SVG variant + header/footer swap) on top of the card-title change you already reviewed.

PR #1016, opened 2026-08-11
fix(content): bump content.lock to 3eb3660 — deploy quest unlocked on your-first-solana-program

thomgabriel:
Owner report: the deploy quest on /courses/btc-to-sol-evolution/lessons/your-first-solana-program never appears. Root cause is content, not code: the block had buildType: buildable but no deployable flag, and the schema defaults it false — so the deploy panel never mounts even though the lesson title promises a devnet deploy. academy-courses#36 adds the one-line flag; this bump activates it (regenerated bundle diff is exactly that flag + the meta stamp).

The only other buildable block (an-anchor-vault) also lacks the flag but its copy doesn't promise a deploy — owner decision pending on whether it should deploy too.


thomgabriel:
@claude re-review the current head please — one commit added since your LGTM: the golden-fixture flip for the deployable delta (documented in the test header's content-wave list), which is what un-broke Frontend CI.

Issue #1004, opened 2026-08-08
enroll: split payer from learner so zero-SOL embedded wallets can enrol


Potolski:
Part of the #983 story, but a PROGRAM change — filed separately because it is a different shape of work from the frontend epic.

## The gap

`enroll` is the one instruction where the learner is both **signer and payer** (`programs/onchain-academy-pinocchio/src/instructions/enroll.rs` — `learner (signer, mut)` pays for the Enrollment PDA). Every other learner-facing instruction (`complete_lesson`, `finalize_course`, `issue_credential`) is backend-cosigned with the backend paying.

So an embedded-wallet learner sails through Phantom Connect sign-up with zero friction — and then hits a wall at their **first enrol**: their freshly-provisioned wallet holds no SOL. The whole "no wallet friction" story dies at the exact moment they try to start learning.

## Proposed change

Split payer from learner in `enroll`: learner stays the signer (consent — enrolment writes their pubkey into the PDA and starts their bitmap), a separate `payer (signer, mut)` funds the account, and the frontend routes enrolment through a backend-cosigned path like the other instructions.

Rent for an Enrollment PDA is small and bounded; at current course counts the cost of sponsoring every enrolment is trivial, and the backend signer is already rotatable via `update_config` if the key needs cycling.

## Why this is not a frontend PR

- Program change → discriminator/CPI-shape review against the Anchor-parity contract (`docs/ANCHOR-VS-PINOCCHIO.md` catalogues every divergence)
- CU re-measurement + drift-gate baseline update (`tests/CU_BASELINE.rust.md`)
- Differential + acceptance suites extend to the new account shape
- A deploy, with the pre-mainnet checklist implications that carries

## Alternatives considered

- **Faucet/airdrop dust to new wallets** — works on devnet, unacceptable on mainnet (costs SOL per learner *and* leaves them holding dust; also a sybil magnet)
- **Keep learner-pays, show a funding UI** — surrenders the entire point of embedded onboarding
- **Backend fully signs enrolment without the learner** — loses the learner's consent signature; enrolment is the one action that SHOULD carry it

## Acceptance

- Embedded-wallet learner with 0 SOL enrols successfully
- Learner signature still required (consent preserved)
- CU baseline updated, differential suite green, byte-compat divergences documented

PR #1018, opened 2026-08-12
feat(enroll): walletless enrolment — custodial Academy Wallet + embedded-wallet support, no install required

thomgabriel:
## Problem

Enrolling requires a wallet: a visitor without a browser-extension wallet who clicks **Enroll** dead-ends at the wallet-adapter "install a wallet" modal. The Phantom Connect embedded-wallet path (#983/#984) cannot close this gap in production yet — it is blocked on Phantom Portal's external domain approval.

## What changed

The Enroll click now works for every account state, choosing the least-friction path available:

### 1. Custodial "Academy Wallet" — the zero-dependency walletless path (`52d7b25`)

For a signed-in account with **no linked wallet** (plain Google/GitHub OAuth), enrollment happens **entirely server-side** with no redirect, extension, or signing prompt:

- `lib/solana/custodial-wallet.ts` — per-account keypair via `HKDF(CUSTODIAL_WALLET_SECRET, userId)`. **Derivation, not storage**: no private key ever touches the database. Rotation hazard (new secret = every derived address changes) documented at the definition and in `CLAUDE.md`.
- `lib/solana/custodial-enroll.ts` — builds, signs (backend sponsor pays fee + rent; custodial key provides the learner consent signature) and confirms `enroll` server-side. Idempotent on an existing Enrollment PDA.
- `POST /api/enroll/custodial` — auth'd, rate-limited, freeze/maintenance-gated. Binds the derived address as `profiles.wallet_address` with provenance `prefs.walletProvider: "custodial"` (guarded against a concurrent SIWS link with a conditional claim + recheck), refuses accounts that linked a different wallet (409 `walletLinked`), and mirrors the `enrollments` row so the UI updates instantly instead of waiting for the Helius webhook.

**Custody trade-off, stated plainly:** these wallets can only ever hold soulbound XP (NonTransferable Token-2022) and frozen Core credential NFTs — nothing transferable — and the master secret lives in the same trust domain as `BACKEND_SIGNER_SECRET`. Feature is **off unless `CUSTODIAL_WALLET_SECRET` is set** (route 503s with a stable key and the client falls through).

### 2. Embedded-wallet (Phantom Connect) enrolment — ready for when the domain is approved

- `PhantomConnectProvider` exposes `signTransaction` (sign-only; the app submits via its own RPC so the SDK's network setting can't route a devnet tx to the wrong cluster).
- Embedded wallets enrol through the existing sponsored route (#1004); sponsored **only**, since a zero-SOL wallet can never self-pay.
- Wallet-mismatch hardening from review: the client verifies the sponsored transaction's `learner` matches the wallet about to sign, and walletless provisioning is refused for accounts whose linked wallet is a different one (`PhantomAuthHandler` would never link a second wallet).

### 3. Click-survives-redirect resume

A TTL-bounded (15 min) `sessionStorage` marker carries the Enroll intent across auth/provisioning redirects. With Phantom off, resume fires immediately after sign-in and completes custodially — so the full booth flow is: *click Enroll → sign in with Google → enrolled*, zero installs, zero SOL, zero prompts.

Fallback order for a walletless click: **custodial → Phantom provisioning → wallet modal**; extension-wallet users are untouched (sponsored when available, self-paid otherwise).

## Testing

- 33 new/updated tests across the hook, the custodial route, key derivation, and the provider (`signTransaction`); full web suite green (2,803 tests). `tsc --noEmit`, ESLint, Prettier clean.
- End-to-end custodial flow needs one config to go live: set `CUSTODIAL_WALLET_SECRET` (≥ 32 chars, server-only) in Vercel.

## Config

| Var | Effect |
| --- | --- |
| `CUSTODIAL_WALLET_SECRET` | Enables walletless custodial enrolment. Unset = off. **Never rotate casually** — see `apps/web/CLAUDE.md`. |
| `NEXT_PUBLIC_PHANTOM_APP_ID` | Enables the Phantom embedded-wallet path (currently blocked on Portal domain approval). Unset = off. |

https://claude.ai/code/session_01P7BRjrUn4p3wjDok3Ugwor


thomgabriel:
Addressed the blocking finding in 086c4f9 — the wallet-mismatch scenario was real, both variants:

- `fetchSponsoredEnrollTransaction` now reads back the `learner` the sponsor route built for. The embedded-wallet path refuses to sign on a mismatch and surfaces "This account is linked to a different wallet. Connect that wallet to enroll." instead of an opaque signature failure at submission. The adapter path only uses the sponsored transaction when the connected wallet is the linked one, otherwise it self-pays for the connected wallet exactly as before this PR.
- The walletless branch now checks `profiles.wallet_address` before provisioning. Extension-linked accounts are sent to the wallet modal to connect their linked wallet (a newly provisioned embedded wallet could never be linked or sign for them). Embedded-linked accounts (`prefs.walletProvider === "phantom-embedded"`, the #987 provenance marker) still go through `connect("google")`, since user-wallet scoping reconnects the same wallet — and the learner check above backstops the wrong-Google-account case.

Also softened the TTL doc comment per the non-blocking note. New regression tests cover all three cases; full web suite green (2782 tests).

---

Issue #1017, opened 2026-08-11
Re-enable Phantom Connect in prod once Portal approves the app


Potolski:
## Status

Phantom Connect is **disabled in production**. `NEXT_PUBLIC_PHANTOM_APP_ID` has been removed from the Vercel production environment so the "Continue with Phantom" button does not render.

## Why it was disabled

The button was live and reaching Phantom, but Phantom rejected every attempt:

```
Auth2 /login/start request failed (400). Bad Request
```

The app submission was rejected because the domain-verification TXT record took too long to propagate. The DNS record (`_phantom_portal-challenge`) now resolves correctly, so the app can be resubmitted — but until Portal approves it, `/login/start` will keep 400ing and the button is a dead end for learners.

## How it is disabled

No code change. The integration was built with an env-var gate:

- `getPhantomAppId()` returns null for unset or whitespace-only (`lib/phantom/config.ts`)
- `createPhantomClient()` then returns null
- the provider's `enabled` stays `false` — no SDK, no listeners, no traffic (`phantom-connect-provider.tsx:70`)
- the modal button is behind `{phantomEnabled && …}` (`auth-modal.tsx:195`)

## To re-enable

1. Confirm the app is **approved** in [Phantom Portal](https://phantom.com/portal/), and that the allowed-domains list contains the production origin **and** `http://localhost:3000` for dev.
2. Set `NEXT_PUBLIC_PHANTOM_APP_ID=d12abb45-9120-4e3c-a5f7-c8e0ad9842c3` in Vercel **Production**.
3. **Redeploy with build cache disabled.** `NEXT_PUBLIC_*` is inlined at build time, so a cache-reusing redeploy silently keeps the old bundle. This bit us when disabling it — the variable was removed and the app id was still present in the served chunk afterwards.
4. Verify from outside:
   ```bash
   curl -sL https://st.academy/en \
     | grep -o '/_next/static/chunks/[^"]*\.js' | sort -u \
     | while read c; do curl -sL "https://st.academy$c" \
         | grep -q d12abb45 && echo "present in $c"; done
   ```
5. Do a real sign-in and confirm `/login/start` returns 200, not 400.

## Blocks / related

- #988 — browser verification matrix (Safari / iOS / Brave). Still unrun; the redirect flow is exactly where those browsers break. Worth doing on the same pass as re-enabling.
- #983 — Phantom Connect epic.

## Note

[...]

PR #1020, opened 2026-08-12
feat(auth): replace Phantom with Dynamic; wallet and email login side by side

### LINKED ISSUE #989: Phantom Connect: React 19 upgrade evaluation (unblocks @phantom/react-sdk)

Potolski:
Replaces Phantom Connect with Dynamic, and makes both ways in first-class: **connect a wallet directly** (unchanged wallet-adapter SIWS) **or sign in with an email/social account** and get a Solana wallet provisioned for you.

Closes #989. Supersedes #1017 — the Phantom re-enable it tracked no longer applies.

## Why

Phantom Portal never approved our app, so `Auth2 /login/start` returned 400 for every learner and "Continue with Phantom" was a live dead end in production. Dynamic requires no per-app approval.

It also clears the blocker that shaped the Phantom integration: `@phantom/react-sdk` required React ≥19.0.1 against our 18.3.1, which is why we hand-wrapped a browser SDK (#989). Dynamic's React SDK declares `"react": ">=18.0.0 <20.0.0"`, so we use the vendor's supported bindings.

## Both paths, one account model

The two entry points converge — this is deliberately **not** a second auth system:

| Path | Wallet from | Session via |
| --- | --- | --- |
| Connect a wallet | wallet-adapter | `WalletAuthHandler` → SIWS |
| Email / social | Dynamic | `DynamicAuthHandler` → SIWS |

Both sign a server-issued nonce and exchange it through the **existing** `/api/auth/wallet` and `/api/auth/link-wallet`. Those routes are **unchanged** — the SIWS bridge was already provider-agnostic behind a `MessageSigner` interface, so it moved from `lib/phantom/siws.ts` to `lib/wallet/siws.ts` (`runPhantomSiws` → `runWalletSiws`) and gained a Dynamic adapter.

## The one real integration subtlety

Dynamic and our bridge disagree on both ends of `signMessage`:

| | our `MessageSigner` | Dynamic's Solana wallet |
| --- | --- | --- |
| input | `string \| Uint8Array` | `string` only |
| output | `{ signature: Uint8Array; publicKey }` | `string \| undefined` (base58) |

`lib/dynamic/siws.ts` adapts it: message in as text, base58 signature decoded to the raw bytes the route verifies, address carried separately because Dynamic's signer returns none.

`undefined` (Dynamic's "no signature" — a declined prompt) is **thrown, not passed on**. `runWalletSiws` treats a throw as a quiet retryable decline; an empty `Uint8Array` would instead reach the server and fail verification as though the learner had forged a signature.

## CSP

Origins were **read out of the installed SDK** (4,165 files across 45 `@dynamic-labs` packages), not guessed:

- `connect-src`: `app.dynamicauth.com` (API), `logs.dynamicauth.com` (telemetry)
- `frame-src`: `webview.dynamicauth.com` — unlike the Phantom SDK, **this one does frame**, so `frame-src` is no longer untouched

Pinned exactly rather than `*.dynamicauth.com`, matching the previous convention.

## Phantom removal

`lib/phantom/`, the provider, auth handler, graduation card, logo and their tests are deleted, along with 8 orphaned i18n keys per locale.

Verified before deleting: `auth.phantomAccountHasWallet` / `auth.phantomWalletTaken` were Phantom-modal copy consumed only by the deleted handler. The keys the auth routes actually emit — `walletAlreadyLinked` / `differentWalletLinked` — are untouched and still have copy.

The settings graduation card is gone rather than ported: Dynamic exposes wallet export in its own account UI, so a bespoke card would duplicate and drift from instructions the vendor owns.

## Gate

Unchanged contract from Phantom: `NEXT_PUBLIC_DYNAMIC_ENVIRONMENT_ID` unset means the whole path is off — no provider, no SDK init, no button — and SIWS with an external wallet stays the guaranteed way in. Whitespace-only counts as unset so a blank dashboard field fails at the gate rather than at connect time.

**Deploy note:** `NEXT_PUBLIC_*` is inlined at build time, so setting this needs a redeploy with **build cache disabled**. A cache-reusing redeploy silently keeps the old value — that cost us a live incident when switching Phantom off.

## Tests

Full web suite green, `tsc --noEmit` clean, i18n key parity verified across en / pt-BR / es (1380 keys each).

## Worth a deliberate decision

The SDK pulls in **331 transitive packages** for what we use of it (an address and a `signMessage`). Not a blocker, but a real dependency and supply-chain surface increase that deserves a conscious yes.


Potolski:
All seven review items addressed, plus one defect the review could not have seen from the diff: **this branch could not produce a production build at all.** Details per item:

**1 (blocking) — orphaned `@phantom/browser-sdk`: removed.** `pnpm remove @phantom/browser-sdk --filter web`; the transitive tree is gone from the lockfile. The framing was fair — dead supply-chain surface in a PR lecturing about dependency weight.

**2 — the hook assumption: you were right, and it was worse than "worth verifying".** From the SDK source: `createContext(undefined)` + `if (context === undefined) { if (isSSR()) return SSR_SAFE_DYNAMIC_CONTEXT; throw ... }`. It falls back **only during SSR** and throws on the client — so with the env var unset, every page server-rendered cleanly and crashed on hydration. `DynamicAuthHandler` had the same bug as a *sibling* of the provider, taking out every page, not just the modal. Fixed with a component boundary (`DynamicSignInButton` owns the hook; the handler now mounts *inside* the provider), and covered by `auth-modal-no-dynamic-provider.test.tsx`, which renders the modal with the gate off and **does not mock the SDK** — the mocked sibling suite is exactly why nothing caught this.

**3 — stale docs/config: all five swept.** `env.ts` no longer validates the dead Phantom var; `apps/web/CLAUDE.md` and `.env.example` now document `NEXT_PUBLIC_DYNAMIC_ENVIRONMENT_ID` including the build-cache warning, placed where a deployer will see it; the `lib/wallet/siws.ts` header now describes the module as provider-agnostic; the `wallet/route.ts` comment names `DynamicAuthHandler`.

**4 — Dynamic test coverage:** `lib/dynamic/__tests__/siws.test.ts` pins the adapter's wire format — base58 → exact bytes, string passthrough, `Uint8Array` → text, and the `undefined` → throw branch (a decline must never reach the server as an empty signature, where it reads as forgery). Plus the unmocked no-provider modal test from item 2.

**5 — CSP tests:** assertions added for `app.dynamicauth.com` / `logs.dynamicauth.com` (`connect-src`, exact hosts, explicitly not `*.dynamicauth.com`) and `webview.dynamicauth.com` (`frame-src`) — and the old test's "frame-src is untouched" premise corrected, since it no longer is. A regression test also asserts no `phantom.app` origin survives.

**6 — pt-BR accent:** "você" fixed.

**7 — leftover "Phantom Connect" comments** in `instructions.ts`, `sponsor-enroll.ts`, `use-on-chain-enroll.ts`: reworded to "embedded wallet (Dynamic)".

**The defect the diff hid: `next build` OOM'd** (`FATAL ERROR: Reached heap limit`, no `routes-manifest.json`), reproduced locally at a 4GB heap. Raising the heap to 8GB surfaced the *real* second failure underneath: `Module not found: Can't resolve 'axios'` in `@irys/bundles` (the Arweave uploader behind `/api/certificates/mint`).

Root cause, verified not guessed: `@irys/bundles` imports `axios` **without declaring it** (neither dependency nor peer, both installed versions), and it only ever resolved through pnpm's hoisted fallback because `@phantom/browser-sdk`'s tree provided axios. Item 1's removal took that away — meaning **main builds by accident today**, and any future removal of the last axios provider would have broken it the same way. Fixed at the root with `pnpm.packageExtensions` granting `@irys/bundles` its missing `axios` dependency.

With axios restored, a **third layer** surfaced: WalletConnect (a `@dynamic-labs/solana` transitive) ships `pino`, whose dynamic require of its *optional* pretty-printer hard-fails webpack the same way. `pino-pretty` / `lokijs` / `encoding` are marked as webpack externals — the canonical WalletConnect/Next.js recipe; they are dev-CLI conveniences, correct to leave unresolved rather than install. Both `@dynamic-labs` packages are also in `optimizePackageImports` (the sdk is ~11MB across 4,302 files behind a single barrel).

**Deploy note:** the compile needs more heap than Vercel's default — set `NODE_OPTIONS=--max-old-space-size=8192` in the Vercel build env (or enable Enhanced Builds) when this merges, or the container-level SIGKILL comes back.

Verified end to end: `tsc --noEmit`, the full web suite, and a completed production build with `routes-manifest.json` emitted — the build being the point, since it is what could not finish before.

Potolski:
@claude check if the blocking issues were fixed

Potolski:
**Live-flow findings** (from testing against the real environment, `12482bcf…`):

- The environment's served settings are verified correct: `automaticEmbeddedWalletCreation: true`, embedded chain `SOL` (enabled, primary), Solana on. Checked via the SDK's public settings endpoint, not the dashboard UI.
- **"Create on sign up" fires at sign-up only.** An account that existed before the toggle gets Dynamic's *Link a new wallet* picker on login instead of an embedded wallet — which looks exactly like a broken integration when you test with your own pre-existing account. Fresh emails get the wallet automatically. v5's React SDK exports no programmatic `createEmbeddedWallet`, so there is no code-side backfill; delete pre-toggle test users in the dashboard (Users tab) rather than debugging the picker.
- `d9d1bcea` adds the `dynamic-static-assets.com` CSP origins (wallet catalogue + icon sprites) that only surface at runtime — the read-the-dist sweep could not find them because the SDK builds those URLs dynamically.

PR #1026, opened 2026-08-13
feat(landing): point "Start building" at the Solana Speedrun course

Potolski:
## Summary

The homepage's "Start building" hero CTA (and the bottom "Get started" CTA) deep-link into the flagship entry course's first lesson. That entry was `course-btc-to-sol-evolution`, so a landing visitor clicking "Start building" opened the blockchain-history course instead of the hands-on **Solana Speedrun**.

## Change

One constant: `ALPHA_ENTRY_COURSE` in `lib/courses/learner-segment.ts` now points at `course-solana-speedrun`. This is the single source of truth for the flagship entry, so:

- both landing CTAs now open the Speedrun's first lesson
- the `/start` funnel retargets with it — by design it converges on the same lesson as the landing deep-link, so changing only the CTA would have broken that invariant

The `entry-course-live` suite (which runs the real resolver against the real committed bundle) has its expectation updated and passes — proving the Speedrun entry resolves to a live first-lesson URL rather than the `/courses` fallback.

## Note for after merge

The deep link stays sync-gated: if `course-solana-speedrun` isn't deployed+synced in the production DB, the CTA safely falls back to `/courses`. Worth a one-click check on the deployed site that the button lands on the Speedrun lesson (i.e., the course is synced).

## Verification

- `vitest`: 284 files / 2758 tests pass (entry-course-live asserts the new target against the real bundle)
- `tsc --noEmit` clean

---


Potolski:
Closing: superseded by #1024, which landed the same retarget on main (`EVENT_ENTRY_COURSE = "course-solana-speedrun"`) minutes before this. After resolving the conflict against main, this branch's diff is empty — the "Start building" CTA already deep-links into the Solana Speedrun on main.

---

PR #1027, opened 2026-08-13
fix(auth): event readiness — Dynamic sign-out + OTP rate-limit copy

thomgabriel:
Two field-critical fixes for today's event plus two riders, all from the 13-08 review of the Dynamic integration.

**Sign-out bug (the real one):** app sign-out killed Supabase and wallet-adapter but never Dynamic. The Dynamic session survives, DynamicAuthHandler mounts on every page, and MPC signing is promptless — so the next page load silently re-signed the learner in. On borrowed/shared devices at an event that makes sign-out meaningless. Fix: `logoutDynamic()` helper (gate-aware, best-effort) called from the sign-out path; Dynamic's docs name `logout()` as the required session end.

**429 copy:** Dynamic's OTP endpoints allow 10 req/min per IP — a venue NAT trips this with a handful of concurrent sign-ups. Rate-limited sends/verifies now say 'too many sign-ups from this network — wait a minute' instead of the wrong-code message.

Riders: aria-label on the mobile lesson Back link (text label is display:none below sm, so the link announced as unnamed), and the device-registration comment corrected (docs: optional and off by default, not required).

Deliberately NOT touched: the speedrun deep-link serving all locales — that's the event configuration working as intended today; the en/es question is a post-event decision.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
@@ -728,6 +728,9 @@ export function LessonPageClient({
         <div className="flex min-w-0 flex-1 items-center gap-2 sm:gap-3">
           <Link
             href={`${linkBase}/${courseSlug}`}
+            // Below sm the text label is display:none, which drops it from the
+            // a11y tree — the aria-label keeps the link named on phones.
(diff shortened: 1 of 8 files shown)


thomgabriel:
@claude re-review the head please — one commit: the timeout you suggested on logoutDynamic (2.5s race; losing it is safe since the redirect reloads and the next OTP send clears any straggler session).

PR #1028, opened 2026-08-13
feat(auth): bridge Dynamic social sign-in to the existing Supabase account

thomgabriel:
Signing in with Google through Dynamic currently lands the learner in a fresh wallet-shaped account, stranding their XP and credentials — prod carries 108 google + 8 github identities to preserve. This adds POST /api/auth/dynamic: verify the Dynamic JWT against our environment's JWKS (RS256 family only, iss + environment_id bound to our env, exp required, scope must include user:basic), take the provider-verified email from the OAuth credential that established the session, and mint a session for that account through the same generateLink → verifyOtp mechanism as /api/auth/wallet.

Email-OTP credentials and oauth_emails are deliberately not matched on — GitHub's list can carry addresses the user never verified. Claim shapes come from the SDK's own JWT model rather than prose docs; aud is not pinned because Dynamic sets it to the client origin.

Inert until a client sends traffic — the Google button wiring is the follow-up PR. 29 route tests.


thomgabriel:
@claude review the head commit 929c30df — it adds the post-login parity block (placeholder-username replacement + retryPendingOnchainActions drain) your first review asked for, plus 5 tests. Focus on whether the parity block matches the wallet/callback chokepoints and introduces no new risk.

PR #1029, opened 2026-08-13
feat(auth): Continue with Google through Dynamic

thomgabriel:
Wires the Google button through Dynamic when the environment id is configured: click → signInWithSocialRedirect (web social sign-in is redirect-only — the popup variant throws on web) → DynamicAuthHandler picks up the return leg, exchanges the Dynamic JWT for a Supabase session via /api/auth/dynamic (#1028), and reloads. The learner lands in their existing email-matched account and walks away with an embedded wallet. With no environment id the Supabase OAuth button renders untouched — that's the kill switch.

Two SDK findings that shaped the implementation: the bridge must send `legacyToken` (the full DynamicJwt with verified_credentials) — the public `token` is the minified JWT, which verifies and then 403s; and a `bridgingSocial` guard holds off the SIWS auto-sign-in during the return leg, which would otherwise mint a second wallet-shaped account before the Supabase cookies land.

**Draft until verified on localhost:3000 with a real Google login** (Vercel previews aren't in Dynamic's CORS allowlist). One thing only that E2E can prove: if the Dynamic environment uses cookie-based auth storage, `legacyToken` is null and the bridge fails closed — the fix is the dashboard's auth-storage toggle, not code.

73 tests across the route + new social/button suites.

### DIFF
--- a/apps/web/src/app/api/auth/dynamic/__tests__/route.test.ts
+++ b/apps/web/src/app/api/auth/dynamic/__tests__/route.test.ts
@@ -87,14 +87,21 @@ let privateKey: CryptoKey;
 let ps256PrivateKey: CryptoKey;
 const fetchMock = vi.fn();
 
-/** The credential Dynamic emits for a completed Google sign-in. */
+/**
+ * The credential Dynamic emits for a completed Google sign-in, as observed in
+ * a LIVE sandbox token (2026-08-13): no `email` field — the address arrives
+ * only in `oauth_emails` (single element) and `oauth_username`. The SDK model
+ * claims `email` can be present; tests for that shape pass it explicitly.
+ */
 function googleCredential(overrides: Record<string, unknown> = {}) {
   return {
     id: GOOGLE_CREDENTIAL_ID,
     format: "oauth",
     oauth_provider: "google",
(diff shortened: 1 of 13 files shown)


thomgabriel:
@claude re-review head 097b82ea — it converts the bridgingSocial guard from a ref to state initialized true, so releasing it re-runs job 3 (your stuck-link-path race). Check the state-based guard has no new gap: first-render claim, the reload path keeping it held, and the deps change on job 3.

thomgabriel:
@claude re-review the two commits since your last pass — ebcd81a8 (matcher: live google credential carries no email field, so single-element oauth_emails is accepted for google only; github still requires the email field — check the provider scoping is airtight and the multi-element refusal holds) and ee9fa129 (social-return handshake guarded to run at most once per page load; module-scope flag claimed synchronously — check the non-redirect path still releases the bridgingSocial state on remounts).

thomgabriel:
@claude final check on f53867ba — implements your in-flight-vs-settled fix: module-level promise, every invocation awaits it, settled outcome (release/hold) decides the guard; plus the StrictMode double-invoke test you asked for (3 cases, mutation-checked against the boolean version). Confirm the gap is closed.

PR #1036, opened 2026-08-17
docs: AUTH-FLOWS.md — every way into a session, mapped from code

thomgabriel:
The definitive auth map for future maintenance: all three ways in, the DynamicAuthHandler's four jobs with the incident history behind each guard, /api/auth/dynamic's verification ladder, identity linking, session model, known seams, and the trust-root decision record.

Three findings from writing it worth a maintainer's attention: the 'Dynamic proves things' invariant only holds on the Dynamic path (the doc states the corrected version); dynamic-email-sign-in.tsx is a ghost component one import away from reintroducing the account-fork; and the unlink 2-methods rule effectively means 'one usable method plus the permanent wallet'.

### DIFF
--- a/docs/AUTH-FLOWS.md
+++ b/docs/AUTH-FLOWS.md
@@ -1,0 +1,306 @@
+# Auth Flows
+
+Every way into a Supabase session, mapped from the code. Paths are relative to
+`apps/web/` unless they start with `docs/` or `onchain-academy/`.
+
+The architecture invariant:
+
+> An external prover proves things (Dynamic for Google handshakes and embedded-wallet
+> ownership, wallet-adapter for external-wallet ownership, Supabase OAuth for GitHub);
+> **Supabase owns identity**; every proof is exchanged **server-side** for a Supabase
+> cookie session.
+
+The one-liner "Dynamic proves things" is true only for the Dynamic path. Two ways in
+never touch Dynamic: wallet-adapter SIWS (the wallet itself signs the proof) and
+Supabase OAuth (GitHub always; Google when Dynamic is unconfigured), where Supabase both
+proves and owns. The load-bearing halves hold everywhere: no client-side session
+minting, and Supabase user ids are the only identity every table and RLS policy hangs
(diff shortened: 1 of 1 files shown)


thomgabriel:
Adversarial validation of every falsifiable claim against main (a505740d): **~88 checked — 82 true, 1 false, 5 imprecise.** Two independent verifiers converged on the false one.

**Must fix before merge** (doc:34-39): "All four server chokepoints share three post-login rituals" — false. The three session-minting routes do (`wallet` 177/217/229, `dynamic` 505/521/535, `callback` 110/124/155); `middleware.ts` carries the tombstone check only (zero placeholder-replace, zero queue-drain), and it inspects sessions rather than minting them. Since "add a new way in, add all three" is the sentence future auth PRs will be reviewed against, the count has to be right: *three minting chokepoints share three rituals; the middleware is a per-request tombstone backstop.*

**Worth a line each (imprecisions/omissions):** avatar refresh applies to any OAuth provider's `user_metadata.avatar_url`, not just Google (and it sits in the GitHub section); the SIWS issued-at/window checks are skipped when `Issued At` is absent (`verify-siws.ts:172` wraps them in `if (issuedAt)`); the ways-in diagram omits the fourth lane — Dynamic embedded-wallet SIWS via handler job 3 signs users *in*, not just links; job 3's real re-prompt guard for linked users is the `wallet_address` early return at `dynamic-auth-handler.tsx:247-254` (`handledAddress` only dedupes within a page load); the 10KB/16KB "caps" are decimal UTF-16-length checks after the body is fully read; the dynamic route's rate limiter fails open on store error (doc flags fail-open for the tombstone, silent here); middleware also gates `/admin` on a separate HMAC cookie — one line saying that's out of scope stops a wrong conclusion.

**Everything else held under hostile checking** — all 11 ladder rungs down to the constants, the guard mechanics (`??=` before first await, `bridgingSocial` in both deps arrays), both 409 keys, the atomic nonce consume, the kill-switch chain, and an independent entry-point enumeration confirming the doc's coverage of Supabase-session mints is complete. PR attributions match the git log. Merge-grade once the chokepoint sentence is corrected.

PR #1039, opened 2026-08-17
feat(auth): auto-merge wallet-shell accounts on Dynamic sign-in

thomgabriel:
Heals the mixed-method account fork AUTH-FLOWS.md §7 documents (until now only prevented): when a Dynamic sign-in's JWT carries a `blockchain` credential matching a wallet-shaped shell account (synthetic `@wallet.superteam-lms.local` email, no OAuth identities), the shell — wallet, XP, enrollments, everything — folds into the account just signed into. The JWT credential is the ownership proof; an email match alone never merges.

The merge is one SECURITY DEFINER transaction (`merge_wallet_shell_account`): keep-target on unique collisions (2026-08-13 manual-merge precedent), XP totals summed, shell tombstoned, and a fail-closed FK sweep that aborts if any table the function doesn't cover still references the shell — a future table can't be silently stranded. Route side is fail-open: any doubt skips the merge and sign-in proceeds.

Execution tests run against the real `enforce_profile_wallet_write` trigger in pglite — which caught that `SET ROLE` is forbidden inside SECURITY DEFINER (the function satisfies the trigger via `request.jwt.claims` instead). Red-proof: merge route tests 2-failed at the pre-change head; execution suite fails wholesale without the migration.

Migration not yet applied to prod — will apply via MCP + ledger reconciliation after gate.

### DIFF
--- a/apps/web/src/app/api/auth/dynamic/__tests__/route.test.ts
+++ b/apps/web/src/app/api/auth/dynamic/__tests__/route.test.ts
@@ -16,6 +16,9 @@ const {
   isAccountDeleted,
   profileSingle,
   profileUpdate,
+  shellLookup,
+  getUserById,
+  rpcMock,
   retryPendingOnchainActions,
   generateWalletName,
 } = vi.hoisted(() => ({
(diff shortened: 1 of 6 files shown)


thomgabriel:
Gate record, sensitive lane: three-round independent adversarial review, all execution-verified. R1 FAIL → F1 (unscoped FK sweep aborted every real merge on GoTrue's auth.identities) + F2 (keep-target destroyed on-chain enrollment/progress). R2 FAIL → R4 (xp_transactions idempotency collision — the exact target users abort), R5 (minted certificates/achievements destroyed, global signature unique trap), R6 (flags partial uniques). R3 PASS at 30fd64fd; three non-blocking notes folded in the head commit. claude-review remains down (usage-wall $0 refusals — same outage recorded on #1037/#1040); the adversarial review is the review of record. Migration applies to prod via MCP after merge, with the DB-MIGRATION-LEDGER cross-reference.

PR #1041, opened 2026-08-17
fix(admin): course create writes bundle XP values; sync no-ops when chain matches

### LINKED ISSUE #993: Course create path sets xpPerLesson/creatorRewardXp needing an immediate update_course to match the bundle
Seen on the alpha go-live (2026-08-04): `POST /api/admin/courses/sync` created both alpha courses (`action: created`), but a follow-up sync from prod found on-chain drift and issued `update_course` with `fieldsUpdated: [newXpPerLesson, newCreatorRewardXp]` on both — meaning the create arm writes those fields differently from the bundle values the drift check expects (txs `KRG7K7GM…`, `2naFgjVb…`). Harmless when caught, but a created-and-never-resynced course would run with wrong XP config. Done-when: the create arm writes bundle values in the create tx (or immediately after), and a created course produces a no-op on the next sync.

thomgabriel:
Closes #993. Two halves of the same defect: the create arm defaulted `xpPerLesson ?? 10` (everything else — projector, drift engine, award path — treats absent as 0), and the update arm pushed `newXpPerLesson`/`newCreatorRewardXp` on every sync without comparing to chain, so a created course always ate an immediate `update_course` and could never no-op. Now the create tx writes bundle values and the update arm only includes fields that differ from the decoded on-chain course (plus an explicit 409 if the account vanishes between the two reads). Done-when evidence is operational per the issue: the next prod sync of the two alpha courses should report `action: noop`.

### DIFF
--- a/apps/web/src/app/api/admin/courses/sync/__tests__/route.test.ts
+++ b/apps/web/src/app/api/admin/courses/sync/__tests__/route.test.ts
@@ -1,0 +1,174 @@
+/* eslint-disable import/order -- vi.mock calls must precede importing the route. */
+// #993 regression: a course whose bundle XP values already match the chain
+// must NO-OP on sync — the old update arm pushed newXpPerLesson /
+// newCreatorRewardXp unconditionally, so every sync issued an update_course
+// and a freshly created course could never settle.
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { NextRequest } from "next/server";
+
+vi.mock("server-only", () => ({}));
+
+const {
+  getAccountInfo,
+  fetchCourse,
+  updateCoursePda,
+  deployCoursePda,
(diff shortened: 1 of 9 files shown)


thomgabriel:
Gate record, on-chain admin lane: three-round independent adversarial review. R1 FAIL — the `?? 0` half was a misdiagnosis (fallback is dead code behind getMissingCourseFields; xpPerLesson convention elsewhere is 10) with two false comments; the differs-from-chain update arm was verified as the real #993 fix. R2 closed R1 + added a route test suite (red-proofed against main), but caught a cast my red-proof restore had resurrected. R3 PASS at 592b9676 — full suite 2846/2846 by the reviewer's own run. Non-blocking note filed in review: omission→Option::None seam covered by inspection only.

PR #1042, opened 2026-08-17
fix(lessons): topic chips show skills.json labels; discussion fetches on expand

### LINKED ISSUE #952: Lesson pane: topic chips render raw slugs instead of skills.json labels
From the gate review of #948 (M3, medium-low). getLessonSkills returns slugs and the chips render them verbatim — learners see 'program-errors', 'account-model' in every locale. skills.json already maps slugs to display labels; one lookup at the render site fixes it. Also fold in L5 (perf): keepMounted makes useThreads fire an uncached /api/community/threads?limit=20 on first paint of EVERY lesson page purely to label a collapsed community row — fetch on expand instead.

thomgabriel:
Closes #952. Chips rendered `getLessonSkills` slugs verbatim ('account-model' in every locale); the skills.json vocabulary now maps them at the query layer, unknown slugs falling back to the slug. And the discussion section's unconditional `keepMounted` made `useThreads` fire an uncached `/api/community/threads` on every lesson first paint just to label a collapsed row — it now mounts on first expand, held mounted only while the composer is open so collapsing can't eat a draft. Tradeoff accepted per the issue: the collapsed row shows no thread count until first expand.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/lesson-client.tsx
@@ -604,7 +604,12 @@ export function LessonPageClient({
         count={threadCount}
         open={isDiscussionListOpen}
         onOpenChange={setIsDiscussionListOpen}
-        keepMounted
+        // Mount on first expand, not first paint: an unconditional keepMounted
+        // made useThreads fire an uncached /api/community/threads on EVERY
+        // lesson load purely to label this collapsed row (#952). The one state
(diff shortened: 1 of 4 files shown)


thomgabriel:
Gate record: independent adversarial review, two rounds. R1 PASS-with-must-fix — literal-typed skills map was a latent build break on two valid content states (label-less vocabulary entry, empty skills.json), LessonSection docstrings contradicted the change, label dedupe missing. R2 confirmed at 6ff509db (reviewer re-ran the type-break fixtures: both exit 0). Count-at-rest drop verified as prescribed by #952 itself; stale-count-after-collapse + refetch-per-toggle nuances surfaced to the owner separately. claude-review remains walled ($0 refusals); this review is the record.

PR #1044, opened 2026-08-17
fix(lint): widen restricted-import globs past single-star; gate dashboard toasts on active

### LINKED ISSUE #976: Widen no-restricted-imports globs: solana + content/generated bans miss relative and nested forms
Follow-up from the #975 gate review (finding 1).

The `no-restricted-imports` guard added in #975 (`apps/web/.eslintrc.json`, `src/lib/gamification/**` override) blocks `@/lib/solana/x` but NOT:
- relative forms: `../solana/pda`
- nested paths: `@/lib/solana/idl/superteam_academy.json` (single `*` doesn't cross `/`; `lib/solana/idl/` and `lib/solana/hooks/` are real dirs)

The pre-existing `src/content/generated` ban has the identical limitation.

Fix: widen both pattern lists to cover `**` and relative segments, e.g. `["@/lib/solana/*", "@/lib/solana/**", "**/lib/solana/**", "../solana/**", "../../lib/solana/**"]` (and the analogous set for content/generated), everywhere the rule appears — note ESLint overrides REPLACE rule options, so the widened list must be repeated in each override that redefines the rule.

Also from the same review (finding 4, optional in same change): in `apps/web/src/hooks/use-dashboard-data.ts` the toast dispatches (`dispatchXpGain`/`celebrate`/`dispatchQuestReward`/`dispatchSurpriseBonus`) are not gated by the `active` cancellation flag — harmless today (session-wide dedupe prevents double-toasts) but inconsistent with the cancellation discipline added in #975.

thomgabriel:
Closes #976. Single `*` doesn't cross `/`, so `@/lib/solana/idl/…` and relative `../../solana/pda` escaped both the content-generated and lib-solana bans; the widened `**`-form groups catch all three escape shapes in a lint fixture while `@solana/*` packages stay unmatched (`@solana` is a different path segment than `solana`). Folds in the same review's finding 4: the dashboard quest/bonus toast dispatches now respect the fetch's `active` cancellation flag.

### DIFF
--- a/apps/web/.eslintrc.json
+++ b/apps/web/.eslintrc.json
@@ -27,7 +27,7 @@
       {
         "patterns": [
           {
-            "group": ["@/content/generated/*", "**/content/generated/*"],
+            "group": ["@/content/generated/**", "**/content/generated/**"],
             "message": "The generated content bundle holds quiz answers, code solutions and hidden tests. Import it ONLY through src/lib/content/store.ts (server-only) — never directly, or those secrets leak into the client bundle."
           }
         ]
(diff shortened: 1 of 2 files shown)


thomgabriel:
Gate record: claude-review still walled (2s $0 refusal). Verification of record: 3-form escape fixture errors 3x under the widened globs (relative depth-2 solana, nested idl path, relative generated) with zero hits on @solana/* consumers, and the repo-wide lint in CI passing green is the exhaustive false-positive check. Hook change covered by 25/25 hook tests.

PR #1045, opened 2026-08-17
fix(review): dashboard badge counts what the session actually serves

### LINKED ISSUE #977: Review queue carries orphaned items from retired courses (bare '6 due for review', degraded session)
Found 04-08 while restyling the dashboard review card.

`review_items` for a real learner (wallet B7o8…) holds item_keys from RETIRED courses — `lesson-anchor-intro`, `lesson-anchor-setup-challenge`, `lesson-amm-concepts`, `lesson-defi-overview`, plus pre-rebuild `lesson-bfsp-*` ids that don't match the recreated C3 lesson ids. None resolve via `getLessonsByIds` (bundle only carries active courses), so:

- the dashboard review card shows a bare count with no lesson titles (owner: 'we don't know what is this');
- `/review` sessions will try to grade against lessons the bundle can't resolve (`resolveReviewItems`) — verify what actually happens: skipped? errored? empty session while the count still says 6?

Decide + implement one of:
1. Filter unresolvable item_keys out of `getDueReviewItems` (display-side fix, leaves dead rows accruing);
2. Garbage-collect review_items whose item_key no longer resolves at content-pin bump time (durable fix, needs a migration/script + admin resync hook);
3. Migrate item_keys where a retired lesson has a successor in the rebuilt course (best UX, needs a mapping that may not exist).

The count shown to the learner must equal the number of items the session can actually serve, whichever option lands.

thomgabriel:
Closes #977. Disposition: option 1 (resolution-side), hardened — instead of two surfaces each re-deriving resolvability with different rules, the new authed `/api/review/due` runs the exact `buildReviewSession` the /review page renders and the strip displays its answer, so count ≡ served by construction (orphaned item_keys from retired courses drop out of both identically). Dead rows still accrue in `review_items` as the issue anticipated for this option; garbage-collection at pin-bump time remains available as a follow-up if accrual ever matters.


thomgabriel:
Gate record: claude-review walled all evening ($0 refusals). Verification of record: new route is session-user-only (401 unauthenticated), takes no params, and returns only count + lesson titles — the session's quiz blocks never cross the boundary. 62/62 dashboard+review tests, full real CI green.

Issue #1047, opened 2026-08-17
No CI guard for i18n key parity — MISSING_MESSAGE is one typo away


thomgabriel:
Found during #1046: nothing in CI or the test suite compares the key structures of en.json / pt-BR.json / es.json, even though apps/web/CLAUDE.md declares identical structures a hard requirement (mismatch = runtime MISSING_MESSAGE). Current state is clean (1524 keys, verified identical by hand on 2026-08-17). Done-when: a test or CI step does a recursive key-set comparison across the three locale files and fails on any missing/extra key.

Issue #436, opened 2026-07-12
admin/status: Supabase failure 500s the whole admin screen while GitHub failure degrades gracefully


thomgabriel:
Asymmetry found across the SP3-C reviews.

- GitHub unreachable → content-drift degrades to an explicit `"unknown"` (logged), route still 200s. ✅
- Supabase unreachable → the deployments read throws → the **entire** `/api/admin/status` route 500s, taking down program liveness, authority-mismatch display, and everything else with it. ❌

**Fix:** degrade the deployments read per-row the same way (unknown state), so a DB blip doesn't blind the operator to on-chain state.

PR #1048, opened 2026-08-17
fix(landing): survive deployment-read failures; localized marketing error boundary

### LINKED ISSUE #931: Landing page hard-crashes on deployment-read failure; (marketing) group has no error.tsx
From the 2026-07-31 UI + build audit; reproduced locally (fetch-failed → whole landing renders the root error boundary). getCourseById (lib/content/queries.ts:313) calls getDeploymentById which throws on any Supabase error (lib/content/deployments.ts:176-179); resolveFlagshipLessonHref sits OUTSIDE page.tsx's try/catch. The graceful getDeploymentByIdSafe already exists (deployments.ts:226, built in #436) — adopt it in getCourseById. Also: app/[locale]/(marketing)/ has no error.tsx, so failures escalate to root app/error.tsx which renders hardcoded English to pt-BR/es visitors; add a localized boundary. Note claude-review on #928 believed resolveEntryLessonHref's sync-gate guard covers this — empirically it does not catch thrown fetch failures. Also fix root error page theming: app/global-error.tsx renders unthemed (white flash for dark-mode users) — add bg-bg text-text + data-theme. P1/P2.

thomgabriel:
`getCourseById` threw on any Supabase error and the landing page reaches it via `resolveFlagshipLessonHref`, which sits outside `page.tsx`'s try/catch — one DB blip took the whole front door to the root error boundary. It now reads through the existing `getDeploymentByIdSafe`; the row gates nothing (it only supplies `trackCollectionAddress`), so degrading cannot leak a hidden course, but the resulting null is ambiguous with "never synced", and `deploymentReadFailed` keeps the credential queue from reporting a transient outage as an operator error.

Also adds a `(marketing)/error.tsx` so a front-door failure stops falling through to the root boundary's hardcoded English — it renders inside the `[locale]` layout, so it reuses the existing `error` namespace like `(platform)/error.tsx` (no new message keys), and themes `global-error.tsx` off the tokens so dark-mode readers no longer get a white page.

Red-proof: both new `getCourseById` tests fail with `Error: connection refused` against the pre-change code.

Closes #931

### DIFF
--- a/apps/web/src/app/[locale]/(marketing)/error.tsx
+++ b/apps/web/src/app/[locale]/(marketing)/error.tsx
@@ -1,0 +1,51 @@
+"use client";
+
+import { useEffect } from "react";
+import Link from "next/link";
+import { useLocale, useTranslations } from "next-intl";
+import * as Sentry from "@sentry/nextjs";
+
+/**
+ * Marketing-group error boundary (#931). Without one, a landing-page failure
+ * escalated to the root `app/error.tsx`, which renders outside the `[locale]`
+ * layout and so hardcodes English — a PT-BR visitor hitting a transient error
+ * on the front door got an English page. This boundary renders inside that
+ * layout, where `NextIntlClientProvider` is mounted, so it uses the same
+ * `error` namespace as the platform boundary rather than inline translations.
+ */
+export default function MarketingError({
+  error,
+  reset,
+}: {
+  error: Error & { digest?: string };
+  reset: () => void;
+}) {
+  const t = useTranslations("error");
+  const locale = useLocale();
+
+  useEffect(() => {
+    Sentry.captureException(error);
+  }, [error]);
+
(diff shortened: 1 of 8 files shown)


thomgabriel:
Gate record: builder + separate gate diff-read. Verified: safe-read swap leaks nothing (getCourseById gates nothing — collection only), deploymentReadFailed mirrors the #436 contract so the mint queue distinguishes outage from never-synced, marketing boundary reuses the existing error namespace (zero new keys — #1047 risk sidestepped), global-error theme read pinned to next-themes' configured defaults with storage-throw guard. Red-proof recorded in the PR (2 tests fail at pre-change head with the exact landing throw). Follow-ups filed: #1049 (sponsor missing await — real dead 404 guard), #1050 (landing slug read). claude-review walled; this is the review of record.

PR #1051, opened 2026-08-17
fix(enroll): await the unknown-course guard — a Promise is always truthy

### LINKED ISSUE #1049: enroll/sponsor: missing await makes the unknown-course 404 guard dead code
Found during the #1048 gate: apps/web/src/app/api/enroll/sponsor/route.ts:68 has `if (!getCourseById(courseId))` with no await — a Promise is always truthy, so the 404 guard never fires and unknown course ids fall through to the enroll build. Its test passes only because the mock is synchronous (route.test.ts:98 mockReturnValue instead of mockResolvedValue). Done-when: awaited, and the test uses the async mock shape so the regression can't hide again.

thomgabriel:
Closes #1049. `if (!getCourseById(courseId))` never 404'd; awaited now, and the test mocks are async-shaped (red against the unawaited route: the unknown-course test fails).

### DIFF
--- a/apps/web/src/app/api/enroll/sponsor/__tests__/route.test.ts
+++ b/apps/web/src/app/api/enroll/sponsor/__tests__/route.test.ts
@@ -72,7 +72,7 @@ beforeEach(() => {
   h.isRateLimited.mockResolvedValue(false);
   h.isPlatformFrozen.mockResolvedValue(false);
   h.isCourseInMaintenance.mockResolvedValue(false);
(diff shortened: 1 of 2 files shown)


thomgabriel:
Gate record: one-word fix + async-shaped mocks, red-proofed against the unawaited route (unknown-course test fails there). Sponsored-enroll lane: the guard change is strictly fail-closed — an unknown course now 404s instead of proceeding to the enroll build. claude-review walled; substituted verification recorded.

PR #1052, opened 2026-08-17
perf(landing): resolve entry-lesson slugs from the bundle, not Supabase

### LINKED ISSUE #1050: Landing pays a service-role deployment read for a slug it discards
From #1048: resolveEntryLessonHref only uses course.slug, but reaches it through getCourseById, which does a service-role Supabase deployment read per landing render for trackCollectionAddress — a field that caller discards. A bundle-only getCourseSlugById would remove the dependency instead of just surviving its failures (#931 made it survive). Done-when: the landing path does zero Supabase reads for slug resolution.

thomgabriel:
Closes #1050. New ungated `getCourseSlugById` (bundle map read, documented as never-for-visibility); `resolveEntryLessonHref` uses it, keeping the F1 sync gate exactly where it was (`getCourseLessons` returns empty for unsynced). Landing now does zero Supabase reads for slug resolution.

### DIFF
--- a/apps/web/src/app/[locale]/(marketing)/start/__tests__/routes.test.ts
+++ b/apps/web/src/app/[locale]/(marketing)/start/__tests__/routes.test.ts
@@ -1,10 +1,10 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { getCourseById, getCourseLessons } from "@/lib/content/queries";
+import { getCourseLessons, getCourseSlugById } from "@/lib/content/queries";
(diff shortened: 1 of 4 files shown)


thomgabriel:
Gate record: bundle-map read replaces a discarded-field Supabase round-trip; F1 sync gate unmoved (entry-lesson + funnel suites both pass against the new shape — the funnel suite was the consumer my local run missed and CI caught, fixed in 1ed8196f). claude-review walled; substituted verification recorded.

PR #1053, opened 2026-08-17
feat(auth): refuse unlinking a wallet-first account's last OAuth identity

thomgabriel:
Owner decision 2026-08-17 (census-grounded): keep the ≥2 rule and permanent wallet; add the one missing guard — a synthetic-email account dropping its last OAuth identity is the sole unrecoverable transition (0 accounts there today, 68 exposed). Guard fires only when the targeted provider is the sole OAuth identity, so wrong-provider 400s aren't masked. Settings shows recovery context both directions (reassurance for real-email accounts, explanation for wallet-first), keys in all three locales (1527-key parity verified). Red-proof: the guard test fails against the old route.


thomgabriel:
Gate record, auth lane: independent adversarial review PASS (guard unbypassable via ordering; email predicate case-safe and anchored; wrong-provider 400 not masked; UI honest incl. null-email loading). Findings folded on the head: F2 double-click TOCTOU → shared isUnlinking flag; F4 AUTH-FLOWS unlink paragraph rewritten (it promised the now-forbidden transition). F3 filed as #1055 — for Google-sole wallet-first accounts the preserved identity is currently inert on Dynamic deployments (dynamic route resolves by email only); the guard still blocks the strictly-worse transition. F1 (GoTrue direct identity DELETE, self-inflicted only) recorded, no code. claude-review down (expired token, zero passes today); this review is the record.

PR #1054, opened 2026-08-17
chore(ci): apply the approved dependabot action bumps directly

thomgabriel:
Executes the owner's go on #1009–#1012 (the PRs were closed Aug 12 with dependabot silenced and their branches deleted — `@dependabot reopen` was ignored, so this applies the same bumps directly; every SHA cross-checked against dependabot's own PR diffs). pnpm/action-setup v4.3.0→v6.0.10, taiki-e/install-action v2.85.2→v2.85.10, Swatinem/rust-cache v2.9.1→v2.9.2, dorny/paths-filter v3.0.3→v4.0.3. This PR's own CI is the live firing for all four (they run in every PR's jobs). #1013 (claude-code-action) deferred until the token re-mint.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
       buildserver: ${{ steps.filter.outputs.buildserver }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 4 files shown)


thomgabriel:
Gate record: this PR's own green CI is the live firing for all four bumped actions (paths-filter v4 + pnpm-setup v6 both ran in every job of this run); SHAs cross-checked against dependabot's own diffs. claude-review dead (expired token).

PR #1056, opened 2026-08-18
feat(auth): route GitHub sign-in through Dynamic when configured

thomgabriel:
Mirror of the Google lane (#1029) for GitHub, now that GitHub login is enabled in the Dynamic dashboard: DynamicGithubSignIn swaps in on dynamicEnabled, Supabase OAuth stays as the kill-switch fallback, return leg unchanged (handler verified provider-agnostic). Bridge-failure copy made provider-neutral since the key now serves both lanes. 4 new tests; zero i18n keys added.

**Draft until a real GitHub login E2E** — the live Dynamic GitHub credential shape is unverified; if it lacks the email field (as Google's live shape did), the bridge 403s and the fix is a server-side GitHub email-trust decision, not this PR. Dynamic requests `user:email` scope, so odds are good, but the E2E decides.


thomgabriel:
Gate note: delta since the clean review is the review's own doc-line fix (dynamic-google-sign-in → dynamic-social-sign-in in AUTH-FLOWS.md) — verified by diff, docs only.

PR #1058, opened 2026-08-18
fix(ui): card-owned course titles, mobile thread rows, neutral signing copy

thomgabriel:
Three owner-reported polish items: course cards render the title as an overlay on the banner (banner art ships text-free — the BTC and Speedrun banners still carry baked-in titles and need text-free re-exports in academy-courses, tracked separately); thread-list rows on mobile crushed the title to zero width (meta was shrink-0, title the only shrinkable item — a flex-basis makes meta wrap to a second row instead); auth.signingIn said 'with your wallet' on Google returns.

Verified: 75 component tests green, tsc/eslint/prettier clean.


thomgabriel:
Added: achievements strip left-aligned (owner call, 18-08) — one-line justify-content flip, drag-scroll unaffected.

PR #1063, opened 2026-08-18
feat(auth): adopt the provider photo on the Dynamic bridge

thomgabriel:
The bridge never set `profiles.avatar_url`, so learners whose only way in is Dynamic kept the placeholder forever. The winning credential in `resolveVerifiedEmail` now surfaces `oauth_account_photos[0]` (https-only — it lands in an img src), and post-mint the route applies the legacy callback's exact rule: refresh rotating provider URLs, never overwrite a Supabase-storage upload. Red-proof: adoption + refresh tests fail against the pre-change route.

### DIFF
--- a/apps/web/src/app/api/auth/dynamic/__tests__/route.test.ts
+++ b/apps/web/src/app/api/auth/dynamic/__tests__/route.test.ts
@@ -1361,3 +1361,105 @@ describe("POST /api/auth/dynamic — subject rung (#1055)", () => {
     );
   });
 });
+
+describe("POST /api/auth/dynamic — avatar adoption on the bridge", () => {
+  const PHOTO = "https://lh3.googleusercontent.com/a/photo=s96-c";
+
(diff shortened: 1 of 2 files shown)


thomgabriel:
Routing record (auth lane): built in the other orchestrator session; adversarially reviewed here by an independent context after the original reviewer died mid-run. Verdict PASS — https-only gate backed by React-escaped sinks, CSP img-src allowlist, and next/image remotePatterns (all three checked); isCustomUpload rule byte-identical to the legacy callback; avatar write keyed on the already-resolved userId, so the subject rung adds no new surface. Red-proof replayed: 2 adoption/refresh tests fail against the pre-change route, 62/62 after restore. The one CI failure was a test/env coupling (hardcoded storage host vs CI's real NEXT_PUBLIC_SUPABASE_URL) — fixed in 0927fdf3, verified under both env variants.

PR #1069, opened 2026-08-18
fix(lesson): derive completion gate from BLOCK_REGISTRY

### LINKED ISSUE #970: Lesson completion gate: derive gateBlocks from BLOCK_REGISTRY (parsons is graded+required but ungated client-side)
From the gate review of #969 (finding 5, latent). lesson-client.tsx hardcodes _type === 'quiz' || _type === 'openEnded' for the client completion gate, while the server derives from BLOCK_REGISTRY where parsons is graded:true, required:true — and ParsonsBlock never calls setBlockDone. A non-code lesson with a parsons block would show an enabled Mark Complete that the server 403s. Latent today (zero parsons blocks in the compiled bundle). Fix: derive gateBlocks from BLOCK_REGISTRY[type].required so client and server can never disagree.

thomgabriel:
The lesson client hardcoded which block types gate Mark Complete (quiz/openEnded) while the server gates on BLOCK_REGISTRY, where parsons is graded — so a parsons-only lesson would show an enabled button the server 403s. The gate set now comes from a shared registry-derived module (graded ∪ required, matching the server's two gate loops exactly).

ParsonsBlock is fully renderable and completable in the UI (Check judges the arrangement client-side), it just never reported done — it now calls setBlockDone on a correct Check and retracts it when the learner edits afterward. A new test pins the client gate set to the registry-derived set so a future block type can't drift silently.

apps/web: eslint clean on touched files (one pre-existing warning), typecheck clean, full vitest suite green (292 files / 2873 tests).

Closes #970

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/lessons/[id]/blocks/__tests__/parsons-block.test.tsx
@@ -213,6 +213,37 @@ describe("ParsonsBlock — arrange interaction", () => {
   });
 });
 
+describe("ParsonsBlock — completion-gate report (#970)", () => {
+  it("reports done after a correct Check", () => {
+    const ctx = makeCtx();
+    renderWithIntl(<ParsonsBlock block={parsonsBlock} ctx={ctx} />);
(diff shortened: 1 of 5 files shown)


thomgabriel:
Gate note: the substantive review covered 804b57d5; the only delta since (f3ad9a4d) is the review's own suggested doc comment in gate-blocks.ts — verified by diff, no code change.

PR #1079, opened 2026-08-18
feat(lesson): anatomy-matched skeletons for the challenge screen

thomgabriel:
The dynamic ChallengeInterface fallback was a flat 32rem slab, so the challenge screen flashed in from nothing. ChallengeSkeleton mirrors the mounted two-card anatomy — rail card with jump-chip pills, prose lines, and the three disclosure rows; editor card with toolbar strip, line-numbered gutter, resizer, and the 120px output strip — at the same lg dimensions and mobile stacking order, so nothing jumps on mount. Shimmer already respects prefers-reduced-motion via the global reduce blanket. Part of #942.

Render test asserts the toolbar/editor/output regions and reserved heights; web suite 2929 tests green, typecheck and lint clean.


thomgabriel:
@claude review the delta — both blocking findings applied: lg:mt-auto removed from the disclosure rows (normal flow like the real component) and an order-4 AI-pane placeholder now reserves that region.

PR #1084, opened 2026-08-19
fix(analytics): pre-init queue, loud Sentry init, identity reset — audit follow-ups

thomgabriel:
Follow-ups from today's analytics audit (builds on #1083).

What was being lost: PostHog init is an async dynamic import, but the first pageview and first-render events fire synchronously and hit the `if (!posthog) return` no-op — that's the 53-pageleave/14-pageview gap. Events are now buffered pre-init (cap 50, drop-oldest) and flushed in order, with a single console.warn when PostHog is unconfigured or failed to load. A malformed Sentry DSN also failed silently in prod (debug is stripped); all three inits now check `Sentry.getClient()` and console.error when a DSN was provided but no client came up, and `sentry.ts` gates on `Sentry.isInitialized()` so the dev-console fallback works with a bad DSN. And `resetPostHogUser` had zero callers, so identity bled across sign-outs — sign-out and account-delete now call a `resetUser()` facade (PostHog reset + `Sentry.setUser(null)`).

Also: `$current_url` now carries the full URL (was a bare pathname; GA4's `page_path` unchanged), the GA4 init no longer skips stub+config when a foreign googletagmanager tag exists (script injection still deduped), and the analytics provider's auth subscription survives a StrictMode remount (split from the ref-guarded init effect).

Red-proof: the new `posthog.test.ts` run against main fails 5/7 — including "event captured before init resolves is delivered after init". Against this branch, all 7 pass; full `src/lib/analytics` + `src/components` suites are green (64 files / 457 tests), eslint and `tsc --noEmit` clean.

### DIFF
--- a/apps/web/sentry.edge.config.ts
+++ b/apps/web/sentry.edge.config.ts
@@ -15,4 +15,13 @@ if (dsn) {
     environment: process.env.NODE_ENV,
     tracesSampleRate: process.env.NODE_ENV === "development" ? 1.0 : 0.1,
   });
+
+  // A malformed DSN makes `init` bail without a client — and with `debug`
+  // statements stripped from production bundles, that failure is otherwise
+  // invisible. Shout, so a broken error pipeline is caught at boot.
+  if (!Sentry.getClient()) {
+    console.error(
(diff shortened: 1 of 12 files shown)


thomgabriel:
Gate note: delta since the clean review is the README paragraph it asked for (pre-init buffer + single-warn semantics) — docs only.

PR #1086, opened 2026-08-19
feat(admin): session-based admin auth via admin_users allowlist

thomgabriel:
Replaces the ADMIN_SECRET/HMAC-cookie admin system with the caller's own Supabase session checked against the `admin_users` allowlist (migration `20260819170000_admin_users.sql`, already applied + ledgered on prod with 4 admin rows; RLS on, zero policies, service-role read only).

`requireAdmin()` in `lib/admin/auth.ts` is the single gate: session user must have an `admin_users` row, fail closed on any error (no session, DB error, missing env → not admin). All 14 `/api/admin` routes keep their existing 401 contract and same-origin CSRF check; the destructive recreate route keeps its confirm/preflight machinery and now logs the acting admin's user id (sync does too). `/admin` pages: no session → middleware redirects to the localized landing; session but not on the allowlist → 404, so the panel is never revealed. The password login form, `/api/admin/auth`, the Edge+Node HMAC verifiers and every ADMIN_SECRET reference are deleted — ADMIN_SECRET can be removed from Vercel after this merges.

New: `GET /api/admin/me` (`{ admin: boolean }`, never 500s) drives a cosmetic Admin entry in the user menu for allowlisted users.

Tests: requireAdmin unit suite, /api/admin/me contract, middleware admin redirects, admin routing (404 for non-admins), user-menu entry, plus the existing admin route suites — all green; tsc and eslint clean (only pre-existing warnings).


thomgabriel:
@claude review the current head (1ba06b0e) — delta since your run on b23686c6 is the rebase over main (user-menu import conflict: fetchIsAdmin + resetUser both kept) plus the doc/config purge you can diff.

thomgabriel:
@claude review the current head (2d548bc5) — delta since your 1ba06b0e review is the rebase over today's main (#1102 middleware getClaims landed there: the doc conflict kept the new matcher + session-based admin wording, the admin gate now checks the getClaims-derived userId, and the new admin middleware test drives the getClaims mock). Middleware 23/23 + admin suites 280/280 green locally.

thomgabriel:
Routing record (auth lane): session-based admin auth touches the auth/admin surface, so it went through the independent adversarial pass in addition to CI + claude-review. Adversarial result: PASS — red-proof by targeted mutation (weakening requireAdmin's fail-closed branch made the probe succeed, confirming the test would catch it). Findings dispositioned: service-token alternative for init-program sync filed as #1087; stale ADMIN_SECRET references purged in 5ff1c6ad; layout/page parallel-render caveat documented in-code. Post-rebase head 2d548bc5 re-reviewed clean (getClaims userId repair only). Merging on green CI + clean review of the current head.

PR #1088, opened 2026-08-19
fix(landing): seamless achievement marquee at any viewport

thomgabriel:
The achievements strip visibly ran out on wide monitors — one 2.5k-px strip translating -25% leaves blank space on the right before each reset (owner repro'd on an ultrawide). Restructured to the standard two-identical-groups marquee, each animating -100% of its own width: seamless at any viewport, hover-pause preserved, second group aria-hidden.


thomgabriel:
@claude re-review the current head (f46c85a2) — the prior blocking finding (stray _exploit-repro.test.ts) was removed; the diff is now only the marquee TSX + globals.css.

thomgabriel:
Gate note: delta since the reviewed head f46c85a2 is 39197d5f — CSS-only, implements this review's own non-blocking finding (duration 40s→160s so one 4-set-width group keeps the pre-PR px/s). No new surface; merging on the standing suggested-delta precedent.

PR #1099, opened 2026-08-19
perf: landing stats via get_platform_stats() RPC (#1091)

### LINKED ISSUE #1091: perf: landing page scans all of public_user_xp per request
`(marketing)/page.tsx:37` selects `total_xp` for every row of `public_user_xp` and sums in JS — one row per user, per visitor, growing forever. Its `revalidate = 300` is currently inert (whole app renders dynamically), so this runs on every landing view. Two other stat queries run in a separate sequential wave.

Fix (researched): one `get_platform_stats()` SECURITY DEFINER function (STABLE, `SET search_path=''`, GRANT service_role only — mirror `get_leaderboard`) returning `{total_xp, builders, credentials}`; landing calls it once. PostgREST aggregates stay disabled (enabling is platform-wide risk for one stat). Collapses 3 round trips + a scan into 1 indexed call.

Migration discipline: follow docs/DB-MIGRATION-LEDGER.md — repo migration file + ledger mapping when applied via MCP.

Done when: landing does one RPC for stats, numbers match the old computation, migration in repo + applied + ledger updated.

thomgabriel:
Landing page was selecting every `public_user_xp` row and summing in JS, plus two more count queries in a second wave — per request. Now one `get_platform_stats()` SECURITY DEFINER RPC (aggregates in SQL, service_role-only, mirrors get_leaderboard's hardening) with a fallback to the old three-query path if the RPC errors, so deploy order can't break the landing.

Migration `20260819180000_get_platform_stats.sql` is repo-only — apply via MCP + extend the ledger per docs/DB-MIGRATION-LEDGER.md before relying on the fast path. pglite execution test asserts old-vs-new number parity (seeds real rows).

Closes #1091


thomgabriel:
Adversarial review (DB lane): PASS. Should-fix applied in 7e39dbce — profiles stub gains deleted_at and a soft-deleted seed, pinned values now document the stale-view inclusion so #1105 can't land without flipping them. schema.sql drift filed as #1105; builders-counts-deleted product question left as-is (parity was the goal). Note for whoever watches the deploy: totalXpMinted may jump upward when the RPC goes live — the old unbounded select was subject to PostgREST row caps; the jump is the fix, not a bug.

PR #1101, opened 2026-08-19
perf: cut Anchor + IDLs + web3.js from the shared layout chunk (#1090)

### LINKED ISSUE #1090: perf: Header ships Anchor + both IDLs + web3.js to all 119 routes (124KB layout chunk)
The `[locale]/layout` client chunk is 124KB and contains both program IDLs inline, on every route including anonymous marketing pages. Two chains: Header → useXpTotal → academy-reads (Anchor BorshCoder + 2 IDL JSONs + spl-token + web3.js), and LowSolBanner → web3.js for `LAMPORTS_PER_SOL` alone. Measured First Load JS: lesson 869kB / course 817kB / landing 647kB.

LowSolBanner is also a live bug: its 30s `getBalance` poll starts before the devnet-only render guard, so mainnet users poll the RPC forever for a banner that renders null (low-sol-banner.tsx:33-42).

Fix (researched 2026-08-19, verified against built chunks):
- XP badge reads `user_xp` from Supabase via the existing AuthProvider userId — the header wants a user total, not a wallet read. Falls back gracefully when signed out. (Keeping it on-chain would need next/dynamic ssr:false; DB is simpler and drops 4 RPC round-trips/page too.)
- Inline `1_000_000_000` in LowSolBanner (verified: web3.js exports it as a plain const) and move the devnet/publicKey guard into the effect.
- After: confirm via build that no `"use client"` module reaches academy-reads; the IDL/server-only split is likely moot then.

Blocks provider demotion (separate issue): marketing routes must not call wallet hooks before providers can move down to (platform).

Done when: layout chunk drops ~124KB (verify in .next/static/chunks), no IDL bytes in any shared chunk, mainnet makes zero getBalance polls, XP badge behavior unchanged for signed-in users.

thomgabriel:
The Header's `useXpTotal` chain pulled Anchor's BorshCoder, both program IDL JSONs, spl-token, and web3.js into the `[locale]/layout` client chunk — 124KB on every route for a badge whose primary source was already Supabase `user_xp`. The on-chain read was only a supplementary max() input; removing it changes nothing visible (the badge still never moves down, and the xp-gain event path is untouched).

LowSolBanner kept web3.js in the chunk a second way for one constant, now inlined — and its 30s `getBalance` poll started before the devnet-only render guard, so mainnet wallets polled forever for a banner that renders null. The guard now lives in the effect (regression test added).

Measured: layout chunk 124KB → 45.9KB; the IDL-bearing chunk now loads only on the 5 routes that use client-side on-chain reads (course, lesson, dashboard, teach previews) instead of all 119.

Closes #1090


thomgabriel:
Gate record: claude-review check passed but the bot stalled mid-review (checklist never completed). Substituting per protocol: line-by-line review done in-session (diff is a pure deletion of the supplementary on-chain XP source — header/sidebar already read user_xp from Supabase), build-verified layout chunk 124KB→45.9KB with IDL chunk demoted to 5 on-chain routes, banner poll regression test green, full CI green. Merging on that record.

PR #1102, opened 2026-08-19
perf: middleware getClaims + drop per-nav tombstone query (#1089)

### LINKED ISSUE #1089: perf: middleware pays 2 network round-trips per authed navigation
Every logged-in navigation runs `supabase.auth.getUser()` (HTTP to Supabase Auth) then a serial `isAccountDeleted` profiles query in middleware (src/middleware.ts:138,146), including on hover-prefetches. Six pages then repeat `getUser()` server-side. The matcher's `.*\..*` also skips middleware for any page slug containing a dot (CSP + locale lost).

Research (2026-08-19):
- `getClaims()` is Supabase's current documented middleware pattern — local WebCrypto verify against cached JWKS; falls back to getUser on HS256, so the swap is safe unconditionally. Prod already publishes an ES256 key (confirm it's the CURRENT signing key in Dashboard → JWT Keys before claiming the win).
- The tombstone query is largely redundant: account deletion already does a global signOut (revokes all refresh tokens); `deleted_at` has one writer. Drop it from middleware, keep checks at existing chokepoints; guarantee stays ≤ JWT expiry. Also removes SUPABASE_SERVICE_ROLE_KEY from the Edge bundle.
- Duplicate page getUser(): React `cache()`-wrapped DAL helper (Next's documented pattern); don't pass identity via headers.
- Matcher: extension-anchored regex per Next/Supabase docs.

Done when: middleware makes zero network calls for a valid asymmetric-key session, the six pages share one cached claims read, dotted slugs hit middleware, and the auth-lane adversarial review passes (expired/deleted-session redirects unchanged).

thomgabriel:
Logged-in navigations paid two network round trips in middleware before rendering: `getUser()` (HTTP to Supabase Auth) then a serial `isAccountDeleted` profiles query — on every request including hover-prefetches. Now `getClaims()` verifies the JWT locally against the cached JWKS (Supabase's current documented middleware pattern; falls back to getUser on HS256 so behavior is safe on either key type), and the tombstone query is gone from the hot path.

Dropping the per-request tombstone check is sound because every `deleted_at` writer now pairs with session revocation: the delete route already global-signs-out, and the wallet-shell merge — found during this work to tombstone shells *without* revoking their live sessions — now permanently bans the shell (supabase-js has no revoke-by-id; a banned user dies at next token refresh, same ≤JWT-expiry bound as deletion). Login chokepoints still refuse deleted accounts.

Also: new `lib/auth/dal.ts` (React cache()-wrapped claims read) replaces six pages' duplicate `getUser()` calls, and the matcher no longer skips middleware for page slugs containing dots (`/en/courses/node.js-basics` previously lost CSP + locale handling).

Note for the win to fully land: confirm in Supabase Dashboard → JWT Keys that the ES256 key is the current signing key (JWKS already publishes one); on legacy HS256 the middleware silently keeps today's getUser behavior.

tsc clean; middleware + dynamic-route suites 86/86; page/auth suites 264/264.

Closes #1089

### DIFF
--- a/apps/web/CLAUDE.md
+++ b/apps/web/CLAUDE.md
@@ -6,13 +6,15 @@ API route reference lives in `src/app/api/CLAUDE.md` (loads when you work on rou
 
 The middleware (`src/middleware.ts`) chains two concerns:
 
-1. **Supabase auth**: Creates server client, calls `getUser()` (may refresh tokens)
+1. **Supabase auth**: Creates server client, calls `getClaims()` — local JWT verify against cached JWKS (no network for asymmetric keys; falls back to `getUser()` on HS256), refreshing the session when needed. Only `claims.sub` is used. No per-request DB queries: soft-deleted accounts are refused at the login chokepoints (`/api/auth/callback`, `/api/auth/wallet`, `/api/auth/dynamic`), and every `profiles.deleted_at` writer pairs with session revocation, so a stale session dies at its next token refresh (bounded by access-token expiry).
 2. **next-intl**: Adds locale prefix to all routes (default: `en`)
+
+Server components share one per-request claims read via `getAuthClaims()` (`lib/auth/dal.ts`, React `cache()`); call `supabase.auth.getUser()` directly only when the full user object is needed.
 
 **Auth-gated routes** (redirect to landing if unauthenticated): `/dashboard`, `/settings`, `/profile` (exact — own profile only)
 **Public routes** (no auth required): `/` (landing), `/courses`, `/leaderboard`, `/community`, `/certificates`, `/profile/[username]`
 **Admin routes**: Checked against HMAC-signed `admin_session` cookie (separate from Supabase auth). Sub-routes redirect to `/admin` login form if cookie is absent or expired.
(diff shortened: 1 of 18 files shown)


thomgabriel:
Adversarial review (auth lane, claude[bot]-independent): FAIL → fixed in 34b62004. F1: delete route's signOut was silent best-effort — now checked, falls back to a permanent ban (shared revokeUserSessions helper, 3 retries), 500s if both fail. F2: merge-path ban now retries before the loud log. F3: .webmanifest (+ other asset extensions) restored to the matcher exclusion with a regression test — the PWA manifest was being 307'd into next-intl and 404ing. Non-blocking F4/F5 filed as #1104/#1103; F8 stale comment fixed. Re-verdict requested.

thomgabriel:
Adversarial re-verdict: PASS (34b62004). Nits applied on top in 4dba28a8: N1 ban runs unconditionally (signOut swallows 401/403 as success), N2 revocation gets its own catch with an honest label, N4 typed client. N3 (request timeout) and N5 (fake timers) noted, not taken. Merging on green CI.

PR #1107, opened 2026-08-19
perf: caching pack — CDN headers, cookieless public reads, request-level dedup (#1094)

### LINKED ISSUE #1094: perf: caching pack — public data served uncacheable
Grab-bag of researched caching fixes, each small and independent (platform facts verified: Vercel CDN refuses responses carrying Set-Cookie; plain s-maxage gets stripped without CDN-Cache-Control; "use cache" is not usable on stable 15.5 — stay on unstable_cache):

1. Leaderboard: cookieless anon client (copy deployments.ts pattern — get_leaderboard never reads auth.uid()), drop force-dynamic, `Cache-Control: public, s-maxage=60, stale-while-revalidate=300` + mirrored CDN-Cache-Control. Server page calls the RPC via unstable_cache directly, not its own API route.
2. certificates/metadata: cookieless client (nft_metadata has public SELECT), keep Cache-Control but drop to s-maxage=86400 unless metadata is provably never updated post-mint (immutable+1y is unrecallable).
3. /api/content/* param routes: `public, s-maxage=300, stale-while-revalidate=3600` + CDN-Cache-Control (cookie-free today, CDN-eligible). Param-less routes already segment-revalidate.
4. Wrap getActiveDeployments in React cache() outside the existing unstable_cache (documented layering; also memoizes the Map rebuild, 3-4 calls/render → 1).
5. next.config images: `formats: ['image/avif','image/webp']`, `minimumCacheTTL: 2592000` (content assets immutable per content.lock; short TTL multiplies Vercel image billing). Add `sizes` to responsive Image uses (currently zero repo-wide). KEEP `unoptimized` on lh3 Google avatars — optimizer adds cost, no gain.
6. AnalyticsProvider: consume useAuth() instead of its own getSession()+subscription (browser client is already a singleton — not a double-client bug).
7. next-intl plugin `experimental.messages.precompile: true` (verified available in installed 4.13.4), then re-measure; skip namespace-splitting unless the flight payload still matters (measured: messages are NOT in First Load JS).

Done when: leaderboard + content routes CDN-HIT on second request (verify x-vercel-cache), cert metadata cacheable, no behavior change anywhere.

thomgabriel:
Closes #1094.

Now CDN-cacheable (cookieless client where needed, `Cache-Control` + `CDN-Cache-Control` on 200s):
- `/api/leaderboard` (s-maxage=60, swr=300) — new shared `lib/leaderboard/global.ts` (unstable_cache 60s, tag `leaderboard`), also used by the leaderboard page instead of the cookie-bound service (same RPC, same shape)
- `/api/referrals/leaderboard` (s-maxage=60, swr=300) — already admin-client; RPC reads no auth.uid()
- `/api/certificates/metadata` (s-maxage=86400, was browser-immutable/1y — metadata is updatable post-mint)
- `/api/content/{courses,course-lessons,lessons-summary,recommended,is-instructor,instructor-wallet}` (s-maxage=300, swr=3600); instructor-wallet switched off the cookie client

**Dropped from the plan:** `/api/community/search` — it reads `auth.getUser()` + the caller's votes (`userVote` per thread), so a public CDN cache would serve one user's vote state to everyone. Left dynamic.

Also:
- `getActiveDeployments` wrapped in React `cache()` outside the existing `unstable_cache` (one Map build per render instead of 3-4)
- `createCookielessClient` extracted to `lib/supabase/cookieless.ts` (deployments.ts now imports it)
- images: avif/webp formats, `minimumCacheTTL` 30d, `sizes` on the course-card grid image (no `fill` images exist repo-wide); lh3 avatars stay `unoptimized`
- next-intl `experimental.messages.precompile` — the 4.13 types require `format`/`path`/`locales` alongside it, so all four are set
- AnalyticsProvider consumes `useAuth()` instead of its own `getSession()` + `onAuthStateChange` (it sits inside AuthProvider); identify still fires on session appearance and on any user object change, sign-out reset stays in the explicit sign-out flows

Tests: content routes suite asserts the new headers on 200 and their absence on 500; touched suites green (74 + 106 tests across content/deployments/leaderboard/referrals/i18n/landing), typecheck + lint clean. Build verified in CI per the one-build-slot rule.


thomgabriel:
Adversarial review (caching lane): PASS on the leak class → 4 findings fixed in b046028a. F1 cache-poisoned empty leaderboard (throw-inside-cache, uncached 500), F2 browser TTL restored on cert metadata, F3 season param canonicalised (regex rejects leading zeros too), F4 ids/exclude sorted+deduped for cache-key sharing. Header assertions added on all nine cached routes incl. no-header-on-error negatives. 198/198 across touched suites. Re-verdict requested on the delta.

thomgabriel:
Re-verdict PASS; N1/N2 applied in a40b5dc6 — RPC message rides in the throw, route catch logs via ERROR_IDS.LEADERBOARD_FETCH_FAILED, and a direct loadLeaderboard test pins reject-not-cacheable-[]. Merging on green CI.

thomgabriel:
Gate note: review of head a40b5dc6 is clean (no blocking findings; CDN-cache cross-user-leak surface explicitly verified against schema grants + cookieless clients). Vercel check still queued — preview is not a merge gate per owner rule. The route-table doc nit lands in a follow-up docs commit together with #1098's date nit.

Issue #1077, opened 2026-08-18
Sign-in surface polish: settings Dynamic-awareness, dedup handlers, one error style


thomgabriel:
Follow-ups from the 2026-08-18 sign-in flow audit (the bugs it found shipped with #1056's ee6c7da8). Remaining: (1) settings account-tab shows 'Not linked' for the very provider a Dynamic user signs in with daily — distinguish 'signed in with Google (via Dynamic)' from a Supabase-linked identity, and reconcile the ≥2-methods count presentation; (2) ~80 lines of copy-paired handlers in account-tab (link/unlink Google/GitHub, wallet-link effect) plus landing-client's hand-copied AuthModal spinner markup — parameterize; (3) the modal renders Supabase-fallback errors at modal level (text-sm) but Dynamic button errors inline (text-xs) — one style. Also fold the C3 note into AUTH-FLOWS.md: avatar adoption now exists on both rails (#1063), document it as the fourth shared ritual.

PR #1109, opened 2026-08-20
perf: mount wallet/Dynamic/gamification providers only on platform routes (#1097)

### LINKED ISSUE #1097: perf: wallet/Dynamic/Query provider stack mounts on marketing pages
SolanaWalletProvider (autoConnect on, Wallet Standard discovery), DynamicWalletProvider (+TanStack Query), and GamificationOverlays (canvas-confetti chain) all sit in [locale]/layout.tsx above the route groups — anonymous marketing visitors initialize all of it. (platform)/layout.tsx is an empty div.

Research verified (wallet-adapter 0.15.39 source): unmount never calls adapter.disconnect(), walletName persists in localStorage, remount takes the silent autoConnect branch — session continuity is safe; the app session is Supabase cookies anyway. Dynamic's SSR tree-shape constraint is orthogonal (platform layout is also server-rendered; client is a module singleton). Query rides along with Dynamic for free. Accept a brief connected:false flicker on marketing↔platform crossings; re-verify WalletAuthHandler's hasTriedAuth ref doesn't re-trigger SIWS on remount.

Move all three (+ their CSS import) into (platform)/layout.tsx. ReferralCapture stays global. Do NOT lazy-mount providers with ssr:false (hydration tree-shape hazard documented in lib/dynamic/client.ts:36).

Blocked by #1090 — Header must stop calling wallet hooks first or every marketing route throws.

Done when: marketing First Load JS drops (measure landing before/after), wallet reconnects silently on platform entry, no repeat SIWS prompt, no hydration mismatch warnings.

thomgabriel:
Closes #1097.

Moves SolanaWalletProvider, DynamicWalletProvider (+its QueryClient), and GamificationOverlays from `[locale]/layout.tsx` into `(platform)/layout.tsx`, same order and constraint comments. ToastContainer is split out of the overlays and stays global — marketing dispatches toasts too (AuthErrorToast on the landing page).

**Landing First Load JS: 647 kB → 315 kB** (baseline main at eb9a4725; platform routes unchanged — dashboard 483 kB, courses 283 kB on the rebased branch). Chunk-group verification: across all 17 marketing first-load chunks, zero hits for wallet-adapter, the Dynamic SDK, TanStack Query, ConnectionProvider, WalletModalProvider — including the `useWallet` default-context module that v1 still carried.

Header renders in `[locale]/layout.tsx` ABOVE the demoted stack, so React context can never reach its wallet consumers (adversarial review of v1). The bridge is a module-level **ambient store** (`lib/solana/ambient-wallet-store.ts`, SDK-free): a registrar inside SolanaWalletProvider publishes `{connected, publicKey, disconnect, openWalletModal}` on mount and on change, with an ownership-guarded unregister so a stale unmount never wipes a newer stack.

- **Sign-out (F1)**: UserMenu disconnects via the store when a stack is live, and always clears wallet-adapter's `walletName` localStorage key (verified default in `WalletProvider.js`) — so autoConnect + WalletAuthHandler cannot silently re-SIWS the account that just signed out, even from a marketing page. Dynamic logout is a dynamic import (static would drag the SDK back into the header chunk).
- **LowSolBanner (F1b)**: moved out of Header into `(platform)/layout.tsx` inside the stack, plain `useWallet` again; its test mocks reverted accordingly.
- **Double-mount guard (F2)**: the modal arms its scoped stack only when the store has no live registration; the scoped stack mounts as a **sibling** of the Dialog (F5 — no reparenting; the Dialog's tree position never changes), sticky after close for the wallet-select handoff, and disarmed on client-side navigation while closed. Test-covered: ambient-live ⇒ never arms; no stack ⇒ arms; sticky after close.
- **Capture window (F4)**: DynamicReturnCatcher sets a synchronous module flag the moment it decides to mount (before its chunk resolves); the modal won't arm and the trigger shows the signing-in state inside that window; any stack registration closes it.
- **Chunk failure (F3)**: both lazies sit under an error boundary that renders a translated in-modal error + retry; retry recreates the lazy components (a rejected `React.lazy` caches its rejection).
- The modal body needs no provider above it now: the Solana button opens the wallet-select modal through the store, and DynamicSocialSignIn reads the stale-session check imperatively via the SDK's `getCore` (same escape hatch `lib/dynamic/social.ts` uses) instead of `useUser`.

**Tradeoff:** the sign-in click on marketing/admin lazy-loads the modal body + provider stack — a brief spinner in the modal on first open.

**No repeat SIWS on platform re-entry:** crossing marketing→platform remounts SolanaWalletProvider and autoConnect silently reconnects, re-firing WalletAuthHandler's connect effect (`hasTriedAuth` is per-mount and does NOT guard this). What does: `authenticate()` calls `supabase.auth.getUser()` and returns before any prompt when a session exists (`wallet-auth-handler.tsx:39`). The ref only prevents double-trigger within a mount.

Tests: 3027 pass, including the ambient store (ownership, capture flag), scoped-vs-ambient/capture modal mounting, and the reworked DynamicSocialSignIn session check.

### DIFF
--- a/apps/web/src/app/[locale]/(marketing)/layout.tsx
+++ b/apps/web/src/app/[locale]/(marketing)/layout.tsx
@@ -1,4 +1,5 @@
 import type { Metadata } from "next";
+import { DynamicReturnCatcher } from "@/components/auth/dynamic-return-catcher";
 
 export const metadata: Metadata = {
   title: "Superteam Academy — Solana Developer Education",
@@ -11,5 +12,12 @@ export default function MarketingLayout({
 }: {
   children: React.ReactNode;
 }) {
-  return children;
+  // The catcher completes a Dynamic social/device redirect that lands on a
+  // marketing page, where the (platform) provider stack is absent (#1097).
+  return (
+    <>
+      <DynamicReturnCatcher />
+      {children}
+    </>
+  );
 }
--- a/apps/web/src/app/[locale]/(platform)/layout.tsx
+++ b/apps/web/src/app/[locale]/(platform)/layout.tsx
@@ -1,11 +1,41 @@
+import { SolanaWalletProvider } from "@/lib/solana/wallet-provider";
+import { DynamicWalletProvider } from "@/components/auth/dynamic-wallet-provider";
+import { GamificationOverlays } from "@/components/gamification/gamification-overlays";
+import { LowSolBanner } from "@/components/layout/low-sol-banner";
+
+/**
+ * (platform) routes carry the wallet/Dynamic provider stack and the
+ * gamification overlays (#1097); marketing/admin routes do not, and reach
+ * wallet features through AuthModal's lazily-mounted scoped stack instead.
+ *
+ * Order matters and mirrors what [locale]/layout.tsx had before the
+ * demotion: SolanaWalletProvider outside DynamicWalletProvider, both inside
+ * the global Auth/Analytics providers. Crossing marketing↔platform unmounts
+ * the stack; wallet-adapter keeps `walletName` in localStorage and never
+ * disconnects the adapter on unmount, so re-entry silently reconnects via
+ * autoConnect, and WalletAuthHandler's existing-session check keeps the
+ * reconnect from re-prompting SIWS.
+ */
 export default function PlatformLayout({
   children,
 }: {
   children: React.ReactNode;
 }) {
   return (
-    <div className="container px-4 pb-20 pt-6 sm:px-6 md:pt-8 lg:px-8 lg:pb-8">
-      {children}
-    </div>
+    <SolanaWalletProvider>
+      {/* DynamicAuthHandler is mounted by the provider itself, not here:
+          its hook throws outside a DynamicContextProvider, so it must never
+          be a sibling. */}
+      <DynamicWalletProvider>
+        {/* Moved out of the Header (#1097): the banner needs useWallet /
+            useConnection, which only resolve under this stack. Devnet-only,
+            learner-facing — marketing routes never showed it anyway. */}
+        <LowSolBanner />
+        <div className="container px-4 pb-20 pt-6 sm:px-6 md:pt-8 lg:px-8 lg:pb-8">
+          {children}
+        </div>
+        <GamificationOverlays />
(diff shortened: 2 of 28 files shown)


thomgabriel:
Coordination note: I started a rebase of this branch onto main (it was ~5 merges behind and conflicting) and backed off when it became clear someone else is actively working the same worktree implementing this PR's review findings. Nothing of mine was pushed — origin is still at aeac24d0 and the branch is untouched. The rebase is yours; I'm out of it.

Three things I verified before stopping that hold regardless of which auth-modal design wins, so they don't get lost:

onError is now required on DynamicSocialSignIn. #1098 made it required and lifted the error into the modal's single role="alert" block, but auth-modal-body.tsx was extracted from a copy predating that — both DynamicSocialSignIn call sites in the body need `onError={setErrorMessage}` or tsc fails.

The two "#1077 one error placement" tests query the modal body synchronously. Once the body is a lazy chunk they have to become findByRole, or they race the Suspense boundary.

admin/layout.tsx resolves without a new branch. #1086 deleted the render-children-without-chrome path, so the only conflict resolution is main's `requireAdmin()` + `notFound()` gate plus this branch's DynamicReturnCatcher inside the rendered admin tree. middleware.ts:112-125 redirects anonymous /admin to the localized landing before the layout runs, and reaching /admin at all requires a session while a signed-in non-admin gets notFound() — so there is no non-admin path left that needs the catcher.

thomgabriel:
Routing record (auth-flow lane, claude[bot]-independent adversarial gate): round 1 FAIL (5 findings — headline: Header sits above the demoted providers, so sign-out lost wallet-disconnect everywhere and the ambient guard was inert) → fixed at 58a851b5 via the ambient-wallet module store. Round 2 PASS conditional on NEW-1 (catcher chunk-failure = permanently stuck sign-in buttons) → closed at 4140a119 with the shared ChunkErrorBoundary + clearWalletReturnCapture, plus owner-keyed registrations (NEW-2/3) and the timing-argument comment. Measured: landing First Load 647→315 kB, platform flat, zero wallet/Dynamic/TanStack bytes in the marketing chunk group. Full suite 3033/3033. Merging on completed-green CI + final reviewer confirm.

PR #1110, opened 2026-08-20
docs: review follow-ups — avatar ruling date, CDN-cached route notes

thomgabriel:
The two non-blocking doc nits from the #1098 and #1107 reviews: AUTH-FLOWS.md's avatar-ruling date now matches the code (2026-08-18, per #1095), and the API route table marks the nine routes #1094 made CDN-cacheable.


thomgabriel:
Gate note: rebased onto main at 1abf2315 to pick up the instructor-wallet test fix — its earlier red Frontend check was that main-wide failure, not this PR. New head 6e908e75 is content-identical to the reviewed 052d4682 (pure rebase), and the review on that content was clean.

Issue #921, opened 2026-07-31
Wallet-placeholder email predicate is case-sensitive + no post-claim consent re-check (pre-existing, all email pipelines)


thomgabriel:
From adversarial review of #919 (F5+F6, nits; pre-existing shape in #779/#869/#896/#899). F5: u.email NOT LIKE '%@wallet.superteam-lms.local' is case-sensitive — an uppercase-domain placeholder would be claimed (not currently exploitable: wallet route + GoTrue lowercase). Fix once across all claim RPCs with lower(u.email) NOT LIKE. F6: consent/placeholder are not re-checked between claim and send — a learner who unsubscribes mid-flight still gets that one email; inherent to claim-then-send, documented, but worth one cheap re-check in the send loop.

thomgabriel:
Status check against prod, since most of this has landed.

F5 (case-sensitive placeholder gate) is done for the claim RPCs — verified live: `claim_due_session_reminders` and `claim_due_reengagement` both carry `lower(u.email) NOT LIKE` with the byte-wise form gone, applied as `20260820011023`… correction, `20260820003828`, via #1111. The only function on prod still matching byte-wise is `list_marketing_recipients`, which is tracked separately as #1118. `merge_wallet_shell_account` also matches byte-wise but is fail-closed — it raises when the email does NOT match, so case-sensitivity there refuses a merge rather than permitting one, and it is correctly out of scope.

F6 (consent re-checked between claim and send) is partly addressed: #1111 asserts consent at claim time inside the same statement as the ledger insert, so a flip between becoming due and being claimed now excludes the row. The cheap re-check in the send loop — for a flip between claim and send — is still open and is what remains of this issue.

Suggest narrowing this to F6 alone and letting #1118 carry the rest.

PR #1111, opened 2026-08-20
fix(email): claim RPCs re-check consent and exclude placeholder emails

### LINKED ISSUE #1075: Consent re-check + placeholder exclusion inside the email claim RPCs
Part 2 of #921 (part 1 = #1074's shared predicate). Consent is enforced only inside the claim RPCs' SQL — no send-time re-check exists in lib/email/{reminders,reengagement-send,campaign}.ts, and the RPCs' email predicate is case-sensitive (the issue's F5). Both fixes live in the same migration: widen the claim RPCs to exclude lower(u.email) LIKE the placeholder domain and to re-assert opt_in/reminder_opt_in at claim time, or return the consent columns for a send-time check with claim-release on mid-flight unsubscribe. Done when a mixed-case placeholder can't be claimed and a user who unsubscribes between claim and send is not mailed.

thomgabriel:
Two guarantees, both enforced in SQL inside the claim RPCs (the only consent gate the send pipelines trust):

1. **No placeholder spelling is ever claimed.** The wallet-placeholder exclusion was a byte-wise `u.email NOT LIKE '%@wallet.superteam-lms.local'`, so `AbC123@Wallet.Superteam-LMS.LOCAL` sailed through both RPCs (#921's F5). It is now `lower(u.email) NOT LIKE ...`, matching `isWalletPlaceholderEmail` from #1074. The domain literal lives in exactly one place per language — `WALLET_PLACEHOLDER_EMAIL_DOMAIN` in `apps/web/src/lib/auth/wallet-placeholder.ts` and the RPC bodies — cross-referenced in both directions.

2. **Consent is asserted at claim time.** `es.reminder_opt_in = true` sits inside the claim's `WHERE`, and the claim CTE plus the ledger `INSERT` are one statement, so a learner who unsubscribes between scheduling and the cron's claim is excluded and consumes no idempotency slot. This PR does not add that predicate — it pins it: the line is marked as the #1075 gate so a future body rewrite cannot silently drop it, and the PGlite suite now proves it behaviourally for both RPCs.

**Byte diff against the current definitions.** `claim_due_session_reminders` (from `20260804120000_recurring_lesson_plan.sql` §2) and `claim_due_reengagement` (from `20260731170000_reengagement_pipeline.sql` §3) are byte-identical except, in each body:

- one line changed: `AND u.email NOT LIKE '%@wallet.superteam-lms.local'` → `AND lower(u.email) NOT LIKE '%@wallet.superteam-lms.local'`
- its one-line comment expanded to four, cross-referencing the TS predicate
- three comment lines added above the existing `WHERE es.reminder_opt_in = true`

Nothing else moves: signatures and OUT columns, SECURITY DEFINER + `SET search_path = ''`, the service_role-only REVOKE/GRANT pairs, the per-(user, kind, São Paulo day) idempotency claim, the N=14 frequency cap and its advisory lock, the #980 `days[]` eligibility read, the #896 reminder-scoped token, and `ORDER BY d.uid` are carried forward unchanged. `list_marketing_recipients` is not a claim RPC and is untouched.

**TS senders need no change.** `lib/email/reminders.ts`, `reengagement-send.ts` and `campaign.ts` take recipients exclusively from the RPCs and never re-derive consent, so the fix is entirely in SQL.

**Tests.** New `apps/web/src/lib/supabase/__tests__/email-claim-consent-recheck-execution.test.ts` runs the real migration SQL in PGlite over the full prod lineage (18 tests). Its RED block replays the same fixtures *without* this migration and shows both RPCs claiming the mixed-case placeholder. GREEN proves: neither placeholder spelling is claimed by either RPC, a row whose `reminder_opt_in` flipped off after scheduling is excluded (and gets no ledger row), and an ordinary consenting learner is still claimed with the #896 token.

The migration awaits MCP apply to prod before this merges.

Closes #1075

### DIFF
--- a/apps/web/src/lib/auth/wallet-placeholder.ts
+++ b/apps/web/src/lib/auth/wallet-placeholder.ts
@@ -11,6 +11,12 @@
  * to know that).
  *
  * Importable from both server and client code — keep it dependency-free.
+ *
+ * SQL mirror (#1075): the email claim RPCs (`claim_due_session_reminders`,
+ * `claim_due_reengagement`) apply the same predicate as
+ * `lower(u.email) NOT LIKE '%@wallet.superteam-lms.local'` — see
+ * supabase/migrations/20260819190000_email_claim_consent_recheck.sql. A domain
+ * change must land in both languages.
  */
 
 /** Domain suffix of the synthetic wallet-auth placeholder email. */
--- a/apps/web/src/lib/supabase/__tests__/email-claim-consent-recheck-execution.test.ts
+++ b/apps/web/src/lib/supabase/__tests__/email-claim-consent-recheck-execution.test.ts
@@ -1,0 +1,443 @@
+// #1075: REAL SQL-execution tests for the claim RPCs' consent re-check and the
+// case-insensitive wallet-placeholder exclusion, replayed against the PROD
+// migration lineage.
+//
+// Lineage replayed here (repo filename → applied prod ledger version):
+//   20260728120000_email_subscriptions.sql        → 20260727200924
+//   20260731120000_reminder_consent.sql           → 20260731120914
+//   20260731150000_kind_scoped_unsubscribe.sql    → 20260731142948
+//   20260731170000_reengagement_pipeline.sql      → 20260731154032
+//   20260804120000_recurring_lesson_plan.sql      → 20260804190632   ← last applied
+//   20260819190000_email_claim_consent_recheck.sql → THIS CHANGE (not yet applied)
+//
+// RED-PROOF: the "pre-fix" block replays the identical fixtures WITHOUT this
+// migration and shows a MIXED-CASE placeholder being claimed by BOTH RPCs —
+// the byte-wise `NOT LIKE` only catches the lowercase spelling (#921 F5).
+// Reverting the migration turns the GREEN placeholder assertions into exactly
+// that failure.
+//
+// TIMEZONE (#956): date anchors are computed from
+// `(now() AT TIME ZONE 'America/Sao_Paulo')::date`, never from
+// `(now() - interval)::date`.
+//
+// @vitest-environment node
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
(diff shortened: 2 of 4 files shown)


thomgabriel:
Gate note: rebased onto main at 154c7c6c (new head d746879c). The earlier red Frontend check was the instructor-wallet failure that was live on main until #1113 — unrelated to this PR, and the rebase picks up the fix. Content is otherwise unchanged from the reviewed b1f1b713.

Migration is APPLIED to prod: `20260819190000_email_claim_consent_recheck.sql` went in as MCP version `20260820003828`, before this merges, per the apply-before-merge rule (#731). Ledger row is #1117.

thomgabriel:
Routing record (DB/email lane). Independent adversarial gate: PASS, no blocking findings.

The risk this PR carries is that CREATE OR REPLACE silently overwrites whatever is live, so a body carried forward from the wrong source would revert prod. Two independent checks closed that. Mine, before applying: reversing the single intended line in each migration body reproduced the LIVE prod definition's normalized hash exactly (ffcb5219f7480e34346de4a16949b29f and 7194f050df30f04ec5634d1f67dd4f77), proving the carry-forward matched what was actually running rather than just the source files. The gate's, independently: the migration lineage swept in timestamp order confirmed both named sources were the newest definitions of their functions and nothing later touched either, and a mechanical body diff showed exactly one differing code line each.

Gate also verified the predicate against isWalletPlaceholderEmail across 14 adversarial spellings with zero divergence, confirmed the consent gate sits in the first CTE so no row can be ledgered before it applies, confirmed SECURITY DEFINER + empty search_path + service_role-only grants, and red-proved both fixes by scratchpad mutation (5 failures each).

Findings dispositioned: F1 (list_marketing_recipients keeps the byte-wise gate, unreachable today) filed as #1118; F2 (ledger row) is #1117; F3 (PGlite parallel-suite contention) is pre-existing and already tracked by #922.

Applied to prod as 20260820003828 before merge, post-apply verified byte-equal to the repo bodies.

PR #1112, opened 2026-08-20
test: instructor-wallet suite mocks the cookieless client (#1094 follow-up)

thomgabriel:
#1107 was merged while its Frontend check was red: the route moved to the cookieless client but its test suite still mocked @/lib/supabase/server — 3 failures now on main. This repoints the mock (and adjusts for the sync client). Suite 4/4 locally.


thomgabriel:
Closing as a mis-created duplicate: this branch was cut at the same commit as #1109, so it carries the provider-demotion diff, not the instructor-wallet fix its title describes. The real fix for the red main is #1113; the provider work stays on #1109.

Issue #997, opened 2026-08-04
Teacher identity: custom display name + verified badge on course pages


Potolski:
Course pages currently show a teacher's **auto-generated wallet nickname** (e.g. `brave-anchor-9000`) or, with no profile, a truncated wallet — see `components/course/instructor-card.tsx`. Neither reads as a credible author, and there is no way to distinguish an official Superteam instructor from any wallet that happens to have created a course.

## Two additions

**Custom display name.** An editorial name ("Ana Souza", "Superteam Brasil") shown instead of the generated username. Falls back to the username, then to the truncated wallet — the existing chain, one rung longer.

**Verified badge.** A mark next to the name on the course page and public profile.

## Why both must be admin-controlled

A teacher can already edit their own `profiles` row. If `display_name` and `verified` were self-writable, any learner could rename themselves "Superteam Official" and self-award a verified badge — the badge would then mean nothing.

So both columns follow the `wallet_address` precedent (#408): writable by `service_role` only, enforced by a BEFORE trigger, not merely by RLS policy. Setting them is an admin action.

## Scope

- [ ] Migration: `profiles.display_name TEXT`, `profiles.verified BOOLEAN NOT NULL DEFAULT false`
- [ ] Trigger extending `enforce_profile_wallet_write` so a learner cannot set either on their own row
- [ ] Add both to the `public_profiles` view (`supabase/migrations/20260714130000_public_profiles_view.sql`) and to `PublicProfile`
- [ ] `InstructorCard`: render `displayName ?? username ?? truncated wallet`, plus the badge
- [ ] Badge on the public profile page too — a verified teacher is verified everywhere
- [ ] `aria-label` on the badge; it must not be icon-only to a screen reader
- [ ] Strings in en / pt-BR / es
- [ ] Extend `lib/supabase/schema-expectations.ts` so `/api/health/schema` catches the merge→apply gap
- [ ] Record the migration in `docs/DB-MIGRATION-LEDGER.md`

## Note

The profile link keeps using `username`, not the display name — usernames are unique and route-safe, display names are neither.

PR #1115, opened 2026-08-20
fix(db): lock profiles.deleted_at to service_role, realign schema.sql

### LINKED ISSUE #1103: security: deleted_at is user-writable — a tombstoned session can un-delete itself
The profiles UPDATE policy is column-agnostic (`USING (auth.uid() = id)`), and `deleted_at` has no write-lock trigger (wallet_address, referral fields, and role do). A deleted user holding a still-valid access token can call PostgREST directly with `UPDATE profiles SET deleted_at = NULL, is_public = true` on their own row and permanently resurrect the account — one write inside the ≤JWT-expiry window is permanent. Pre-existing (middleware never gated PostgREST), surfaced by the #1102 adversarial review (F5).

Done when: a service_role-only write-lock trigger covers deleted_at (mirror the escalation-lockdown trigger pattern), with a pglite test proving an authenticated own-row update can't clear it.

### LINKED ISSUE #1105: db: schema.sql drifted from prod — public_user_xp and get_leaderboard lost the deleted_at guard
Migration 20260704140000_account_deletion.sql rewrote `public_user_xp` to `WHERE p.is_public = true AND p.deleted_at IS NULL` and added the same guard to `get_leaderboard` — its own comment calls the pair "the sole access guard". schema.sql (public_user_xp ~:909, get_leaderboard ~:860-885) never picked either up: anyone rebuilding from schema.sql exposes soft-deleted users' XP publicly. Live prod is unaffected; the snapshot is what's wrong. Surfaced by the #1099 adversarial review, which is the first code to consume schema.sql's view definition programmatically.

Done when: schema.sql matches the migration for both objects, and the #1099 pglite suite (which extracts the view from schema.sql) still proves RPC-vs-queries parity with a soft-deleted profile seeded.

thomgabriel:
Locks `profiles.deleted_at` to service_role with a BEFORE trigger, the same escalation-lockdown shape already used for `wallet_address` and the referral columns. Without it a soft-deleted user holding a still-valid access token could hit PostgREST with `UPDATE profiles SET deleted_at = NULL, is_public = true` on their own row and permanently resurrect the account — the self-service UPDATE policy gates the row, never the columns.

The schema.sql half is snapshot-only: prod's `public_user_xp` and `get_leaderboard` have carried the `deleted_at IS NULL` guard since `20260704140000_account_deletion.sql`, but the snapshot never picked it up, so anything rebuilt from schema.sql exposed soft-deleted users' XP. Both objects now match the migration byte-for-byte.

A new pglite suite installs the real RLS policies alongside the trigger, so the exploit is proven blocked and — with the trigger dropped — proven to succeed against the same policies. 3035/3035 tests pass; the new suite is 16 (8 cases × migration and schema.sql copies), and `platform-stats-execution` drops from 925 to 425 total_xp, the change #1099 left a TODO for.

Two decisions the issues didn't specify: a separate `enforce_profile_deleted_at_write` rather than a third column on `enforce_profile_wallet_write` (schema.sql's copy of that function is stale — it predates #997's display_name/verified locks — so replacing prod's body from the snapshot would silently revert two live guards); and `deletion_requested_at` left writable, since only `deleted_at` is load-bearing for access.

**Not applied to any database.** The migration file is `supabase/migrations/20260819200000_deleted_at_write_lock.sql`, awaiting the usual apply-before-merge + ledger entry.

Closes #1103
Closes #1105

### DIFF
--- a/apps/web/src/lib/supabase/__tests__/deleted-at-write-lock-execution.test.ts
+++ b/apps/web/src/lib/supabase/__tests__/deleted-at-write-lock-execution.test.ts
@@ -1,0 +1,281 @@
+// REAL SQL-execution proof of the profiles.deleted_at write lock (#1103), run
+// in in-process Postgres (pglite) against BOTH copies of the DDL: the migration
+// that gets applied, and the supabase/schema.sql mirror that fresh environments
+// are built from.
+//
+// THE EXPLOIT. The self-service profiles policies are column-agnostic —
+// `FOR UPDATE USING (auth.uid() = id)` says which ROW you may write, never which
+// COLUMNS. deleted_at is the account tombstone every public read and every login
+// chokepoint keys on, so a soft-deleted user still holding a valid access token
+// could go straight to PostgREST with
+//
+//   UPDATE profiles SET deleted_at = NULL, is_public = true WHERE id = <self>
+//
+// and resurrect the account permanently — a single write inside the token's
+// remaining lifetime, long after the deletion route revoked the session.
+//
+// The RLS policies are installed here for real (with a stub auth.uid() reading
+// the JWT claim), so the "RLS alone does not stop it" case below is a live
+// red-proof: drop the trigger and the exploit succeeds against exactly the same
+// policies. That is what makes the rest of this suite a security test rather
+// than a trigger test.
+//
+// @vitest-environment node
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { PGlite } from "@electric-sql/pglite";
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+
+function findRepoRoot(): string {
+  let dir = dirname(fileURLToPath(import.meta.url));
+  while (!existsSync(resolve(dir, "supabase/schema.sql"))) {
+    const parent = dirname(dir);
+    if (parent === dir)
+      throw new Error("repo root (supabase/schema.sql) not found");
+    dir = parent;
+  }
+  return dir;
+}
+
+const repoRoot = findRepoRoot();
+const migration = readFileSync(
+  resolve(
+    repoRoot,
+    "supabase/migrations/20260819200000_deleted_at_write_lock.sql"
+  ),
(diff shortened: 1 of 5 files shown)


thomgabriel:
@claude review the current head (798cb44c). The prior review covered 4d8e2d1-era content; since then this PR added `deleted_at`/`deletion_requested_at` to schema.sql's profiles declaration, corrected two now-false comments in that declaration, added an operator note to the migration header, added a self-coherence test that builds profiles from schema.sql's own CREATE TABLE, and restructured the new suite to share one PGlite instance per describe. Please focus on the schema.sql declaration change and the new coherence test.

thomgabriel:
Gate note: both non-blocking findings from the 798cb44c review are fixed in 51436419.

The operator-note citation was mine — I gave the builder that line pointer in its brief and it was wrong. 20260817180000:44-49 documents SET ROLE being forbidden inside a SECURITY DEFINER function, not an INSERT-coercion incident, so the note now names enforce_profile_wallet_write's identical coercion branch instead of sending an operator to unrelated lines.

idx_profiles_deleted_at is now mirrored in schema.sql. I confirmed against prod that it exists exactly as 20260704140000 defines it (btree on deleted_at, partial WHERE deleted_at IS NOT NULL), so this is snapshot fidelity with no DDL implication.

Delta is one comment block plus one CREATE INDEX IF NOT EXISTS mirroring an already-applied migration. Suites re-run after the edit: the deleted_at suite 18/18, and src/lib/supabase 25 files / 397 tests green — the coherence test parses schema.sql directly, so it covers the index addition.

Migration is APPLIED to prod as MCP version 20260820011023, verified live: a non-privileged postgres session attempting the #1103 exploit on a real tombstoned row returned BLOCKED_42501, probe aborted, tombstoned count unchanged. Ledger row is #1119.

The is_public gap the earlier round raised is real — I verified on prod that community_stats, get_referral_leaderboard and is_public_profile lack the deleted_at filter while the other three surfaces have it — and it is filed as #1120 (P1). Not holding this PR for it: this change strictly removes the worse hole (permanent resurrection), and #1120 is pre-existing and independent.

Issue #1104, opened 2026-08-19
ops: operator-initiated tombstones need paired session revocation


thomgabriel:
Post-#1102, middleware no longer checks profiles.deleted_at per-request; both code writers of deleted_at pair with session revocation, but an operator tombstoning an account via Studio/psql gets none — the account's sessions live until token refresh… which succeeds, forever. From the #1102 adversarial review (F4).

Fix ideas (either closes this): a DB trigger on deleted_at transitions that bans the user via GoTrue, or a documented runbook step ("also ban in Dashboard → Auth") + admin-panel action that does both.

MayurK-cmd:
Hello can i work on this?

PR #1121, opened 2026-08-20
perf: dashboard server shell — one render instead of 19 client requests (#1096)

### LINKED ISSUE #1096: perf: dashboard is fully client-side — ~19 requests in 4 waves after hydration
dashboard/page.tsx is "use client"; nothing renders until hydration + auth resolve, then ~19 HTTP requests across 4 sequential waves. Five hit /api/content/* routes whose data is a synchronous in-memory bundle read server-side.

Researched target shape (Supabase's own App Router guidance): server-component shell — cookie client + cached claims read, direct @/lib/content/queries imports (no HTTP hop) — for XP/level, streak, quests, achievements, activity, courses, continue target, recommendations; each slow section in its own Suspense. Client only for wallet-connected on-chain reads, quest-claim mutation, celebration layer. Skip TanStack hydration — server-rendered values need no client refetch. Follow the existing courses/page.tsx server-fetch shape.

Gotcha: any app-wide TanStack useQuery needs the QueryClientProvider hoisted unconditionally first — today it only mounts when NEXT_PUBLIC_DYNAMIC_ENVIRONMENT_ID is set.

Depends on: #1090 (header cut), the loading/skeleton primitives from #1092. Do after those merge.

Done when: dashboard first paint is server-rendered, /api/content/* calls from dashboard drop to zero, total requests <6, review-strip/cohort-strip either server-rendered or folded into one wave.

thomgabriel:
Closes #1096.

dashboard/page.tsx is now a server component on the #1102 DAL (`getAuthClaims()`). The old `useDashboardData` burst — 11 Supabase reads, the quests poll, 5 `/api/content/*` fetches, plus the review/cohort strip fetches — moved into `lib/dashboard/loaders.ts`: per-user reads on the cookie-bound server client (RLS unchanged), content lookups as direct bundle imports, review/quests/cohort calling their lib functions instead of HTTP. Five Suspense boundaries (hero + main column share one `cache()`d loader pass; review / quests / cohort stream independently) with the app spinner as fallback.

Client islands kept: the section components (props-fed as before), NextLessonPlan (owns its prefs read/write), and two render-nothing celebration islands — the #790 quest-reward / surprise-bonus toast dedupe lives in sessionStorage, so only the pick+dispatch stays in the browser. review-strip/cohort-strip are now presentational; `use-dashboard-data.ts` is deleted (its only consumers were the page and two type imports, moved to `lib/dashboard/types`). No TanStack, no QueryClientProvider hoist — server-rendered values need no client refetch.

Client-visible requests on dashboard load: ~19 → 1 (NextLessonPlan's own prefs read; header/sidebar xp-gain listeners unchanged). Route table: dashboard 27.3 kB / 468 kB First Load JS; same-method raw-chunk measure vs the 19-08 main build: 27 client chunks / 1624 kB → 5 chunks / 585 kB (baseline predates #1109, so part of that cut is the provider demotion).

vitest 3069/3069, tsc clean, one production build green. Hook tests migrated to `dashboard-celebrations.test.tsx` (same #790/#796 assertions).


thomgabriel:
Routing record (XP/data lane, claude[bot]-independent adversarial gate): PASS — prefetch cannot render the page (loading boundary + force-dynamic), double-award closed at the DB (xp_granted row-lock + idempotency keys), no anonymous evaluation, no cross-user caching. Five non-blocking findings → three fixed in 93d003cc (1500ms bound on the on-chain read in the service, structured logs restored on both rail loaders, null fallbacks on the empty-case rails), two documented in-code (the ≤JWT-expiry banned-token window in front of the quest write — #1089 invariant, operator closure #1104; the cohort RPC's lazy weekly assignment). Merging on completed-green CI.

PR #1124, opened 2026-08-20
fix(auth): free the SIWS overlay and the OAuth buttons from the lazy chunks

### LINKED ISSUE #1097: perf: wallet/Dynamic/Query provider stack mounts on marketing pages

thomgabriel:
Post-merge review of #1109 (closes #1097) found four real defects and a set of comments the merge invalidated. Fixed forward here.

**The SIWS overlay was trapped in the header.** On marketing routes the scoped stack mounts inside the header bar, whose `backdrop-blur-md` makes it a containing block for `position: fixed` descendants — verified on production: that div computes `backdrop-filter: blur(12px)`, is 57px tall, and a `fixed; inset:0` probe injected inside it measures 57px. So an anonymous visitor using the header's Sign in got a "Signing in…" band across the nav with the page still interactive, and on the error branch the failure message plus Retry/Dismiss were squeezed into the same strip. The overlay now portals to `document.body` the way the wallet-select modal does, which also moves it to the dialog layer — as a body child, `z-100` would have painted under the header's `z-200`.

**Google and GitHub were hostage to the wallet chunk.** The modal gated its whole body on a live ambient registration, so the two methods that need nothing but Supabase waited on a ~300 kB stack download and then shifted layout — and because `LazyBody` only rendered after the stack registered, the two chunks downloaded serially. The body now renders as soon as its own chunk lands; only the Solana button tracks the registration, and it reports `authFailed` rather than spinning if the stack chunk dies. Stack and body failures are tracked separately with their own retries, and a dead body chunk falls back to `oauth-fallback-button.tsx`, a static module with no wallet or Dynamic import, so the OAuth paths survive a blocked CDN or a stale deploy.

The #1097 goal is intact and now enforced: `provider-free-import-graph.test.ts` walks the static import graph from `header.tsx`, `auth-modal.tsx`, `user-menu.tsx` and `landing-client.tsx` and fails if wallet-adapter, the Dynamic SDK, or TanStack Query becomes statically reachable. It carries its own guard against passing vacuously, and I checked it fails as intended by flipping the modal's `import type` of the body to a value import.

**The two flaky auth tests** raced the lazy chunk behind a 10s `findBy`. Each suite now statically imports `auth-modal-body`, so the module is in the registry at collect time and the lazy import resolves in a microtask; the inflated per-test and per-query timeouts are gone rather than raised.

`docs/AUTH-FLOWS.md` gets a new §1.1 covering the scoped stack, the ambient wallet store, `DynamicReturnCatcher` and which route group mounts what, plus corrections to the three stale claims and to the middleware bullet still describing a retired HMAC `admin_session` cookie.

Twelve full `pnpm --filter web test` runs on this branch: 324 files / 3080 tests, five fully green. Every failure in the other seven was a PGlite `*-execution` test (`Hook timed out in 10000ms` at `PGlite.create()`), and that reproduces on main unchanged — one of two baseline runs failed the same way, in the same files. Flagged separately; no auth test failed in any of the twelve. `tsc --noEmit` and `next lint` clean.

Left alone deliberately: the `LowSolBanner` position change, and `ambient-wallet-store.ts`'s unused `publicKey` field.

Refs #1097, #1109.

### DIFF
--- a/apps/web/src/__tests__/provider-free-import-graph.test.ts
+++ b/apps/web/src/__tests__/provider-free-import-graph.test.ts
@@ -1,0 +1,129 @@
+import { readFileSync, existsSync, statSync } from "fs";
+import path from "path";
+import { describe, it, expect } from "vitest";
+
+/**
+ * #1097's actual deliverable, as a test: marketing first loads must not carry
+ * wallet-adapter, the Dynamic SDK, or TanStack Query. Those live behind
+ * `React.lazy` boundaries — `scoped-auth-providers` and `auth-modal-body` —
+ * and one careless static import from anything the Header renders undoes the
+ * whole demotion silently, since the app still works, just 300 kB heavier.
+ *
+ * Walks the STATIC import graph (dynamic `import()` is the boundary, so it is
+ * deliberately not followed; `import type` is erased at build and likewise
+ * skipped) from the four entry points a marketing page mounts, and fails if
+ * any forbidden package is reachable.
+ */
+
+const SRC = path.resolve(__dirname, "..");
+
+const ENTRY_POINTS = [
+  "components/layout/header.tsx",
+  "components/auth/auth-modal.tsx",
+  "components/auth/user-menu.tsx",
+  "app/[locale]/(marketing)/landing-client.tsx",
+];
+
+const FORBIDDEN = [
+  "@solana/wallet-adapter-base",
+  "@solana/wallet-adapter-react",
+  "@solana/wallet-adapter-react-ui",
+  "@solana/wallet-adapter-wallets",
+  "@dynamic-labs-sdk/client",
+  "@dynamic-labs-sdk/react-hooks",
+  "@dynamic-labs-sdk/solana",
+  "@tanstack/react-query",
+];
+
+/** `import … from "x"` / `export … from "x"`, minus `import type`. */
+const IMPORT_RE =
+  /(?:^|\n)\s*(?:import|export)(?!\s+type\s)(?:[\s\S]*?from\s*)?["']([^"']+)["']/g;
+
+function readImports(file: string): string[] {
+  const source = readFileSync(file, "utf8");
+  const found: string[] = [];
+  for (const match of source.matchAll(IMPORT_RE)) {
+    const specifier = match[1];
+    if (specifier) found.push(specifier);
+  }
+  return found;
+}
+
+function resolveLocal(specifier: string, fromFile: string): string | null {
+  const base = specifier.startsWith("@/")
+    ? path.join(SRC, specifier.slice(2))
+    : specifier.startsWith(".")
+      ? path.resolve(path.dirname(fromFile), specifier)
+      : null;
+  if (base === null) return null; // a package, not a local module
+
+  for (const candidate of [
+    base,
+    `${base}.ts`,
+    `${base}.tsx`,
+    path.join(base, "index.ts"),
+    path.join(base, "index.tsx"),
+  ]) {
+    if (existsSync(candidate) && statSync(candidate).isFile()) return candidate;
+  }
+  return null; // css, json, or something with no module body to walk
+}
+
+/** Every package name statically reachable from `entry`, with one path each. */
+function reachablePackages(entry: string): Map<string, string[]> {
+  const packages = new Map<string, string[]>();
+  const seen = new Set<string>();
+  const queue: { file: string; trail: string[] }[] = [
+    { file: path.join(SRC, entry), trail: [entry] },
+  ];
+
+  while (queue.length > 0) {
+    const { file, trail } = queue.shift()!;
+    if (seen.has(file)) continue;
+    seen.add(file);
+
+    for (const specifier of readImports(file)) {
+      const local = resolveLocal(specifier, file);
+      if (local) {
+        queue.push({
+          file: local,
+          trail: [...trail, path.relative(SRC, local)],
+        });
+      } else if (!specifier.startsWith(".") && !specifier.startsWith("@/")) {
+        if (!packages.has(specifier)) packages.set(specifier, trail);
+      }
+    }
+  }
+  return packages;
+}
+
+describe("marketing entry points stay free of the wallet/Dynamic stack (#1097)", () => {
+  it.each(ENTRY_POINTS)("%s", (entry) => {
+    expect(existsSync(path.join(SRC, entry))).toBe(true);
+    const packages = reachablePackages(entry);
+
+    const offenders = FORBIDDEN.filter((pkg) =>
+      [...packages.keys()].some(
+        (found) => found === pkg || found.startsWith(`${pkg}/`)
+      )
+    ).map((pkg) => {
+      const via = [...packages.entries()].find(
+        ([found]) => found === pkg || found.startsWith(`${pkg}/`)
+      );
+      return `${pkg} via ${via?.[1].join(" -> ")}`;
+    });
+
+    expect(offenders).toEqual([]);
+  });
+
+  it("still finds the forbidden packages where they legitimately live", () => {
+    // Guards the walker itself: if the regex or the resolver silently stopped
+    // matching, every assertion above would pass vacuously.
+    const packages = reachablePackages(
+      "components/auth/scoped-auth-providers.tsx"
+    );
+    const names = [...packages.keys()];
+    expect(names).toContain("@solana/wallet-adapter-react");
+    expect(names.some((n) => n.startsWith("@dynamic-labs-sdk/"))).toBe(true);
+  });
+});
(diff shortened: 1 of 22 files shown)


thomgabriel:
@claude review the current head (5b41918d). This is the fix-forward for two defects that shipped in #1109: the SIWS overlay was clipped into the 57px header by the header's backdrop-filter containing block, and the OAuth buttons were gated behind the wallet chunk. Please focus on whether any sign-in path can now hang, fail silently, or become unreachable when either lazy chunk dies, and on the deliberate z-index change on the portalled overlay.

thomgabriel:
Adversarial gate: BLOCK. Holding the merge. Fixes in progress on this branch.

The blocking one is a real regression, and it is on this PR's own axis. The comment justifying the z-index bump says the sign-in dialog and the SIWS overlay never coexist. That stopped being true at #1097: opening the dialog is what mounts the wallet stack (auth-modal.tsx:130-134 → scoped-auth-providers → wallet-provider.tsx:88-89, autoConnect true at :74), so a returning signed-out learner with walletName still in localStorage reconnects the moment the stack mounts and SIWS auto-fires with the dialog open. Radix sets body pointer-events:none for a modal dialog and the body-portalled overlay inherits it, so the scrim covers the dialog and its own Retry/Dismiss are inert — only Escape gets out. In that window this is worse than before the PR, when the overlay painted under the dialog at 57px and the learner kept a working dialog.

The other one that matters is the import-graph guard, since it is the only thing stopping #1097 regressing. It was fooled twice, both proved by mutation: bare side-effect imports never match IMPORT_RE (a smuggling local module passed 5/5 green with its whole subtree invisible), and the layouts that render the Header on every marketing route are not entry points — which is exactly where the stack lived before #1097.

Also fixing: the bodyFailed branch announces "Wallet authentication failed" when a chunk failed to download and no wallet was involved, with the wrong string regression-locked by a test; unguarded handoff timers that can leave a learner with no error and no feedback, or pop the wallet picker 600ms after they navigate away; and a latent loading-state orphan that shares a root cause with #1126.

Credit where due: the tests here are honest — all three mutation checks kill exactly the intended tests, and the guard is non-vacuous for the cases it does cover.

thomgabriel:
@claude review the current head (7ac069b4) — the delta since 5b41918d addresses the adversarial BLOCK. The load-bearing change: instead of fighting Radix for pointer-events, WalletAuthHandler now raises a SIWS-active flag in the ambient store and AuthModal closes on it, so the dialog and the overlay genuinely cannot coexist. The handoff timers also moved from the lazy body up into AuthModal so unmount cleanup does not fire on the happy path. Please focus on whether closing the dialog can now lose or hide an error the learner needs, and on the new store flag's lifecycle across overlapping stacks.

thomgabriel:
Adversarial re-check on 7ac069b4: PASS WITH FINDINGS. The BLOCK is cleared — F1 is proved dead by reproducing the takeover tests on the parent commit (all three fail there, all pass here), Radix's focus trap is torn down rather than relocated, and the flag lifecycle survived seven attacks looking for a stuck flag, including an error boundary replacing the subtree and a client-side nav mid-SIWS. No path makes the modal permanently un-openable, which was the highest-severity risk this design could have introduced.

Three proved findings before merge, two of them regressions in the same returning-learner population this PR is for.

The handoff never checks getSiwsActive() before opening the wallet picker, so on the likely ordering — body chunk first, learner clicks Solana before the stack lands, stack registers as autoConnect reconnects — the picker opens on top of a live SIWS overlay, burying the Retry and Dismiss this PR exists to surface. The PR's own bug re-created through a different modal.

And declining the signature now leaves nothing at all: silent dismiss composed with the new auto-close means the dialog vanishes with no overlay and no message. Before this delta the dialog stayed up behind the overlay, so a rejection left Google and GitHub in front of the learner.

Third: a comment anywhere inside an import clause blinds the import-graph guard — a commented wallet-adapter import in header.tsx passes 14/14, uncommented it fails 2. Latent today (an AST diff over all 47 reachable files found zero misses) but that guard is the only thing standing between #1097 and a silent regression, and comments in import blocks are ordinary.

Fixes in progress on this branch.

thomgabriel:
All four findings fixed at 17af070a. The builder ran out of session mid-way, so I finished it.

The picker now checks getSiwsActive() before opening, and both handoff branches refuse to write an error into a dialog that has already closed. Declining a signature says so and offers Retry — Retry specifically, because hasTriedAuth only resets on disconnect, so re-opening and re-picking the same connected wallet would not re-fire SIWS. New auth.signatureDeclined key in all three locales. The import guard strips comments before matching, quote-aware and newline-preserving so line anchoring survives.

Two things worth flagging beyond the brief.

The leaked SIWS claim the builder was chasing when it stopped is real, and the mechanism is ordering: RTL registers its cleanup at import time, so under LIFO this file's own afterEach ran first and restored the wallet mock before unmount — tearing the mock out from under the effect that releases the claim, which is module-level state that would then contaminate the next test. It now unmounts explicitly first.

And a test failed in the combined run while passing alone. Both failure suites wait on a lazy import that is deliberately made to REJECT, so they cannot be pre-registered the way the other auth suites statically import the body, and under parallel load the module resolution alone outran the 1s default. Those waits now carry an explicit budget with the reason written down. Three consecutive clean runs of the auth set afterwards.

Results observed: auth suites plus the guard 13 files / 69 tests green, three runs in a row. tsc --noEmit exit 0 (it caught me putting the timeout in findByRole's options slot instead of its third waitForOptions slot — fixed). Lint clean on every touched file. Full suite 3105 tests with 3 failures, all 'Hook timed out in 10000ms' in PGlite *-execution suites — the pre-existing load-dependent flake both prior reviews independently confirmed. Zero auth failures, and none in any file this PR touches.

thomgabriel:
On the discarded OAuth error from the 7ac069b4 review — real, and accepted deliberately rather than fixed. Reasoning written into the code at be4c7839 so it is checkable rather than tribal.

From the moment the SIWS overlay takes the screen it owns error reporting, and every SIWS terminus surfaces its own outcome: success hard-redirects, and every failure — including the declined signature this PR just gave a message to — renders the overlay's own text with Retry and Dismiss. So nothing is hidden and nobody is stranded; what is lost is a Google error the learner has already read, and resurfacing it underneath a wallet flow would misattribute the failure to the wrong method. Reopening and retrying Google is available throughout.

The comment states that as an invariant a future SIWS ending has to keep: if one can ever end silently, this reset starts hiding a real error. That is the condition under which this decision flips.

Also noting the review's out-of-scope catch, since it is worth knowing and is not mine to fix here: two overlapping handlers can each pass their own getUser() check and race two real SIWS attempts, two nonce fetches and two prompts, if both see a null user before either resolves. That is inherent to the per-stack handler design from #1097, not introduced by this diff, and it is the same family as #1127.

Head is be4c7839. Auth suites plus the guard: 13 files, 69 tests green. tsc clean.

thomgabriel:
@claude review the current head (be4c7839). Delta since the 7ac069b4 you last reviewed is two commits. 17af070a fixes the three findings from the adversarial round plus the one you raised: startWalletHandoff now checks getSiwsActive() before opening the picker and refuses to write an error into a closed dialog; a declined signature gets its own message and Retry (new auth.signatureDeclined in all three locales); the import guard strips comments before matching, quote-aware and newline-preserving; the two failure suites' waits carry explicit budgets because a deliberately-rejecting lazy import cannot be pre-registered; and the takeover suite unmounts before restoring mocks so the SIWS claim's release effect still has its wallet mock. be4c7839 is comment-only, recording why the close-reset discarding a visible OAuth error is intended.

Please focus on the two new early-return guards in startWalletHandoff — whether either can now swallow a click and leave the learner with no picker, no error, and no spinner — and on whether the declined-signature path interacts correctly with hasTriedAuth only resetting on disconnect.

PR #1125, opened 2026-08-20
chore(ui): remove the low-SOL banner

thomgabriel:
Owner decided to drop it rather than keep either placement.

It was devnet-only, required a connected wallet, fired under 0.001 SOL, and was dismissible — a nudge toward the faucet. #1109 had moved it out of the fixed header into the page flow, which raised the question of where it belonged; the answer is nowhere.

Removes the component and its test, the `(platform)` mount, the Header comment that pointed at the old location, and `nav.lowSolBalance` / `nav.getDevnetSol` from en, pt-BR and es. Locale key parity verified at 1553 keys each; no references remain.


thomgabriel:
Good catch on nav.dismissBanner — it was the dismiss button's aria-label, and my sweep grepped the component name plus the two visible strings, so it slipped through. Removed from all three locales; parity re-verified at 1552 keys each, and no reference to it remains anywhere in apps/web/src.

Issue #1116, opened 2026-08-20
db: schema.sql's enforce_profile_wallet_write is missing the display_name/verified locks that are live on prod


thomgabriel:
Same family as #1105, found while building #1115.

Prod's `enforce_profile_wallet_write` locks `wallet_address`, `display_name` and `verified` — verified live: `prosrc` matches all three. schema.sql's copy (around line 790) mentions only `wallet_address`; grepping the function body for `display_name` or `verified` returns zero hits.

The guards were added by `20260805120000_teacher_display_name_and_verified.sql`, which the migration ledger records as applied out-of-band, so the snapshot never picked them up. The risk is the usual one for this drift family, and it is worse than a stale comment: anyone rebuilding from schema.sql, or issuing a `CREATE OR REPLACE` built from it, silently drops two deployed admin-only guards — at which point a learner can set their own `verified` badge and `display_name`. #1115 hit exactly this and deliberately chose a separate trigger rather than extending the stale function.

Also worth checking in the same pass whether `profiles.display_name` and `profiles.verified` themselves are present in schema.sql's table definition.

Done when: schema.sql's `enforce_profile_wallet_write` matches the live definition for all three columns, with a test that would catch the next divergence.

thomgabriel:
Two additions from the #1115 adversarial review.

A green test is holding the staleness in place: `wallet-write-lock-guard.test.ts:34` pins schema.sql's copy against the older `20260710120000` migration rather than the live definition, so the missing display_name/verified locks pass CI. Whatever fixes this should re-point that test, or the drift just comes back.

The same file's `profiles` CREATE TABLE also declares neither `display_name` nor `verified` (nor `deleted_at`/`deletion_requested_at` — that pair is being fixed in #1115 since its new trigger references them). All four exist on prod; I confirmed via information_schema: display_name text nullable, verified boolean NOT NULL, both timestamps timestamptz nullable. So the table declaration needs realigning alongside the trigger body.

PR #1128, opened 2026-08-20
fix(db): filter deleted_at on the three public surfaces that skipped it

### LINKED ISSUE #1120: security: is_public is user-writable and three public surfaces don't filter deleted_at
Found by the #1115 review, verified live on prod.

#1115 locks `profiles.deleted_at` to service_role, which kills the permanent-resurrection exploit in #1103. But `is_public` is still user-writable under the same column-agnostic RLS policies, and it is the column the remaining public surfaces gate on. Before #1115 this was subsumed by the full exploit (clear `deleted_at` and set `is_public` together); now that `deleted_at` is locked, `is_public` is the one lever left, so it stops being theoretical.

Checked each object's definition on prod:

| object | filters deleted_at |
| --- | --- |
| `public_profiles` | yes |
| `public_user_xp` | yes |
| `get_leaderboard` | yes |
| `community_stats` | **no** |
| `get_referral_leaderboard` | **no** |
| `is_public_profile` | **no** |

So a tombstoned account that flips its own `is_public` back to true resurfaces through those three, which contradicts the invariant the account-deletion migration states for itself — that the `is_public` + `deleted_at` pair is the sole access guard.

Two ways to close it, not mutually exclusive: add `p.deleted_at IS NULL` to the three objects, matching the pattern already applied to the other three; and/or extend the new `enforce_profile_deleted_at_write` trigger to cover `is_public` for a tombstoned row. The first is the more complete fix, since it also covers any future writer.

Done when: all six public surfaces agree on the guard, with a test that seeds a soft-deleted profile with `is_public = true` and proves it appears in none of them.

thomgabriel:
Closes #1120.

`community_stats`, `get_referral_leaderboard` and `is_public_profile` now carry
`deleted_at IS NULL` next to the `is_public` check, matching what
`public_profiles`, `public_user_xp` and `get_leaderboard` have had since
`20260704140000_account_deletion.sql`. Each body is carried forward from the
latest migration that defines it — `20260624181348`, `20260818150000` and
`20260726130000` respectively — and every grant is re-stated rather than left to
`CREATE OR REPLACE` to carry.

I did not add a write-side guard on `is_public`. The read-side fix makes the
write moot, and it covers a case a trigger could not:
`merge_wallet_shell_account` tombstones the merged shell with a bare
`UPDATE profiles SET deleted_at = now()` and never touches `is_public`, which
DEFAULTs to true — so merged shells are exposed through these three objects
today with no write to `is_public` anywhere.

On `community_stats` the filter sits on the public branch only —
`(is_public AND deleted_at IS NULL) OR id = auth.uid()` — because
account_deletion's rule is that public read paths gain the guard while own-row
access stays intact. A pinned test covers that.

The migration is not applied. `supabase/schema.sql` is mirrored in the same
commit.

The test executes both copies of all six objects in pglite, extracted from the
real .sql files rather than hand-written, so a schema.sql that drifts fails
rather than passes. It installs the pre-fix bodies first and applies the
migration over them, so `CREATE OR REPLACE VIEW community_stats` is exercised
against the shape prod actually has. 12 tests pass; reverting the guard in both
files turns 8 of them red, with exactly `community_stats`,
`get_referral_leaderboard`, `is_public_profile` and the enrollments RLS path
flipping back to visible while the three already-guarded surfaces stay hidden.

Unrelated drift found while wiring the test, not fixed here: schema.sql's
`public_profiles` is missing the `display_name` and `verified` columns that
`20260805120000` added to prod, and neither column appears anywhere in
schema.sql — the same family as #1116, which asks that exact question.


thomgabriel:
Pre-apply verification done from my side, against live prod rather than against the source migrations — that is the check the reviewers cannot run.

Carry-forward is faithful. For both functions, reversing the added guard reproduces the LIVE prod definition hash exactly (is_public_profile 5ed276b4…, get_referral_leaderboard 854074e7…), with exactly one guard added each. For community_stats I compared structurally against pg_get_viewdef, since Postgres re-prints views: column list, aliases, aggregates, FILTER clauses, joins and GROUP BY are identical, and the only semantic delta is the guard on the public branch. So nothing silently reverts.

All three confirmed unguarded on prod, and the exposure is not theoretical — one tombstoned profile currently has is_public true and is surfacing through community_stats today. That corroborates your reasoning for fixing readers rather than writers: nobody wrote is_public on that row, so a write-side guard would never have fired.

One correction pushed as 0f3e18b3. The header said no grants move; one does. community_stats carries Supabase's default arwdDxtm for anon and authenticated, and the REVOKE narrows it to SELECT. That is the right change — it matches public_profiles and public_onchain_deployments, which were already tightened — and it is inert, because the view aggregates with GROUP BY so it is not auto-updatable. But it is a real privilege change on prod and the ledger has to record it rather than gloss it.

Worth knowing, checked while verifying and deliberately not fixed here: public_user_xp carries the same default grants. Also inert for the same reason — it is non-updatable. The two views that ARE auto-updatable, public_profiles and public_onchain_deployments, are exactly the two already locked to SELECT, so the posture is coherent; public_user_xp is just an untidied leftover.

PR #1155, opened 2026-08-21
fix(lesson): prev/next split the mobile bottom nav 50/50

thomgabriel:
On mobile the three stacked full-width buttons read as one column of equal
actions. The completion CTA keeps the full-width top row; Previous and Next
share the second. Desktop flex row unchanged (the CTA wrapper dissolves via
sm:contents).


thomgabriel:
@claude review the three commits after the first (quiz auto-fold removal, referral stats Promise.all, mini-league CSS) — the earlier review predates them.

PR #1157, opened 2026-08-21
feat(ui): unify learner progress bars on the XP-bar construction

thomgabriel:
The dashboard XP bar is the progress-bar standard (muted track, 2.5px ink outline, solid mint fill, pill radius). The remaining learner-facing bars each had their own construction — a 1px soft-bordered track, a 7px gradient sliver, an SVG donut. They now all derive from the same three custom properties (`--xpbar-ink`, `--xpbar-track`, `--xpbar-fill`), declared once on every construction that uses them, the way `.sidebar-prog` already did.

Lesson counts are discrete, so course and lesson bars render one cell per lesson inside the ink outline rather than a continuous fill. Past 24 lessons the ticks stop reading as countable units, so `ProgressBar` falls back to a smooth fill on its own (`SEGMENT_CAP`). Continuous shares stay smooth: the learning-path bar spans whole courses, so ticks there would read as lessons and mislead.

The Current Courses donut becomes a mono fraction over a 64px micro strip — narrower than the donut plus its label, so the card doesn't widen. `deriveEndowedProgress` is preserved: with nothing completed, the endowed fill lights exactly one cell, and `aria-valuenow` still reports the honest count, never the endowed fill.

Also dropped the redundant "60%" next to the course-detail fraction, and deleted `.progress-fat*` / `.progress-fill-{teal,amber,green}` plus their `PROGRESS_STYLES` fill entries — grep found no consumers. Admin bars (`components/ui/progress.tsx`), the XP bar itself, `.sidebar-prog`, `.lb-*` and the "Plan your next lesson" card are untouched.

**Numbers**: 75 vitest tests pass across `components/course` + `components/dashboard` (13 files), including 7 new segment-math cases; `tsc --noEmit` clean; eslint on the changed files reports only the two pre-existing warnings.

**Red-proofs**: replacing the endowed minimum with `lit = done` fails 2 tests (the component case and the dashboard card case); rendering `total - 1` cells fails 4.

**Verification**: the dashboard is auth-gated so I could not browser-probe it — I verified by token construction instead, rendering the copied CSS block in both themes (ink outline, mint fill, hairline ticks legible at card/slim/micro, smooth cap fallback, smooth path bar).

**Overlap**: #1156 also edits `globals.css` (`.chip`/`.mark` near the `.patch` section). This change sits next to `.dash-xp-track` and in the `.cc-*` / `.path-*` blocks, so the two should not collide; whichever lands second may still need a trivial rebase.

### DIFF
--- a/apps/web/src/app/[locale]/(platform)/courses/[slug]/course-detail-client.tsx
+++ b/apps/web/src/app/[locale]/(platform)/courses/[slug]/course-detail-client.tsx
@@ -283,8 +283,8 @@ export function CourseDetailClient({
                 {isEnrolled && (
                   <div className="mb-5 space-y-2">
                     <div className="flex items-center justify-between text-sm">
-                      <span className="font-display text-[13px] font-bold text-text-2">
-                        {completedCount} / {totalLessons} {t("lessons")}
+                      <span className="font-mono text-xs font-semibold uppercase tabular-nums tracking-wider text-text-2">
+                        {completedCount}/{totalLessons} {t("lessons")}
                       </span>
                       <span className="font-display text-[13px] font-bold text-text-3">
                         {isComplete ? t("completed") : t("inProgress")}
@@ -293,7 +293,7 @@ export function CourseDetailClient({
                     <ProgressBar
                       value={completedCount}
                       max={totalLessons}
-                      showLabel
+                      segmented
                     />
                   </div>
                 )}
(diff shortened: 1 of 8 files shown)


thomgabriel:
Consolidating into #1156 per owner — one UI PR so the whole set (glyphs + bars) tests together on one dev server. The branch merges in there; nothing is lost.

PR #1161, opened 2026-08-23
feat(referrals): share a course with a friend, and reach referrals from the menu

Potolski:
Two additions to the referral system.

## Share a course with a friend

The course action rail gains a share control under the CTA. The link it shares is the course page with `?ref=<code>` appended — which the app-wide `ReferralCapture` already claims on any page load. So recommending a course **is** a referral, and the friend lands on the thing they were told about rather than a generic home page.

Two behaviours worth naming:

- **Signed out still works.** The code comes from `/api/referrals/me`, which 401s without a session. Without one the button shares the plain course link. "I can't share this" is a worse answer than "shared, but uncredited", and nothing about the control changes between the two states — so a signed-out visitor is never shown a dead affordance.
- **Native share sheet where there is one.** That's what "share with a friend" means on a phone. Clipboard is the desktop fallback, and also the fallback when the sheet is dismissed.

## Reach referrals from the user menu

A Referrals entry sits directly under Certificates.

It needed one supporting change: the leaderboard's board was local state with no way in, so the menu entry would have dropped the reader on the **League** board rather than Referrals. `?board=` now wins over the default. It's validated against a hoisted id list — the labelled `BOARDS` array is built inside the component (it needs `t`) and can't be read from a state initialiser that runs before it.

## Testing

New coverage:
- the shared link carries the viewer's referral code
- it still shares a plain course link when there is no code
- the native share sheet is preferred where available
- the clipboard is used when the sheet is dismissed
- `?board=referrals` opens the Referrals tab; no query and an unknown id both fall back to League

`pnpm typecheck` clean, `pnpm lint` clean, full suite **3374 passing** across 339 files. The three new message keys were added to all three locales and structural parity was asserted (1547 keys each).

## Note

I verified the deep link against a local instance, but that copy has no database, so the course detail page 404s there — the share control's behaviour is covered by unit tests rather than an end-to-end check, and its visual fit was confirmed by rendering the rail from the real stylesheet in both themes.

https://claude.ai/code/session_01MJMV9QpjXQ42PVjgEeMQp4

---


Potolski:
Good catch — confirmed and fixed in 1537825d.

`?board=` now goes through `useSearchParams()` with an effect, matching the `?linked=` pattern in `account-tab.tsx` you pointed at.

**One addition beyond the report.** The tab click also *writes* the param back. Reading reactively fixes the direction you described, but the reverse case is the same bug wearing a different hat: a visitor sitting on `?board=referrals` who clicks League leaves the URL and the tab disagreeing — and the menu entry then no-ops again, because the param it would set is already there. With the URL as the single source of truth that state is unreachable. `replace` rather than `push`, so flipping tabs isn't something the back button has to walk through, and `scroll: false` keeps the reader in place.

**On the test gap.** You were right that `renderClient` always mounted fresh. The suite now drives `next/navigation` instead of `window.location` and covers both missing paths — the param changing *without* a remount, and a tab click writing it back. I checked the first one actually catches the bug rather than just passing: reverting the effect turns it red.

```
× follows ?board= when it changes without a remount
  Tests  1 failed | 22 passed (23)
```

**No Suspense boundary needed** — worth stating since `useSearchParams` can force a static-render bailout. `ReferralCapture` already calls it from the root layout on every page without one, and this page is cookie-bound through `getAuthClaims`, so it renders dynamically regardless.

Full suite green: 3376 passing across 339 files.

---

Potolski:

PR #1170, opened 2026-08-24
feat(db): exclude the retired early-adopter grant from ranking

thomgabriel:
Owner decision 2026-08-24 (narrow-exclusion option). New `is_rank_excluded_reason()`; `get_leaderboard`'s alltime branch now sums `xp_transactions` (as weekly/monthly already did) so the 1000-XP early-adopter grant — 34.7% of all XP, worth more than the entire catalog's learning XP — stops scoring; same predicate added to league tier placement and score refresh. Holders keep XP and level (`user_xp` untouched).

Already applied to prod (MCP version `20260824150235`, ledger row included) per the #731 zero-gap ordering. Live probe: the four early-adopter holders in the top 5 each dropped exactly 1000 on the board, everyone else byte-identical.


thomgabriel:
Superseded same-day by the owner's standardization call: every surface shows the one true on-chain XP total, so the scoring exclusion was rolled back on prod (restored pre-exclusion bodies verbatim, predicate dropped; board verified to match user_xp on all 82 rows). Both prod migration stamps (apply 20260824150235, rollback 20260824172511) are being recorded in the DB migration ledger — neither has a repo file, by design. The gate's CI finding (supabase/** missing from the paths filter) is being filed separately.

PR #1171, opened 2026-08-24
feat(content): activate the launch achievement ladder + quest repricing

thomgabriel:
Bumps `content.lock` to academy-courses@958e7877 (academy-courses#41). Achievements go 7 → 24 and the four live quests get repriced; no course, lesson, path or slot moves, so the course/lesson/path goldens are byte-unchanged.

Two behavioral notes:

Every one of the 17 new achievement docs needs an admin on-chain sync (deploy) before Helius can award it — the content bump alone does nothing for learners.

`achievement-early-adopter` flips from `user-number lte:100` to `manual`, so it no longer auto-fires on signup. Day one now yields one badge (first-steps, 25 XP) plus the lesson quest (20) and the lesson itself (20), instead of the old early-signup windfall. `quest-challenge` goes inactive and drops out of the active set.

Goldens regenerated through the real projectors; full web suite green (3396 tests).

### DIFF
--- a/apps/web/content.lock
+++ b/apps/web/content.lock
@@ -1,4 +1,4 @@
 {
   "repo": "solanabr/academy-courses",
-  "sha": "4a6e4c6f8ac1c19566a3de85a19012809b28b050"
+  "sha": "958e78779ab4d3a309c12a1ff8ae30887331e9ff"
 }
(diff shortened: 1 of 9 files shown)


thomgabriel:
@claude re-review the delta (cceebaa3): community achievement unlocks now gate on getDeployedAchievements() with a two-way test (undeployed qualifying doc → no unlock; deployed → unlocks). Confirm this closes your blocking finding.

PR #1202, opened 2026-09-06
fix(layout): fluid page gutter so laptop widths stop running edge-to-edge

thomgabriel:
On a 1324px laptop window the landing page kept only ~36px of gutter and the platform shell ~54px, so content ran almost edge to edge — the container cap (1400px at 2xl) sits at or above that viewport, leaving a fixed 16-32px padding as the only side margin. This adds a `--page-gutter: clamp(1rem, 5vw, 6rem)` token next to `--header-h` plus a `.page-gutter` utility, and swaps the hardcoded `px-*` on every page-level shell for it: the platform layout, the three landing sections, roadmap, the footer, and the header's inner bar (so the wordmark now tracks the content gutter instead of sitting 16px in). The lesson workspace is exempt — it already cancels the shell's vertical padding at lg, and now cancels the gutter horizontally the same way via `.page-gutter-bleed` before re-applying its own `lg:px-8`, so the editor keeps exactly the width it has today; below lg it picks up the normal gutter. Config `container.padding` is left as-is since it is already overridden everywhere. Browser screenshots at 1324 and 1440 follow in a comment.

### DIFF
--- a/apps/web/src/app/[locale]/(marketing)/landing-client.tsx
+++ b/apps/web/src/app/[locale]/(marketing)/landing-client.tsx
@@ -446,7 +446,7 @@ export function LandingPageClient({
             />
           </div>
 
-          <div className="container w-full px-4 pb-10 pt-8 sm:pb-12 sm:pt-10 md:pb-14 md:pt-14">
+          <div className="page-gutter container w-full pb-10 pt-8 sm:pb-12 sm:pt-10 md:pb-14 md:pt-14">
             <div className="mx-auto grid max-w-6xl items-center gap-12 md:grid-cols-2 md:gap-16">
               <div>
                 <div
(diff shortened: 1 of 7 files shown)


thomgabriel:
Browser probe at the two widths from the report (content edge measured from the viewport's left; production `main` vs this branch):

| Page | Width | Before | After |
|---|---|---|---|
| Catalog | 1324 | 33px | 66px |
| Catalog | 1440 | 52px | 92px |
| Header wordmark | 1324 / 1440 | 16px | 66px / 72px (now aligned with content) |
| Landing hero text | 1324 / 1440 | 86px / 144px | unchanged (own 1152px cap) |
| Code-lesson workspace | 1324 | 32px | 32px (exempted) |
| Prose lesson | 1324 | 278px | 278px |

Why the old numbers were so tight: the Tailwind container's `screens` override leaves only a 1400px cap, so below 1400 the fixed padding was the entire gutter. Full before/after screenshots with the measured edge marked: https://claude.ai/code/artifact/ac67a5f0-88ee-491a-b0f2-d83274a1d2bc

PR #1218, opened 2026-09-09
fix(challenge): grade buildable submissions against the canonical harness

thomgabriel:
A buildable Rust lesson was graded on `data.success` alone, and the build server copies a template Cargo.toml that makes almost any `lib.rs` compile — so replacing `your-first-solana-program`'s starter with `pub fn nothing() {}` compiled, passed every test and awarded XP. What actually carries the requirement is the verification harness the lesson ships at the bottom of its starter, which the grader never required to still be there.

`lib/challenge/harness.ts` splits a source at the `VERIFICATION HARNESS` marker; `buildGradeFiles` strips whatever harness region a submission carries (deleted, edited, duplicated, CRLF) and returns two files: the learner's body under a prepended `#[allow(dead_code)] mod _verify;`, and the starter's canonical harness as `/src/_verify.rs`. `runBuildableSubmission` compiles that instead of the raw submission, and fails closed (503) on a lesson whose starter carries no harness rather than falling back to bare compilation. The browser `runBuildChallenge` sends the identical file list (it only pins `declare_id!`, in `lib.rs`) so an in-editor "Compilar" cannot go green on something the server will fail. New content gate 22 requires every `buildType: buildable` starter to carry the marker plus a non-empty harness — it caught the `good` template fixture, which shipped a two-word starter.

The first version *appended* the harness, and the gate broke it twice: a submission ending in a dangling `#[cfg(any())]` bound that attribute to the appended `mod verify` and deleted it, and a crate-level `#![cfg(any())]` stripped the whole crate. Both compiled and paid XP. With the declaration on line 1 instead, a trailing attribute dangles at EOF and an inner attribute sits below an item — each a hard error, and a learner who redefines `_verify` collides with the declaration.

## Compiles

`cargo build-sbf --manifest-path Cargo.toml --tools-version v1.54` in a scratch dir, `apps/build-server/programs/Cargo.toml` verbatim, files exactly as `buildGradeFiles` emits them. Same verdict on both live buildable lessons (`lesson-b2s-your-first-solana-program`, `lesson-b2s-an-anchor-vault`).

**(a) `pub fn nothing() {}` + a dangling `#[cfg(any())]`** — was a pass, now:

```
error: expected item after attributes
 --> src/lib.rs:6:1
  |
6 | #[cfg(any())]
  | ^^^^^^^^^^^^^

error: could not compile `academy-program` (lib) due to 1 previous error
```

**(b) `#![cfg(any())]` then `pub fn nothing() {}`** — was a pass, now:

```
error: an inner attribute is not permitted in this context
 --> src/lib.rs:4:1
  |
4 | #![cfg(any())]
  | ^^^^^^^^^^^^^^
5 | pub fn nothing() {}
  | ------------------- the inner attribute doesn't annotate this function
```

**(c) `pub fn nothing() {}` alone** — fails on the harness, as before the change:

```
error[E0412]: cannot find type `Ping` in this scope
  --> src/_verify.rs:12:22
   |
12 |     fn accounts() -> Ping {
   |                      ^^^^ not found in this scope
```

and on the vault lesson:

```
error[E0412]: cannot find type `VaultState` in this scope
  --> src/_verify.rs:12:31
   |
12 |     fn vault_state_fields(s: &VaultState) -> (Pubkey, u64, u8, u8) {
   |                               ^^^^^^^^^^ not found in this scope
```

**(d) both lessons' reference solutions** (which carry the harness inline — the region is stripped and re-supplied as the module) and **(e) the same solutions with the harness region removed**: `Finished \`release\` profile [optimized] target(s)`, exit 0. So real work still passes, with or without the harness left in the buffer.

These ten compiles are `src/lib/challenge/__tests__/harness.sbf.integration.test.ts`, opt-in behind `SBF_INTEGRATION=1` plus the toolchain (skipped otherwise). String-level unit tests cover the same shapes on `buildGradeFiles` and on the grader's request body, and a jsdom test asserts the browser runner sends byte-identical files.

Suites: 235 web tests across `components/editor`, `lib/challenge`, `lib/grading` (+10 opt-in SBF); 140 content-lint tests; `tsc --noEmit` and `pnpm lint` clean.

### DIFF
--- a/apps/web/src/components/editor/__tests__/challenge-runner.grade-files.test.tsx
+++ b/apps/web/src/components/editor/__tests__/challenge-runner.grade-files.test.tsx
@@ -1,0 +1,91 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import { NextIntlClientProvider } from "next-intl";
+import type { TestCase } from "@superteam-lms/types";
+import messages from "@/messages/en.json";
+import {
+  buildGradeFiles,
+  HARNESS_MARKER,
+  LIB_PATH,
+  VERIFY_PATH,
+} from "@/lib/challenge/harness";
+import { ChallengeRunner } from "../challenge-runner";
+
+const h = vi.hoisted(() => ({ buildProgram: vi.fn() }));
+
+vi.mock("@/lib/build-server/client", () => ({ buildProgram: h.buildProgram }));
+vi.mock("@superteam-lms/deploy", () => ({ setCachedBinary: vi.fn() }));
+
+const STARTER = [
+  "use anchor_lang::prelude::*;",
+  'declare_id!("Fg6");',
+  "// TODO (p1): add the `ping` handler here.",
+  "// ─────",
+  HARNESS_MARKER,
+  "mod verify {",
+  "    const PING: fn() = first_program::ping;",
+  "}",
+  "",
+].join("\n");
+
+/** The bypass shape: harness deleted, trailing attribute to eat what follows. */
+const CODE = 'declare_id!("Fg6");\npub fn nothing() {}\n\n#[cfg(any())]\n';
+
+const tests: TestCase[] = [
+  {
+    id: "t1",
+    description: "Program compiles successfully",
+    input: "",
+    expectedOutput: "true",
+  },
+];
+
+beforeEach(() => {
+  h.buildProgram.mockReset();
+  h.buildProgram.mockResolvedValue({ success: true, stderr: "", uuid: "u" });
+});
+
+describe("browser runner sends the same file layout as the server grader", () => {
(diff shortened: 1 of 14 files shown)


thomgabriel:
Gate fix pushed as 1f77e7b9. The harness is no longer appended to the learner’s file — it ships as `/src/_verify.rs` with `#[allow(dead_code)] mod _verify;` prepended to the body, so a trailing attribute dangles at EOF and a crate-level inner attribute lands below an item.

Real `cargo build-sbf --tools-version v1.54` against the build-server Cargo.toml, on both live lessons: (a) `pub fn nothing() {}` + `#[cfg(any())]` → `expected item after attributes`; (b) `#![cfg(any())]` → `an inner attribute is not permitted in this context`; (c) `pub fn nothing() {}` alone → `E0412` on the harness; (d) both reference solutions → `Finished release profile`, exit 0; (e) same solutions with the harness region stripped → also clean. Those ten compiles live in `harness.sbf.integration.test.ts` behind `SBF_INTEGRATION=1`.

PR #1219, opened 2026-09-09
fix(ai-partner): applied diffs stay applied, no billed turns on a passing solution

thomgabriel:
The diff card re-derived its apply from the live editor buffer on every render, so accepting a proposal consumed its own `search` text and the card immediately fell into the red "couldn't be applied" branch — `diff.applied` was effectively unreachable in the app, and the existing test only passed because it re-rendered with a static buffer. The accepted result is now latched from a snapshot taken at accept time, an edit whose replacement is already in the buffer (reload, typed by hand, sibling card) renders a neutral "already in your code" state, and the red banner is left for real drift; the apply itself is still evaluated fail-closed against the live buffer.

On the billing side, "Me mostre uma mudança" is withdrawn once the solution passes (the schema forces the model to invent an edit, which is either a no-op or a 502 — both billed), and "Revisar minha solução" is disabled until the buffer changes, which is what let ten reviews of one unchanged solution burn ten turns in prod.

I did **not** refund the no-op propose. The learner sends the buffer the model reasons about, so "already-correct code makes the model propose nothing" is learner-reachable and a refund there would be an unmetered-call lever rather than a courtesy. Instead the route tags that 502 `reason: "no_change"` and tags every billed failure `billed: true`; the client advances its meter on those (it previously returned before `setCounts`, so a billed 502 silently desynced the counter until reload) and says the honest thing — the turn was used, the partner had nothing to change, ask for a review.

Also: one sentence in the header explains the metered → Socratic switch (the counter changing unit and restarting at 1/20 read as a reset), keys added in en/pt-BR/es, and card dismissal is keyed by the proposal's seal token instead of its list index — dismissals still don't survive a reload, which would need a field on the persisted log entry.

Verification: `vitest run src/lib/ai src/components/editor/ai-partner src/app/api/ai` — 297 passed; `tsc --noEmit` clean; lint clean on the touched files.

### DIFF
--- a/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
+++ b/apps/web/src/app/api/ai/__tests__/partner-route.test.ts
@@ -294,6 +294,22 @@ describe("POST /api/ai/partner", () => {
     const res = await POST(makeRequest(VALID_BODY));
     expect(res.status).toBe(502);
     expect(refundAssistTurn).not.toHaveBeenCalled();
+    // The client needs to know the turn is gone (meter) and why (copy) — the
+    // no-op is NOT refunded, because the learner controls the buffer that
+    // provokes it and a refund would be an unmetered-call lever.
+    const body = (await res.json()) as { billed?: boolean; reason?: string };
+    expect(body.billed).toBe(true);
+    expect(body.reason).toBe("no_change");
+  });
+
+  it("marks a billed 502 as billed so the client meter can follow", async () => {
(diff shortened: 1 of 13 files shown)


thomgabriel:
Fixed: a deletion edit (`replace: ""`) now counts as already applied when its `search` text is no longer in the buffer, instead of falling through to the "couldn't be applied" banner. Added deletion/mixed-edit cases to diff-card.test.tsx; full suite (66 tests), tsc, and lint all pass.

PR #1220, opened 2026-09-09
feat(deploy): sign and fund through the embedded wallet

thomgabriel:
`DeployPanel` and `WalletFundingCard` took their signer from `useWallet()` alone, so a learner who signed in with Google had no key, Deploy stayed disabled forever, and the funding card had no address to show — the same dead end enrol had before #1004, with the extra problem that an embedded wallet starts at zero SOL. A new `useDeploySigner()` resolves adapter-first then the Dynamic session (four states, `isDynamicEnabled()` kill switch), `signAllWithDynamicWallet` batches MPC signing, and the panel gates Deploy on a real cost estimate (`estimateDeployCost`) with the funding card inline when the balance is short. An expired session — before or mid-deploy — gets the reauth card, not a generic error; uploaded chunks survive via the existing resume path.

Measurement pending: the embedded batch size is `EMBEDDED_BATCH_SIZE = 15`, provisional. MPC `signAllTransactions` latency for N transactions cannot be measured without a real Dynamic session, so the panel logs per-batch signing time via `console.debug` outside production for the owner to measure and replace the number. The adapter path keeps `BATCH_SIZE = 50` unchanged.

Spec PR: #1213 (sections C1–C7). Content (PR A) and lint (PR B) are separate; do not merge this ahead of the lock bump order in the spec's rollout.

Verified: `packages/deploy` 23/23 green, `pnpm --filter web exec tsc --noEmit` clean, `pnpm --filter web lint` clean (warnings only, matching the existing files), and the new suites for the hook, `signAllWithDynamicWallet`, `estimateDeployCost`, `batchSize`, the panel's embedded/resolving/expired/funding paths and the funding card all pass. The existing `deploy-panel.test.tsx` still passes (its `@superteam-lms/deploy` mock gained the new exports). Two failures in the full web run — `reminder-consent-execution` and one `open-ended-block` timeout — reproduce on stock main and are unrelated.

Sensitive lane: wallet signing and on-chain spend, so this wants the independent adversarial gate plus the browser probe on a fresh embedded account before merge.

### DIFF
--- a/apps/web/src/components/deploy/__tests__/deploy-panel-funding-gate.test.tsx
+++ b/apps/web/src/components/deploy/__tests__/deploy-panel-funding-gate.test.tsx
@@ -1,0 +1,112 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { render, screen, waitFor } from "@testing-library/react";
+import { NextIntlClientProvider } from "next-intl";
+import { Keypair, LAMPORTS_PER_SOL } from "@solana/web3.js";
+import messages from "@/messages/en.json";
+import { DeployPanel } from "../deploy-panel";
+
+/**
+ * The funding gate on the ADAPTER path. The main adapter suite switches the
+ * gate off wholesale (`getCachedBinaryLength: () => null`), so without this the
+ * one behaviour this PR changed for extension learners had no coverage.
+ */
+const ADAPTER = Keypair.generate().publicKey;
+const COST_LAMPORTS = 1.5 * LAMPORTS_PER_SOL;
+
+const h = vi.hoisted(() => {
+  const getBalance = vi.fn();
+  return {
+    getBalance,
+    connection: { getBalance },
+    balanceLamports: 0,
+    estimateDeployCost: vi.fn(),
+  };
+});
+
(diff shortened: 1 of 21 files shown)


thomgabriel:
Fixed the render loop the gate found. `useDeploySigner` was keyed on `parseWalletAddress`'s result (a fresh `PublicKey` per render) and on the Dynamic account object (rebuilt per read), so `signer` was new on every render and `WalletFundingCard.refreshBalance` re-fired forever; it now keys on the address string and reaches the live account through a ref. `WalletFundingCard` re-derives its own key from the address too, so a future caller can't re-break it.

Measured with a new test that renders the card against the real hook: 11,874 `getBalance` calls in the 5 s before it timed out, now 1 (asserted ≤ 2 across five re-renders). The hook test also pins `signer` identity across re-renders.

Two follow-ups in the same change: the funding gate on the adapter path now has a test (short balance → card shown, Deploy disabled; sufficient → enabled), and `generic-program-explorer` routes `isDynamicSessionExpiredError` during execute to the re-auth card instead of a raw error, the way the panel does.

`src/hooks` + `src/components/deploy` 66/66, typecheck 6/6, lint clean on the touched files. `EMBEDDED_BATCH_SIZE = 15` left as is, still pending the C3 measurement.

PR #62, opened 2026-03-04
Superteam Academy — On-chain Solana Developer Education Platform

killa3241:
## Superteam Academy

Superteam Academy is an on-chain developer education platform for Solana that transforms learning progress into verifiable developer reputation.

Instead of a traditional LMS, the platform connects developer learning progress to their wallet identity, enabling portable credentials within the Solana ecosystem.

---

## Key Features

• Wallet-based authentication using Solana Wallet Adapter  
• Google OAuth login via NextAuth  
• XP-based developer reputation system  
• Interactive coding challenges using Monaco Editor  
• Course progression and lesson tracking  
• Leaderboard ranking system  
• Verifiable course certificates  
• Multi-language support (English / Spanish / Portuguese)

---

## Tech Stack

Next.js 14 (App Router)  
TypeScript  
Tailwind CSS  
shadcn/ui  
TanStack React Query  
Solana Wallet Adapter  
NextAuth (Google OAuth)  
Monaco Editor  
Google Analytics & Microsoft Clarity

---

## Links

Live Demo:  
https://superteam-academy-git-main-anikait-s-projects.vercel.app/

Demo Video:  
[VIDEO_DEMO](https://youtu.be/GnSEaxrlzbA)

GitHub Repository:  
https://github.com/killa3241/superteam-academy

PR #1222, opened 2026-09-09
ci(content-lint): compile Rust starters and solutions through the real toolchain

thomgabriel:
Gate 6 ran the real oracle for TypeScript blocks and emitted a notice for every `buildType: buildable` Rust block, so nothing in CI checked that a Rust starter fails and its solution passes — part of why the stale build-server image and compile-only grading both went unnoticed (audit P1 #4). It now compiles both sides with `cargo-build-sbf --tools-version v1.54` against `apps/build-server/programs/Cargo.toml`, splicing the starter's canonical harness onto each source exactly as the runtime grader does (that splice is duplicated from #1218's `harness.ts` with a keep-in-sync note, since content-lint cannot depend on the app). One build directory and one `CARGO_TARGET_DIR` serve every block in a run, each build is capped by `CONTENT_LINT_RUST_TIMEOUT_MS` (300s) and a timeout is an error rather than a hang; without the toolchain on PATH the old notice returns, saying so, and `CONTENT_LINT_RUST_ORACLE=require` — what CI sets — turns that absence into an error so the gate cannot silently stop running.

The reusable `validate-content` workflow now installs the Solana CLI + platform-tools with the same `solana-${{ runner.os }}-...-tools-v1.54` cache key as the on-chain jobs, plus a cargo/target cache keyed on the template `Cargo.lock`.

The content repo's workflow lives in `solanabr/academy-courses` and is a standalone copy rather than a caller of the reusable one, so I could not land the change there. `docs/ci/academy-courses-validate-content.yml` is the drop-in replacement that calls this workflow instead, and `docs/ci/academy-courses-template-harness.patch` fixes `courses/_template`, whose `// starter program` / `// solution program` placeholders both compile and which the oracle therefore fails until it ships a real harness.

Verified against a fresh clone of academy-courses with `CONTENT_LINT_RUST_ORACLE=require`: both live buildable lessons pass, and inserting a compile error into `an-anchor-vault`'s solution (above the harness marker) produces `solution does NOT compile against the build-server template + its own harness`. 144 package tests pass, `tsc --noEmit` clean; the real-compile tests skip where `cargo-build-sbf` is absent.


thomgabriel:
Merged main and rebuilt the oracle on top of #1218: it now compiles the same two files the grader sends (`/src/lib.rs` = `mod _verify;` + body, `/src/_verify.rs` = `use super::*;` + the starter's harness) instead of the append this PR copied from #1218's first shape. Gate 22 now imports `HARNESS_MARKER` from the shared module rather than keeping a second copy, and `docs/ci/` drops the template patch since academy-courses #62 merged.

Verified against a fresh clone of the content repo with `CONTENT_LINT_RUST_ORACLE=require`: 0 errors, both live buildable lessons compile (starter fails, solution passes). Tampering one solution to `// nothing implemented` + a trailing `#[cfg(any())]` — which the old append layout would have passed — now fails the gate.

PR #61, opened 2026-03-04
feat(app): add Next.js frontend with all core pages and services

StaySafe020:
- 10 pages: landing, courses, course detail, lessons, dashboard, profile, leaderboard, settings, certificates
- Wallet auth via Solana Wallet Adapter (Phantom, Solflare, etc.)
- i18n support (EN, ES, PT-BR) with next-intl
- Dark/light theme with full design token system
- Gamification UI: XP, levels, streaks, achievements, leaderboard
- Service layer with clean interfaces and stub implementations
- Code editor component (placeholder for Monaco/CodeMirror)
- Responsive design with Tailwind CSS + Radix UI components
- Container centering fix and gamification color utilities
- Fix Credential type alignment across stubs and components

PR #64, opened 2026-03-04
feat(app): Production-ready LMS frontend with Solana on-chain integration, gamification, code editor, i18n, and analytics — Bounty Submission

ikarn-dev:
- Live Demo: https://superteam-academy-lyart.vercel.app/en

**Overview**
Production-ready LMS for Solana development built with Next.js 16 + TypeScript (strict) + Tailwind CSS 4 + Solana Wallet Adapter + Anchor + Metaplex Core. Interactive courses, on-chain soulbound credentials, gamification (XP/streaks/achievements), code editor with execution, and i18n (EN/PT-BR/ES) — backed by a self-deployed Anchor program on Devnet.

**What's Implemented**
All 10 Required Pages

- / Landing, 
- /courses Catalog, 
- /courses/[slug] Detail, 
- /courses/[slug]/lessons/[id] Lesson + Code Editor,
- /dashboard, /profile + /profile/[username], 
- /leaderboard, 
- /settings, 
- /certificates/[id], 
- /achievements

**Bonus**

- /login 
- /onboarding 
- /challenges 
- /community 
- /community/new 
- /community/[id] 
- /admin 
- /admin/users 
- /admin/courses 
- /admin/community 
- /offline

**Authentication**
- Solana Wallet (Ed25519 + nonce + lockout) / Google OAuth / GitHub OAuth
- Account linking (wallet + Google + GitHub to one profile)
- JWT sessions (7-day), RBAC (student/admin), audit logging
- Solana Integration (Devnet)
- Live: Wallet auth, XP balance (Token-2022), credential NFT display (Helius DAS), leaderboard (XP indexing), course enrollment (user-signed tx)
- Stubbed with clean interfaces: Lesson completion, course finalization, credential issuance, achievement claiming — 14 service modules with executeWithRetry and isTransientError patterns, ready for on-chain swap
- Gamification
- XP: Soulbound SPL Token-2022, 10-level progression (0-120K XP)
- Streaks: Dual model (login + activity), freeze system, milestone rewards (7/30/100 day)
- 13 Achievements: Progress, Streak, Skill, Special categories — each a soulbound Metaplex Core NFT
- 5 Credential Tracks: Anchor, DeFi, Mobile, Pinocchio, Token Engineer — upgrade-in-place NFTs
- Code Editor
- CodeMirror 6 (Rust/TS/JSON) + Judge0 CE execution, split resizable layout, test cases with pass/fail, starter code, XP on completion

**Analytics**
[...]

PR #1226, opened 2026-09-09
feat(content): lock bump to academy-courses @f2d97edb

thomgabriel:
Bumps the content pin to academy-courses main and activates the deploy lesson's fund + deploy-card blocks, now that #1220 put the embedded-wallet signing path on main. In b2s `your-first-solana-program` a `wallet-funding` block produces `funded-wallet`, the deployable code block consumes it and produces `deployed-program`, and a `deployed-program-card` consumes that (#61); #62 gives the template scaffold a real verification harness, and the same lesson picks up a paragraph on what the deploy actually costs.

Counts and slots are unchanged (4 courses / 36 lessons, `slots.json` byte-identical), so only that one lesson moves — lessons.json and course-by-slug.json are the only goldens regenerated. The capstone-gate dormancy test asserted the alpha catalog hosts no deploy panel at all; that was never the load-bearing invariant, since the gate keys on the course, so it now asserts no panel sits inside a gated course.


thomgabriel:
Re-bumped to `f2d97edb` (academy-courses #64) — the ping-program solution's harness moves to match the starter byte-for-byte, comments only. Only that block's solution text changes; slots.json byte-identical, counts unchanged (4 courses / 36 lessons). Goldens + tests updated, `tsc --noEmit` clean.

PR #1230, opened 2026-09-09
feat(deploy): session-key upload for embedded wallets

thomgabriel:
Implements the session-key addendum of the deploy design (#1227) so an embedded-wallet learner can finish the deployable lesson from #1220: the wallet signs one transaction — a transfer of `estimateSessionKeyDeployCost` into a keypair generated in the tab — and that key is payer and authority for the buffer, the writes and the finalize, all signed locally with no batch ceiling. Immediately after the deploy confirms, a Loader v3 `SetAuthority` (variant 4, new authority non-signing) moves the upgrade authority to the learner, so `verifyProgramDeployment` and `/api/deploy/save` keep attributing the program exactly as today, and the session key then sweeps its balance minus the sweep fee back to the learner. The secret is persisted in the wallet-scoped sessionStorage record under AES-GCM with an HKDF key derived from a per-page-load nonce plus the buildUuid, so resume continues from the saved offset with the same key and no second transfer; the threat model (a storage dump, not code on the page) is documented on the module. Start over closes the buffer to the session key and sweeps before regenerating, a failed `SetAuthority` is retried with backoff and otherwise surfaces a "claim ownership" button that re-runs only that step, and a failed refund is a non-blocking warning with a retry.

Saved state is validated before any resume — build id, offset bounds, chunk count against the cached binary, well-formed keypair secrets, the program keypair matching this build's `declare_id!` injection, and the on-chain buffer's authority being the session key — so a crafted sessionStorage record cannot make a learner pay into a foreign buffer or deploy to someone else's program id. The adapter path is untouched: extension wallets still sign their own batches at the existing batch size.

Tests: 34 in `packages/deploy` (funding amount equals the estimate and the helper is called once, every chunk and finalize tx signed by the session key, SetAuthority accounts and new authority, sweep amount, resume reuses key and offset, start over closes then sweeps, crafted-state rejection) and 128 across `src/hooks src/components/deploy src/lib/dynamic src/lib/deploy` including the encryption round trip and the panel's one-signature funding path. `vitest.setup.ts` restores Node's `Uint8Array` under jsdom, without which every web3.js instruction encode in a component test fails with "b must be a Uint8Array". `tsc --noEmit` and lint are clean; the twelve PGlite suites that time out in a full parallel run pass individually and fail the same way on a clean tree.

Devnet: UNVERIFIED. This machine has no default `solana` keypair and the public devnet faucet refused every airdrop, so the end-to-end run is for the gate. What was verified locally is the artefact it would deploy: the lesson solution built with `cargo build-sbf --tools-version v1.54` against `apps/build-server/programs` is 66,704 bytes — 75 chunks, the size that produced the rate-limit failure this PR removes.

Expect a small conflict with the MPC-backoff/airdrop-RPC branch; this PR touches neither `airdrop.ts` nor the retry logic in `signAllWithDynamicWallet`.

### DIFF
--- a/apps/web/src/components/deploy/__tests__/deploy-panel-signer.test.tsx
+++ b/apps/web/src/components/deploy/__tests__/deploy-panel-signer.test.tsx
@@ -1,6 +1,12 @@
 // @vitest-environment jsdom
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
-import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import {
+  cleanup,
+  render,
+  screen,
+  fireEvent,
+  waitFor,
+} from "@testing-library/react";
 import { NextIntlClientProvider } from "next-intl";
 import {
   Keypair,
@@ -32,6 +38,7 @@ const h = vi.hoisted(() => ({
   deployProgram: vi.fn(),
   runSessionKeyDeploy: vi.fn(),
   startOverSessionKey: vi.fn(),
+  sweepSessionKey: vi.fn(),
   binaryLength: null as number | null,
   estimateDeployCost: vi.fn(),
   balanceLamports: 0,
@@ -84,6 +91,7 @@ vi.mock("@superteam-lms/deploy", async (importOriginal) => ({
   resumeDeployment: vi.fn(),
   runSessionKeyDeploy: h.runSessionKeyDeploy,
   startOverSessionKey: h.startOverSessionKey,
+  sweepSessionKey: h.sweepSessionKey,
   getCachedBinaryLength: () => h.binaryLength,
   estimateDeployCost: h.estimateDeployCost,
   estimateSessionKeyDeployCost: h.estimateDeployCost,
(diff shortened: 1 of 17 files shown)


thomgabriel:
Gate follow-ups, all seven:

**F1** a reload no longer deletes the record — the session key’s address survives, is shown with a copy button, and outlives a start over, and `sessionKeyNotice` now says a reload during the upload strands the funds until you recover them by hand. **F2** the funding signature is persisted *before* the send (`fund` gets an `onSignature` callback), and a carried-over key is checked against the chain — balance first, then the signature’s status — before anything is transferred again, so a `fund` that throws after broadcasting can never be paid for twice. **F3** the retry after a failed start-over sweep re-runs the close-and-sweep with the key the reset was about to discard, instead of returning immediately; the banner is now rendered in the ready and paused states too, which is where a start over actually lands. **F5** `sessionKeyLost` is reachable (it is the F1 notice); every other new string is used. **F6** the single funding signature goes through `signAllWithRateLimitBackoff` in the hook’s embedded signer, blockhash refresh included — it is the only remote signature in this flow. **F7** `OwnershipTransferError` carries the session ADDRESS and the program id, never the secret.

Gate’s devnet run stands: program `7T4BtECMRFvRFCdA5NsYZhoCmNMATyZG47F8qr3sNbaX`, 75 chunks in 13.4 s, one wallet signature. `packages/deploy` 45/45, `apps/web` deploy/hooks/dynamic suites 154/154, `tsc --noEmit` clean in both packages, lint unchanged.

Gate F7 (the program-id pin only binding when `programKeypairSecret` is passed) is untouched: with no injected keypair there is nothing to pin against, and the build that sets it is the same one that caches the binary a resume requires.

PR #1231, opened 2026-09-09
feat(deploy): a deploy flow a learner can follow

thomgabriel:
The deploy lesson didn't answer "I pressed Compilar, now what?": the deploy card sits in the left column far from the editor, the funding card printed `Airdrop failed after 3 attempts: 403 : {"jsonrpc":...}` under a bar that said "Pronto para o deploy!", and Submit lived in the editor toolbar rather than on the card that had just said the program was live.

A `Build → Fund → Deploy → Submit` stepper now sits above the deploy panel and as a slim strip beside the Build button; both read one module store (`lib/deploy/flow-store.ts`), clicking a step scrolls to its card, and a finished build scrolls the panel into view and focuses it (reduced motion jumps). Every failure goes through the new `lib/deploy/friendly-error.ts`, which maps airdrop/wallet throttles, dead sessions, gone builds, thin balances and stale blockhashes to one sentence and one action; an unmapped message is the only thing that keeps its raw text, behind a "show details" toggle. Progress is now a bar with the phase, elapsed and a rough remaining estimate, with the signature log (explorer links intact) folded under a disclosure — and an automatic wallet-service retry reads as "waiting, retrying automatically" instead of "paused". The success card carries the program id, rent, deploy time and the lesson's own Submit (it asks the editor via `superteam:request-submit`, so enrollment and grading are untouched), then swaps to XP earned plus a next-lesson link. Funding shows required / balance / shortfall, keeps "ready to deploy" for when the balance actually covers it, and offers "Open faucet" which copies the address.

Copy is in en / pt-BR / es. 48 new tests cover step derivation, every friendly-error mapping (including that raw text never renders outside the disclosure), the progress math, each panel state's copy and actions, the success-card submit wiring and both funding routes; `pnpm --filter web exec vitest run src/components/deploy src/components/editor src/lib/deploy src/hooks src/messages` is 289/289, tsc and lint are clean. Probed live at `/en/courses/btc-to-sol-evolution/lessons/your-first-solana-program` in light and dark: the toolbar strip renders next to Build with BUILD active, and the panel shows the stepper over the anonymous funding card.

Four unrelated suites (admin courses page, instructor courses, admin freeze-gate, lesson numbering) time out at vitest's 5s default in this cold worktree and pass individually — `--testTimeout=60000` turns the freeze-gate red green with no code change.


thomgabriel:
Addressed the gate follow-ups in d9971be6: Submit on the success card now gates on `saveStatus` (disabled + "Recording your deploy…" while saving, enabled once saved, hidden in favor of the existing retry affordance if the save fails) instead of `isComplete` alone — a 500 from `/api/deploy/save` no longer lets a learner bank XP with no `deployed_programs` row. Also dropped the duplicate "Back to the editor" button on the build-expired card and bumped `--label-disabled` so the dark-mode "Not enough SOL" button clears 3:1 contrast (was 2.27:1).

`vitest run src/components/deploy` (63/63), `tsc --noEmit`, and lint are clean; i18n key parity holds across en/pt-BR/es (1533 keys each).

PR #1233, opened 2026-09-09
fix(deploy): drop the stepper, explain harness errors, carry quiz answers on submit

thomgabriel:
Three things from the owner's run through `your-first-solana-program`.

The Build → Fund → Deploy → Submit stepper is removed — the toolbar strip, the panel header, `lib/deploy/steps.ts`, the flow store and the hook. Only the scroll-to-panel-after-a-build survives. Its "build" tick came from "a buildUuid exists", which outlives the failed re-build that follows, so a red compile still showed a green step; the comment where the scroll now lives says not to derive any future indicator that way.

Since #1218 the verification harness compiles as `src/_verify.rs`, so a submission that doesn't define what the lesson asks for fails inside a file the learner has never seen, and the test row printed `error[E0425]: cannot find function ping … --> src/_verify.rs:12:22` verbatim. Diagnostics located in that file now become one sentence naming the symbol (or a fallback when the message names none), in en/pt-BR/es, with the raw stderr still behind the details toggle; `src/lib.rs` errors are the learner's own code and are untouched. The server grader's 403 `detail` uses the same mapping so the two verdicts agree.

Root cause of the P0: submitting was never gated on the lesson's other graded blocks, because `isGateBlock` includes `code` — which never reports done, since the editor's submit IS its proof — so the client gate was false on every challenge lesson and could not be applied to a code submit at all. On a deploy lesson the quiz sits BELOW the deploy card, so the natural order is deploy → Submit → a POST with no quiz proof → the server's `quiz_failed`, whose copy blames answers the learner never gave; the editor then kept that rejection on screen while the quiz was finished above it, which is the "2/2 CONCLUÍDO and still Não aceito" the owner saw. The gate now excludes `code`, both submit entry points (toolbar and success card) go through one payload builder and that one gate, the success card's Submit waits and says which block is outstanding, and a gate-blocked rejection retires itself when the block is done. The toolbar Submit had the same hole — both funnel through `superteam:lesson-complete`.

`pnpm --filter web exec vitest run src/components/deploy src/components/editor src/lib/deploy src/lib/challenge src/app src/messages` is 1158 passing, tsc and lint clean, three locales at 1528 keys each. The new `submit-payload` suite renders the lesson, answers the quiz and asserts the POST body from both entry points, plus the blocked and the un-blocked cases.


thomgabriel:
Dropped the XP chip and next-lesson button from the deploy success card — submit now just shows the same "Lesson Complete!" text the editor toolbar uses elsewhere. `nextLessonHref` is removed end to end (BlockContext, DeployPanel, DeploySuccessCard) since the deploy card was its only caller; matching i18n keys pruned from en/pt-BR/es. Tests, tsc, and lint all green.

PR #1234, opened 2026-09-10
feat(profile): verified badge kinds — Superteam member vs verified partner

Potolski:
`profiles.verified` (#997) answers "is this author vouched for" with a boolean, and every course page printed the same green seal for every yes. The Academy now publishes from two kinds of author — Superteam members and outside partner organisations (Forge College first) — and a learner deciding whether to trust a course should be able to tell which they are looking at.

## Data model

`verified_kind` carries the **which**; `verified` keeps carrying the **whether**. Splitting them rather than widening `verified` into an enum means every existing reader keeps working untouched, and a verified author nobody has classified yet keeps today's badge instead of losing one.

Admin-only, following the `wallet_address` precedent (#408): service_role writes, enforced by the **same** BEFORE trigger the other privileged columns use, so one place still answers "which profile columns are privileged" (the third such extension, same documented exception as `wallet_kind` #1183). A self-awardable "Superteam member" badge would be worth nothing.

Two CHECKs back it:
- the closed set `('superteam','partner')`, mirrored in TypeScript by `toVerifiedKind` so a typo'd column value can't reach the UI as a kind;
- `verified_kind IS NULL OR verified` — a kind can never sit on a profile that shows no badge at all.

The trigger body is copied from the **live** #997 definition, not from `schema.sql`, per the #1116 trap.

## Look

| kind | token | light | dark | tooltip |
|---|---|---|---|---|
| superteam | `--primary` | `rgb(10,112,85)` deep green | `rgb(46,204,142)` mint | Superteam member |
| partner | `--text` | `rgb(28,25,23)` near-black | `rgb(230,237,243)` near-white | Verified partner |

`--text` **is** the theme's ink, which is what "white or black depending on the mode" asks for — no third colour invented.

Colour is never the whole message (colour-blind readers, greyscale screenshots, screen readers): the label is the tooltip *and* the accessible name, and the badge is `tabIndex=0`, so the label is reachable by keyboard and by tap, not hover-only. Uses the house Radix tooltip pattern already used by the achievements strip and leaderboard.

Strings added in all three locales.

## The three authors

The migration backfills them:

| author | kind | matched by |
|---|---|---|
| Kaue Kano | superteam | wallet `3WECquw…AJzQ` — creator of btc-to-sol-evolution, pilula-solana-superteam, solana-speedrun |
| Artur Moura (Forge College) | partner | wallet `Em8D6Xyu…qxVZ` — creator of visao-geral-solana, the Forge College pilot |
| David Potolski Lafetá | superteam | **account sign-in email** |

Wallet is the right key: it's what a course's on-chain `Course.creator` is, so it's the only identifier tying an author to the courses they publish. **David has no course in the bundle yet, so there is no creator wallet to match on** — his row is matched by `auth.users.email` instead. If his account signed in with a wallet only, or a different address, that UPDATE matches nothing and silently no-ops. The ledger entry carries the VERIFY-ON-APPLY check:

```sql
SELECT username, verified, verified_kind FROM public.profiles WHERE verified_kind IS NOT NULL;
-- must return all three rows
```

If his is missing, one `UPDATE … WHERE id = '<his uuid>'` fixes it — tell me the wallet or username and I'll pin it properly instead.

## Two things a reviewer should know

**`schema.sql` is deliberately not touched.** I did add `verified_kind` to its `public_profiles` view, and the embedded-Postgres coherence tests caught it immediately: `column p.display_name does not exist`. The snapshot's `profiles` table predates #997 entirely, so its view cannot reference any of these columns. Backporting #997 into the snapshot is the separate piece of work #1116 explicitly warns against doing casually, so I reverted my edit and left the snapshot as the ledger already documents it — stale for these columns, with the migration as the truth.

**The migration is not applied.** Per the ledger's standing rule this repo ships SQL and docs only. The ledger row is added with the post-apply checks, and a `public_profiles.verified_kind` expectation is registered in `schema-expectations.ts` so `/api/health/schema` catches the merge→apply gap (#731) rather than letting a 400 reach a course page.

## Verification

- `pnpm typecheck` ✓ · `pnpm lint` ✓
- New `verified-badge.test.tsx` (6 cases): each kind's label and colour class, the no-kind fallback keeps the pre-#1234 badge, keyboard reachability, and `toVerifiedKind` rejecting `"Superteam"`/`"SUPERTEAM"`/`""`/`null`/numbers.
- Full `apps/web` suite: **3697 passed, 0 failed** (10 skipped).
- Rendered at 1100px in both themes on both real course pages, with the deployment gate and profile resolution locally stubbed (Supabase is unreachable here). Both stubs are reverted and verified absent from the diff. Measured colours are the table above; hovering opened the correct tooltip in all four combinations.

## Commit

Authored by David Potolski Lafetá; no co-author trailers.

https://claude.ai/code/session_01MJMV9QpjXQ42PVjgEeMQp4

---


Potolski:
Loose end closed in `0f4a8c61`. Owner supplied David's wallet, so all three authors are now keyed the same way.

| author | kind | wallet |
|---|---|---|
| Kaue Kano | superteam | `3WECquw…AJzQ` |
| Artur Moura (Forge College) | partner | `Em8D6Xyu…qxVZ` |
| David Potolski Lafetá | superteam | `8kMziL5e…whbSo` |

All three base58-decode to exactly 32 bytes (checked, not assumed).

David keeps a **second** UPDATE keyed on sign-in email rather than having it replaced: `wallet_address` is only on the profile row if the account has linked a wallet, and a Google-only sign-in carries none — so the wallet statement alone could still match nothing. Both statements are idempotent and target the same person, so whichever identifier his row actually holds, the badge lands.

The VERIFY-ON-APPLY caveat the ledger carried for him is removed. The post-apply check is unchanged and still the thing to run:

```sql
SELECT username, verified, verified_kind FROM public.profiles WHERE verified_kind IS NOT NULL;
-- expect all three rows
```

---

Potolski:
Confirmed and fixed in `8505658d`. You were right about the cause too — the body was written from #997's text, which is why it matched #997's column set exactly.

## The same bug is already live

Auditing the function's lineage while fixing this turned up the identical mistake, in the other direction, already applied to production:

| migration | applied | guarded columns |
|---|---|---|
| `20260710120000` (#696) | yes | `wallet_address` |
| `20260805120000` (#997) | yes | `wallet_address`, `display_name`, `verified` |
| `20260824120000` (#1183) | **yes — `20260824213846`** | `wallet_address`, `wallet_kind` ← dropped #997's two |
| `20260910120000` (this PR) | pending | all five |

#1183 rebased the body onto the #710 text and lost `display_name` and `verified` on the way. Since it's applied and the trigger is the only thing narrowing the policy:

```sql
CREATE POLICY "Users can update their own profile"
  ON profiles FOR UPDATE USING (auth.uid() = id);   -- no column list
```

…and there are no column-level `REVOKE UPDATE`s on `profiles` (checked across every migration), an authenticated learner can currently `PATCH /rest/v1/profiles?id=eq.<self>` with `{"verified": true, "display_name": "Superteam Official"}` and it lands. That is verbatim the escalation #997's own header describes: *"could rename themselves 'Superteam Official' and self-award a verified badge."*

So this migration is now a fix, not only an addition, and I'd treat its apply as time-sensitive rather than routine. The replaced body guards all five: `wallet_address`, `wallet_kind`, `display_name`, `verified`, `verified_kind` — on `UPDATE` (raise) and on `INSERT` (coerce), since a column guarded only on `UPDATE` is still squattable at row creation.

## The invariant that was missing

`wallet-write-lock-guard.test.ts` couldn't catch either regression, and it isn't a gap in that file: it asserts a guard is *present*, and every version of this body has always guarded the column its own migration cared about. What was missing is the cumulative property — that the newest body is a superset of every earlier one.

`apps/web/src/lib/supabase/__tests__/profile-write-lock-union.test.ts` pins that, derived from the migrations rather than a hardcoded list, so it holds for whatever column gets locked next. Mutation-checked by re-deleting the `wallet_kind` guard:

```
× the newest body guards every column any earlier body guarded
  AssertionError: 20260910120000_verified_kind.sql drops UPDATE guards that earlier
  migrations added: wallet_kind. CREATE OR REPLACE swaps the whole body — re-add
  them, do not rebase the body onto an older copy.
```

It also pins #1183's set as a historical fact, so the reasoning above can't rot into a stale comment.

## What I did not touch

`supabase/schema.sql`. Its copy of the function is still at the #1183 state, and its `profiles` table has no `display_name`/`verified`/`verified_kind` columns at all — it is coherent with itself at that older point, and moving the function forward without the table would break every environment built from the snapshot. That's the documented #1116 drift; papering over half of it here would be worse than leaving it. Flagging it as a separate piece of work rather than silently widening this PR.

## Post-apply verification, revised

The ledger row now asks for more than the badge query, because a green apply is not evidence the guards came back — only the errors are. As an **authenticated** (not `service_role`) session:

```sql
SELECT username, verified, verified_kind FROM profiles WHERE verified_kind IS NOT NULL;
-- expect all three authors

-- then, one per column, each must fail with 42501:
UPDATE profiles SET verified      = true      WHERE id = auth.uid();
UPDATE profiles SET verified_kind = 'superteam' WHERE id = auth.uid();
UPDATE profiles SET display_name  = 'x'       WHERE id = auth.uid();
UPDATE profiles SET wallet_kind   = 'external' WHERE id = auth.uid();
UPDATE profiles SET wallet_address = 'x'      WHERE id = auth.uid();
```

`pnpm typecheck`, `pnpm lint`, and the full `src/lib/supabase/__tests__` suite (29 files, 454 tests) pass, embedded-Postgres suites included.

---

Potolski:

Potolski:
Applying this hit a second failure, fixed in `ec4846ca`:

```
ERROR:  42501: permission denied for table users
HINT:   Grant the required privileges to the current role with:
        GRANT SELECT ON auth.users TO service_role;
```

`service_role` is not "postgres with more rights" — it has **less** reach into `auth`. Supabase grants it the public schema, not the auth schema. So the two halves of the backfill need different privileges and neither role can do both:

| statement | needs | why |
|---|---|---|
| `... JOIN auth.users` | `postgres` | service_role can't see the auth schema |
| `UPDATE profiles SET verified …` | `service_role` | the trigger blocks everyone else |

The `SET ROLE service_role` from `6b86383f` fixed the second and broke the first.

Fixed by resolving the email to a profile id **before** the role switch and carrying only that id across in a `TEMP` table, dropped at the end:

```sql
CREATE TEMP TABLE _verified_kind_targets (id UUID PRIMARY KEY);
INSERT INTO _verified_kind_targets (id)
SELECT p.id FROM public.profiles p JOIN auth.users u ON u.id = p.id
 WHERE lower(u.email) = '…';
GRANT SELECT ON _verified_kind_targets TO service_role;

SET ROLE service_role;
  …
  UPDATE public.profiles SET verified = true, verified_kind = 'superteam'
   WHERE id IN (SELECT id FROM _verified_kind_targets);
RESET ROLE;
DROP TABLE _verified_kind_targets;
```

**Deliberately not the hint's fix.** `GRANT SELECT ON auth.users TO service_role` clears the error by permanently widening service_role's reach into `auth` — to run a one-off badge backfill. There's a test asserting that grant does *not* exist after the migration.

## Why the execution test didn't catch it

It did the same thing the hint suggests. The pglite harness I added in `6b86383f` contained:

```sql
GRANT SELECT ON auth.users TO service_role;   -- "to model production"
```

That models a database Supabase does not issue. It made the harness agree with the migration instead of testing it, and masked a file that could not run on production at all. The grant is gone, its absence is now commented as load-bearing, and the suite reproduces the exact failure when the fix is removed.

It's the same failure mode as the original review finding, one level up: the first bug was a body copied from the wrong source, this was a fixture copied from the wrong source. Both looked right in review.

## Added coverage

- **an author with no linked wallet** badged via the email statement alone — the Google-only sign-in case that is the entire reason that second UPDATE exists, and which nothing exercised before
- the temp table not outliving the migration
- `has_table_privilege('service_role', 'auth.users', 'SELECT')` is `false` afterwards
- **a second apply is a no-op, not an error** — relevant to anyone whose first attempt died part-way

14 execution cases; `pnpm typecheck`, `pnpm lint`, and the full `src/lib/supabase/__tests__` suite (30 files, 468 tests) pass.

## For anyone whose apply already failed

Re-run the whole corrected file. Every ADD is `IF NOT EXISTS`, every constraint `DROP`-then-`ADD`, every function and view `CREATE OR REPLACE`, and the backfill UPDATEs are absolute assignments — so a partial apply is repaired by re-running, never by running the leftover half. A failed run may also leave the *session* as `service_role` (its `RESET ROLE` never executed); that dies with the connection, but `RESET ROLE;` or a fresh editor tab settles it.

---

